PR #137, opened 2020-09-23
Add API: is_in_space()

qinsoon:
Add an API: `pub fn is_in_space(mmtk, obj, allocator) -> bool`. It checks if the object is located in a space with the given allocation type (allocation semantics). Its implementation should be consistent with the `alloc()` implementation for each plan (i.e. it should check whether an object is in a space in the same way as how `alloc()` maps allocation types to each space). 

Currently no VM binding is using it, so it is not covered in any tests. The V8 binding needs this API.


qinsoon:
Closing this for now. This PR has been inactive for a while.

PR #159, opened 2020-11-02
Document top-level modules and VM interfaces

qinsoon:
This PR adds documentation to top-level modules and VM interfaces. You can view it here: https://mmtk.github.io/mmtk-core/mmtk/index.html

* Adds rustdoc-style comments to a few files.
* Adds a script `ci-doc.sh` that builds the doc.
* Adds a workflow to publish rustdoc as github page for the mmtk-core repo when a commit is pushed to master.
* Removed `unboxed_size_constants` from the `vm` module.


qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=pdate-mmtk-core-159

qinsoon:
jikesrvm-perf-compare
JIKESRVM_BINDING_BRANCH_REF=update-mmtk-core-159

src/lib.rs
```diff
@@ -22,7 +22,8 @@
 //! Logically, this crate includes these major parts:
 //! * GC components:
 //!   * [Allocators](util/alloc/allocator/trait.Allocator.html): handlers of allocation requests which allocate objects to the bound space.
-//!   * [Spaces](policy/space/trait.Space.html): memory regions in the heap with defined semantics and behaviors.
+//!   * [Policies](policy/space/trait.Space.html): definitions of semantics and behaviors for memory regions.
+//!      Each space is an instance of a policy, and takes up a unique proportion of the heap.
```
javadamiri, line 26: I don't understand the meaning of _unique proportion_ here.
Does it mean two spaces can not have the same proportion of the heap (e.g. have the same size), or that spaces occupy non-overlapping virtual memory address ranges?

qinsoon: It means spaces are non-overlapping in the heap.

javadamiri: Not sure, but I think it's vague.
Maybe write it as:
Each space is an instance of a policy. Spaces take up non-overlapping pieces (chunks? pages? ...) of the heap.

src/mm/memory_manager.rs
```diff
@@ -1,3 +1,16 @@
+//! VM-to-MMTk interface: safe Rust APIs.
+//!
+//! This module provides a safe Rust API for mmtk-core.
+//! We expect the VM binding to inherit and extend this API by:
+//! 1. adding their VM-specific functions
+//! 2. exposing the functions to native if necessary. And the VM binding needs to manage the unsafety
```
javadamiri, line 6: Maybe rewrite as?:
exposing the functions to native if necessary (The VM binding needs to manage the unsafety for exposing this safe API to FFI).

src/lib.rs
qinsoon: Yeah. Sounds good. I will change it.

src/mm/memory_manager.rs
qinsoon: Sure. I will address your comments all the together when you finish the review.

qinsoon: Yeah, sounds good. Feel free to do so. But probably be cautious about the time spent on this. I expect the open PRs will be merged today, and I plan to do another PR to rename `enum Allocator` to `AllocationSemantics` (and get it merged).

src/mm/memory_manager.rs
```diff
@@ -20,46 +33,64 @@ use crate::util::heap::layout::vm_layout_constants::HEAP_START;
...
-// How the VM gets mutator/collector/tracelocal handles:
-// * Mutator: from bind_mutator() as Box<Mutator>
-// * Collector: from Collection::spawn_worker_thread() as &mut Collector
-// * TraceLocal: Scanning::* as &mut TraceLocal
-
+/// Run the main loop for the GC controller thread. This method does not return.
+///
+/// Arguments:
+/// * `mmtk`: The reference to an MMTk instance.
+/// * `tls`: The thread that will be used as the GC controller.
```
javadamiri, line 40: `tls` is not clear.
What does thread mean here?
Is it a pointer to a thread-local storage (e.g. an `int`) that my be used by the current function?
Or is it a `pthread_t *` pointing to the thread?

qinsoon: It doesnt matter. It is an opaque pointer to mmtk-core. mmtk-core never assumes what it is. It is merely an identifier for a thread.

src/mm/memory_manager.rs
```diff
@@ -20,46 +33,64 @@ use crate::util::heap::layout::vm_layout_constants::HEAP_START;
...
 
+/// Allocate memory for an object. For performance reasons, a VM should
+/// implement the allocation fast-path on their side rather than just calling this function.
+///
+/// Arguments:
+/// * `mutator`: The mutator to perform this allocation request.
+/// * `size`: The number of bytes required for the object.
+/// * `align`: Required alignment for the object.
+/// * `offset`: Offset associated with the alignment.
+/// * `allocator`: The allocation semantic required for the allocation.
```
javadamiri, line 93: If I remember correctly, we should use `semantics` instead of `semantic` in most cases, because the `Allocator` type includes more than one semantic.
In the rest of this file, `semantics` and `semantic` are both used.

src/mm/memory_manager.rs
```diff
@@ -97,61 +149,121 @@ pub fn start_worker<VM: VMBinding>(
...
 
+/// Is the address in the mapped memory?
+///
+/// Arguments:
+/// * `address`: The address to query.
 pub fn is_mapped_address(address: Address) -> bool {
     address.is_mapped()
 }
 
+/// Check that if a garbage collection is in progress and if the given
```
javadamiri, line 250: It's not clear what this function does.
Does it exit if `(a collection is in progress) && (object is movable)`?

If so, maybe it should be rewritten as:
Exit and log error messages if a garbage collection is in progress and the object is movable.

src/mm/memory_manager.rs
qinsoon: Yeah.

qinsoon: But we won't have multiple semantics for a while.

src/mm/memory_manager.rs
qinsoon: I am not clear about this function either. It is ported from JikesRVM.

src/mm/memory_manager.rs
javadamiri: If it is an identifier, it should be constant and unique, which is not clear here.

qinsoon: I dont quite get why it is constant. It uniquely identifies an object, which makes it an identifier. A thread pointer can also identify a thread.

qinsoon: Though `tls` is confusing, it is used everywhere in the code. I don't suggest change that, unless you plan to change every occurrences of it from the code base - which I believe we dont have time to at the moment.

javadamiri: There should be something constant about an identifier (either the pointer address, or the content the pointer points to).
If your name/staff-id changes, it doesn't help identifying you

src/plan/trace.rs
```diff
@@ -1,6 +1,9 @@
 use crate::util::queue::SharedQueue;
 use crate::util::{Address, ObjectReference};
 
+/// This trait describes the core functionality for a transitive
+/// closure over the heap. Any class implementign this trait holds the global state, TraceLocal
```
javadamiri, line 5: I don't understand the last sentence here.
Does the "," need to be changed to "."?
Probably rewriting it as:

Any class implementing this trait holds the global state (per-thread state is handled by TraceLocal and its implementers).

src/policy/mod.rs
```diff
@@ -1,3 +1,16 @@
+//! Memory policies that can be used for spaces.
+
+/// This class defines and manages spaces.  Each policy is an instance
```
javadamiri, line 3: Each policy is ...

to:

Each space is an instance of a policy?

src/plan/trace.rs
qinsoon: Yeah. That should be a period `.`

src/policy/mod.rs
qinsoon: Correct. That was copied from Java MMTk, and it is incorrect.

src/vm/scanning.rs
```diff
@@ -4,33 +4,67 @@ use crate::util::ObjectReference;
...
     fn scan_object<T: TransitiveClosure>(
         trace: &mut T,
         object: ObjectReference,
         tls: OpaquePointer,
     );
+
+    #[deprecated]
     fn reset_thread_counter();
+
+    /// MMTk callls this method at the first time during a collection that thread's stacks
```
javadamiri, line 34: This is not clear to me.
Does it mean?:
MMTk calls this method when the thread's stacks scanning is completed for the first time during a collection.

(completion may happen multiple times during a collection and this function is called on the first occurrence?)

javadamiri (approved):
LGTM, only minor comments

src/vm/scanning.rs
qinsoon: Yeah.

PR #168, opened 2020-11-06
fix bug in calculating the number of chunks in grow_space

javadamiri:
This PR fixes a bug in how the total number of chunks was calculated in the grow_space function.
E.g. the following function call:
```
Space::grow_space("nogc_space", 0x200002f0000, 3440640, true)
```
was calculating the number of chunks as:
```
let chunks = conversions::bytes_to_chunks_up(bytes);
```
which was returning `1` by only considering the total growth size.

With the number of chunks being `1`, `SFTMap::update` was only setting the space (e.g. `Space::set`) for chunk index `0x80000`, which covered addresses `0x200002f0000` to `0x20000400000 - 1`, and left the rest of the grown space (e.g. `0x20000400000` to `0x20000638000`) which mapped to chunk index `0x80001` uncovered.

To solve this, this PR calculates the number of chunks based-on start/end addresses of the grown space. So in cases such as above, the correct number of chunks in calculated.

### DIFF
--- a/src/policy/copyspace.rs
+++ b/src/policy/copyspace.rs
@@ -25,6 +25,9 @@ pub struct CopySpace<VM: VMBinding> {
 }
 
 impl<VM: VMBinding> SFT for CopySpace<VM> {
+    fn name(&self) -> &str {
+        self.get_name()
+    }
     fn is_live(&self, object: ObjectReference) -> bool {
         !self.from_space() || ForwardingWord::is_forwarded::<VM>(object)
     }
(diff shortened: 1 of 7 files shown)


qinsoon (approved):
LGTM. Thanks for investigating and fixing the bug. It should be ready to merge once all the binding tests pass.

qinsoon:
I will look at this further for the JikesRVM fails when we finish running all the tests. That is not a known bug for JikesRVM. We should not merge until we figure out the issue.

qinsoon:
The PR now is now quite different from the original PR. I am closing this PR, and will restart a new one and redo code review.

Issue #59, opened 2020-03-26
Miscellaneous Code Cleanup


qinsoon:
In GitLab by @qinsoon on Feb 4, 2020, 12:30

This tracks miscellaneous code cleanups we should do before public release.
* [x] Remove `OOM_LOCK` (use CAS on `allocation_success`)
* [x] Remove compiler warnings.
* [x] Tidy up 'sync'/'unsync'.
* [x] Move to Rust 2018 Edition.
* [x] Clarify `tls` variable
* [x] Replacing the use of `*const (dyn SFT + Sync)` with `&'static dyn SFT + Sync`

Cleanup API:
* [x] `tls` in `enable_collection()`
* [x] `type_refer` in `post_alloc()`
* [x] `scan_region()`

qinsoon:
For 'clarifying `tls`', this is what I propose to do. I think this will clarify our code around `tls`. Please let me know what you think. @javadamiri @caizixian @wenyuzhao 
* Stop using the variable name `tls`, which I think is quite confusing. We should differentiate what is `worker_tls` and what is `mutator_tls`.
* Introduce two new types (e.g. `VMMutatorThread(OpaquePointer)` and `VMWorkerThread(OpaquePointer)`), so they are typed.

qinsoon:
This is a draft PR for refactoring TLS I was talking about: https://github.com/mmtk/mmtk-core/pull/306. You are welcome to provide any feedback.

PR #174, opened 2020-11-13
Fix SFT map bugs

qinsoon:
This PR fixes two issues for SFT map.
1. Previously there were cases where we overwrote entries in SFT map to a different value. 
2. Previously, for contiguous spaces, we relied on `grow_space()` to set SFT map. However, we should not 'grow space' for contiguous space, as its address range is known at creation.

This PR:
* Adds some debug logging for SFT map (guarded by `DEBUG_SFT_MAP`).
* Adds some assertions for SFT map
  * We only allow these transitions for entries in SFT map: from empty to a valid value (when we grow space to the chunks), from a valid value to empty (when clearing the chunks), or from a valid value to the same valid value (when we have initialized the map, but `ensuer_mapped()` is called)
  * `grow_space()` should only be called when the `start` is chunk aligned. Otherwise, we may overwrite entries in SFT map.
* In `MonotonePageResource.alloc_pages()`, `new_chunk` is only set to true when a space is discontiguous and we used up the current chunk, in which case, we actually acquire new chunks. 
* For contiguous spaces, we initialise SFT map at init time, as its address range is known at creation time.

### DIFF
--- a/src/policy/copyspace.rs
+++ b/src/policy/copyspace.rs
@@ -25,6 +25,9 @@ pub struct CopySpace<VM: VMBinding> {
 }
 
 impl<VM: VMBinding> SFT for CopySpace<VM> {
+    fn name(&self) -> &str {
+        self.get_name()
+    }
     fn is_live(&self, object: ObjectReference) -> bool {
         !self.from_space() || ForwardingWord::is_forwarded::<VM>(object)
     }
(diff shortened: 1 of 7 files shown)


qinsoon:
v8-binding-test
V8_BINDING_REF=semispace

wenyuzhao (approved):
LGTM.

Also, we may need to consider replacing the use of `*const (dyn SFT + Sync)` with `&'static dyn SFT + Sync` or `Box<dyn SFT + Sync>` in the future.

qinsoon:
> LGTM.
> 
> Also, we may need to consider replacing the use of `*const (dyn SFT + Sync)` with `&'static dyn SFT + Sync` or `Box<dyn SFT + Sync>` in the future.

Thanks. I added this to the cleanup list here: https://github.com/mmtk/mmtk-core/issues/59

Issue #145, opened 2020-10-15
Remove the uses of set-env in workflows


qinsoon:
It is deprecated due to https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

PR #175, opened 2020-11-15
Fix deprecated set-env

qinsoon:
This fixes https://github.com/mmtk/mmtk-core/issues/145.
* Use comment-env-vars 1.0.2 (with the updated action-core of v1.2.6)
* Replace the use of set-env to env files.


javadamiri (approved):
LGTM

PR #177, opened 2020-11-20
Use env-logger. Allow logger initialisation to fail.

### LINKED ISSUE #22: Fine grained logging
In GitLab by @caizixian on Mar 23, 2018, 23:15

### LINKED ISSUE #169: MMTk should not init logger
According to https://docs.rs/log/0.4.11/log/#in-libraries, Rust library (such as MMTk) should only use the `log` crate but should not initialize a logger. It is the lib consumer's duty to choose their desired logger implementation. 

Currently MMTk initializes a logger. This causes problems to a VM or a VM binding if they have to initialize the logger on the VM side (e.g. Zebu).

We could solve this by either of the approaches:
1. MMTk does not initialize logger. This is more Rust idiomatic. But it seems not necessary as most VMs are not implemented in Rust, and do not have to initialize the logger themselves. 
2. MMTk attempts to initialize logger, but allow fails if the VM has initialized a logger.

qinsoon:
This PR addresses https://github.com/mmtk/mmtk-core/issues/22 and https://github.com/mmtk/mmtk-core/issues/169.
* Replace the simple `MMTkLogger` implementation with `env-logger` (https://crates.io/crates/env_logger), which allows logging at different levels for different modules.
* Allow logger initialisation to fail. For a VM binding or a VM that uses their logger and has initialized the logger, previously calling `gc_init()` will cause a crash as MMTk cannot initialize MMTk's logger. Now calling `gc_init()` is fine. 
* Adds a log level check before iterating SFT map.

This This closes #22, closes #169.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ downcast-rs = "1.1.1"
 atomic-traits = "0.2.0"
 atomic = "0.4.6"
 spin = "0.5.2"
+env_logger = "0.8.2"
 
 [dev-dependencies]
 crossbeam = "0.7.3"
--- a/src/mm/memory_manager.rs
+++ b/src/mm/memory_manager.rs
@@ -43,13 +43,19 @@ pub fn start_control_collector<VM: VMBinding>(mmtk: &MMTK<VM>, tls: OpaquePointe
 }
 
 /// Initialize an MMTk instance. A VM should call this method after creating an [MMTK](../mmtk/struct.MMTK.html)
-/// instance but before using any of the methods provided in MMTk.
+/// instance but before using any of the methods provided in MMTk. This method will attempt to initialize a
+/// logger. If the VM would like to use its own logger, it should initialize the logger before calling this method.
 ///
 /// Arguments:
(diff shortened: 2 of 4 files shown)


qinsoon:
I cancelled redundant performance runs, so some checks were shown as not successful. However, they actually ran, and we have the perf results in the comments.

javadamiri (approved):
LGTM

caizixian:
https://crates.io/crates/env_logger
It seems that we shouldn't use `env_logger` in libraries/

qinsoon:
> https://crates.io/crates/env_logger
> It seems that we shouldn't use `env_logger` in libraries

Yeah. See the description here https://github.com/mmtk/mmtk-core/issues/169. Usually a library should not initialize a logger, and should let the application that uses the lib to initialize the logger. I think MMTk is a bit different, as MMTk mostly is used by non-Rust applications (VMs) and they do not have their own logger. So it is fine for MMTk to initialize the logger (i.e. use env-logger). And this PR also allows the initialization to fail if the VM has initialized a logger.

Issue #172, opened 2020-11-12
Work packet cleanup


qinsoon:
From unresolved comments in https://github.com/mmtk/mmtk-core/pull/136:
* [x] Allocator mapping for GenCopy is incorrect. 
* [x] `TransitiveClosure` and `TraceLocal` can be removed. 
* [x] We may want a general bucket map such as `Map<Stage, WorkBucket>`.

k-sareen:
https://github.com/mmtk/mmtk-core/pull/607

PR #179, opened 2020-12-06
Cache work-packets locally

wenyuzhao:
This PR adds a local work-packet buffer to each worker thread. Work packets can be added to the local buffers before flushing to the global pool, to reduce synchronization cost.

Workers will consume local work-packets first before polling from the global pool.


qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=fix-work-packets

qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=fix-work-packets

qinsoon:
openjdk-microbm
OPENJDK_BINDING_BRANCH_REF=fix-work-packets

qinsoon:
openjdk-perf-compare
OPENJDK_BINDING_BRANCH_REF=fix-work-packets

qinsoon:
jikesrvm-perf-compare
JIKESRVM_BINDING_BRANCH_REF=fix-work-packets

src/scheduler/work_bucket.rs
```diff
@@ -143,3 +144,12 @@ impl<C: Context> WorkBucket<C> {
         false
     }
 }
+
+#[derive(Debug, Enum, Copy, Clone, Eq, PartialEq)]
+pub enum WorkBucketId {
```
qinsoon, line 149: Should we call it `WorkBucketStage`?

src/scheduler/worker.rs
```diff
@@ -74,6 +95,10 @@ impl<C: Context> Worker<C> {
         self.local().init(tls);
         self.parked.store(false, Ordering::SeqCst);
         loop {
+            while let Some((bucket, mut work)) = self.local_work_buffer.pop() {
+                debug_assert!(self.scheduler.work_buckets[bucket].is_activated());
```
qinsoon, line 99: Do we allow adding work to the local buffer for buckets that are not yet activated? If we allow that, we have no way to guarantee that the buckets are activated when we pop the work, and the assertion here will fail.

src/scheduler/worker.rs
```diff
@@ -39,6 +42,24 @@ impl<C: Context> Worker<C> {
             stat: Default::default(),
             context: None,
             is_coordinator,
+            local_work_buffer: Vec::with_capacity(LOCALLY_CACHED_WORKS),
+        }
+    }
+
+    #[inline]
+    pub fn add_work(&mut self, bucket: WorkBucketId, work: impl Work<C>) {
+        self.local_work_buffer.push((bucket, box work));
```
qinsoon, line 51: Do we allow adding work to any bucket (active and inactive buckets)?

qinsoon: I feel we may only want to push works to local buffer for currently activated buckets. For works in other buckets, we should not put them to the local buffer, as they take up place in local buffer, and can't be done immediately.

qinsoon (changes requested):
This PR looks good. It also did some great refactoring for https://github.com/mmtk/mmtk-core/issues/172.

My main question is that our bucket rules for the local buffer. I feel we should only allow works in the local buffer for buckets that are already activated. Please let me know if my understanding is wrong. 

Please update the V8 binding for the changes.

src/plan/gencopy/gc_works.rs
```diff
@@ -186,15 +186,15 @@ impl<VM: VMBinding> GCWork<VM> for GenCopyProcessModBuf {
         if mmtk.plan.in_nursery() {
             let mut modified_nodes = vec![];
             ::std::mem::swap(&mut modified_nodes, &mut self.modified_nodes);
-            worker.scheduler().closure_stage.add(
-                ScanObjects::<GenCopyNurseryProcessEdges<VM>>::new(modified_nodes, false),
-            );
+            worker.scheduler().work_buckets[WorkBucketId::Closure].add(ScanObjects::<
```
javadamiri, line 189: This code style [lines 189-193] looks a bit weird.

qinsoon:
Also the JikesRVM tests timed out. Can you take a look at that? I assume we should not have timeout for JikesRVM after https://github.com/mmtk/mmtk-jikesrvm/pull/43.   @wenyuzhao

src/scheduler/worker.rs
steveblackburn: Ideally we allow pushing to local buckets under all circumstances, and block consuming local buckets if they're not globally open.

Perhaps one way to do that:   have two lists of references to local buckets, one for consuming, one for producing.   The producing ones are always available (since it always safe to produce locally).   The consuming ones are empty and only get populated when the respective bucket is noticed as being open according to the existing protocols.   When it is open, consuming and producing refer to the same local bucket.

qinsoon:
> Also the JikesRVM tests timed out. Can you take a look at that? I assume we should not have timeout for JikesRVM after [mmtk/mmtk-jikesrvm#43](https://github.com/mmtk/mmtk-jikesrvm/pull/43). @wenyuzhao

The JikesRVM timeout is probably not an issue from this PR.

wenyuzhao:
v8-binding-test
V8_BINDING_REF=fix-work-packets

src/plan/gencopy/gc_works.rs
wenyuzhao: Done

src/scheduler/worker.rs
wenyuzhao: Fixed. Now it will cache work locally only when the corresponding bucket is open.

src/scheduler/work_bucket.rs
wenyuzhao: Fixed.

src/scheduler/worker.rs
wenyuzhao: Now only activated work packets can be cached locally.

@qinsoon I think this assertion holds now, given the fix above.

Since:
1. A work packet has to be activated before caching it locally.
2. Locally-cached work packets are executed with the highest priority.
3. A worker will never yield unless it drains all its local work packets.
4. A stop-the-world bucket will never close until the end of the GC.
5. A bucket with open only if (bucket of the previous stage is empty && all workers are yielded)

So ideally,
1. Based on [1,4], when a local-work is being executed, its bucket is already open and will never close (until the end of GC)
2. Based on [2,3,5], a worker can never yield before it drains all the local works. Thus no more buckets will open.

src/scheduler/worker.rs
```diff
@@ -7,6 +7,8 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc::Sender;
 use std::sync::{Arc, Weak};
 
+const LOCALLY_CACHED_WORKS: usize = 1;
```
qinsoon, line 10: This is changed from 10 to 1. Is this intentional?

qinsoon (approved):

src/scheduler/worker.rs
wenyuzhao: Yes. My benchmark results show that setting `LOCALLY_CACHED_WORKS` > 1 won't make performance better.

qinsoon:
@wenyuzhao Can you create a PR for fix-work-packets for V8?

src/scheduler/worker.rs
qinsoon: But 1 is better than 0 (no local cache)? That's interesting...

wenyuzhao: This local cache is specifically to address the "tail recursion" problem. i.e. A work packet `W` only creates another `W`,  which results a long `W->W->W->W->...` chain. So caching one packet locally is enough to make a worker consume all items in this chain, instead of pushing each item to the global pool.

PR #182, opened 2020-12-09
Assert minimum object size

javadamiri:
As a remainder from the object model cleanup


qinsoon (approved):
LGTM. I have added some comments to the added code to explain the assertion.

PR #184, opened 2020-12-10
Side-metadata supported object remembering barrier

wenyuzhao:
Related to https://github.com/mmtk/mmtk-core/pull/183

* Metadata memory range:
    * **Metadata start**: `HEAP_END`
    * **Metadata pages**: `MAX_CHUNKS * SelectedConstraints::METADATA_PAGES_PER_CHUNK`
* Lazily initialize metadata memory for each chunk, during its SFT initialization.

[**[Performance Comparison Results]**](https://squirrel.anu.edu.au/plotty/wenyuz/rust-g1/#0|shrew-2020-12-10-Thu-214900&benchmark^build^hfac^invocation^iteration&time^time.gc^time.mu&|10&iteration^1^4&benchmark^2^luindex|20&1^invocation|30&1&benchmark&build;jdk-mmtk-barrier-hashset|70&build^jdk|41&Histogram%20(with%20CI)^build^benchmark&)

# TODO

* [x] ~Performance comparison~
* [x] Clear metadata as a whole
* [x] Finalize side-metadata design with @javadamiri


src/policy/space.rs
```diff
@@ -574,15 +574,6 @@ impl<VM: VMBinding> CommonSpace<VM> {
...
         if self.contiguous {
-            if !try_map_metadata_space(
-                self.start,
-                self.extent,
-                VM::VMActivePlan::global().global_side_metadata_per_chunk(),
-                space.local_side_metadata_per_chunk(),
-            ) {
-                // TODO(Javad): handle meta space allocation failure
-                panic!("failed to mmap meta memory");
-            }
```
wenyuzhao, line 585: This code eagerly mmaps all the metadata memory for the whole 64-bit contiguous space, which will most likely panic due to OOM.

src/util/heap/monotonepageresource.rs
```diff
@@ -157,12 +157,24 @@ impl<VM: VMBinding> PageResource<VM> for MonotonePageResource<VM> {
...
+                    if !try_map_metadata_space(
+                        chunk,
+                        BYTES_IN_CHUNK,
+                        VM::VMActivePlan::global().global_side_metadata_per_chunk(),
+                        local_side_metadata_per_chunk,
+                    ) {
+                        // TODO(Javad): handle meta space allocation failure
+                        panic!("failed to mmap meta memory");
+                    }
+                });
```
wenyuzhao, line 177: Map metadata memory only when the heap chunk is not mmaped before. See fragmented_mapper.rs

src/util/heap/layout/fragmented_mapper.rs
```diff
@@ -107,6 +107,7 @@ impl Mmapper for FragmentedMapper {
                         let mmap_start = Self::chunk_index_to_address(base, chunk);
                         let _guard = self.lock.lock().unwrap();
                         crate::util::memory::dzmmap(mmap_start, MMAP_CHUNK_BYTES).unwrap();
+                        notify_mmap(mmap_start);
```
wenyuzhao, line 110: Call the callback function only when the chunk is not mapped before.

javadamiri: I suggest adding two arguments to `Mmapper::ensure_mapped` e.g. change it to `Mmapper::ensure_mapped(start, pages, global_metadata_per_chunk: usize, local_metadata_per_chunk: usize)`, and probably do the same for `dzmmap()`.

javadamiri: This will be slightly different if we switch from per-chunk side metadata, but is conceptually the same

qinsoon:
@wenyuzhao Can you fix the check?

wenyuzhao:
> @wenyuzhao Can you fix the failed check?

Yes I'll do that asap. I think a more urgent thing is that we need to figure out a better way to replace my metadata fix in this PR, as I pointed out in the comments.

qinsoon:
> > @wenyuzhao Can you fix the failed check?
> 
> Yes I'll do that asap. I think a more urgent thing is that we need to figure out a better way to replace my metadata fix in this PR, as I pointed out in the comments.

I agree. I was asking just to avoid a deadlock issue - I am waiting for your fix for the test while you are waiting for my review.

qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=gencopy-object-remembering-barrier-fastpath

src/plan/gencopy/mod.rs
```diff
@@ -5,4 +5,18 @@ pub(super) mod global;
 pub(super) mod mutator;
 
 pub use self::global::GenCopy;
+
+pub const NO_BARRIER: bool = false;
```
qinsoon, line 9: Are those 3 constants only used for you to run our experiments? If my understanding is correct, they change the behavior of gen copy, and they should not be set to `true` for a normal user. If that's the case, I would suggest removing them from the code. It is more about the methodology you used for the experiments.

src/plan/barriers.rs
```diff
@@ -30,69 +32,100 @@ impl Barrier for NoBarrier {
...
+    _nursery: &'static S,
+    modbuf: Vec<ObjectReference>,
+    meta: SideMetadataSpec,
 }
 
-impl<E: ProcessEdgesWork, S: Space<E::VM>> FieldRememberingBarrier<E, S> {
+impl<E: ProcessEdgesWork, S: Space<E::VM>> ObjectRememberingBarrier<E, S> {
     #[allow(unused)]
-    pub fn new(mmtk: &'static MMTK<E::VM>, nursery: &'static S) -> Self {
+    pub fn new(mmtk: &'static MMTK<E::VM>, nursery: &'static S, meta: SideMetadataSpec) -> Self {
```
qinsoon, line 44: Though you pass in a constant for `SideMetadataSpec`, it probably is not treated as a constant when you use the field `meta`. Does it matter for performance?

src/plan/gencopy/gc_work.rs
```diff
@@ -209,29 +212,3 @@ impl<VM: VMBinding> DerefMut for GenCopyMatureProcessEdges<VM> {
         &mut self.base
     }
 }
-
-#[derive(Default)]
```
qinsoon, line 213: Was this code unused?

src/plan/gencopy/global.rs
```diff
@@ -172,11 +172,8 @@ impl<VM: VMBinding> Plan for GenCopy<VM> {
     }
 
     fn global_side_metadata_per_chunk(&self) -> usize {
-        if !VM::VMObjectModel::HAS_GC_BYTE {
-            meta_bytes_per_chunk(3, 1)
-        } else {
-            0
-        }
+        debug_assert!(VM::VMObjectModel::HAS_GC_BYTE);
+        meta_bytes_per_chunk(3, 0)
```
qinsoon, line 176: Can you explain this change? Previously the plan needed 2 bits global side metadata if we don't have GC bytes, and needed 0 if we have GC bytes. Now the plan needed 1 bit and assumed we always have GC bytes.  Is the 1 bit for `LOGGING_META`? What's the reason we need to assume we always have GC bytes for GenCopy?

src/scheduler/gc_work.rs
```diff
@@ -446,39 +450,36 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for ScanObjects<E> {
     }
 }
 
-#[derive(Default)]
 pub struct ProcessModBuf<E: ProcessEdgesWork> {
-    modified_nodes: Vec<ObjectReference>,
-    modified_edges: Vec<Address>,
+    modbuf: Vec<ObjectReference>,
     phantom: PhantomData<E>,
+    meta: SideMetadataSpec,
```
qinsoon, line 456: This `meta` also will not be treated as a constant.

src/scheduler/gc_work.rs
```diff
@@ -446,39 +450,36 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for ScanObjects<E> {
...
-            ::std::mem::swap(&mut modified_edges, &mut self.modified_edges);
-            worker.scheduler().work_buckets[WorkBucketStage::Closure].add(E::new(
-                modified_edges,
-                true,
-                mmtk,
-            ));
+            if !self.modbuf.is_empty() {
+                let mut modbuf = vec![];
+                ::std::mem::swap(&mut modbuf, &mut self.modbuf);
+                GCWork::do_work(&mut ScanObjects::<E>::new(modbuf, false), worker, mmtk)
```
qinsoon, line 481: This will do the work with the current thread and the current work packet. What is the reason for this change? Does  this affect the correctness of work packet statistics? Basically the `ScanObject` work is counted in a `ProcessModBuf` work.

src/util/heap/layout/byte_map_mmapper.rs
```diff
@@ -206,7 +206,7 @@ mod tests {
     use crate::util::heap::layout::vm_layout_constants::MMAP_CHUNK_BYTES;
     use std::sync::atomic::Ordering;
 
-    const MEGABYTE: usize = 1 << 20;
+    const MEGABYTE: usize = 1 << 22;
```
qinsoon, line 209: This does not make sense? You can rename it to `CHUNK_SIZE`.

src/util/heap/freelistpageresource.rs
```diff
@@ -114,12 +114,11 @@ impl<VM: VMBinding> PageResource<VM> for FreeListPageResource<VM> {
...
-        self.common()
-            .space
-            .unwrap()
-            .common()
-            .mmapper
-            .ensure_mapped(rtn, required_pages);
+        self.common().space.unwrap().common().mmapper.ensure_mapped(
+            rtn,
+            required_pages,
+            &|_| unreachable!(),
```
qinsoon, line 120: Why is the callback not invoked for `FreelistPageResource`?

src/policy/space.rs
qinsoon: In the `init()` for `LockFreeImmortalSpace`, we also have code like this.

src/util/heap/layout/vm_layout_constants.rs
```diff
@@ -87,7 +87,7 @@ pub const AVAILABLE_END: Address = HEAP_END;
 pub const AVAILABLE_BYTES: usize = AVAILABLE_END.get_extent(AVAILABLE_START);
 
 /** Granularity at which we map and unmap virtual address space in the heap */
-pub const LOG_MMAP_CHUNK_BYTES: usize = 20;
+pub const LOG_MMAP_CHUNK_BYTES: usize = LOG_BYTES_IN_CHUNK;
```
qinsoon, line 90: What is the reason for changing this?

src/util/heap/layout/fragmented_mapper.rs
qinsoon: Why is this callback only used in `FragmentedMapper` (64-bit only)? You removed the side metadata mmap code from `Space.init()` which is used for both 32bit and 64bit. Does this mean we are not mmapping data for contiguous spaces in 32 bits?

src/util/heap/layout/fragmented_mapper.rs
```diff
@@ -352,11 +353,12 @@ mod tests {
                     i ^ ((i + 1) << 1)
                 );
 
-                let c = b + ((i + 2) << (LOG_MMAP_SLAB_BYTES + LOG_SLAB_TABLE_SIZE * 2 + 2));
-                assert_eq!(
-                    FragmentedMapper::hash(Address::from_usize(c)),
-                    i ^ ((i + 1) << 1) ^ ((i + 2) << 2)
-                );
+                // TODO(wenyuzhao): Looks like hashing is incorrect with a different mmap chunk size?
```
qinsoon, line 356: This is related to my question: why we need to change the mmap chunk size. But generally, it is not good to just comment out failed assertions. It usually reveals some problems if we have failed assertions - it could be the test itself that was right.

src/util/heap/layout/vm_layout_constants.rs
wenyuzhao: `LOG_MMAP_CHUNK_BYTES = 20` means you have to do 4 mmap calls to map a 4mb chunk. And Mmappers (e.g. FragmentedMapper) have a table to track the mapping state for each MMAP_CHUNK. So it can map the 1mb MMAP_CHUNK only if it is not mapped before.

By setting the size of a MMAP_CHUNK equals to a 4MB CHUNK, we can directly use the mapping state table to check whether we need to map metadata memory as well.

src/util/heap/freelistpageresource.rs
wenyuzhao: This is just my temporary metadata quick fix. I only made it work with 64bit contiguous space, so this code will never be executed when running gencopy with openjdk. Anyway, we need to fix this after figuring out a better approach for the fix.

src/util/heap/layout/vm_layout_constants.rs
qinsoon: @steveblackburn Is there any reason that a mmap chunk is 1MB and a chunk is 4MB? Is it okay if we change the mmap chunk size to 4MB as well?

src/plan/gencopy/global.rs
wenyuzhao: If openjdk turns off gc-byte support, then the gc byte has to be moved to side-metadata, which means the barrier also needs to be updated to adapt to the new layout... So I just add an assertion here to check this can never happen.

src/util/heap/freelistpageresource.rs
qinsoon: Alright. You requested reviews so I assumed it was ready for reviewing.

src/plan/gencopy/global.rs
qinsoon: If it is for openjdk, you probably need to keep the assertions at the openjdk side.

src/plan/gencopy/mod.rs
wenyuzhao: I suggest we keep these flags. These flags provide a way to run GenCopy without any barriers, or with a completely different barrier. With these flags, one can directly clone mmtk-core and openjdk binding, and start doing barrier overhead measurements without further changes to the GenCopy GC, which is quite convenient.  We also need this feature later for barrier fast path debugging.

src/util/heap/layout/fragmented_mapper.rs
wenyuzhao: One design problem of this fix is, `Space` is the caller of this mmapper code. When a space is notified by a mmaper (using `notify_mmap`), it needs to map both space-specific and global metadata. But spaces shouldn't access/manage global metadata by design.

wenyuzhao: Anyway, we still need to refactor this mmap code at some point in the future.  When a mmtk heap chunk is mapped by some third-party other than mmtk, we should not do metadata memory mapping.

src/plan/barriers.rs
wenyuzhao: I guess it is possible to treat them as constants if we pass them as a set of associated constants wrapped in a generic type parameter. I'll fix this (and the other one).

src/plan/gencopy/gc_work.rs
wenyuzhao: The code is unused at the very beginning (when GenCopy is merged for the first time). GenCopy uses `scheduler::gc_work::ProcessModBuf` instead.

src/plan/gencopy/mod.rs
qinsoon: Okay. If we have to keep them, I suggest adding some more comments:
1. For where the constants are defined, explain how to config those constants to measure overhead for different components
2. For where those constants are used, be very clear that the constant is only used for special methodology for measuring overhead of different components. The reason why I suggest removing them is those conditional checks are quite disruptive when I tried to read the code, as some essential code will be skipped based on the constant - which is confusing.

src/scheduler/gc_work.rs
wenyuzhao: This work-packet only produces one (instead of `N`) ScanObjects packets. So I think it makes no sense to push it to the local work-packet cache and pull it pack before executing it.

Yes it will affect the statistics. Thanks for pointing this out. I'll fix this.

src/plan/barriers.rs
qinsoon: I was just pointing out this could potentially be an issue. If in your measurement, this is not an issue then we dont need to do anything for it.

src/plan/barriers.rs
```diff
@@ -30,69 +32,100 @@ impl Barrier for NoBarrier {
...
-    modified_nodes: Vec<ObjectReference>,
-    modified_edges: Vec<Address>,
-}
-
-pub struct FieldRememberingBarrier<E: ProcessEdgesWork, S: Space<E::VM>> {
+pub struct ObjectRememberingBarrier<E: ProcessEdgesWork, S: Space<E::VM>> {
     mmtk: &'static MMTK<E::VM>,
-    nursery: &'static S,
-    mod_buffer: ModBuffer,
+    _nursery: &'static S,
```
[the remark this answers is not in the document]
wenyuzhao: `nursery` is not used anymore. I've removed it.

src/plan/barriers.rs
```diff
@@ -30,69 +32,100 @@ impl Barrier for NoBarrier {
...
+
+/// Note: Please also disable vm-binding's barrier fast-path.
+pub const ENABLE_BARRIER_COUNTER: bool = false;
+
+pub static BARRIER_COUNTER: BarrierCounter = BarrierCounter {
+    total: AtomicUsize::new(0),
+    slow: AtomicUsize::new(0),
+};
+
+pub struct BarrierCounter {
```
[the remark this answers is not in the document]
wenyuzhao: Done. I removed all the barrier counter code.

src/plan/barriers.rs
```diff
@@ -30,69 +32,100 @@ impl Barrier for NoBarrier {
...
             }
+            _ => unreachable!(),
+        }
+    }
+}
+
+/// Note: Please also disable vm-binding's barrier fast-path.
+pub const ENABLE_BARRIER_COUNTER: bool = false;
+
+pub static BARRIER_COUNTER: BarrierCounter = BarrierCounter {
```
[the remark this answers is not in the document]
wenyuzhao: Removed

src/scheduler/gc_work.rs
wenyuzhao: I double checked the code. By calling `GCWork::do_work(&mut ScanObjects...)` directly, the `ScanObjects` packet is not accounted when collecting statistics.

src/util/heap/layout/byte_map_mmapper.rs
wenyuzhao: Done

src/policy/space.rs
wenyuzhao: I'm not quite sure how to deal with `LockFreeImmortalSpace`. It maps all the heap memory ahead of time. But mapping metadata memory ahead of time as well will most likely fail. Probably we shouldn't have any metadata when using  `LockFreeImmortalSpace`.

src/util/heap/layout/vm_layout_constants.rs
steveblackburn: I don't know the history behind that particular decision.

However, the danger is that this will cause us to significantly overstate our physical memory use.

src/util/heap/layout/fragmented_mapper.rs
wenyuzhao: > I suggest adding two arguments to `Mmapper::ensure_mapped` e.g. change it to `Mmapper::ensure_mapped(start, pages, global_metadata_per_chunk: usize, local_metadata_per_chunk: usize)`, and probably do the same for `dzmmap()`.

Fixed

src/util/heap/freelistpageresource.rs
wenyuzhao: Fixed. Now support FreelistPageResource as well.

src/policy/space.rs
wenyuzhao: We don't need to do anything to `LockFreeImmortalSpace`. It maps metadata only for the heap that the VM actually uses.

src/plan/gencopy/mod.rs
wenyuzhao: Done.

src/plan/gencopy/global.rs
wenyuzhao: Done. The check is now moved to openjdk binding

src/util/heap/layout/fragmented_mapper.rs
wenyuzhao: This test is fixed.

qinsoon (approved):

qinsoon:
Is this PR ready to merge? @wenyuzhao

wenyuzhao:
> Is this PR ready to merge? @wenyuzhao

Yes it's ready to merge, after the mmtk-openjdk PR is approved.

PR #189, opened 2020-12-15
Gencopy allocator fix

qinsoon:
This PR fixes a few issues with gencopy. 
* Fixes the wrong allocator config for gencopy (as in https://github.com/mmtk/mmtk-core/issues/172)
   * We don't need allocators for mature spaces if we do not have pre-maturing.
   * Adds the allocator for large object space.
   * This should not affect OpenJDK, as it only uses the default allocator.
* We just need to reset nursery allocator rather than rebinding it to the same nursery space.


wenyuzhao (approved):
LGTM

Issue #40, opened 2020-03-26
SharedQueue correctness bugs


qinsoon:
In GitLab by @caizixian on Jul 5, 2019, 19:10

Reported by @wenyuzhao 

A queue should be initialized as locally done https://gitlab.anu.edu.au/mmtk/mmtk/blob/9224c23f73b4713fd972395d7a7d0e1f4f3a1987/src/util/queue/shared_queue.rs#L66

Should continue the outer loop https://gitlab.anu.edu.au/mmtk/mmtk/blob/9224c23f73b4713fd972395d7a7d0e1f4f3a1987/src/util/queue/shared_queue.rs#L47

PR #190, opened 2020-12-15
Remove Trace and queue implementation

qinsoon:
Our queue implementation (`SharedQueue` and `LocalQueue`) is no longer needed. Neither does `Trace`. This PR removes those types from the code base.

This PR closes https://github.com/mmtk/mmtk-core/issues/40.


wenyuzhao (approved):
LGTM. Probably I'll have to bring these work-stealing code back in the future...

qinsoon:
> LGTM. Probably I'll have to bring these work-stealing code back in the future...

Yeah. I think so. But I assume you would use other queue implementation, right? This implementation is not work stealing. However, if you plan to use this implementation, I feel there is no point to remove them now.

wenyuzhao:
> > LGTM. Probably I'll have to bring these work-stealing code back in the future...
> 
> Yeah. I think so. But I assume you would use other queue implementation, right? This implementation is not work stealing. However, if you plan to use this implementation, I feel there is no point to remove them now.

OK. Please remove them if this is not work stealing :-)

PR #191, opened 2020-12-16
Mutex-guarded VM Map

qinsoon:


qinsoon:
We generally would prefer finer grained lock. It is undesirable to put a lock on the entire VM map. I am closing this PR.

Issue #120, opened 2020-07-25
Audit busy waiting loop conditions


caizixian:
https://github.com/mmtk/mmtk-core/blob/521b9e40e47c0570e3d11ed0f6ed522b7ebd7cc0/src/util/heap/layout/map32.rs#L135

Loops like this might NOT terminated. There was a bug, fixed in 1c07109 , that arose from the compiler caching the loop condition, because it is correct to do so for immutable values. It can result in some very hard to debug bugs.

The fact that we need to suppress linting warnings with `#[allow(clippy::while_immutable_condition)]` is a bad sign.

We will need to go through places like this and either fix them or ensure them are correct.

PR #192, opened 2020-12-16
Audit VM map unsafe code

qinsoon:
This PR clarifies unsafe code in `VMMap` (mainly `Map32`).
* adds `mut_self_with_sync()` to `Map32`, which acquires a lock before getting a mutable reference from the immutable `self` reference.
* clarifies the cases of using unsafe mut ref cast due to 1. single threads, 2. access mutually exclusive data, 3. with lock.
* removes `#[allow(clippy::while_immutable_condition)]` for `free_all_chunks()`. This closes https://github.com/mmtk/mmtk-core/issues/120.


wenyuzhao (approved):
LGTM

Issue #71, opened 2020-04-04
Discontiguous SpaceDescriptor is initialized to zero


wenyuzhao:
`SpaceDescriptor(0)` is used to represent uninitialized descriptor value. i.e. `SpaceDescriptor(0) == SpaceDescriptor::UNINITIALIZD`.

The creation of discontiguous space-descriptors is simply done by [bump incrementing a counter & return the old value](https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/space_descriptor.rs#L87).

**But the counter's [initial value](https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/space_descriptor.rs#L29) is _zero_**.

The consequence is, the first created discontiguous space gets the zero descriptor -- an incorrect value. Objects in this space is considered as not in mmtk heap, since it's region's corresponding space-descriptor is `SpaceDescriptor::UNINITIALIZED` ([code](https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/layout/map32.rs#L251)).

The reason why the code currently does not crash is, we never test if an object is mapped in mmtk heap or not...

**Potential fix:**

[space_descriptor.rs line 29](https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/space_descriptor.rs#L29)

```diff
- static DISCONTIGUOUS_SPACE_INDEX: AtomicUsize = AtomicUsize::new(0);
+ static DISCONTIGUOUS_SPACE_INDEX: AtomicUsize = AtomicUsize::new(DISCONTIG_INDEX_INCREMENT);
```

Also, we may need to add this to the unit tests after this bug is fixed.

PR #193, opened 2020-12-17
Cleanup space descriptor, remove force_32bit_heap_layout

qinsoon:
This PR
* Adds unit tests for space descriptor.
* Fixes https://github.com/mmtk/mmtk-core/issues/71 so that the first descriptor for discontiguous space will not be considered as empty.
* Removes the `force_32bit_heap_layout` feature. Whether we use 32bit heap or 64bit depends on the compilation target.

### DIFF
--- a/src/util/heap/space_descriptor.rs
+++ b/src/util/heap/space_descriptor.rs
@@ -1,6 +1,5 @@
 use super::vmrequest::HEAP_LAYOUT_64BIT;
 use crate::util::constants::*;
-#[cfg(target_pointer_width = "64")]
 use crate::util::heap::layout::heap_parameters;
 use crate::util::heap::layout::vm_layout_constants;
 use crate::util::Address;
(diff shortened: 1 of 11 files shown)


qinsoon:
v8-binding-test
V8_BINDING_REF=update-mmtk-core-193

src/util/heap/space_descriptor.rs
```diff
@@ -108,7 +99,7 @@ impl SpaceDescriptor {
         unsafe { Address::from_usize(self.get_index() << heap_parameters::LOG_SPACE_SIZE_64) }
     }
 
-    #[cfg(target_pointer_width = "32")]
+    #[cfg(taget_pointer_width = "32")]
```
javadamiri, line 102: typo in `taget` -> `target`

javadamiri (approved):
LGTM,

The typo in the cfg statement should be fixed as it may cause bugs

src/util/heap/space_descriptor.rs
qinsoon: Sorry for the typo. Will fix this.

PR #194, opened 2020-12-18
fixing some stress tests affected by the bug in clear_forwarding_bits

javadamiri:
This PR fixes a bug in `clear_forwarding_bits` by replacing a sequence of atomic read/write ops with a compare and exchange.


qinsoon (approved):
Looks good

Issue #42, opened 2020-03-26
Run-time flag for GC plan


qinsoon:
In GitLab by @qinsoon on Nov 29, 2019, 17:39

Currently MMTk requires specifying a GC plan at build time, and only one GC plan is built into the library. As a few VMs allow users to specify which GC to use at run-time, we should allow MMTk to be built with all its GC plans and allow specifying the plan at run time.

PR #210, opened 2021-01-13
Run-time plan selection

qinsoon:
This PR allows MMTk to choose a GC plan through command line options. It closes https://github.com/mmtk/mmtk-core/issues/42. The basic idea behind the refactoring is 1) plan now is a dynamic type, so outside a plan, we refer to plan-related types (e.g. `Plan`, or `CopyContext`) as trait objects which requires dynamic dispatch, and 2) within a plan, we refer to other types of the same plan as concrete types (e.g. `SemiSpace`, or `SSCopyContext`) to avoid overhead from dynamic dispatch. 

Changes:
* Removes build-time features for plans and spaces
* Removes re-export of `SelectedPlan` and `SelectedConstraints`
* Removes any use of associate types through `SelectedPlan::Mutator/CopyContext`. 
* Slightly refactors `Plan` so it is object safe and can be used as trait objects.
* Introduces a `plan` option. It is the same as other run-time options. However, currently the only way to set a plan is to use env vars (as the `process()` function for processing options can only be used after creating an MMTk instance, and we need to choose a plan before creating an MMTk instance - it is a problem with the implementation of command line options. We should fix it, but I suggest we do it in another PR).
* Introduced `struct PlanConstraints`. Each plan defines its own constraints as a constant, and the constraints are referred to as a constant within the plan. However, when used for plan-neutral types, constraints are not constants any more. E.g. immortal space and large object space would need to know whether a plan uses log bits, and it is not a constant any more - this would occur overhead that is unavoidable.
* Refactors `HeaderByte` so it requires a `PlanConstraints` to create and behaves different based on the constraints.
* Adds `WorkerLocalPtr`, which is basically an untyped pointer used to store pointers to `CopyContext`. In `trace_object()`, it needs to be used with a concrete type (each plan knows the concrete type of its own `CopyContext`)

Bindings:
* JikesRVM (https://github.com/mmtk/mmtk-jikesrvm/pull/49). JikesRVM does not support run-time plan. A build-time plan feature is still required.
* V8 (https://github.com/mmtk/mmtk-v8/pull/16). Run-time plan selection is possible, but we only have nogc with V8 at the moment.
* OpenJDK (https://github.com/mmtk/mmtk-openjdk/pull/42). We allow run-time plan selection if no build-time plan is provided. A build-time plan can be optionally provided (to make it easier for perf testing scripts). 

Performance:
The following were executed on a reserved machine (Core i9-9900K Coffee Lake (Family 6 Model 158), 8/16 cores, 3.6GHz) using perf CI scripts (but with 100 invocations on each benchmark). It shows less 1% overhead for JikesRVM and OpenJDK. 
OpenJDK
openjdk-ss
JikesRVM
jikesrvm-ss


qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=run-time-plan

qinsoon:
openjdk-perf-compare
OPENJDK_BINDING_BRANCH_REF=run-time-plan

qinsoon:
openjdk-microbm
OPENJDK_BINDING_BRANCH_REF=run-time-plan

qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=update-mmtk-pr-210

qinsoon:
jikesrvm-perf-compare
JIKESRVM_BINDING_BRANCH_REF=update-mmtk-pr-210

qinsoon:
v8-binding-test
V8_BINDING_REF=update-mmtk-pr-210

src/plan/plan_constraints.rs
```diff
@@ -1,20 +1,40 @@
...
-pub const NEEDS_CONCURRENT_WORKERS: bool = false;
-
-pub const GENERATE_GC_TRACE: bool = false;
-
-pub const MAX_NON_LOS_COPY_BYTES: usize = MAX_INT;
+/// This struct defines plan-specific constraints.
+/// Most of the constraints are constants. Each plan should declare a constant of this struct,
+/// and use the constant wherever possible. However, for plan-neutral implementations,
+/// these constraints are not constant.
+pub struct PlanConstraints {
```
wenyuzhao, line 7: Is it possible we have another `PlanConstraintsTrait` like the following

```rust
trait PlanConstraintsTrait {
    const MOVES_OBJECTS: bool;
    const GC_HEADER_BITS: usize;
    const GC_HEADER_WORDS: usize;
    // ... other constants
    fn constraints() -> PlanConstraints {
        PlanConstraints {
            moves_objects: Self::MOVE_OBJECTS,
            gc_header_bits: Self::GC_HEADER_BITS,
            gc_header_words: Self::GC_HEADER_WORDS,
            // ... other constants
        }
    }
}
```

And then say, for a GenCopy plan:

```rust
impl PlanConstraintsTrait for GenCopy {
    const MOVES_OBJECTS: bool = false;
    const GC_HEADER_BITS: usize = 0;
    const GC_HEADER_WORDS: usize = 0;
    // ... other constants
}

impl Plan for GenCopy {
    fn constraints(&self) -> PlanConstraints {
        <Self as PlanConstraintsTrait>::constraints()
    }
}
```

Although we're now having two interfaces for accessing these constants, as long as some code can known the type of a plan, it can guarantee to inline these constants if they are using `const` defined constants.

Probably `fn constraints` is also possible to be inlined properly. So this may not be a problem...

src/scheduler/worker.rs
```diff
@@ -54,9 +74,15 @@ impl<C: Context> Worker<C> {
         &self.scheduler
     }
 
+    /// # Safety
+    /// The user needs to guarantee that the type supplied here is the same type used to create this pointer.
     #[inline]
-    pub fn local(&mut self) -> &mut C::WorkerLocal {
-        self.local.as_mut().unwrap()
+    pub unsafe fn local<W: WorkerLocal>(&mut self) -> &mut W {
+        self.local.as_type::<W>()
```
wenyuzhao, line 81: Probably we can remove this unsafe, and
* For debug build, use dynamic downcast (`std::any::Any`), to ensure the code is correct
* For release build, simply use this force downcast in a unsafe block.

src/scheduler/gc_works.rs
```diff
@@ -131,12 +145,18 @@ impl<VM: VMBinding> GCWork<VM> for ReleaseMutator<VM> {
...
+        ReleaseCollector(PhantomData)
+    }
+}
 
-impl<VM: VMBinding> GCWork<VM> for ReleaseCollector {
+impl<VM: VMBinding, W: CopyContext + WorkerLocal> GCWork<VM> for ReleaseCollector<W> {
     fn do_work(&mut self, worker: &mut GCWorker<VM>, _mmtk: &'static MMTK<VM>) {
         trace!("Release Collector");
-        worker.local().release();
+        unsafe { worker.local::<W>() }.release();
```
wenyuzhao, line 159: We can simply use dynamic dispatch here and remove the `W: CopyContext + WorkerLocal` type argument in `ReleaseCollector`, `Release`, `PrepareCollector` and `Prepare`. The performance is probably not a problem. We only do the dynamic call twice for each worker, in a single gc.

wenyuzhao (approved):
The changes looks good to me.

Just one question I mentioned in last week's meeting, why not change `MMTK<VM>` to `MMTk<VM, Plan>` and manually specialising it for each plan. In this way we don't even need to take care of any constant-inlining, unsafe down-casting, dynamic dispatching and trait objects involved in this PR. Instead, we only need to add a switch statement in each c-api definition to select the target mmtk instance. And this can be no slower than the ffi call itself.

javadamiri (approved):
I did a quick review and it looks good to me

src/plan/plan_constraints.rs
qinsoon: This should be possible. What is the benefit of doing this?

wenyuzhao: Just to ensure the code can inline these constants if possible.
Probably it may not worth doing so... I think if the specific type of a `Plan` is known at compile time, the compiler can also inline the `fn constraints(&self)` call, and eventually inline those values defined in `const GENCOPY_CONSTRAINTS: PlanConstraints`.

qinsoon:
> Just one question I mentioned in last week's meeting, why not change `MMTK<VM>` to `MMTk<VM, Plan>` and manually specialising it for each plan. In this way we don't even need to take care of any constant-inlining, unsafe down-casting, dynamic dispatching and trait objects involved in this PR. Instead, we only need to add a switch statement in each c-api definition to select the target mmtk instance. And this can be no slower than the ffi call itself.

I thought about that approach. I couldn't figure out a way to make that work. Say we have `MMTK<VM, Plan>`. It means the binding will have a variable such as `let mmtk: MMTK<OpenJDK, SemiSpace> = ....`. We can only create this variable after knowing which plan to use, which means this variable cannot be a Rust static/global (as we don't know its type statically) or a field of a static. So `mmtk` definitely outlives the method in which it is created, and it cannot be a static. The only way I can think up to deal with this case is to store it as a static untyped pointer or `Any`, but that means before calling to mmtk-core, we need to check the plan and find out the type (basically a branching). I don't think it is better than just using plan as trait objects.

If you think there are other ways to make it work, please let me know.

src/plan/plan_constraints.rs
qinsoon: For the plan-specific code, they can always use the constant like `GENCOPY_CONSTRAINTS` - I think the compiler can do the right job. For plan-neutral code, I don't expect they can use plan constraints as a constant. For `plan.constraints()` where `plan` is a `dyn Plan`, I *assume* there is always a dynamic dispatch. It is not about the implementation of `constraints()`, it is the `dyn Plan`. 

Do you think we can improve on this with your proposal?

wenyuzhao:
> > Just one question I mentioned in last week's meeting, why not change `MMTK<VM>` to `MMTk<VM, Plan>` and manually specialising it for each plan. In this way we don't even need to take care of any constant-inlining, unsafe down-casting, dynamic dispatching and trait objects involved in this PR. Instead, we only need to add a switch statement in each c-api definition to select the target mmtk instance. And this can be no slower than the ffi call itself.
> 
> I thought about that approach. I couldn't figure out a way to make that work. Say we have `MMTK<VM, Plan>`. It means the binding will have a variable such as `let mmtk: MMTK<OpenJDK, SemiSpace> = ....`. We can only create this variable after knowing which plan to use, which means this variable cannot be a Rust static/global (as we don't know its type statically) or a field of a static. So `mmtk` definitely outlives the method in which it is created, and it cannot be a static. The only way I can think up to deal with this case is to store it as a static untyped pointer or `Any`, but that means before calling to mmtk-core, we need to check the plan and find out the type (basically a branching). I don't think it is better than just using plan as trait objects.
> 
> If you think there are other ways to make it work, please let me know. @wenyuzhao

I think it is possible to have multiple static/global instances of mmtk. As long as we wrap them with `std::sync::SyncLazy` or `lazy_static!`, then none of them will be initialized until you explicitly access one of them in `openjdk_gc_init`.

e.g.

```rust
lazy_static! {
    pub static ref MMTK_NOGC: = MMTK<OpenJDK, NoGC> = MMTK::new();
    pub static ref MMTK_SEMISPACE: = MMTK<OpenJDK, SemiSpace> = MMTK::new();
}

extern fn openjdk_gc_init(selected_gc: GCName) {
    match selected_gc {
        GCName::NoGC => MMTK_NOGC.gc_init(...),
        GCName::SemiSpace => MMTK_SEMISPACE.gc_init(...),
    }
}
```

qinsoon:
> > > Just one question I mentioned in last week's meeting, why not change `MMTK<VM>` to `MMTk<VM, Plan>` and manually specialising it for each plan. In this way we don't even need to take care of any constant-inlining, unsafe down-casting, dynamic dispatching and trait objects involved in this PR. Instead, we only need to add a switch statement in each c-api definition to select the target mmtk instance. And this can be no slower than the ffi call itself.
> > 
> > 
> > I thought about that approach. I couldn't figure out a way to make that work. Say we have `MMTK<VM, Plan>`. It means the binding will have a variable such as `let mmtk: MMTK<OpenJDK, SemiSpace> = ....`. We can only create this variable after knowing which plan to use, which means this variable cannot be a Rust static/global (as we don't know its type statically) or a field of a static. So `mmtk` definitely outlives the method in which it is created, and it cannot be a static. The only way I can think up to deal with this case is to store it as a static untyped pointer or `Any`, but that means before calling to mmtk-core, we need to check the plan and find out the type (basically a branching). I don't think it is better than just using plan as trait objects.
> > If you think there are other ways to make it work, please let me know. @wenyuzhao
> 
> I think it is possible to have multiple static/global instances of mmtk. As long as we wrap them with `std::sync::SyncLazy` or `lazy_static!`, then none of them will be initialized until you explicitly access one of them in `openjdk_gc_init`.
> 
> e.g.
> 
> ```rust
> lazy_static! {
>     pub static ref MMTK_NOGC: = MMTK<OpenJDK, NoGC> = MMTK::new();
>     pub static ref MMTK_SEMISPACE: = MMTK<OpenJDK, SemiSpace> = MMTK::new();
> }
> 
> extern fn openjdk_gc_init(selected_gc: GCName) {
>     match selected_gc {
>         GCName::NoGC => MMTK_NOGC.gc_init(...),
>         GCName::SemiSpace => MMTK_SEMISPACE.gc_init(...),
>     }
> }
> ```

This theoretically could work. But performance-wise, I don't think it will be better. If you search `SINGLETON` in the openjdk binding, it is used in object scanning fast path. 

I think as long as we allow plan selection at run-time, we will have some kind of dynamism and some overhead - there is no way around it. What you proposed tries to push the dynamism out of mmtk-core. As a result, the binding will have to manage the dynamism, and the question is whether they can handle it with lower overhead.

src/scheduler/worker.rs
qinsoon: Do you mean we use `Any` instead of `WorkerLocalPtr` for the `Worker.local` for debug build? 

The reason I put `unsafe` for this is `local()` is called from plan specific code, and the caller needs to provide the same `CopyContext` as they used to initiate `GCWorker` - we cannot guarantee this in this method.

src/scheduler/gc_works.rs
qinsoon: `worker.local()` is a `void*` pointer, and we cannot do dynamic dispatch on it. We rely on the `WorkerLocal` type the plan tells us when it creates this work packet, like here:
https://github.com/mmtk/mmtk-core/blob/0fc3a168a0f97bac9659e6fe856a5bdf8b3cc294/src/plan/semispace/global.rs#L92

src/util/heap/space_descriptor.rs
```diff
@@ -23,6 +22,7 @@ const MANTISSA_BITS: usize = 14;
 const BASE_EXPONENT: usize = BITS_IN_INT - MANTISSA_BITS;
 
 /* 64-bit */
+#[cfg(target_pointer_width = "64")]
```
caizixian, line 25: Maybe explain why this (and perhaps other) constants relates to the pointer width of the target architecture

src/scheduler/worker.rs
caizixian: I agree with Yi, although it would be nice to perform a type check for debug build to save people from debugging headaches.

caizixian (approved):
LGTM. Only some minor comments.

src/scheduler/worker.rs
qinsoon: Currently `WorkerLocalPtr` does not carry type info so we cannot use it to check type. I could try to see if I can add another field only for debugging build, and use it to assert the type.

src/util/heap/space_descriptor.rs
qinsoon: Done

src/scheduler/worker.rs
qinsoon: For debug build, I added a field `ty: &'static str` for `WorkerLocalPtr` to save the ptr type from `std::any::type_name()`, and check against it in `as_type()`. This is only for debug build so performance doesnt matter.

Issue #214, opened 2021-01-20
Pages are mapped without checking the budget when collection is not enabled


wangbrenda:
https://github.com/mmtk/mmtk-core/blob/bb9fd6c/src/policy/space.rs#L237

When a space acquires pages, it only `poll`s if `allow_poll` is true. `allow_poll` can only be true if the plan is initialized, where the flag is set by https://github.com/mmtk/mmtk-core/blob/4e09cfb2078add72ed81c45779609d376a667fe2/src/mm/memory_manager.rs#L176

Considering the following scenario. If a VM port only uses `NoGC`, they might only call `gc_init` but not `enable_collection`.
So even if the heap size is 0, the acquire of pages will still succeed.

So we should either make sure `acquire` behave as expected (i.e. reporting OOM if the heap size is 0 even if collection is not enabled) or be explicit about what `enable_collection` does.

cc @caizixian

qinsoon:
Thanks for submitting the issue. 

Can we modify this line to only check `is_mutator && poll()` which will properly check heap size? If it fails (which means we want to trigger a GC), we do a check to see if mmtk is `initialized` - it panics if not initialized because we cannot do a GC without calling `enable_collection()`.
https://github.com/mmtk/mmtk-core/blob/bb9fd6c9acfd685af729ca6be9a9d3cb9b26435f/src/policy/space.rs#L246

If this works, it should be a small PR and I can do it tomorrow. As for 0 or negative heap size, we can have a check in `gc_init()` to make sure heap size is valid.

PR #215, opened 2021-01-21
Fix poll condition when enable_collection() is not called

qinsoon:
This PR fixes https://github.com/mmtk/mmtk-core/issues/214. We no longer allow allocating without checking budget if the user does not call `enable_collection()`.

### DIFF
--- a/src/policy/space.rs
+++ b/src/policy/space.rs
@@ -243,7 +243,9 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
 
         trace!("Polling ..");
 
-        if unsafe { VM::VMActivePlan::is_mutator(tls) } && VM::VMActivePlan::global().poll(false, self.as_space()) {
+        if unsafe { VM::VMActivePlan::is_mutator(tls) }
+            && VM::VMActivePlan::global().poll(false, self.as_space())
+        {
             debug!("Collection required");
             if !VM::VMActivePlan::global().is_initialized() {
                 panic!("Collection is not enabled.");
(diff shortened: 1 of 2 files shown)


javadamiri (approved):
LGTM

src/policy/space.rs
```diff
@@ -243,8 +243,13 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
 
         trace!("Polling ..");
 
-        if allow_poll && VM::VMActivePlan::global().poll(false, self.as_space()) {
```
caizixian, line 246: This is now dead code. Should be removed
```Rust
let allow_poll = unsafe { VM::VMActivePlan::is_mutator(tls) }
            && VM::VMActivePlan::global().is_initialized();
```

caizixian (changes requested):

src/policy/space.rs
qinsoon: It is not actually dead code. It is still used in the `else` block.

qinsoon: I will refactor this part to make it cleaner.

caizixian: Ah, sorry, I missed that.

qinsoon: Refactored the code a bit, and added some comments to make it clearer.

qinsoon: You can take a look and let me know if it is better now (if so, also remove the change request so we can merge this when tests pass).

caizixian (approved):

src/policy/space.rs
caizixian: Thanks Yi. It's much clearer now.

PR #217, opened 2021-01-27
Malloc/Free Implementations

paigereeves:
This PR includes code for using the implementations of malloc/free that work correctly when tested with mark-sweep, which are libc, mimalloc, jemalloc and Hoard.


qinsoon:
@paigereeves For the CI error, can you try use `https://github.com/paigereeves/hoard-sys.git` instead of `ssh://`? It should fix the build issue.

Cargo.toml
```diff
@@ -62,3 +65,9 @@ nogc_lock_free = ["nogc", "lockfreeimmortalspace"]
 nogc_no_zeroing = ["nogc_lock_free"]
 
 single_worker = []
+
+# malloc implementations
+malloc_libc = []
```
qinsoon, line 70: This can removed as it is not used.

Cargo.toml
```diff
@@ -62,3 +65,9 @@ nogc_lock_free = ["nogc", "lockfreeimmortalspace"]
 nogc_no_zeroing = ["nogc_lock_free"]
 
 single_worker = []
+
+# malloc implementations
```
qinsoon, line 69: Probably explain only one of the malloc features can be selected. Otherwise we use default malloc from libc. And be clear that this does not replace the global allocator for Rust, but only provides those libraries for GC implementation.

src/util/malloc.rs
```diff
@@ -0,0 +1,19 @@
+#[cfg(feature = "malloc_jemalloc")]
+pub use jemalloc_sys::{calloc, free, malloc_usable_size};
+
+#[cfg(feature = "malloc_mimalloc")]
+pub use mimalloc_sys::{
+    mi_calloc as calloc, mi_free as free, mi_malloc_usable_size as malloc_usable_size,
+};
+
+// Hoard links dynamically; you must define LD_LIBRARY_PATH
```
qinsoon, line 9: Is there any reason we have to deal with hoard-sys differently? Do the other two sys crates link to their native libraries statically? Can we link hoard statically?

qinsoon (approved):

qinsoon:
The pre-code-review check has a timeout for 15mins. This PR probably makes the build longer, and now it times out. I will push a change to mmtk-core master to allow a longer timeout for that check.

src/util/malloc.rs
paigereeves: Thanks for your feedback. I wasn't able to figure out how to link Hoard statically myself, so I asked Wenyu for help today and now hoard-sys works in the same way as the other malloc implementations.

PR #220, opened 2021-02-01
Fix DummyVM mmtk.h interface to match recent changes

angussidney:
[This commit](https://github.com/mmtk/mmtk-core/commit/f9ebe772b12ea4654dfbe31605b3fa8fc5b6893b#diff-25f32b23421651ab5171b50fda1c441c4ed86f74db168a3cdc656234de99e2c4) made some changes to the MMTk API provided by the DummyVM binding. However, the corresponding C/C++ header file has not been updated to match. This PR fixes this discrepancy. 

It also fixes a small syntactical typo.


qinsoon (approved):
Thanks for the PR. Looks good.

PR #222, opened 2021-02-02
Porting guide

angussidney:
This guide was produced through a collaborative effort between myself, Brenda and Steve.

Comments welcome!


docs/portingguide/Porting_Guide.md
```diff
@@ -0,0 +1,80 @@
...
+    - It may be easier to simply build a static and/or dynamic binary for MMTk and link it to the language directly, manually building new binaries as necessary. 
+        1. `cd binding_repo/mmtk`
+        2. `cargo build --features nogc` to build in debug mode or add `--release` for release mode
+        3. Copy the shared or static library from `target/debug` or `target/release` to your desired location
+    - Later, you can edit the language build process to build MMTk at the same time automatically.
+3. Replace VM allocation with calloc
+    - Change all alloc calls in the GC to calloc (https://www.tutorialspoint.com/c_standard_library/c_function_calloc.htm). Note: calloc is used instead of malloc as it zero-initialises memory.
+    - The purpose of this step is simply to help you find all allocation calls.
+4. Single Threaded MMTk Allocation
+    1. Include the [`mmtk.h` header file](https://github.com/mmtk/mmtk-core/blob/master/vmbindings/dummyvm/api/mmtk.h)
```
qinsoon, line 63: The binding implementer should have their own `mmtk.h` based on API functions exposed to native. `mmtk.h` provided in dummyvm is just an example.

docs/portingguide/Porting_Guide.md
```diff
@@ -0,0 +1,80 @@
...
+    2. Initialise MMTk by calling `gc_init`, with the size of the heap. Note that the heap size limit is currently disregarded when using NoGC, but this will be important when you implement more advanced GCs. In the future, you may wish to make this value configurable by a command line argument or environment variable.
```
qinsoon, line 64: The heap size is no longer disregarded for NoGC. The issue has been fixed.

qinsoon (commented):
I would suggest we have another section after the overview about requirements for the runtime. Ideally, we expect the runtime to meet the following requirements. If the VM does not meet these requirements, we can explain what they should do, and the consequences if they can't meet the requirements.
* Have a clear GC interface
* Allow building with external source code
* Written in a language that provides efficient FFI with native code, or written in Rust.

qinsoon (approved):

qinsoon:
@angussidney Do you plan to add a VM requirement section for this guide? If not, I am happy to merge it. I can do a PR for that later.

docs/portingguide/Porting_Guide.md
```diff
@@ -0,0 +1,80 @@
...
+
+The diagram above illustrates a port of MMTk to OpenJDK with the binding in the center.    The code colored brown is logically part of MMTk and is written in Rust.   The code colored white is logically part of OpenJDK and is written in C++.   The rightmost box is entirely free of any OpenJDK-specific code.   The leftmost box should be entirely free of any MMTk-specific code.
+
+> Note: we do currently maintain a fork of OpenJDK which includes some necessary changes to their code base, but this is not MMTk-specific and ideally this will be upstreamed.  Our port to V8 is a cleaner example, where we’ve managed to work closely with the V8 team to upstream all of the refactoring of the V8 code base that was necessary for it to support a third party heap.
+
+We structure the code into three repos.  The [MMTk core](https://github.com/mmtk/mmtk-core), the [binding repo](https://github.com/mmtk/mmtk-openjdk) containing both parts of the binding, and the OpenJDK repo, which is currently [a fork](https://github.com/mmtk/openjdk) we maintain.
+
+## How to Undertake a Port
```
steveblackburn, line 32: ## Things to Consider Before Starting a Port

In principle, a port to MMTk is not particularly difficult.   It presents itself as a standard library and the core of the API is relatively simple.   

However, porting a runtime to a different GC (*any* GC) can be difficult and time consuming.   Key questions include: How well encapsulated is the runtime's existing collector?   Does the runtime make tacit assumptions about the underlying collector's implementation?   How many places in the runtime codebase reference some part of the GC?   If the runtime has a JIT, how good is the interface between the JIT and the GC (for write barriers and allocations, for example)?   Does the runtime support precise stack scanning?   Etc.   Thinking through these questions should give you a sense for how big a task a GC port will be.

qinsoon: > It presents itself as a standard library

`mmtk-core` is a Rust crate/library. But a binding can easily expose it as a standard library. If 'it' refers to mmtk-core + mmtk-side of a binding, then it is fine.

qinsoon: @angussidney Can you add Steve's paragraph to the doc? I tried to do it myself, but it seems I can't push the changes to your fork.

angussidney: Thanks for the suggested wording, I've added that now.

PR #225, opened 2021-02-08
Select barrier based on active plan

wenyuzhao:
Each plan configures its required barrier in `PlanConstraints`, and vm-bindings will enable the selected barrier if supported.

Note: currently only OpenJDK binding support mutator barriers.

https://github.com/mmtk/mmtk-openjdk/pull/44

# TODO

- [ ] Benchmark


qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=runtime-barrier-selection

qinsoon:
openjdk-perf-compare
OPENJDK_BINDING_BRANCH_REF=runtime-barrier-selection

qinsoon:
openjdk-microbm
OPENJDK_BINDING_BRANCH_REF=runtime-barrier-selection

src/plan/plan_constraints.rs
```diff
@@ -13,6 +14,7 @@ pub struct PlanConstraints {
     pub needs_write_barrier: bool,
     pub needs_log_bit_in_header: bool,
     pub needs_log_bit_in_header_num: usize,
+    pub barrier: BarrierSelector,
```
qinsoon, line 17: Does this `barrier` refer to write barrier only? Should we merge `needs_write_barrier` with `barrier`? For example, remove `needs_write_barrier` - basically `BarrierSelector::NoBarrier` means `false` for `needs_write_barrier`.

qinsoon: Should the `barrier` selector be consistent with the `barrier` field in `Mutator`? I would suggest adding a function `fn new_barrier(selector: BarrierSelector) -> Box<dyn Barrier>` to `barriers.rs` so that `Mutator.barrier` is created from the selector, and making constructors of the barrier implementations as private so that people can only create barrier from a selector. This could ensure the selector is always consistent with the actual barrier used.

qinsoon (approved):
Looks good to me. But I am concerned that we currently have a few fields about barrier, and we should try make them consistent. I suggest merging `PlanConstraints.needs_write_barrier` with `PlanConstraints.barrier`, and create `Mutator.barrier` from `PlanConstraints.barrier`.

src/plan/plan_constraints.rs
wenyuzhao: I've removed `needs_write_barrier`.

For the second suggestion, I think it is hard to use a factory method for all Barrier struct creation. Different barrier type requires different constructor arguments. Some barriers require a specific type of nursery space while some barriers do not require any arguments (e.g. NoBarrier).

javadamiri:
> Looks good to me. But I am concerned that we currently have a few fields about barrier, and we should try make them consistent. I suggest merging `PlanConstraints.needs_write_barrier` with `PlanConstraints.barrier`, and create `Mutator.barrier` from `PlanConstraints.barrier`.

Isn't it more informative to replace `barrier` with something like `barrier_type` or `barrier_selector`?

javadamiri (approved):
LGMT

qinsoon:
We have two approvals for this PR. I think it is good to merge.
@caizixian Do you plan to review this?
@wenyuzhao Do you have anything you would like to address before merging?

qinsoon:
> > Looks good to me. But I am concerned that we currently have a few fields about barrier, and we should try make them consistent. I suggest merging `PlanConstraints.needs_write_barrier` with `PlanConstraints.barrier`, and create `Mutator.barrier` from `PlanConstraints.barrier`.
> 
> Isn't it more informative to replace `barrier` with something like `barrier_type` or `barrier_selector`?

I think `barrier` is not ambiguous in the context. However, it is up to you and Wenyu.

wenyuzhao:
> We have two approvals for this PR. I think it is good to merge.
> @caizixian Do you plan to review this?
> @wenyuzhao Do you have anything you would like to address before merging?

I don't have anything else to address

PR #227, opened 2021-02-09
Add generalized analysis trait

k-sareen:
This PR adds a generalized analysis trait that can be used to perform some analysis routines. I've also implemented two toy examples: (i) a routine that counts the number of objects; and (ii) a routine the counts the number of objects for each size class. Currently only an allocation hook is provided with future PRs adding more hooks such as for collection.

The PR adds a new option to MMTk -- `analysis_factor` -- which works the same way as `stress_factor`, but controls the frequency of executing the analysis routine. Note that as it stands, `stress_factor` and `analysis_factor` are mutually exclusive, i.e. there is no mechanism to run stress and analysis at the same time.


qinsoon:
The checks were cancelled as there are changes to the `.github` folder. CI jobs are cancelled if PRs from non-org members changes any CI checks (changes in `.github` folder). I will see if I can push a change to master to allow you add new features without modifying the script.

qinsoon:
@k-sareen Can you merge with master? With https://github.com/mmtk/mmtk-core/pull/228, I don't think you will need to modify anything in the `.github` folder and the CI can run properly.

k-sareen:
Yikes; looks like git rebase didn't happen nicely. Sorry :/

k-sareen:
Is it still broken? Could it be because my editor removed extra whitespace from `.github/scripts/ci-common.sh`?

qinsoon:
> Is it still broken? Could it be because my editor removed extra whitespace from `.github/scripts/ci-common.sh`?

Possibly.

javadamiri (approved):
Looks good to me,

I only have a general comment:

I suggest using longer type names in the mmtk projects.
E.g. change `ObjSize` to something like `PerSizeClassObjectCounter`.

I think the current naming is too general in some cases, and becomes confusing as the software size grows.

k-sareen:
While I'm at it, I've also quickly added a simple toy example for the collection hook. Would you prefer I do it in a separate PR or just add it to the current one (since I'll be refactoring sanity checker to use this interface quite soon)?

javadamiri:
> While I'm at it, I've also quickly added a simple toy example for the collection hook (it counts the number of GCs that occur -- I know we technically already have this in the `Stats` implementation, but it was to simply demonstrate how to use the interface). Would you prefer I do it in a separate PR or just add it to the current one (since I'll be refactoring sanity checker to use this interface quite soon)?

If it's simple, then probably add it to the current PR

qinsoon:
There are a few things, but I am not suggesting them to be fixed in this PR. This is more like a note for future cleanup.

1. I feel the hooks in `RtAnalysis` are not behaving like hooks. What I would expect is that after a counter is created, the framework can make sure its hooks are invoked at the right time. But for what is in the PR, those hooks need to be explicitly called for all the counters created. This could be an easy refactoring: have a list of `dyn RtAnalysis`, invoke all the `alloc_hook()` in where you invoke `obj_size.alloc_hook()`, and invoke all the `gc_hook()` in a GC packet (we possibly may need later need `pre_gc_hook()` and `post_gc_hook()`). Each hook should may take the same arguments, e.g. for `alloc_hook()`, you always get size, offset, align, etc. For `gc_hook()`, you may directly pass `&MMTK<VM>` or something. The generic type parameter on `RtAnalysis` seems unnecessary. If those general hooks do not work for a specific counter, it is always possible to call a specific function for the counter.
1. I suggest wrapping all the analysis counters into something like `AnalysisManager`, and put that to the `BasePlan` rather each single counter. 
1. It is a bit strange that the constructors for `PerSizeClassObjectCounter` and `GcCounter` are different. I assume you can assume fixed bins for `PerSizeClassObjectCounter`, and you can create all the counters at initialization. And you don't need `&stats` in the hook call.
1. It will be good if we can select which counters to use. If we have a `AnalysisManager`, we should be able to config which counter to use (set each one to on or off at build time).

qinsoon:
I started perf checks for this PR. If the checks show there is no slowdown, we can merge this PR.

k-sareen:
Hmm. Yep I see what you mean. I'll quickly refactor it for a new PR.

> It is a bit strange that the constructors for PerSizeClassObjectCounter and GcCounter are different. I assume you can assume fixed bins for PerSizeClassObjectCounter, and you can create all the counters at initialization. In that case, you don't need &stats in the hook call.

The reason why it's different is you need to have access to the stats instance to create counters as well as if we get a randomly large size that we haven't seen before, then we have to create a counter to keep track of it. At least for these reasons I thought it'll be useful to pass the stats instance.

PR #228, opened 2021-02-09
Allow PRs to add new features without modifying CI scripts

qinsoon:
We have features that are mutually exclusive, such as `malloc_jemalloc`, `malloc_mimalloc`, etc. Only one from each group should be allowed (though it is possible that none is chosen). 

However, cargo does not support mutually exclusive features, and we cannot use `--all-features` for running cargo command on all the features. Thus we have some scripts to run all features considering mutually exclusive ones. We rely on hard coded feature list in `ci-common.sh`. 

This PR removes the hard coded feature list, and instead extracts features from `Cargo.toml`.

### DIFF
--- a/.github/scripts/ci-build.sh
+++ b/.github/scripts/ci-build.sh
@@ -1,5 +1,3 @@
-set -xe
-
 . $(dirname "$0")/ci-common.sh
 
 # Execute this script under the root folder of this repo. Otherwise it will fail.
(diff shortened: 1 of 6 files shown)


javadamiri (approved):
Looks good to me,

My only concern is that our bash scripts are getting more complicated, and they are not easy to read/follow. 
Maybe we could consider replacing or complementing them with Python, to make them easier to read and maintain.

qinsoon:
> Looks good to me,
> 
> My only concern is that our bash scripts are getting more complicated, and they are not easy to read/follow.
> Maybe we could consider replacing or complementing them with Python, to make them easier to read and maintain.

Yes, I totally agree. I was thinking about implementing this with Python. But existing scripts are bash script, so I think probably I will just follow what we have. I would suggesting rewriting all CI related scripts in Python.

PR #234, opened 2021-02-11
fix bug in address calculation for side metadata

javadamiri:
This PR fixes a bug found by Paige in address calculation of the side metadata module


qinsoon (approved):
LGTM

Issue #202, opened 2021-01-04
MallocMS (a simple non-moving malloc-based GC)


qinsoon:
We will introduce a new GC plan `MallocMS`, a mark sweep collector with a malloc-based free list allocator. This should give us a usable non-moving GC.

PR #235, opened 2021-02-11
Malloc MarkSweep

paigereeves:
This is the main PR for MallocMS, addressing issue #202 

This PR includes a new plan called `MarkSweep` and an allocator called `MallocAllocator`.


src/plan/marksweep/gc_works.rs
```diff
@@ -0,0 +1,50 @@
+use crate::plan::marksweep::metadata::is_marked;
+use crate::plan::marksweep::metadata::set_mark_bit;
+use crate::scheduler::gc_works::*;
```
javadamiri, line 3: This module is renamed from `gc_works` to `gc_work` in the last commit,
You need to consider this before being able to rebase to the latest master commit

src/util/alloc/malloc_allocator.rs
```diff
@@ -0,0 +1,73 @@
...
+        unsafe {
+            let ptr = calloc(1, size);
+            let address = Address::from_mut_ptr(ptr);
+            if !meta_space_mapped(address) {
+                self.plan.poll(true, self.space.unwrap());
+                let chunk_start = conversions::chunk_align_down(address);
+                map_meta_space_for_chunk(chunk_start);
+            }
+            set_alloc_bit(address);
+            HEAP_USED.fetch_add(malloc_usable_size(ptr), Ordering::SeqCst);
```
wenyuzhao, line 51: Looks like the metadata initialization, and os memory allocation is part of `Space.acquire`'s job. But probably it is also fine to put them here as MallocAllocator can only be used together with MallocSpace.

wenyuzhao (approved):
LGTM

wenyuzhao:
For the `{` file in the project root directory, what is this? looks redundant. Can we remove this?

src/plan/marksweep/metadata.rs
```diff
@@ -0,0 +1,85 @@
...
+use crate::util::side_metadata::SideMetadataScope;
+use crate::util::side_metadata::SideMetadataSpec;
+use crate::util::Address;
+use crate::util::ObjectReference;
+use conversions::chunk_align_down;
+use std::collections::HashSet;
+use std::sync::RwLock;
+
+lazy_static! {
+    pub static ref ACTIVE_CHUNKS: RwLock<HashSet<Address>> = RwLock::default();
```
qinsoon, line 16: Is `ACTIVE_CHUNKS` still used? I thought you moved to side metadata. In this case, all code related to `ACTIVE_CHUNKS` can be removed.

src/plan/marksweep/global.rs
```diff
@@ -0,0 +1,183 @@
...
+    fn get_allocator_mapping(&self) -> &'static EnumMap<AllocationSemantics, AllocatorSelector> {
+        &*ALLOCATOR_MAPPING
+    }
+
+    fn prepare(&self, _tls: OpaquePointer) {
+        // Do nothing
+    }
+
+    fn release(&self, _tls: OpaquePointer) {
+        let mut released_chunks = HashSet::new();
```
qinsoon, line 108: This is policy-specific. You should encapsulate policy-specific code to your policy `MallocSpace`. `Plan` here should simply call all the spaces for the plan to release pages.

qinsoon (changes requested):
Thanks for the PR. The major concern for the PR is the encapsulation of policy-specific code. MMTk has a design that allows you to use different allocators/policies to construct a new GC plan. That means we need to clearly distinguish plan vs policy. However, for the current implementation, `MarkSweep` is kind of mixed with `MallocSpace`. You can think about this question - can I use a malloc allocator and a malloc space in other plans easily (like any existing allocator/space)? That could help you understand what clean up is needed.

qinsoon:
Given the timeframe now, I suggest Paige try address the comments as much as possible, then submit a PR (or edit this PR) to merge into a branch in our repo (rather than master). Then we can collaboratively working on the branch. Also we need to change our CI tests to include testing for marksweep. @steveblackburn @javadamiri

src/policy/mallocspace.rs
```diff
@@ -0,0 +1,123 @@
...
+        self.pr.common().release_reserved(PAGES_IN_CHUNK * released_chunks.len());
+
+        ACTIVE_CHUNKS
+            .write()
+            .unwrap()
+            .retain(|c| !released_chunks.contains(&c.as_usize()));
+    }
+}
+
+impl<VM: VMBinding> MallocSpace<VM> {
```
qinsoon, line 116: We should have a `trace_object()` function about how to trace object in this policy.

src/plan/marksweep/global.rs
```diff
@@ -0,0 +1,183 @@
...
+};
+
+impl<VM: VMBinding> Plan for MarkSweep<VM> {
+    type VM = VM;
+
+    fn collection_required(&self, _space_full: bool, _space: &dyn Space<Self::VM>) -> bool
+    where
+        Self: Sized,
+    {
+        unsafe { HEAP_USED.load(Ordering::SeqCst) >= HEAP_SIZE }
```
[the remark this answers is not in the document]
javadamiri: Considering the policy-specific nature of `trace_object`, shouldn't we have it as a method in the `Space<>` trait?

src/plan/marksweep/metadata.rs
javadamiri: `ACTIVE_CHUNKS` is being used to check whether the metadata space for an address has already been mapped before the current metadata content for the address is read/written.
This is not needed for the rest of MMTk.

To remove this, side metadata should provide an additional method to perform this check.
We had this method in the side metadata before, but decided to remove it because:

1 - It is not required by the rest of MMTk.
2 - it can be implemented with `mmap(MAP_FIXED_NOREPLACE)` and `munmap()`, which is very expensive and requires assumptions on how the address space is being used, or
3 - it can be implemented with a pre-allocated metadata map over the metadata itself, which is again not needed for the rest of MMTk.

javadamiri: I don't think keeping `ACTIVE_CHUNKS` is a good choice here, but I don't have a suitable fix which can be done in a reasonable time before the release.

javadamiri:
> Given the timeframe now, I suggest Paige try address the comments as much as possible ~then submit a PR (or edit this PR) to merge into a branch in our repo (rather than master)~. It seems we can modify the PR (even if it is in the fork) - thanks @javadamiri for testing this. We need to change our CI tests to include testing for marksweep. @steveblackburn @javadamiri

I have added marksweep tests to the `mmtk-openjdk` CI, but I'm not sure how to do it on `mmtk-core`

qinsoon:
> > Given the timeframe now, I suggest Paige try address the comments as much as possible ~then submit a PR (or edit this PR) to merge into a branch in our repo (rather than master)~. It seems we can modify the PR (even if it is in the fork) - thanks @javadamiri for testing this. We need to change our CI tests to include testing for marksweep. @steveblackburn @javadamiri
> 
> I have added marksweep tests to the `mmtk-openjdk` CI, but I'm not sure how to do it on `mmtk-core`

mmtk-core will use `ci-test.sh` in the bindings. As long as we include tests in the binding, it will be fine.

javadamiri:
> Thanks for the PR. The major concern for the PR is the encapsulation of policy-specific code. MMTk has a design that allows you to use different allocators/policies to construct a new GC plan. That means we need to clearly distinguish plan vs policy. However, for the current implementation, `MarkSweep` is kind of mixed with `MallocSpace`. You can think about this question - can I use a malloc allocator and a malloc space in other plans easily (like any existing allocator/space)? That could help you understand what clean up is needed.

One general concern I have is that

qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=paigereeves:master

src/util/alloc/malloc_allocator.rs
qinsoon: I have refactored the code. The allocation now is in the space.

src/plan/marksweep/global.rs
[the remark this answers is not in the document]
qinsoon: We plan to move it to SFT.

src/plan/marksweep/metadata.rs
qinsoon: We should at least move it to the policy code, as it is per space. We should not create it as a global. 

`ACTIVE_CHUNKS` is not only for check whether metadata space is mapped. It is also used in the release phase. A MallocSpace will iterate through all the active chunks, do a linear scan through its address range for each chunk, check if there is an object for the address, then decide to free (if dead) or unset the mark bit (if alive). 

I am not sure if we care about performance for this `MallocSpace`. Using `ACTIVE_CHUNKS` to release malloc'd memory can be really bad for performance. A chunk will not be removed from `ACTIVE_CHUNKS` unless the whole chunk contains no live object. That means that we may have chunks with only a few survived objects, and we have to linearly scan through the entire chunk to find those objects.

src/plan/marksweep/global.rs
qinsoon: This is now moved to the policy.

qinsoon:
I am merging this PR to the branch `marksweep`, as our CI script cannot nicely deal with PRs from a fork. I will later create a PR to merge `marksweep` to `master`.

PR #238, opened 2021-02-12
Finalization

qinsoon:
This PR adds finalization support to mmtk-core. 
* introduces two API calls, `add_finalizer()` and `get_finalized_object()`.
* adds two GC work buckets `RefClosure` and `RefForwarding`.

Related PRs:
* OpenJDK
  * Binding: https://github.com/mmtk/mmtk-openjdk/pull/48
  * OpenJDK fork: https://github.com/mmtk/openjdk/pull/5
* JikesRVM
  * Binding: https://github.com/mmtk/mmtk-jikesrvm/pull/52
  * JikesRVM fork: https://github.com/mmtk/jikesrvm/pull/5

Performance:

This PR makes `xalan` on JikesRVM about 5-6% slower. The slowdown is mostly due to the extra scanning of finalizable objects. I ran `xalan` with JikesRVM's stock SemiSpace - with `noFinalizer` on or off, there is a similar performance difference. So I assume the slowdown is due to the extra work introduced by finalization.


qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=finalization

qinsoon:
jikesrvm-perf-compare
JIKESRVM_BINDING_BRANCH_REF=finalization

qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=finalization

qinsoon:
openjdk-perf-compare
OPENJDK_BINDING_BRANCH_REF= finalization

qinsoon:
openjdk-microbm
OPENJDK_BINDING_BRANCH_REF=finalization

src/util/finalizable_processor.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+#[derive(Default)]
+pub struct ForwardFinalization<E: ProcessEdgesWork>(PhantomData<E>);
+
+impl<E: ProcessEdgesWork> GCWork<E::VM> for ForwardFinalization<E> {
+    fn do_work(&mut self, _worker: &mut GCWorker<E::VM>, mmtk: &'static MMTK<E::VM>) {
+        trace!("Forward finalization");
+        let mut finalizable_processor = mmtk.finalizable_processor.lock().unwrap();
```
wenyuzhao, line 130: If we hold this lock during the whole do_work, then this ForwardFinalization work packets will never execute in parallel if we have multiple instances of this work packet. But currently, I'm not sure whether there are any better options...

src/util/finalizable_processor.rs
```diff
@@ -0,0 +1,140 @@
...
+        // We should go through ready_for_finalize objects and keep them alive.
+        // Unlike candidates, those objects are known to be alive. This means
+        // theoratically we could do the following loop at any time in a GC (not necessarily after closure phase).
+        // But we have to iterate through candidates after closure.
+        self.candidates.append(&mut self.ready_for_finalize);
+
+        for reff in self
+            .candidates
+            .drain(start..)
+            .collect::<Vec<ObjectReference>>()
```
wenyuzhao, line 61: This code copy `candidates` to another vector, clear the old storage, iterate the new copy to construct a new `candidates` vector. I think it is possible to rewrite the loop using [`Vec::drain_filter`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter) here to reduce the copy of this vector.

wenyuzhao (approved):
LGTM

src/util/finalizable_processor.rs
wenyuzhao: Looks like this work packet (as well as `Finalization`) can only have one instance during a GC cycle. So it should be fine...

qinsoon: `ForwardFinalization` is in the `RefForwarding` bucket, and that bucket won't open until the previous bucket `RefClosure` is done.

src/util/finalizable_processor.rs
qinsoon: Thanks. I will fix this.

caizixian: `drain_filter` is not stable yet, and also I'm not sure how `drain_filter` can be efficiently used to get the a range of elements because the filter is applied on individual elements.

caizixian (changes requested):

src/util/finalizable_processor.rs
qinsoon: I quickly tested those two. It seems `drain_filter()` is better. I will do this refactoring. `drain_filter()` is not stable, but it doesn't seem to be an issue for us.

```rust
#![feature(test)]
#![feature(drain_filter)]

extern crate test;
fn main() {
    println!("Hello, world!");
}

pub fn drain(x: &mut Vec<usize>) -> Vec<usize> {
    let mut ret = vec![];
    for i in x.drain(0..).collect::<Vec<usize>>() {
        if i % 2 == 0 {
            x.push(i);
        } else {
            ret.push(i);
        }
    }

    ret
}

pub fn drain_filter(x: &mut Vec<usize>) -> Vec<usize> {
    let mut ret = vec![];
    for i in x.drain_filter(|x| *x % 2 == 0) {
        ret.push(i);
    }
    ret
}

#[cfg(test)]
mod tests {
    use super::*;
    use test::Bencher;

    fn data() -> Vec<usize> {
        (0..100000).collect()
    }

    #[bench]
    fn bench_drain(b: &mut Bencher) {
        b.iter(|| { let mut x = data(); drain(&mut x) });
    }

    #[bench]
    fn bench_drain_filter(b: &mut Bencher) {
        b.iter(|| { let mut x = data(); drain_filter(&mut x) });
    }
}
```

Result
```
yilin@rat:~/tests/drain-filter/drain-bm/src$ cargo +nightly bench
    Finished bench [optimized] target(s) in 0.00s
     Running /home/yilin/tests/drain-filter/drain-bm/target/release/deps/drain_bm-bbd2ae3ecd458477

running 2 tests
test tests::bench_drain        ... bench:     419,649 ns/iter (+/- 5,855)
test tests::bench_drain_filter ... bench:     207,699 ns/iter (+/- 2,138)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured; 0 filtered out; finished in 0.67s
```

caizixian: Right. But `drain(start..).collect::<Vec<ObjectReference>>()` gives you a vector containing the slice `[start:]`. However, `drain_filter(predicate)` gives you a vector containing all elements satisfy the predicate. I'm not sure how to translate this.

qinsoon: Umm, that's right. It would depend on what function you apply for each item. In our case, we trace objects for each item. Probably for now, I will just leave it as it is.

src/util/finalizable_processor.rs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            trace!("Pop {:?} for finalization", reff);
+            if reff.is_live() {
+                let res = FinalizableProcessor::get_forwarded_finalizable(e, reff);
```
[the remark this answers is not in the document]
qinsoon: I made the changes as suggested.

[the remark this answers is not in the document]
qinsoon: Umm, it turns out we cannot do that refactoring. `trace_object()` has side effect. We need to check `is_live()` before `trace_object()`. I tried the refactoring in https://github.com/mmtk/mmtk-core/pull/238/commits/67cf05f1c2ee6378be34e5bdeaf80e796c6173b8, and it causes quite a few tests to run out of memory.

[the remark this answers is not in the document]
caizixian: Oh, right. So if `trace_object` copies the object, and sets the mark bit, it could keep everything alive.

caizixian (approved):

qinsoon:
I am addressing this comment here: https://github.com/mmtk/jikesrvm/pull/5#discussion_r579917646. I will add `schedule_collection()` in the `VMBinding::VMCollection` trait.

PR #247, opened 2021-02-19
move try_map_metadata_space next to SFT_MAP.update

javadamiri:
This PR moves the main call to `try_map_metadata_space` deeper, next to the `SFT_MAP.update` call, potentially reducing the number of calls.


qinsoon:
For contiguous spaces, `grow_space()` is not called with `new_chunk=true`. How do we make sure that we still map metadata properly?

qinsoon (approved):
LGTM

PR #248, opened 2021-02-20
Add AnalysisManager and refactor analysis framework

k-sareen:
I've changed the framework as per the discussion here [1]. Further comments and recommendations are welcome.

~Kunal
\[1]: https://github.com/mmtk/mmtk-core/pull/227#issuecomment-780270882

### DIFF
--- a/src/plan/global.rs
+++ b/src/plan/global.rs
@@ -381,7 +381,7 @@ pub struct BasePlan<VM: VMBinding> {
     pub allocation_bytes: AtomicUsize,
     // Wrapper around analysis counters
     #[cfg(feature = "analysis")]
-    pub analysis_manager: Mutex<AnalysisManager<VM>>,
+    pub analysis_manager: AnalysisManager<VM>,
 }
 
 #[cfg(feature = "base_spaces")]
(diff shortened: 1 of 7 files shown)


src/plan/global.rs
```diff
@@ -437,7 +432,11 @@ impl<VM: VMBinding> BasePlan<VM> {
         constraints: &'static PlanConstraints,
     ) -> BasePlan<VM> {
         let stats = Stats::new();
-        let ctr = stats.new_event_counter("gc.num", true, true);
+        // Initializing the analysis manager and routines
+        #[cfg(feature = "analysis")]
+        let mut analysis_manager = Mutex::new(AnalysisManager::new());
+        #[cfg(feature = "analysis")]
+        analysis_manager.lock().unwrap().initialize_routines(&stats);
```
qinsoon, line 439: Is there any reason we need to `new()` then `initialize_routines()`? You could make `initialize_routines()` a private function, and call it in `new()`.

src/util/analysis/mod.rs
```diff
@@ -1,22 +1,86 @@
...
+
 pub mod gc_count;
 pub mod obj_num;
 pub mod obj_size;
 
+use self::gc_count::GcCounter;
+use self::obj_num::ObjectCounter;
+use self::obj_size::PerSizeClassObjectCounter;
+
 /**
```
qinsoon, line 15: Probably use `///` for rustdoc.

qinsoon (approved):
Good refactoring. I only have a few minor comments.

k-sareen:
Updated in latest commit.

PR #259, opened 2021-02-25
Marksweep

qinsoon:
This PR is mainly from https://github.com/mmtk/mmtk-core/pull/235 (@paigereeves's work) with some bug fixes and clean up. It closes https://github.com/mmtk/mmtk-core/issues/202. 

Changes:
* Introduces with a new allocator `MallocAllocator`
* Introduces a new policy `MallocSpace`. 
* Introduces a new plan `MarkSweep` that uses the above allocator and policy for the default allocation.
* Adds a method `is_in_space()` to `SFTMap`. It checks if an object reference is in our space by using SFT map. If not, it will further check if the object is allocated by malloc. The API `object.is_mapped()` now uses this method internally.
* Adds a new constant `MIN_OBJECT_SIZE` which is word size by default.
* Lowers the worst ratio for side metadata on 32 bit to avoid metadata address clashing with our heap address range.
* Allows longer time for openjdk tests.


qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=marksweep

qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=marksweep

qinsoon:
I think this PR is not yet ready. I removed the reviewers that I assigned before. Sorry for this.

qinsoon:
openjdk-perf-compare
OPENJDK_BINDING_BRANCH_REF=marksweep

qinsoon:
openjdk-microbm
OPENJDK_BINDING_BRANCH_REF=marksweep

src/mm/memory_manager.rs
```diff
@@ -240,18 +240,24 @@ pub fn is_live_object(object: ObjectReference) -> bool {
...
 /// * `object`: The object reference to query.
 pub fn is_mapped_object(object: ObjectReference) -> bool {
     object.is_mapped()
 }
 
-/// Is the address in the mapped memory?
+/// Is the address in the mapped memory? The runtime can use this function to check
+/// if an address is mapped by MMTk. Note that this is different than is_mapped_object().
+/// For malloc spaces, MMTk does not map those addresses (malloc does the mmap), so
+/// this function will return false, but is_mapped_object will may return true if the address
```
javadamiri, line 255: change: "but is_mapped_object will may return true"
to
"but is_mapped_object() will return true"

src/mm/memory_manager.rs
```diff
@@ -240,18 +240,24 @@ pub fn is_live_object(object: ObjectReference) -> bool {
...
+/// is actually a valid object in malloc spaces.
 ///
 /// Arguments:
 /// * `address`: The address to query.
+// TODO: Do we really need this function? Can a runtime always use is_mapped_object()?
```
javadamiri, line 260: From the difference between the two functions, `is_mapped_object()` and `is_mapped_address()`,
it seems we are exposing some policy-level (e.g. mallocspace) information to the client.

src/plan/marksweep/gc_work.rs
```diff
@@ -0,0 +1,56 @@
...
+        Self { plan, base }
+    }
+
+    #[inline]
+    fn trace_object(&mut self, object: ObjectReference) -> ObjectReference {
+        if object.is_null() {
+            return object;
+        }
+        trace!("Tracing object {}", object);
+        if self.plan.ms_space().in_space(object) {
```
javadamiri, line 33: Not relevant to this PR, but I think the `in_space(...)` should be renamed to `contains_object(...)` or simply `contains(...)`, and `address_in_space(...)` should be renamed to `contains_address(...)`.

src/policy/mallocspace/metadata.rs
```diff
@@ -0,0 +1,177 @@
...
+    if ASSERT_METADATA {
+        // Need to make sure we atomically access the side metadata and the map.
+        let lock = ALLOC_MAP.read().unwrap();
+        let check =
+            lock.contains(&unsafe { address.align_down(BYTES_IN_WORD).to_object_reference() });
+        let ret = load_atomic(ALLOC_METADATA_SPEC, address) == 1;
+        debug_assert_eq!(
+            check,
+            ret,
+            "is_alloced_object(): alloc bit does not match alloc map, address = {} (aligned to {}), meta_start = {}",
```
javadamiri, line 96: I don't understand the goal of `meta_start` here.
This will not work in the new metadata implementation as is.

src/plan/marksweep/gc_work.rs
qinsoon: Yeah. Good point. Maybe I will do this renaming separately.

src/policy/mallocspace/metadata.rs
qinsoon: I was just using it to know where the metadata is stored for debugging. Why would it not work for the new implementation? We still could have a way to know where the metadata is, right?

src/mm/memory_manager.rs
javadamiri: Yes, I think we should remove it

src/policy/mallocspace/metadata.rs
javadamiri: From what I see in `meta_start(addr)`'s implementation, it returns the starting address of the metadata chunk which contains the metadata for `addr`.
This is specific to the per-chunk metadata.
In the new implementation, we only have per-chunk metadata for the policy-specific side metadata of the 32-bit targets.

If you use `address_to_meta_address()` instead of `address_to_meta_chunk_addr()` in the implementation of `meta_start(addr)`, that will work in the new implementation.

qinsoon: I see. I can make that change.

src/scheduler/work.rs
```diff
@@ -18,7 +18,7 @@ pub trait Work<C: Context>: 'static + Send + Sync {
 
 /// A special kind of work that will execute on the coorddinator (i.e. controller) thread
```
javadamiri, line 19: coorddinator -> coordinator

src/util/constants.rs
```diff
@@ -101,6 +101,10 @@ pub const BITS_IN_PAGE: usize = 1 << LOG_BITS_IN_PAGE;
 /* Assume byte-addressability */
 pub const LOG_BYTES_IN_ADDRESS_SPACE: u8 = BITS_IN_ADDRESS as u8;
 
+// TODO: Should this be VM specific?
+pub const LOG_MIN_OBJECT_SIZE: u8 = LOG_BYTES_IN_WORD;
```
javadamiri, line 105: In the `SIDE_GC_BYTE_SPEC` case, the min object size mmtk can support is one word (two words is the minimum size mmtk can support in a specific combination of `HAS_GC_BYTE` and `GC_BYTE_OFFSET`).

However, allowing larger min object size for various VMs has benefits such as better side metadata space efficiency.
So, probably we should consider this as a future goal?

javadamiri (approved):
Only some minor comments,
otherwise looks good to me

src/util/constants.rs
javadamiri: Yes, but how do we use it in compile-time constants such as constant instances of `SideMetadataSpec`?

qinsoon: Yeah, I see.

wenyuzhao: Yes I think this should be VM-specific constant. To use it in compile-time constants, we need to have access to `VM` type when constructing `SideMetadataSpec`.

src/policy/space.rs
```diff
@@ -221,6 +221,19 @@ impl SFTMap {
...
     }
+
+    pub fn is_in_space(&self, object: ObjectReference) -> bool {
+        let not_in_space = object.to_address().chunk_index() >= self.sft.len()
+            || self.get(object.to_address()).name() == EMPTY_SPACE_SFT.name();
+
+        if not_in_space {
+            // special case - we do not yet have SFT entries for malloc space
+            use crate::policy::mallocspace::is_alloced_by_malloc;
+            is_alloced_by_malloc(object)
```
wenyuzhao, line 232: Probably we need a branch here to call `is_alloced_by_malloc` only when MallocMS is the selected plan. Otherwise, I'm not sure if it is safe to inspect malloc related metadata when the object is neither in mmtk heap nor in malloc space.

wenyuzhao (approved):
LGTM

src/util/constants.rs
qinsoon: `SideMetadataSpec` is a constant. I don't see a way to access `VM` when creating a constant.

src/policy/space.rs
qinsoon: I think it's fine. In `is_alloced_by_malloc()`, there is a check to see if the side metadata is mapped. If not, it won't access the side metadata.

src/mm/memory_manager.rs
qinsoon: It seems JikesRVM is using `is_mapped_address()`. We cannot simply remove it.

src/policy/mallocspace/metadata.rs
qinsoon: Removed `SideMetadataSpec.meta_start()`. Use `address_to_meta_address()` for those debug prints.

src/mm/memory_manager.rs
qinsoon: I added a comment to state that the runtime should avoid using this method, as it may get removed in the future. Otherwise than that, I don't know what I can do for it.

PR #260, opened 2021-02-27
fix a bug in forward_object

YiluoWei:
Write the forwarding word before writing the GC byte, otherwise other threads may read a wrong forwarding word (which is the header of the object, not the new address).


src/util/forwarding_word.rs
```diff
@@ -81,8 +81,8 @@ pub fn forward_object<VM: VMBinding, CC: CopyContext>(
             );
         }
         None => {
-            gc_byte::write_gc_byte::<VM>(object, FORWARDED);
             write_forwarding_word::<VM>(object, new_object.to_address().as_usize());
+            gc_byte::write_gc_byte::<VM>(object, FORWARDED);
```
wenyuzhao, line 85: I think the order doesn't matter here. When `gc_byte_offset_in_forwarding_word()` returns None, this means the gc_byte does not overlap with the forwarding_word (Correct me if I am wrong @javadamiri). Under such case writing the forwarding word later won't overwrite the gc byte.

wenyuzhao (changes requested):

src/util/forwarding_word.rs
wenyuzhao: Ok... This is a concurrency issue. Then this change makes sense.

wenyuzhao (approved):

src/util/forwarding_word.rs
javadamiri: Yes it may cause a concurrency issue between `forward_object` and `spin_and_get_forwarded_object`.

qinsoon: The method `set_forwarding_pointer()` (which is just below this method) has a similar sequence of `write_gc_byte()` and then `write_forwarding_word()`. Should that method be changed as well? That method is not used though.

javadamiri: Right, I think we should change that one as well, in case we need it in the future.
If we're not sure whether it will be used or not, we should remove it

caizixian: We need to check atomic ordering we used.
```
write_forwarding_word::<VM>(object, new_object.to_address().as_usize());
gc_byte::write_gc_byte::<VM>(object, FORWARDED); // release here, so that forwarding word is visible to other threads if they read `FORWARDED`
```

```
spin // acquire here, so that all operations below won't be reordered before spin reads FORWARDED
```

caizixian: Just checked that we have `SeqCst` almost everywhere. This will result in emitting a full fence, which is quite expensive. We should try avoiding it if possible.

caizixian: But if `SeqCst` is used properly, it is correct and we can address the issue later.

qinsoon: @YiluoWei Can you make the same change to the code in `set_forwarding_pointer()`?

YiluoWei: done

Issue #97, opened 2020-06-04
Unsafe Code Cleanup


qinsoon:
The current code base pervasively uses unsafe code, especially to get mutable references from immutable reference/objects. This could include potential race condition and bugs. We should clean up the unsafe code to maximize the benefits of thread safety we can get from Rust's checks. 

We should diagnose each of the unsafe cases, and categorize them: either 1) we need to use unsafe (mostly for performance reasons), or 2) we can refactor them into safe code. For 1), we should always encapsulate the unsafe code and make sure that they are correct.

qinsoon:
A quick summary of unsafe code in the current MMTk: https://docs.google.com/presentation/d/1X4unAoPeEPbnYoK95H2Fw54d0L139f8aYviDcoFFdqg/edit?usp=sharing

qinsoon:
From our discussion, there are a few notes about unsafe code:
* report some unsafe statistics from our CI for each PR.
* properly document unsafe functions (`# Safety` in rustdoc) and document for the uses of those unsafe functions.
* document our design around unsafe code in a separate markdown file in docs.
* for intrinsic unsafety (e.g. accessing memory), we should think about how we can defend our code. They are unsafe, but we may provide some help for debugging if we can (e.g. in debug build, assert if an address is sane before accessing it?)

qinsoon:
Related issues:
* https://github.com/mmtk/mmtk-core/issues/853
  * https://github.com/mmtk/mmtk-core/pull/953
  * https://github.com/mmtk/mmtk-core/pull/951
* https://github.com/mmtk/mmtk-core/issues/852
  * https://github.com/mmtk/mmtk-core/pull/965

PR #268, opened 2021-03-21
Send/Sync cleanup

qinsoon:
This PR cleans up our code around the `Send` and `Sync` traits. It is one step for https://github.com/mmtk/mmtk-core/issues/97.
* Remove wrong supertraits of `Send`/`Sync` for certain types. For example:
  * `Plan` is `Sync` as it is shared across thread, but it is not `Send` as we do not send its ownership to another thread.
  * `Work` is `Send` as it will be sent to worker threads, but it is not `Sync` as we should not share it across threads.
* Moving to `unsafe impl Send/Sync` to the exact type where it is needed. 
  * For example, `GenCopy` is `Sync`, but it uses `CommonPlan`, and `CommonPlan` uses `UnsafeCell` which prevents Rust deriving `Sync` for `CommonPlan` safely. We should `unsafe impl Sync` for `CommonPlan` rather than `GenCopy`.
* Remove unnecessary `unsafe impl Send/Sync`.

This PR should only be merged after we release 0.3.


wenyuzhao (approved):
LGTM

javadamiri (approved):
LGTM

PR #269, opened 2021-03-22
Move triplespace backup instructions to new chapter, underneath task

rouanebannister:
I think the current placement of the backup is a bit too hidden, especially without a reference to it in the task instructions. So this pull request is just moving it into a new chapter, and placing it just after the task instructions. I also added a short paragraph emphasising attempting the exercise first. My intent with its old positioning was just to keep it from being immediately visible under the task, and putting it in a separate chapter achieves that.

I'm not totally sure I've got the formatting right, however, since I don't know how to view this as a book to debug and I haven't worked with `mdbook` before.


qinsoon (approved):
Looks good. Thanks for the PR.

PR #272, opened 2021-03-24
Remove Option for space in allocators

qinsoon:
It is unnecessary to use `Option` for `space` in allocators. Also the `Allocator` trait now return a space reference from `get_space()` rather than an `Option`.


javadamiri (approved):
LGTM

PR #273, opened 2021-03-24
Update to comment-env-vars 1.0.3

qinsoon:
The action `comment-env-vars` extract values as Github aciton env vars from issue comments. We use this action to specify binding branches. Previosly 1.0.2 had a bug that only fetch the first 30 comments from an issue. Now it will properly fetch all the comments.


javadamiri (approved):
LGTM

Issue #279, opened 2021-03-31
Build/stylecheck for release build, stylecheck for bindings


qinsoon:
* [ ] check release build
* [ ] check style for release build
* [ ] check style for bindings

PR #286, opened 2021-04-07
Add build and stylecheck for release build in ci-build and ci-style

qinsoon:
This PR adds build and style check for the release build of mmtk-core. It also adds style check for the bindings, and will run the binding's style checks for mmtk-core. It closes https://github.com/mmtk/mmtk-core/issues/279.

### DIFF
--- a/src/policy/mallocspace/metadata.rs
+++ b/src/policy/mallocspace/metadata.rs
@@ -3,6 +3,8 @@ use crate::util::constants;
 use crate::util::constants::BYTES_IN_WORD;
 use crate::util::conversions;
 use crate::util::heap::layout::vm_layout_constants::BYTES_IN_CHUNK;
+#[cfg(debug_assertions)]
+use crate::util::side_metadata::address_to_meta_address;
 use crate::util::side_metadata::load_atomic;
 #[cfg(target_pointer_width = "32")]
 use crate::util::side_metadata::meta_bytes_per_chunk;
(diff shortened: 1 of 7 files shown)


qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=stylecheck

qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=stylecheck

qinsoon:
v8-binding-test
V8_BINDING_REF=stylecheck

src/policy/mallocspace/metadata.rs
```diff
@@ -3,7 +3,6 @@ use crate::util::constants;
 use crate::util::constants::BYTES_IN_WORD;
 use crate::util::conversions;
 use crate::util::heap::layout::vm_layout_constants::BYTES_IN_CHUNK;
-use crate::util::side_metadata::address_to_meta_address;
```
javadamiri, line 6: I don't have a strong opinion about this, but you could also add `#[cfg(debug_assertions)]` here, instead of copying the `use` statement twice.

javadamiri (approved):
Only a minor suggestion,
otherwise LGTM

src/policy/mallocspace/metadata.rs
qinsoon: I was just thinking that if we put the `use` to where it is actually used and guarded by their flag, that makes the code easier to read. I don't mind doing either though.

javadamiri: If the `use` statement was only used in one place, I'd prefer the current implementation, but with two or more copies, I'd slightly prefer guarding the `use` statement directly.

qinsoon: Sure. I have made the change.

Issue #209, opened 2021-01-06
collection_required() for GenCopy does not check for stress test GC


qinsoon:
`Plan.collection_required()` is overwritten by GenCopy, but the GenCopy implementation of `collection_required()` does not take stress test GC into account.

PR #292, opened 2021-04-12
Fix stress gc triggering in GenCopy

qinsoon:
This PR refactors `collection_required()` from `Plan` to `BasePlan`, and allows each plan implementation to reuse `BasePlan.collection_required()`. This refactoring is along the line of the idea of 'composition over inheritance'. This fixes https://github.com/mmtk/mmtk-core/issues/209 where gencopy does not check stress test GC.
* Moved `collection_required()` and `stress_test_gc_required()` to `BasePlan`.
* Updated tutorial.
* Updated `ci-doc.sh` to allow re-execute the script (easier for developers to run ci-doc.sh in their dev environment)


javadamiri (approved):
LGTM

Issue #257, opened 2021-02-24
Reserve/quarantine address range for contiguous spaces


qinsoon:
Currently we eagerly mmap the entire address range for contiguous spaces, and the address range for their side metadata (if side metadata is used). We also eagerly initialize the SFT map for the chunks. 

We should instead only reserve/quarantine the address range (for spaces and the side metadata). We mmap the pages and update the SFT map when we use the pages.

qinsoon:
Our requirements are:
1. Our mapped memory should not be replaced by others (if they are sane), and we should not replace others' mapped memory.
2. Our implementation should allow the OS to show the real memory used (RSS). If we reserve memory but do not use them, the OS should not show their usage.

For 1, we need to replace our `dzmmap()` uses with `dzmmap_noreplace()`. For 2, we are fine for linux, as Linux supports overcommitting and we do not need to separate reserve/commit operations [1]. However, for other OS/s, we may still need to do a refactoring of separating reserve vs. commit.

If we want to separate reserve vs commit, for Linux, we have two options. 1) reserve = mmap, commit = nop, or 1) reserve = mmap with PROT_NONE, commit = mprotect [2].

[1] https://www.gamasutra.com/blogs/NiklasGray/20171107/309071/Virtual_Memory_Tricks.php (keyword: overcommitting)
[2] https://mjmwired.net/kernel/Documentation/vm/overcommit-accounting#57

qinsoon:
Another option is: reserve = mmap_noreserve, commit = mmap (basically replace the previous reserved memory). Is this the way you did for side metadata? @javadamiri

javadamiri:
> Another option is: reserve = mmap_noreserve, commit = mmap (basically replace the previous reserved memory). Is this the way you did for side metadata? @javadamiri

Yes, that's how it's being done for side metadata.

javadamiri:
> Our requirements are:
> 
> 1. Our mapped memory should not be replaced by others (if they are sane), and we should not replace others' mapped memory.
> 2. Our implementation should allow the OS to show the real memory used (RSS). If we reserve memory but do not use them, the OS should not show their usage.
> 
> For 1, we need to replace our `dzmmap()` uses with `dzmmap_noreplace()`. For 2, we are fine for linux, as Linux supports overcommitting and we do not need to separate reserve/commit operations [1]. However, for other OS/s, we may still need to do a refactoring of separating reserve vs. commit.
> 
[...]

Issue #244, opened 2021-02-17
Potential bug in dzmmap


javadamiri:
The use of `MAP_FIXED` flag in the `dzmmap` functions means we overwrite any previous mapping that may overlap with our current request:

https://github.com/mmtk/mmtk-core/blob/a4110d917c758b5261eeefbc6e2b00a58233e215/src/util/memory.rs#L15

Although it is very unlikely to happen in 64-bits targets, it is still possible.
For example, we do not know how various `malloc` implementations use the address space, and we may overwrite their mappings.

An alternative is to remove the `MAP_FIXED` flag and compare the result of mmap with our requested starting address.
AFAIK, they will be the same if there is no overlaps.

qinsoon:
This is done in https://github.com/mmtk/mmtk-core/pull/296. All uses of `MAP_FIXED` in mmtk-core is replaced with `MAP_FIXED_NOREPLACE`.

PR #296, opened 2021-04-22
Mmap refactoring

qinsoon:
This PR refactors our mmap implementation and uses. It addresses https://github.com/mmtk/mmtk-core/issues/244, and it works towards https://github.com/mmtk/mmtk-core/issues/257 (though more work is needed for #257).

This PR:
* Replaces most of our uses of `dzmmap()` with `dzmmap_noreplace()`.
* Changes mmap flags in `mmap_noreserve()` from read/write to `PROT_NONE` so it serves the purpose of reserving address range. The memory is not usable before another mmap or mprotect. 
* Marks `dzmmap()` unsafe, as it may overwrite existing mappings. The only place it is now used is in side metadata where we reserve address range with `mmap_noreserve()` and then use `dzmmap()` to map it.
* Renames `check_is_mapped()` to `panic_if_unmapped()`. The function is more than just checking mapping - it may actually mmap memory as a side effect. We should simply panic if we mmap any memory in the function.
* Refactors `util::memory` so they deal with errors in the same way.
* Adds unit tests for `util::memory`.
* Refactors unit tests that uses mmap so 1. they run in serial, 2. they use a memory range that is not in our heap, and 3. they will properly clean up the mapping they create.

### DIFF
--- a/src/util/address.rs
+++ b/src/util/address.rs
@@ -204,6 +204,15 @@ impl Address {
         Address(self.0 + size)
     }
 
+    // We implemented the Sub trait but we still keep this sub function.
+    // The sub() function is const fn, and we can use it to declare Address constants.
+    // The Sub trait function cannot be const.
+    #[allow(clippy::should_implement_trait)]
+    #[inline(always)]
+    pub const fn sub(self, size: usize) -> Address {
+        Address(self.0 - size)
+    }
+
     /// loads a value of type T from the address
     /// # Safety
     /// This could throw a segment fault if the address is invalid
--- a/src/policy/lockfreeimmortalspace.rs
+++ b/src/policy/lockfreeimmortalspace.rs
@@ -86,7 +86,7 @@ impl<VM: VMBinding> Space<VM> for LockFreeImmortalSpace<VM> {
         );
         self.limit = AVAILABLE_START + total_bytes;
         // Eagerly memory map the entire heap (also zero all the memory)
-        crate::util::memory::dzmmap(AVAILABLE_START, total_bytes).unwrap();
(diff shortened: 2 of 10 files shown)


src/util/address.rs
```diff
@@ -204,6 +204,15 @@ impl Address {
         Address(self.0 + size)
     }
 
+    // We implemented the Sub trait but we still keep this sub function.
```
javadamiri, line 207: Have you tested that the non-const `Sub` trait doesn't work for Address constants?
(E.g. we can use the `Sub` trait for int constants)

qinsoon: Umm, it seems we can. Thanks. We need to include `#![feature(const_trait_impl)]` and `impl const Sub for Address {}`. I will change this.

javadamiri (approved):
LGTM

src/util/address.rs
qinsoon: I gave this a try, but `const_trait_impl` seems buggy at the moment. I can replace our definition for `Address` constants  in mmtk-core with `+` or `-`, but I can't do that for our binding crates (I have to use `const fn add/sub()` in the bindings). It could be a bug in nightly rust, and I have submitted an issue for them (https://github.com/rust-lang/rust/issues/84846). Probably for now, we have to keep both const traits and the const function `add/sub()`.

qinsoon: It seems if I declare a constant (e.g. `AVAILABLE_START = HEAP_START + VM_SPACE_SIZE`), and a binding uses `AVAILABLE_START`, the compiler also gives an error. This means we can't use this trait for now. @javadamiri

javadamiri: Right

PR #298, opened 2021-04-28
Side metadata page accounting

### LINKED ISSUE #242: Page accounting for side metadata
We should consider space used for side metadata. We can calculate their usage by multiplying a ratio on the space used for the data.

qinsoon:
This PR refactors side metadata and add page accounting for side metadata. This closes #242.

This PR
* refactors the existing page accounting code in `CommonPageResource` into a separate `PageAccounting` type.
* adds a struct `SideMetadata`, which includes `SideMetadataContext` (all the global and local specs for a policy) and `PageAccounting`. Each space has a field for `SideMetadata` in `CommonSpace` (for spaces that do not use `CommonSpace`, they include `SideMetadata` directly). With this change, a space has all the information it needs to allocate metadata and it does not need to query the plan any more. Also plans do not need to store information about side metadata - they provide those information to spaces when creating spaces.
* moves map/unmap functions to `SideMetadata` and updates pages when mapping/unmapping is successful. Side metadata helper functions will return the size they mapped to help with page accounting.


javadamiri:
I'm still reviewing this PR, but one general comment up-to here is that I think you need to update the module documentation for side metadata in `src/util/side_metadata/mod.rs`, specially the `# How to Use` section.

src/util/heap/accounting.rs
```diff
@@ -0,0 +1,63 @@
...
+
+impl PageAccounting {
+    pub fn new() -> Self {
+        Self {
+            reserved: AtomicUsize::new(0),
+            committed: AtomicUsize::new(0),
+        }
+    }
+
+    pub fn reserve_and_commit(&self, pages: usize) {
```
javadamiri, line 23: Probably a short comment for each of these methods will be useful

javadamiri (approved):
LGTM

src/util/heap/accounting.rs
qinsoon: Done.

Issue #299, opened 2021-04-28
Support pinning of GC threads


caizixian:
If we implement this, we would be able to use per core performance counter to understand the behaviour of a particular GC thread (that execute a particular type of work packet for example).

PR #301, opened 2021-05-03
Supporting different modes of thread affinity

caizixian:
Closing #299


qinsoon:
One thing you may need to consider is that all the threads (including worker threads) are created by the VM. So binding processors to threads may need to call into VM as well.

caizixian:
Superseded by #670

PR #303, opened 2021-05-05
Change sft from pointers to static refs

qinsoon:
This PR changes `sft` in `SFTMap` from a raw pointer to a static ref. This removes some unsafe code, and it is a task in https://github.com/mmtk/mmtk-core/issues/59.

### DIFF
--- a/src/policy/space.rs
+++ b/src/policy/space.rs
@@ -173,12 +173,7 @@ impl SFTMap {
 
     /// Update SFT map for the given address range.
     /// It should be used in these cases: 1. when a space grows, 2. when initializing a contiguous space, 3. when ensure_mapped() is called on a space.
-    pub fn update(
-        &self,
-        space: &'static (dyn SFT + Sync + 'static),
-        start: Address,
-        chunks: usize,
-    ) {
+    pub fn update(&self, space: &(dyn SFT + Sync + 'static), start: Address, chunks: usize) {
         if DEBUG_SFT {
             self.log_update(space, start, chunks);
         }
@@ -195,7 +190,7 @@ impl SFTMap {
         self.set(chunk_idx, &EMPTY_SPACE_SFT);
     }
 
-    fn set(&self, chunk: usize, sft: &'static (dyn SFT + Sync + 'static)) {
+    fn set(&self, chunk: usize, sft: &(dyn SFT + Sync + 'static)) {
         /*
          * This is safe (only) because a) this is only called during the
          * allocation and deallocation of chunks, which happens under a global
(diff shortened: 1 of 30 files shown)


src/policy/space.rs
```diff
@@ -222,7 +214,9 @@ impl SFTMap {
                 new
             );
         }
-        self_mut.sft[chunk] = sft;
+        // We have an assumption that a Space is static that we never drop a space. Under this assumption, the following cast is correct.
```
javadamiri, line 217: How about changing this argument?:

```
sft: &(dyn SFT + Sync + 'static)
```

to

```
sft: &'static (dyn SFT + Sync + 'static)
```

javadamiri: This needs changes to some other functions' arguments to compile, but it doesn't need the unsafe code here

qinsoon: I tried this. If we keep propagating the `static` requirement, eventually we need the `MMTK`->`Plan`->`Space` all to be static. But `MMTK` is instantiated in the binding, and we cannot safely know `MMTK` is static. We would also need unsafe somewhere. In that case, I think it might be better to do the unsafe code here, just in one place. Let me know what you think.

javadamiri: I tried doing this before writing the comment and I could do it without any unsafe.
I don't have the CL right now, but I will try to do it again and see if I did anything wrong, or it is possible without unsafe

qinsoon: Sure. It will be good if we can remove this unsafe code.

javadamiri: I pushed my proposed changes, take a look and revert changes if you think anything is wrong

qinsoon: That looks great. I will work on your changes. I will ask you for another review/approval when I am done.

qinsoon:
@javadamiri I think the reason why the borrow checker is unhappy in your commit e1d3473f6713bfb38dbfef4a0e97c2bf3a10b2cc is that borrowing a mutable static reference basically means the reference will not expire and it is mutable, which does not sound like what we want. I have figured out a way to remove that unsafe cast in SFT, and do not need to require all references as static. I declare a lifetime for SFTMap `'a` (rather than `'static`), and only when declaring `SFTMap` in `lazy_static`, I use `'static' as lifetime for SFTMap. So Rust should be able to figure out the lifetime requirements, and It seems working. I will push the new change. Let me know if you think it is okay.

javadamiri:
> @javadamiri I think the reason why the borrow checker is unhappy in your commit [e1d3473](https://github.com/mmtk/mmtk-core/commit/e1d3473f6713bfb38dbfef4a0e97c2bf3a10b2cc) is that borrowing a mutable static reference basically means the reference will not expire and it is mutable, which does not sound like what we want. I have figured out a way to remove that unsafe cast in SFT, and do not need to require all references as static. I declare a lifetime for SFTMap `'a` (rather than `'static`), and only when declaring `SFTMap` in `lazy_static`, I use `'static' as lifetime for SFTMap. So Rust should be able to figure out the lifetime requirements, and It seems working. I will push the new change. Let me know if you think it is okay.

Yes, it is ok

javadamiri (approved):

javadamiri (approved):

PR #304, opened 2021-05-05
Remove OOM lock

qinsoon:
This PR removes `oom_lock` which is ported from Java MMTk and used to protect access to `allocation_success`. Since `allocation_success` is `AtomicBool`, we can completely replace the use of a lock with proper atomic operations. This also addresses a task in https://github.com/mmtk/mmtk-core/issues/59.


javadamiri (approved):
LGTM

Issue #221, opened 2021-02-01
Gracefully handle out of memory errors


angussidney:
If the function `dzmmap` fails to mmap a section of memory, it does one of two things depending on the code you are looking at:
 - On the current master branch, [this assert statement fails](https://github.com/mmtk/mmtk-core/blob/master/src/util/memory.rs#L26) and Rust panics
 - With [this change by Javad in the now-closed side metadata PR](https://github.com/mmtk/mmtk-core/commit/8cbd277d5565a749c84f114fd4e4ef16ef993414#diff-68e0bc9c71e86736439c4423d8a513b834a0b58ac39cb4ba28ceb76f9151d36e), a better error is thrown, but as you can see in the call stack below, the [parent function](https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/layout/fragmented_mapper.rs#L109) simply `.unwrap()`s the result, causing Rust to panic anyway.

```
thread '<unnamed>' panicked at 'mmap with MAP_FIXED has unexpected behavior: demand zero mmap with MAP_FIXED on 0x20025f00000 returned some other address 0xffffffffffffffff', /home/angusa/mmtk-ruby/repos/mmtk-core/src/util/memory.rs:26:9
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: std::panicking::default_hook::{{closure}}
   4: std::panicking::default_hook
   5: std::panicking::rust_panic_with_hook
   6: rust_begin_unwind
   7: std::panicking::begin_panic_fmt
   8: mmtk::util::memory::dzmmap
   9: <mmtk::util::heap::layout::fragmented_mapper::FragmentedMapper as mmtk::util::heap::layout::mmapper::Mmapper>::ensure_mapped
  10: <mmtk::util::heap::monotonepageresource::MonotonePageResource<VM> as mmtk::util::heap::pageresource::PageResource<VM>>::alloc_pages
  11: mmtk::util::heap::pageresource::PageResource::get_new_pages
  12: mmtk::policy::space::Space::acquire
  13: <mmtk::util::alloc::bumpallocator::BumpAllocator<VM> as mmtk::util::alloc::allocator::Allocator<VM>>::alloc_slow_once
  14: mmtk::util::alloc::allocator::Allocator::alloc_slow
[...]

PR #307, opened 2021-05-07
Handle mmap failure properly

qinsoon:
This PR closes https://github.com/mmtk/mmtk-core/issues/221. 
* `Mmapper.ensure_mapped()` returns a `Result` to indicate whether the mmap succeeds.
* Adds a function `handle_mmap_error()`. Besides simply panicking, it deals with two types of errors differently: 1. For OOM, it calls `Collection::out_of_memory()` to inform the VM, 2. For conflicting with existing mapping, it prints a different error message.
* Add tests for `handle_mmap_error()` in DummyVM.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,6 +12,7 @@
 #![feature(associated_type_defaults)]
 #![feature(specialization)]
 #![feature(trait_alias)]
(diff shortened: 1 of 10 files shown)


javadamiri (approved):
LGTM

PR #308, opened 2021-05-10
Clean up unsync in Base/CommonPlan

qinsoon:


src/scheduler/gc_work.rs
```diff
@@ -85,6 +96,13 @@ impl<VM: VMBinding, W: CopyContext + WorkerLocal> GCWork<VM> for PrepareCollecto
     }
 }
 
+/// The global GC release Work
+/// This work packet invokes release() for the plan (which will invoke release() for each space), and
+/// pushes work packets for releasing mutators and collectors.
+/// We should only have one such work packet per GC, before any actual GC work starts.
+/// We assume this work packet is the only work packet that accesses plan, and there should
```
javadamiri, line 103: You wrote the same thing for `Prepare`.
It's not clear how they can both the only work packet that ...

qinsoon: What I meant was there should be no other work packet that accesses the plan when any of these two packets is being executed. I should say 'this work packet is the only running/active work packet that accesses the plan'.

wenyuzhao (approved):
LGTM

Issue #241, opened 2021-02-17
Expose fewer public types


qinsoon:
Currently `mmtk-core` exposes a lot of modules/types/methods/fields as `public`. Some are careless decisions. This brings the burden to us that some internal refactoring would end up changing the bindings as well. We should do a systematic pass and revisit each public item carefully, and only expose what is necessary. 
* Ideally a binding should only need to use API functions we provide.
* We should only expose types that are used in those API functions.
* We should hide details in a module, and only expose certain types from a module.
* We should try not expose fields.

qinsoon:
We should see if https://crates.io/crates/depub is helpful for our goal.

PR #312, opened 2021-05-13
Reduce public visibility

qinsoon:
This PR aims to reduce the number of public modules we expose from mmtk-core. It closes https://github.com/mmtk/mmtk-core/issues/241. 

This PR
* turns each public module to private or pub crate if possible
* adds some comments
* removes some unused code/macros


qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=limit-public-api

qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=limit-public-api

qinsoon:
v8-binding-test
V8_BINDING_REF=limit-public-api

src/memory_manager.rs
```diff
@@ -372,3 +375,42 @@ pub fn get_finalized_object<VM: VMBinding>(mmtk: &'static MMTK<VM>) -> Option<Ob
         .unwrap()
         .get_ready_object()
 }
+
+/// Get the number of workers. MMTk spawns worker threads for the 'threads' defined in the options.
+/// So the number of workers is derived from the threads option.
```
javadamiri, line 380: It also depends on the `single_worker` feature

src/policy/lockfreeimmortalspace.rs
```diff
@@ -18,6 +18,10 @@ use crate::vm::*;
 use std::marker::PhantomData;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
+/// This type implements a lock free version of the immortal collection
+/// policy. This is close to the OpenJDK's epsilon GC.
+/// Different from the normal ImmortalSpace, this version should only
+/// be used by NoGC plan, and it assumes to take up the rest of the heap range.
```
javadamiri, line 24: I don't quite understand the last sentence.
Does it mean that this policy assumes it can take up the whole heap range?

src/memory_manager.rs
qinsoon: Sure. I will make the change.

src/policy/lockfreeimmortalspace.rs
qinsoon: Yes. I assume so. You can take a look at its init function.

javadamiri (approved):
LGTM

PR #321, opened 2021-05-17
Update to Rust nightly 2021-05-12

qinsoon:


qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=rust-nightly-2021-05-12

qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=rust-nightly-2021-05-12

qinsoon:
v8-binding-test
V8_BINDING_REF=rust-nightly-2021-05-12

src/lib.rs
```diff
@@ -13,6 +12,11 @@
 #![feature(associated_type_defaults)]
 #![feature(specialization)]
 #![feature(trait_alias)]
+// Clippy favors assert!(a) rather than assert_eq!(a, true).
+// Yi: I do not completely agree with this. Sometimes, assert_eq!() is clearer
+// the original value and the expected value, especially when the expression a is long.
+// Currently we allow this pattern.
+#![allow(clippy::bool_assert_comparison)]
```
javadamiri, line 19: This lint is only about using `assert_eq!` on booleans.
For the lint errors we get from this case, I think we should apply clippy's suggestion of for example replacing `assert_eq!(l.is_free(FIRST_UNIT), true);` with `assert!(l.is_free(FIRST_UNIT));`.

I couldn't find any example where the `assert_eq!` is cleaner than `assert!`.
Anyway, we can defer making these changes if it takes too much time but we should apply them at some point

javadamiri (approved):
LGTM

src/lib.rs
qinsoon: For example,
```
assert_eq!(Address::from_usize(0x10).is_aligned_to(0x10), true);
assert_eq!(Address::from_usize(0x11).is_aligned_to(0x10), false); 
assert_eq!(Address::from_usize(0x10).is_aligned_to(0x8), true);
assert_eq!(Address::from_usize(0x10).is_aligned_to(0x20), false);
```
vs
```
assert!(Address::from_usize(0x10).is_aligned_to(0x10));
assert!(!Address::from_usize(0x11).is_aligned_to(0x10)); 
assert!(Address::from_usize(0x10).is_aligned_to(0x8));
assert!(!Address::from_usize(0x10).is_aligned_to(0x20));
```

I think the former is cleaner. However, I can make the change as the linter requires - it should be just a trivial find/replace with regex. Just let me know.

PR #315, opened 2021-05-14
Add perf event support

caizixian:
This PR supports measuring perf event for work packets.

The added code is guarded behind a feature `perf` due to an upstream bug in bindgen.
https://github.com/rust-lang/rust-bindgen/issues/1538
https://github.com/rust-lang/rust/issues/59154

PR #324, opened 2021-05-18
Make it easy to add more stat types to work packet statistics

caizixian:
This PR refactors work-packet level statistics collection.
Work packet execution time measuring was hardcoded. It's now abstracted through a `WorkCounter` trait interface.
This PR also makes #315 easier.


wenyuzhao (approved):
LGTM

qinsoon (changes requested):
As a maintainer of the code base, I have no idea what the code tries to achieve with no description in the PR and no comment in the code.

src/scheduler/work_counter.rs
```diff
@@ -0,0 +1,135 @@
...
+impl<T: 'static + WorkCounter + Clone> WorkCounterClone for T {
+    fn clone_box(&self) -> Box<dyn WorkCounter> {
+        Box::new(self.clone())
+    }
+}
+
+/// An abstraction of work counters
+///
+/// Use for trait objects, as we have might have types of work counters for
+/// the same work packet and the types are not statically known.alloc
```
qinsoon, line 34:
```suggestion
/// the same work packet and the types are not statically known.
```
typo

src/scheduler/work_counter.rs
```diff
@@ -0,0 +1,135 @@
...
+/// Stores the total, min and max of counter readings
+#[derive(Copy, Clone, Debug)]
+pub(super) struct WorkCounterBase {
+    pub(super) total: f64,
+    pub(super) min: f64,
+    pub(super) max: f64,
+}
+
+/// Make [`WorkCounter`] trait objects cloneable
+pub(super) trait WorkCounterClone {
```
qinsoon, line 20: Do we need this? Can we just implement `Clone` for WorkCounter, or require `WorkerCounter: Clone`?

src/scheduler/stat.rs
```diff
@@ -1,13 +1,24 @@
...
     work_counts: HashMap<TypeId, usize>,
-    work_durations: HashMap<TypeId, Vec<Duration>>,
+    /// Collect work counters from work threads.
+    /// Two dimensional vectors are used.
+    /// The first dimension is for different types of work counters.
+    /// The second dimension if for work counters of the same type but from
+    /// different threads.
+    /// We assume different threads have the same set of work counters
+    /// (in the same order).
+    work_counters: HashMap<TypeId, Vec<Vec<Box<dyn WorkCounter>>>>,
```
qinsoon, line 21: Even with comments, I find this hard to understand. 
* For the values in the hashmap, `value[0]` is counters for a certain type, and `value[0][0]` is a counter for a certain type for a certain thread. Is it correct?
* Is `value[i][0]` always from the same thread? How does the thread get mapped to the array index?

qinsoon (approved):
LGTM

qinsoon:
CI reported that `lusearch` on JikesRVM is slower by 12%. Is that just noise? Can you check it? Maybe run that benchmark alone to verify. @caizixian

src/scheduler/work_counter.rs
caizixian: https://doc.rust-lang.org/book/ch17-02-trait-objects.html#object-safety-is-required-for-trait-objects

The `Clone` trait is not object safe.

src/scheduler/stat.rs
caizixian: 1. Yes.

2. The order of insertion  is determined by when `merge` is called for each local stat.

caizixian: I will try to make it clearer

PR #326, opened 2021-05-20
Add explanation in the tutorial about using local mmtk-core

qinsoon:
This PR adds some explanation for how to use a local mmtk-core repo for our binding development in the tutorial. It resolves the issues raised by the conversation on Zulip: https://mmtk.zulipchat.com/#narrow/stream/262679-General/topic/Tutorial.20plan.3DMyGC.20fails/near/239504858


docs/tutorial/src/preliminaries/set_up.md
```diff
@@ -21,4 +21,17 @@ dependencies by following the instructions in the
    fastest debug variant to build, and allows for easier debugging and better 
    testing. The rest of the tutorial will assume you are using `slowdebug`.
    * You can use the env var `MMTK_PLAN=[PlanName]` to choose a plan to use at run-time.
-   The plans that are relevant to this tutorial are `NoGC` and `SemiSpace`.
\ No newline at end of file
+   The plans that are relevant to this tutorial are `NoGC` and `SemiSpace`.
+   * Make sure you only use `MMTK_PLAN` when you run the generated `java` binary
+   (`./build/linux-x86_64-normal-server-$DEBUG_LEVEL/jdk/bin/java`). Do not set `MMTK_PLAN`
```
javadamiri, line 26: It's more like:
unset `MMTK_PLAN` when you want to build...

javadamiri: If you build without setting `MMTK_PLAN`, then set it for running tests, you have to unset it before you can rebuild

javadamiri (approved):
The part about `MMTK_PLAN` can be made more clear.
Otherwise LGTM

docs/tutorial/src/preliminaries/set_up.md
qinsoon: > If you build without setting `MMTK_PLAN`, then set it for running tests, you have to unset it before you can rebuild

I am not assuming they would `export MMTK_PLAN=...` when running. If they use `MMTK_PLAN=...`, that would be fine. I will make this clear.

PR #333, opened 2021-05-26
V8 binding test now uses the perf runner

qinsoon:


.github/workflows/post-review-ci.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   v8-binding-test:
-    runs-on: [self-hosted, linux, freq-scaling-on]
+    runs-on: [self-hosted, linux, freq-scaling-off]
```
javadamiri, line 11: What happens to our performance tests?
Should we manually check that nothing else is running on the perf runner?

javadamiri (approved):
LGTM

PR #334, opened 2021-05-28
Use mmapper for side metadata

qinsoon:
This PR refactors the mmapper and uses mmapper to map memory for side metadata. This fixes the bug that our side metadata implementation may overwrite the its own memory mapping and cause data loss.

This PR:
* Moves the memory mapping of side metadata from within mmapper to `Space`. `Space` manages mapping its data memory and the side metadata memory.
* Changes mmapping for side metadata to use mmapper.
* Extracts `MapState`, and use it for both fragmented mmapper and byte map mmapper. `MapState` takes care of transition between states.
* Adds `MapState::Quarantined` and adds `quarantine_address_range()` to mmapper.
* Unmap is not supported. All unmap code in side metadata is either only used for test or is replaced with a panic (we should not unmap for those cases). 
* Increases `LOG_MAPPABLE_BYTES` in fragmented mapper from 36 (128G) to 40 (2T).
* No longer uses `PageAccounting` for side metadata. It was inaccurate to account side metadata at page granularity. Previously we mmapped side metadata at mmap chunk granularity (4M) and account side metadata at page granularity (4K), if the compression ratio is smaller than 1:1K, we may overcount our usage. Now we mmapped at page granularity, the overcount is more significant. Now we calculate side metadata usage based on the page usage of the data memory and the compression ratio.

### DIFF
--- a/src/util/heap/layout/mmapper.rs
+++ b/src/util/heap/layout/mmapper.rs
@@ -64,6 +64,7 @@ pub trait Mmapper {
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub(super) enum MapState {
     Unmapped,
+    Quarantined,
     Mapped,
     Protected,
 }
--- a/src/util/heap/layout/mmapper.rs
+++ b/src/util/heap/layout/mmapper.rs
@@ -59,3 +59,11 @@ pub trait Mmapper {
      */
     fn protect(&self, start: Address, pages: usize);
 }
+
+#[repr(u8)]
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub(super) enum MapState {
+    Unmapped,
+    Mapped,
+    Protected,
+}
--- a/src/util/heap/layout/mmapper.rs
+++ b/src/util/heap/layout/mmapper.rs
@@ -4,57 +4,58 @@ use crate::util::Address;
 use atomic::{Atomic, Ordering};
 use std::io::Result;
 
+/// Generic mmap and protection functionality
 pub trait Mmapper {
-    /****************************************************************************
-     * Generic mmap and protection functionality
-     */
-
-    /**
-     * Given an address array describing the regions of virtual memory to be used
-     * by MMTk, demand zero map all of them if they are not already mapped.
-     *
-     * @param spaceMap An address array containing a pairs of start and end
-     * addresses for each of the regions to be mappe3d
-     */
+    /// Given an address array describing the regions of virtual memory to be used
+    /// by MMTk, demand zero map all of them if they are not already mapped.
+    ///
+    /// Arguments:
+    /// * `spaceMap`: An address array containing a pairs of start and end
+    ///   addresses for each of the regions to be mapped
     fn eagerly_mmap_all_spaces(&self, space_map: &[Address]);
 
-    /**
-     * Mark a number of pages as mapped, without making any
-     * request to the operating system.  Used to mark pages
-     * that the VM has already mapped.
-     * @param start Address of the first page to be mapped
-     * @param bytes Number of bytes to ensure mapped
-     */
+    /// Mark a number of pages as mapped, without making any
+    /// request to the operating system.  Used to mark pages
+    /// that the VM has already mapped.
+    ///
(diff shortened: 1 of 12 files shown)


src/util/heap/layout/mmapper.rs
```diff
@@ -25,6 +27,8 @@ pub trait Mmapper {
      */
     fn mark_as_mapped(&self, start: Address, bytes: usize);
 
+    fn quarantine_address_range(&self, start: Address, pages: usize) -> Result<()>;
```
javadamiri, line 30: comment plz

src/util/side_metadata/helpers_32.rs
```diff
@@ -165,8 +178,8 @@ pub fn try_mmap_metadata_chunk(
     let policy_meta_start = address_to_meta_chunk_addr(start);
     if !no_reserve {
         // We have reserved the memory
-        unsafe { memory::dzmmap(policy_meta_start, local_per_chunk) }
+        MMAPPER.ensure_mapped(policy_meta_start, local_per_chunk >> LOG_BYTES_IN_PAGE)
     } else {
-        memory::mmap_noreserve(policy_meta_start, local_per_chunk)
+        MMAPPER.quarantine_address_range(policy_meta_start, local_per_chunk >> LOG_BYTES_IN_PAGE)
```
javadamiri, line 183: What if `local_per_chunk` is less than one page?
E.g. 1-bit per chunk?

qinsoon: `local_per_chunk` here is the bytes needed for the chunk, right? We should align up - I will fix this. But I don't understand your example of 1-bit per chunk. Do I misunderstand anything?

src/util/side_metadata/helpers.rs
```diff
@@ -72,43 +77,17 @@ pub(super) fn try_mmap_contiguous_metadata_space(
     let mmap_size =
         address_to_meta_address(*spec, start + size).align_up(BYTES_IN_PAGE) - mmap_start;
     if mmap_size > 0 {
-        // FIXME - This assumes that we never mmap a metadata page twice.
-        // While this never happens in our current use-cases where the minimum data mmap size is a chunk and the metadata ratio is larger than 1/64, it could happen if (min_data_mmap_size * metadata_ratio) is smaller than a page.
-        // E.g. the current implementation detects such a case as an overlap and returns false.
         if !no_reserve {
-            try_mmap_metadata(mmap_start, mmap_size)
+            MMAPPER.ensure_mapped(mmap_start, mmap_size >> LOG_BYTES_IN_PAGE)
```
javadamiri, line 81: This works ok if `mmap_size` is always a multiple of page size.
Otherwise we may use not-mmaped memory

qinsoon: I think the calculation above makes sure `mmap_size` is always page aligned.

src/util/side_metadata/helpers_32.rs
javadamiri: Right, it is the number of side metadata bytes per chunk.
My example is the case where we want n-bit of metadata per chunk (e.g. to store chunk mapping state).
In that case, we only need n-bit per chunk, and if we mmap an aligned up size of 1 page for it, we will re-mmap the same page for the future chunks

src/util/side_metadata/helpers.rs
javadamiri: Oh sorry you're right

src/util/side_metadata/helpers_32.rs
qinsoon: The mmapper ensures that won't happen. If mmapper finds the chunk is quarantined or mapped, it will not do anything. This is exactly the reason for this PR.

qinsoon: I meant the mmapper ensures we won't re-mmap the same page, as it knows the state of each chunk.

javadamiri (approved):
LGTM

src/util/heap/layout/mmapper.rs
qinsoon: Done.

Issue #337, opened 2021-06-03
GenCopy does not trigger full heap GC correctly


qinsoon:
Our implementation of GenCopy did not port some code in [`Gen`](https://github.com/JikesRVM/JikesRVM/blob/master/MMTk/src/org/mmtk/plan/generational/Gen.java), and missed some cases where we have to do a full heap GC. Currently running GenCopy would result in a false OOM without doing a full heap GC to collect mature spaces.

qinsoon:
https://github.com/mmtk/mmtk-core/pull/338 added some conditions to trigger full heap GC. More work is still needed. The place to look at is `BasePlan.last_collection_was_exhaustive()` and `BasePlan.force_full_heap_collection()`. We should put those back to `Plan`, and have a base implementation in `BasePlan`. GenCopy/Gen should implement these two methods considering full heap collection.

PR #338, opened 2021-06-04
Fix Gencopy full heap GC

qinsoon:
This PR adds conditions to properly trigger full heap GC for GenCopy. Those conditions were missing when we first ported GenCopy from Java MMTk. This PR addresses https://github.com/mmtk/mmtk-core/issues/337.

In our current code base, there is only one condition to trigger a full heap GC: at the start of a GC, the number of reserved pages is more than total pages. Otherwise, we only do nursery GCs. However, this is incorrect, and it could result in a case where we  report OOM without attempting full heap GCs. For example, if the total pages are 5000, our tospace has 2499 pages. We would never do a nursery GC (2499x2 < 5000). But when we try to allocate more than 2 pages for nursery, the allocation failed (2499x2 +2 >= 5000). So a GC is triggered and the newly reserved pages will be cleared (reserved pages back to 2499x2), and no full heap GC will be triggered.

This PR ports back more conditions from Java MMTk to trigger full heap GCs
* Adds `next_gc_full_heap` so in some situations we may force the next GC to be full heap.
* Renames `in_nursery` and `Plan.in_nursery()` to `gc_full_heap` and `Plan.is_current_gc_full_heap()` to be consistent with Java MMTk.
* Adds conditions in `collection_required()` and `request_full_heap_collection()` to trigger full heap GCs.
* `get_pages_avail()` for generational plans should be 'half of (total - reserved)'.
* Adds `max_nursery`, `min_nursery` and `full_heap_system_gc` to command line options.
* Adds comments for command line options. Removes `verbose` and `vm_space` from command line options which are no longer used.

### DIFF
--- a/src/plan/gencopy/global.rs
+++ b/src/plan/gencopy/global.rs
@@ -173,6 +173,12 @@ impl<VM: VMBinding> Plan for GenCopy<VM> {
             + self.common.get_pages_used()
     }
 
+    /// Return the number of pages avilable for allocation. Assuming all future allocations goes to nursery.
+    fn get_pages_avail(&self) -> usize {
+        // super.get_pages_avail() / 2 to reserve pages for copying
+        (self.get_total_pages() - self.get_pages_reserved()) >> 1
+    }
+
     fn base(&self) -> &BasePlan<VM> {
         &self.common.base
     }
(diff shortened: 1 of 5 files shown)


javadamiri (approved):
LGTM

wenyuzhao (approved):
LGTM

PR #343, opened 2021-06-09
Support String options

qinsoon:
This PR
* Allows using `String` as a command line option. 
* Fixes a bug that if the value of an option is invalid, we will still set the value.
* Refactors our tests to properly clean up after each test.


caizixian (approved):
LGTM

caizixian:
I think this PR is sound. The issues in the JikesRVM binding test are likely due to the JikesRVM rebase, and @qinsoon is investigating. See also mmtk/mmtk-jilesrvm#70

Issue #346, opened 2021-06-15
Wrong check in FreelistPageResource.alloc_pages()


qinsoon:
https://github.com/mmtk/mmtk-core/blob/dd2af22f94e57fd62fbcb9fb9393aa22d64bd48f/src/util/heap/freelistpageresource.rs#L101-L102

This condition checks if we cross chunk boundary. However, `sync.highwater_mark = 0` seems to be a wrong condition here. I am guessing it is intended to check if `highwater_mark` is uninitialized (which is 0), however, if we allocate for the first time ever, the `highwater_mark` is also `0`. That means we may return `new_chunk = true` when we are not actually at a new chunk. 

The exact scenario that triggers this bug:
* First allocation: we get `page_offset = 0`, and `highwater_mark` is updated to 0 (from 0). This is a new chunk.
* The following allocation: we get a `page_offset = N`. Then we check `page_offset > highwater_mark` which is true.  We further check `highwater_mark = 0 || (page_offset ^ highwater_mark) > PAGES_IN_CHUNK`. This is true regardless the second part which is the actual check to see if we cross chunk boundary. The FreeListPageResource will return `new_chunk=true` which can be incorrect.

PR #347, opened 2021-06-15
Use -1 as uninitialized value for highwater_mark in FreeListPageResource

qinsoon:
This PR tries to solve this issue: https://github.com/mmtk/mmtk-core/issues/346. Please read the issue for the details of the bug. The related code is the same as in Java MMTk, which means this could be a bug for Java MMTk as well. 

This bug does not affect our bindings right now, as 1. for JikesRVM, we have allocated through `MonotonePageResource` before allocating with `FreeListPageResource` (so `page_offset` will not be 0 for the first allocation), and 2. for OpenJDK, it happens that we do not use `FreeListPageResource`. But with the planned changes about using large object space for OpenJDK, if the first allocation is a large object, we will see this bug.

### DIFF
--- a/src/util/heap/freelistpageresource.rs
+++ b/src/util/heap/freelistpageresource.rs
@@ -19,6 +19,8 @@ use crate::vm::*;
 use std::marker::PhantomData;
 use std::mem::MaybeUninit;
 
+const UNINITIALIZED_WATER_MARK: i32 = -1;
+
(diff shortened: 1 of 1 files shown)


javadamiri (approved):
LGTM

wenyuzhao (approved):
LGTM

PR #350, opened 2021-06-17
Use aligned_alloc for malloc marksweep where possible

qinsoon:


javadamiri (approved):
LGTM

k-sareen:
So mimalloc has an aligned `calloc` function as well. We may want to conditionally compile the `memset` instead?

qinsoon:
> So mimalloc has an aligned `calloc` function as well [1, 2]. We may want to conditionally compile the `memset` instead?
> 
> Also I think that Hoard has an `aligned_alloc` [3]. But I might be mistaken.
> 
> [1]: https://github.com/microsoft/mimalloc/blob/master/include/mimalloc.h#L170
> [2]: https://github.com/microsoft/mimalloc/blob/master/src/alloc-aligned.c#L111
> [3]: https://github.com/emeryberger/Heap-Layers/blob/master/wrappers/wrapper.cpp#L213

Yeah. You are right. To summarise:
* mimalloc: aligned calloc (https://docs.rs/mimalloc-sys/0.1.6/mimalloc_sys/fn.mi_calloc_aligned.html)
* jemalloc: aligned alloc (https://docs.rs/jemalloc-sys/0.3.2/jemalloc_sys/fn.aligned_alloc.html)
* hoard: aligned alloc (but is not exposed by hoard_sys).

I don't think we will merge this PR as what it is now. There is more work needed. And we should get the offset right as well.

src/policy/mallocspace/global.rs
```diff
@@ -152,16 +152,19 @@ impl<VM: VMBinding> MallocSpace<VM> {
...
             VM::VMCollection::block_for_gc(VMMutatorThread(tls));
             return unsafe { Address::zero() };
         }
 
-        let raw = unsafe { calloc(1, size) };
+        let raw = unsafe { aligned_alloc(align, size) };
         let address = Address::from_mut_ptr(raw);
+        // zero
+        // malloc+memset is always slower than calloc. But to support alignment, we do not have a better choice.
+        crate::util::memory::zero(address, size);
```
k-sareen, line 167: Should we not check if the address is not a null ptr before zeroing it? I worry that this might segfault otherwise.

qinsoon: We should do a null check. Thanks for pointing this out.

qinsoon:
I am closing this PR, as this PR only addresses alignment for MallocAllocator. @YiluoWei is working on a solution that addresses both alignment and offset for MallocAllocator.

PR #354, opened 2021-06-23
Alloc into LOS for OpenJDK

qinsoon:
This PR adds a new field to `PlanConstraints`: `max_non_los_default_alloc_bytes`. It describes the max object size that can be allocated with the default allocator for a plan. For copying plans, we can set this to a lower number to avoid copying large objects. 

OpenJDK uses the constraint, and check object size against it during allocation: https://github.com/mmtk/mmtk-openjdk/pull/81


qinsoon:
openjdk-perf-compare
OPENJDK_BINDING_REF=use-los

qinsoon:
openjdk-microbm
OPENJDK_BINDING_REF=use-los

qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=use-los

javadamiri:
My concern about this PR is its performance results.
If I understand correctly, we're experiencing considerable overhead for the run-time size-check performed for LOS slow-path.
In some benchmarks, this overhead outweighs the performance gain, and in others it does not.
Is this acceptable for us?

qinsoon:
> My concern about this PR is its performance results.
> If I understand correctly, we're experiencing considerable overhead for the run-time size-check performed for LOS slow-path.
> In some benchmarks, this overhead outweighs the performance gain, and in others it does not.
> Is this acceptable for us?

I think the overhead is mostly gone. I made a mistake for constant size, and that was fixed. My recent measurement shows there is little overhead. However, my measurement does not include many benchmarks. I rerun the measurement with more benchmarks. The build with 'master' is the baseline, and the other is this PR.

SemiSpace
Screen Shot 2021-07-14 at 12 51 35 pm

GenCopy
Screen Shot 2021-07-14 at 12 51 47 pm

For SemiSpace, I have set `max_non_los_default_alloc_size` to be 16 pages so larger objects will be directly allocated into LOS. We expect to see some performance improvements for certain benchmarks. But for GenCopy, there are other bugs in GenCopy that we cannot allocate nursery objects into LOS (I have fixed it, but I plan to do a separate PR). So I haven't set `max_non_los_default_alloc_size` and no object will be directly allocated into LOS for GenCOpy. For GenCopy we do not expect to see speedup, but we should not see overhead as well.

src/util/rust_util.rs
```diff
@@ -0,0 +1,8 @@
+/// Const funciton for min value of two usize numbers.
+pub const fn min_of_usize(a: usize, b: usize) -> usize {
```
wenyuzhao, line 2: Is this function used anywhere?

wenyuzhao (approved):
LGTM

src/util/rust_util.rs
qinsoon: It is in the commented out code for gencopy. I can remove it for now.

qinsoon:
I have measured all the benchmarks from the dacapo bach suite. The results are as expected. 

SemiSpace: improvements for certain benchmarks, no overhead for others
WeChat Screenshot_20210715105529

GenCopy: no overhead (as we are not using a proper max_non_los_default_alloc_bytes - I will do another PR to fix this, and we should expect improvement similar to SemiSpace in this PR)
WeChat Screenshot_20210715105548

javadamiri (approved):
LGTM

PR #357, opened 2021-06-29
Support measuring process-wide perf events

caizixian:
PR #315 added support for measuring perf events for work packets.
This PR support process-wide measurements of perf events, so we can know, say the total cycles, spent in mutators/GC.

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -26,6 +26,8 @@ use crate::util::{Address, ObjectReference};
 use crate::vm::Collection;
 use crate::vm::VMBinding;
 use std::sync::atomic::Ordering;
+use std::fs::File;
+use std::io::Read;
 
 /// Run the main loop for the GC controller thread. This method does not return.
 ///
@@ -44,6 +46,15 @@ pub fn start_control_collector<VM: VMBinding>(mmtk: &MMTK<VM>, tls: VMWorkerThre
 /// * `mmtk`: A reference to an MMTk instance to initialize.
 /// * `heap_size`: The heap size for the MMTk instance in bytes.
 pub fn gc_init<VM: VMBinding>(mmtk: &'static mut MMTK<VM>, heap_size: usize) {
+    let mut status = File::open("/proc/self/status").unwrap();
+    let mut contents = String::new();
+    status.read_to_string(&mut contents).unwrap();
+    for line in contents.lines() {
+        let split: Vec<&str> = line.split('\t').collect();
+        if split[0] == "Threads:" {
+            warn!("Total number of threads {}", split[1].parse::<i32>().unwrap());
+        }
+    }
     match crate::util::logger::try_init() {
         Ok(_) => debug!("MMTk initialized the logger."),
         Err(_) => debug!(
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -25,8 +25,6 @@ use crate::util::opaque_pointer::*;
 use crate::util::{Address, ObjectReference};
 use crate::vm::Collection;
 use crate::vm::VMBinding;
-use std::fs::File;
-use std::io::Read;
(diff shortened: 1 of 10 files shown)


src/memory_manager.rs
```diff
@@ -50,6 +50,23 @@ pub fn gc_init<VM: VMBinding>(mmtk: &'static mut MMTK<VM>, heap_size: usize) {
             "MMTk failed to initialize the logger. Possibly a logger has been initialized by user."
         ),
     }
+    #[cfg(feature = "perf_counter")]
+    {
+        use std::fs::File;
+        use std::io::Read;
+        let mut status = File::open("/proc/self/status").unwrap();
```
qinsoon, line 57: Maybe use something like [`libproc`](https://crates.io/crates/libproc), rather than parsing that yourself.

src/util/options.rs
```diff
@@ -189,7 +189,8 @@ options! {
     // Perf events to measure
     // Semicolons are used to separate events
     // Each event is in the format of event_name,pid,cpu (see man perf_event_open for what pid and cpu mean)
-    perf_events:           PerfEventOptions     [always_valid] = PerfEventOptions {events: vec![]}
+    work_perf_events:       PerfEventOptions     [always_valid] = PerfEventOptions {events: vec![]},
```
qinsoon, line 192: Is it necessary to allow reading different perf events for work and for phase? It would be more straightforward to juse the same set of events for both. Please add comments for each option.

src/util/statistics/stats.rs
```diff
@@ -51,29 +52,45 @@ pub struct Stats {
 }
 
 impl Stats {
-    pub fn new() -> Self {
+    #[allow(unused)]
+    pub fn new(options: Arc<UnsafeOptionsWrapper>) -> Self {
```
qinsoon, line 56:
```suggestion
    pub fn new(options: &Options) -> Self {
```

src/util/statistics/counter/mod.rs
```diff
@@ -22,10 +26,12 @@ pub trait Counter {
...
 }
 
+/// A Diffable object could be stateless (e.g. a timer that reads the wall
+/// clock), or stateful (e.g. holds reference to a perf event fd)
 pub trait Diffable {
     type Val;
-    fn current_value() -> Self::Val;
-    fn diff(current: &Self::Val, earlier: &Self::Val) -> u64;
+    fn current_value(&mut self) -> Self::Val;
+    fn diff(&mut self, current: &Self::Val, earlier: &Self::Val) -> u64;
```
qinsoon, line 34: I think `diff` does not require `self`. But it is okay to include a `self` param here.

qinsoon (approved):
LGTM. Did you confirm that this implementation works? I did a quick implementation almost the same as this, but the read value I got for gc/mu is 0 - so I gave up on that.

caizixian:
> LGTM. Did you confirm that this implementation works? I did a quick implementation almost the same as this, but the read value I got for gc/mu is 0 - so I gave up on that.

Yes, I did test it.
image

src/util/options.rs
caizixian: If we use the same set of events, it might create too many counters (2x the number of events, one for process-wide and one for work packets) than what the CPU can support. So sometimes you might want more counters for the process-wide measurement than for work packets or vice versa.

qinsoon: You can consider using `work_perf_events` and `phase_perf_events` to differentiate them.

src/memory_manager.rs
caizixian: The functionality I need has not been implement for OS other than macOS yet. https://github.com/andrewdavidmackenzie/libproc-rs/blob/fe49a59dbb893b69f2a474a48d69e875893d2785/src/libproc/proc_pid.rs#L307

PR #361, opened 2021-06-30
Add page accounting metadata for MallocSpace

k-sareen:
This PR adds page accounting metadata to MallocSpace. I'll list the major changes to the codebase:
 
- Parallelize the sweep phase of MarkSweep
- Add a page accounting side metadata
- Add an active chunk accounting side metadata
- Use bulk load/xor of 128-bits of the alloc and mark bits in the case when the mark bits are in a side metadata
- Add check for `analysis` for the malloc allocator
- Lastly, add key insights into performance of code on different microarchitectures

There is a known bug on debug builds with the `malloc_mimalloc` feature wherein after a benchmark has finished running, `mimalloc` tries to collect its heap and somehow ends up in MMTk. It asks the SFT map for the space for a random pointer that was not mapped by us. The default behaviour of the SFT map is to ask `MallocSpace` for any pointer it is not aware of which results in the `ensure_metadata_is_mapped()` assertion to fail. I tried to fix it but ended up making the release builds 5x slower.

The other issue is that there is no nice way to count the bits/bytes set for the page metadata -- I tried to do it with essentially a map-reduce in `reserved_pages()` but that took too long as well. Would you have any better ideas than that?

Sorry for the horrible git history -- don't know why the merge resulted in it (I think the previous rebase(s) might be to blame).


k-sareen:
For context what I tried was this using the [`rayon` data parallelism crate](https://docs.rs/rayon/1.5.1/rayon/iter/trait.ParallelIterator.html):
```rust
fn reserved_pages(&self) -> usize {
    let start = self.chunk_addr_min.load(Ordering::Relaxed);
    let end = self.chunk_addr_max.load(Ordering::Relaxed) + BYTES_IN_CHUNK;
    let data_pages: usize = (start..end).into_par_iter()
        .step_by(BYTES_IN_PAGE)
        .map(|page| unsafe { load(ACTIVE_PAGE_METADATA_SPEC, Address::from_usize(page)) })
        .sum();

    let meta_pages = self.metadata.calculate_reserved_pages(data_pages)
        + CHUNK_METADATA.calculate_reserved_pages(data_pages);
    data_pages + meta_pages
}
```
which I thought was reasonable given that _at worst_ the number of chunks used are around 200 or so.

qinsoon:
The test failed. Have you checked that? I assume the PR is not ready for review as the test failed. @k-sareen

k-sareen:
Ah didn't realize it failed -- I don't see a reason for it to fail though. I'll investigate

k-sareen:
So the error lies [here](https://github.com/mmtk/mmtk-core/pull/361/files#diff-cdc1b56966139dcff8d3cc22c88746f0d16767b5e27ef37fb81bd5b08e5f375aR74), as in the test we get an absurdly low address for the chunk start so we can't subtract from it without overflow. I've fixed it by making a new function for `Address` called `saturating_sub()`.

qinsoon:
Both JikesRVM and OpenJDK tests failed at running marksweep. Is there any change you need to make for the binding? @k-sareen

qinsoon:
If it is not straightforward to figure out why the binding does not work, one thing you can consider doing is to break down this PR into smaller ones. In that way, each incremental change will be tested and reviewed separately. However, if you know what the issue is, then it's fine. Let me or Javad know if you need any help with the bindings.

src/vm/object_model.rs
```diff
@@ -48,17 +48,21 @@ pub trait ObjectModel<VM: VMBinding> {
...
     // MMTk reserved PolicySpecific side metadata offsets:
     //
-    //  1 - MarkSweep Alloc bit:
+    //  1 - MarkSweep Chunk Mark byte:
     //      - Offset `0x0` on 32-bits
     //      - Offset `LOCAL_SIDE_METADATA_BASE_ADDRESS` on 64-bits
+    //  2 - MarkSweep Alloc bit:
+    //      - Offset `Chunk Mark`.offset + `Chunk Mark`.metadata_address_range_size()
+    //  3 - MarkSweep Page Mark byte:
+    //      - Offset `Alloc bit`.offset + `Alloc bit`.metadata_address_range_size()
```
qinsoon, line 57: For the side metadata specs here, you would need to update: https://github.com/mmtk/mmtk-core/blob/a8499062f9df749a6cf7acf1e626d642913206bb/src/util/metadata/side_metadata/constants.rs#L61

I am doing a PR, and moving these comments to the constants. We may need a better way to deal with this to make sure when we use new side metadata, the base address is updated properly. But now, we have to manually update it.

qinsoon: @k-sareen

qinsoon: Oh it seems you have done it.

k-sareen: I should have done it yeah. This is something I did want to bring up because it's currently a bit obtuse and maybe we need to replace it with something like a `lazy_static!{}` that can do this during compilation for us

qinsoon: Those needs to be constants so `lazy_static!` won't work. We may need either macros or custom build scripts.

k-sareen: Oh of course yes, my bad. Hmm yeah macros could work

k-sareen: Would it not just be easier to pick a relatively safe base address for the VM metadata instead? Though there is still a chance for overlap I guess

qinsoon: There are assertions to make sure side metadata specs don't overlap. I don't know if it is a hard requirement that the side metadata specs need to be placed adjacent to each other. @javadamiri But I think as long as we have clear documents on how to define side metadata specs for bindings, and have assertions to prevent invalid specs, we are okay. We just need to improve the process so it is easier and less error prone to add new specs.

k-sareen: To be honest I am still a bit confused regarding how the 32-bit metadata is supposed to work. I think that might be the cause for the JikesRVM crash (I haven't looked into it yet). I'm always confused where to use `metadata_address_range_size()` and where to use `metadata_bytes_per_chunk()`

javadamiri: It's not a requirement.
The pros of doing this are:
1 - In chunked side_metadata, the adjacency saves us some memory space.
2 - In contiguous side_metadata, it may save us some address range.

Otherwise, as long as the spec offsets are constants, there shouldn't be any other concerns

javadamiri: `metadata_address_range_size()` is used for contiguous side metadata and it calculates the whole address range size.
We use it for:
1 - global and local side metadata in 64-bits,
2 - global side metadata in 32-bits.

`metadata_bytes_per_chunk()` is used for chunked metadata, and calculates the side metadata space size required for 1 chunk (`2 ^ 22` bytes).
We use it for local side metadata in 32-bits.

The reason why we considered chunked side metadata is that in 32-bits targets the address range size is limited and we won't be able to provide enough address space for all local side metadata.
By making it chunked, we use the fact that each chunk is only managed by one policy, and only the local side metadata of that policy will be sufficient.

k-sareen: Ah so then I should conditionally compile the offset for the local side metadata here?

https://github.com/mmtk/mmtk-core/pull/361/files#diff-cdc1b56966139dcff8d3cc22c88746f0d16767b5e27ef37fb81bd5b08e5f375aR49

javadamiri: Yes, based on `#[cfg(target_pointer_width = "32")]` or `#[cfg(target_pointer_width = "64")]`.

k-sareen: I'll update that -- it's likely the JikesRVM issue then

k-sareen:
@qinsoon The OpenJDK CI fails at `GenCopy` (likely) because of a [hardcoded side metadata base address](https://github.com/mmtk/mmtk-openjdk/blob/8bd1d0a07e32e70c5e87d6ee21ce432b04c5add2/openjdk/barriers/mmtkObjectBarrier.hpp#L16). I would have thought this used the constant `GLOBAL_SIDE_METADATA_VM_BASE_ADDRESS`.

Should I put in a PR?

qinsoon:
> @qinsoon The OpenJDK CI fails at `GenCopy` (likely) because of a [hardcoded side metadata base address](https://github.com/mmtk/mmtk-openjdk/blob/8bd1d0a07e32e70c5e87d6ee21ce432b04c5add2/openjdk/barriers/mmtkObjectBarrier.hpp#L16). I would have thought this used the constant `GLOBAL_SIDE_METADATA_VM_BASE_ADDRESS`.
> 
> Should I put in a PR?

Yes. Thanks. Probably also put an assertion in the initialization of the MMTkHeap to check the value is consistent with the VM base address.

k-sareen:
I honestly don't know why the `hsqldb` benchmark always fails for the runner. I am unable to reproduce the bug locally. The only place where I can see there may be an issue is `is_chunk_marked()` where it may go out of bounds for the eagerly mmaped active chunk space. But that should not be possible unless `malloc()` is spraying allocations everywhere.

qinsoon:
> I honestly don't know why the `hsqldb` benchmark always fails for the runner. I am unable to reproduce the bug locally. The only place where I can see there may be an issue is `is_chunk_marked()` where it may go out of bounds for the eagerly mmaped active chunk space. But that should not be possible unless `malloc()` is spraying allocations everywhere. I can increase the mmap from 16Gb worth of active chunks to 32Gb worth of active chunks if we want to check this out

The OpenJDK binding seems still panicking on invalid specs with `extreme_assertions`. That should be something to look into.

k-sareen:
> > I honestly don't know why the `hsqldb` benchmark always fails for the runner. I am unable to reproduce the bug locally. The only place where I can see there may be an issue is `is_chunk_marked()` where it may go out of bounds for the eagerly mmaped active chunk space. But that should not be possible unless `malloc()` is spraying allocations everywhere. I can increase the mmap from 16Gb worth of active chunks to 32Gb worth of active chunks if we want to check this out
> 
> The OpenJDK binding seems still panicking on invalid specs with `extreme_assertions`. That should be something to look into.

Are you sure about that? I can't trigger the invalid spec panic locally.

qinsoon:
> > > I honestly don't know why the `hsqldb` benchmark always fails for the runner. I am unable to reproduce the bug locally. The only place where I can see there may be an issue is `is_chunk_marked()` where it may go out of bounds for the eagerly mmaped active chunk space. But that should not be possible unless `malloc()` is spraying allocations everywhere. I can increase the mmap from 16Gb worth of active chunks to 32Gb worth of active chunks if we want to check this out
> > 
> > 
> > The OpenJDK binding seems still panicking on invalid specs with `extreme_assertions`. That should be something to look into.
> 
> Are you sure about that? I can't trigger the invalid spec panic locally.

Sorry my bad. I was running a wrong build.

k-sareen:
The extreme assertions build does complete the `hsqldb` benchmark locally. So I'm very confused. Can we run with RUST_BACKTRACE=1 to figure out if my hunch regarding `is_chunk_maked()` is correct?

k-sareen:
> I honestly don't know why the `hsqldb` benchmark always fails for the runner. I am unable to reproduce the bug locally. The only place where I can see there may be an issue is `is_chunk_marked()` where it may go out of bounds for the eagerly mmaped active chunk space. But that should not be possible unless `malloc()` is spraying allocations everywhere. I can increase the mmap from 16Gb worth of active chunks to 32Gb worth of active chunks if we want to check this out

Yeah so apparently `malloc` is giving us an address that is way out of range of the 16Gb we have accounted for. I think I have a fix for this.

k-sareen:
I don't agree with the my change honestly -- there's a race condition there wherein multiple mutators can try to mmap the active chunk metadata. If this is not an issue then that's fine, I guess. However if it is an issue, then I can't figure out a nice way to account for the case when we have a dumb malloc that is spraying allocations all over the heap.

In such a case we either need to mmap a much larger active chunk metadata or somehow have a way to synchronize multiple mutators such that only one will mmap the new active chunk metadata without race conditions.

qinsoon:
> I don't agree with the my change honestly -- there's a race condition there wherein multiple mutators can try to mmap the active chunk metadata. If this is not an issue then that's fine, I guess. However if it is an issue, then I can't figure out a nice way to account for the case when we have a dumb malloc that is spraying allocations all over the heap.
>
> In such a case we either need to mmap a much larger active chunk metadata or somehow have a way to synchronize multiple mutators such that only one will mmap the new active chunk metadata without race conditions.

I don't quite understand the issue you were talking about. If it is a race condition, it is a concurrency bug - there would be a way to fix it. But if it is the assumption that malloc would return addresses that are always close to each other which is not always true, then we would need a new design for the active chunks.

Anyway, if all the issues are about the new active chunk metadata, I would suggest reverting the active chunk part back to the current implementation (using hashset as before).

k-sareen:
That'll murder performance :(

I don't understand why the malloc for the runner is this uncooperative. The allocations are all over the heap.

k-sareen:
Okay I've fixed the race condition as well as the failing CI. It should be good to go now!

src/vm/object_model.rs
```diff
@@ -48,17 +48,21 @@ pub trait ObjectModel<VM: VMBinding> {
     //
     // MMTk reserved PolicySpecific side metadata offsets:
     //
-    //  1 - MarkSweep Alloc bit:
+    //  1 - MarkSweep Chunk Mark byte:
     //      - Offset `0x0` on 32-bits
```
javadamiri, line 52: This is global, right?

javadamiri: Indeed, if bindings use `.._SIDE_METADATA_VM_BASE_ADDRESS` instead of `.._SIDE_METADATA_BASE_ADDRESS`, they don't have to know these offsets.
So, we'd better move these two lists to where `.._SIDE_METADATA_VM_BASE_ADDRESS` are defined (e.g. `src/util/metadata/side_metadata/constants.rs`)

src/policy/mallocspace/global.rs
```diff
@@ -214,45 +249,180 @@ impl<VM: VMBinding> MallocSpace<VM> {
...
+                Err(x) => max = x,
+            }
+        }
+    }
+
+    pub fn sweep_chunk(&self, chunk_start: Address) {
+        // Call the relevant sweep function depending on the location of the mark bits
+        match VM::VMObjectModel::LOCAL_MARK_BIT_SPEC {
+            MetadataSpec::OnSide(local_mark_bit_side_spec) => {
+                unsafe { self.sweep_chunk_mark_on_side(chunk_start, local_mark_bit_side_spec) };
```
javadamiri, line 309: I think marking both `sweep_chunk_mark_on_side()` and `sweep_chunk_mark_in_header()` functions as `unsafe` is pointless.
Rather, we should narrow-down the unsafety to more specific areas.

Non-atomic side metadata access functions are good examples of using unsafe for a whole function.
There, we have the atomic access functions which are the safer alternatives. So, there are two points in declaring non-atomic functions as unsafe:
1 - atomic access is a safer and expected alternative,
2 - non-atomic access functions only wrap around an unsafe non-atomic operation on an `Address`, which means we are not widening the unsafe scope.

In the case of `sweep_chunk_mark_on_side()` and `sweep_chunk_mark_in_header()` functions, none of the above reasons hold true.
Most importantly, there are several locations in these functions where unsafe operations are being used.
So, we are widening the unsafe scope.
Additionally, the caller of these functions can not do anything about their unsafety (while for non-atomic side metadata access functions, they have the atomic alternative).

javadamiri: @qinsoon,
What do you think about these cases?

src/policy/mallocspace/metadata.rs
```diff
@@ -1,61 +1,137 @@
...
-use std::collections::HashSet;
-use std::sync::RwLock;
+use std::sync::atomic::Ordering;
+use std::sync::Mutex;
 
 lazy_static! {
-    pub static ref ACTIVE_CHUNKS: RwLock<HashSet<Address>> = RwLock::default();
+    pub(super) static ref CHUNK_METADATA: SideMetadataContext = SideMetadataContext {
+        global: metadata::extract_side_metadata(&[MetadataSpec::OnSide(
+            ACTIVE_CHUNK_METADATA_SPEC
```
javadamiri, line 24: `ACTIVE_CHUNK_METADATA_SPEC` is always on-side.
So, you can directly write `global: vec![ACTIVE_CHUNK_METADATA_SPEC]` here

src/policy/mallocspace/metadata.rs
```diff
@@ -1,61 +1,137 @@
...
+        )]),
+        local: vec![],
+    };
+    static ref CHUNK_MAP_LOCK: Mutex<()> = Mutex::new(());
 }
 
+/// Metadata spec for the active chunk byte
```
javadamiri, line 31: I think a short explanation of what `active chunk byte` is, is useful here.
For example you can simply say it keeps track of which chunks are used by mallocspace and which are empty.

src/policy/mallocspace/metadata.rs
```diff
@@ -1,61 +1,137 @@
...
+    offset: if cfg!(target_pointer_width = "64") {
+        LOCAL_SIDE_METADATA_BASE_ADDRESS.as_usize()
+    } else {
+        0
+    },
     log_num_of_bits: 0,
     log_min_obj_size: constants::LOG_MIN_OBJECT_SIZE as usize,
 };
 
+/// Metadata spec for the active page byte
```
javadamiri, line 64: Again, probably a short explanation of what `active page byte` is, is useful here

src/policy/mallocspace/metadata.rs
```diff
@@ -1,61 +1,137 @@
...
+    log_min_obj_size: constants::LOG_BYTES_IN_PAGE as usize,
 };
 
 pub fn is_meta_space_mapped(address: Address) -> bool {
     let chunk_start = conversions::chunk_align_down(address);
-    ACTIVE_CHUNKS.read().unwrap().contains(&chunk_start)
+    is_chunk_mapped(chunk_start) && is_chunk_marked(chunk_start)
+}
+
+fn map_chunk_mark_space(chunk_start: Address) {
```
javadamiri, line 95: It's clear to me what this function does, but I think a comment such as the ones inside the function body will be helpful here

src/policy/mallocspace/metadata.rs
```diff
@@ -1,61 +1,137 @@
...
+    if CHUNK_METADATA.try_map_metadata_space(start, size).is_err() {
+        panic!("failed to mmap meta memory");
+    }
 }
 
 pub fn map_meta_space_for_chunk(metadata: &SideMetadataContext, chunk_start: Address) {
-    let mut active_chunks = ACTIVE_CHUNKS.write().unwrap();
-    if active_chunks.contains(&chunk_start) {
+    {
+        let _lock = CHUNK_MAP_LOCK.lock().unwrap();
```
javadamiri, line 122: This lock needs comments somewhere

src/policy/mallocspace/metadata.rs
```diff
@@ -1,61 +1,137 @@
...
+        start,
+        chunk_start + (size / 2)
+    );
+
+    if CHUNK_METADATA.try_map_metadata_space(start, size).is_err() {
+        panic!("failed to mmap meta memory");
+    }
 }
 
 pub fn map_meta_space_for_chunk(metadata: &SideMetadataContext, chunk_start: Address) {
```
javadamiri, line 120: Comment,
e.g. a short list of the map ops this function does

src/policy/mallocspace/metadata.rs
```diff
@@ -72,19 +148,52 @@ pub fn is_alloced(object: ObjectReference) -> bool {
     is_alloced_object(object.to_address())
 }
 
+#[allow(unused)]
 pub fn is_alloced_object(address: Address) -> bool {
     side_metadata::load_atomic(ALLOC_SIDE_METADATA_SPEC, address, Ordering::SeqCst) == 1
 }
 
-pub fn is_marked<VM: VMBinding>(object: ObjectReference) -> bool {
+#[allow(unused)]
```
javadamiri, line 156: Why do we have so many unused functions?

src/util/address.rs
```diff
@@ -215,6 +215,11 @@ impl Address {
         Address(self.0 - size)
     }
 
+    // Perform a saturating subtract on the Address
+    pub const fn saturating_sub(self, size: usize) -> Address {
```
javadamiri, line 219: This function seems unsafe to me because it creates an address from an integer, same as `Address::from_usize`.

src/policy/mallocspace/metadata.rs
```diff
@@ -1,61 +1,137 @@
...
 pub fn is_meta_space_mapped(address: Address) -> bool {
     let chunk_start = conversions::chunk_align_down(address);
-    ACTIVE_CHUNKS.read().unwrap().contains(&chunk_start)
+    is_chunk_mapped(chunk_start) && is_chunk_marked(chunk_start)
+}
+
+fn map_chunk_mark_space(chunk_start: Address) {
+    // We eagerly map 16Gb worth of space for the chunk mark bytes on 64-bits
+    #[cfg(target_pointer_width = "64")]
+    let start = chunk_start.saturating_sub(2048 * BYTES_IN_CHUNK);
```
javadamiri, line 98: Comment on why saturating subtraction is being used

src/vm/object_model.rs
javadamiri: ignore this comments, as it's been fixed in Yi's next PR

k-sareen: Ah sorry -- I didn;t update the comments to reflect that.

src/policy/mallocspace/global.rs
k-sareen: Alright -- I'll reduce the unsafe scope.

src/policy/mallocspace/metadata.rs
k-sareen: Not sure of what you mean by "map ops"?

src/policy/mallocspace/metadata.rs
k-sareen: We don't -- I just put it on all of those functions out of habit. I'll fix it up.

src/policy/mallocspace/metadata.rs
javadamiri: I mean the metadata areas you mmap in this function

src/util/address.rs
k-sareen: Hm... Isn't that true for `Address::sub` just above it as well?

javadamiri: Oh, sorry I misunderstood what it does.
Still, I'm not sure about the safety of `Address::sub`, but it's fine for now

src/policy/mallocspace/metadata.rs
```diff
@@ -1,61 +1,137 @@
...
 };
 
 pub fn is_meta_space_mapped(address: Address) -> bool {
     let chunk_start = conversions::chunk_align_down(address);
-    ACTIVE_CHUNKS.read().unwrap().contains(&chunk_start)
+    is_chunk_mapped(chunk_start) && is_chunk_marked(chunk_start)
+}
+
+fn map_chunk_mark_space(chunk_start: Address) {
+    // We eagerly map 16Gb worth of space for the chunk mark bytes on 64-bits
```
qinsoon, line 103: Is this still necessary? It seems for every malloc result, you will check if it is mapped. If not, you will invoke this method. And when you create sweep packets, you also check if chunks are mapped, and only mapped chunks will be swept. If that's the case, it seems you do not need the eager mapping here.

qinsoon: But it is fine to do this even if it is not necesary. It only allocates 4KB here I guess (16GB/4MB).

qinsoon: What is the reason that you need to map N chunks both before and after the address?

qinsoon (approved):
I suggest that we need an OpenJDK PR for this to allow mark bit on the side (maybe two features for inheader mark bit and side mark bit, and use one for default). Otherwise, some of this PR is untested. The PR itself looks good to me.

src/plan/marksweep/gc_work.rs
```diff
@@ -54,3 +60,61 @@ impl<VM: VMBinding> DerefMut for MSProcessEdges<VM> {
...
+    }
+}
+
+// Work packet that generates sweep jobs for gc workers. Each chunk is given its own work packet
+impl<VM: VMBinding> GCWork<VM> for MSSweepChunks<VM> {
+    #[inline]
+    fn do_work(&mut self, _worker: &mut GCWorker<VM>, mmtk: &'static MMTK<VM>) {
+        let ms = self.plan.ms_space();
+        let mut work_packets: Vec<Box<dyn Work<MMTK<VM>>>> = vec![];
+        let mut chunk = unsafe { Address::from_usize(ms.chunk_addr_min.load(Ordering::Relaxed)) }; // XXX: have to use AtomicUsize to represent an Address
```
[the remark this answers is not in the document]
k-sareen: Ah yes thanks!

src/policy/mallocspace/global.rs
```diff
@@ -129,8 +150,10 @@ impl<VM: VMBinding> Space<VM> for MallocSpace<VM> {
     }
 
     fn reserved_pages(&self) -> usize {
+        // TODO: figure out a better way to get the total number of active pages from the metadata
         let data_pages = conversions::bytes_to_pages_up(self.active_bytes.load(Ordering::SeqCst));
-        let meta_pages = self.metadata.calculate_reserved_pages(data_pages);
+        let meta_pages = self.metadata.calculate_reserved_pages(data_pages)
+            + CHUNK_METADATA.calculate_reserved_pages(data_pages);
```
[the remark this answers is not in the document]
k-sareen: Oh true -- I didn't update this when I added it to the MallocSpace metadata context

src/policy/mallocspace/global.rs
```diff
@@ -214,45 +249,180 @@ impl<VM: VMBinding> MallocSpace<VM> {
...
+            // which was counterintuitive to our beliefs.
+            //
+            // The performance tradeoffs on Intel and older AMD microarchitectures were as expected, i.e.
+            // wherein the performance of the hot loop decreased if more work was done in the loop.
+            if address - page >= BYTES_IN_PAGE {
+                if page_is_empty {
+                    unset_page_mark_unsafe(page);
+                }
+                page = conversions::page_align_down(address);
+                page_is_empty = !on_page_boundary;
```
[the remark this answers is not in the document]
k-sareen: No that's not the case? `on_page_boundary` is set to false before right after this

src/policy/mallocspace/global.rs
```diff
@@ -261,63 +431,178 @@ impl<VM: VMBinding> MallocSpace<VM> {
...
+        debug!("Check active chunk {:?}", chunk_start);
+        let mut chunk_is_empty = true;
+        let mut address = chunk_start;
+        let chunk_end = chunk_start + BYTES_IN_CHUNK;
+        let mut page = conversions::page_align_down(address);
+        let mut page_is_empty = true;
+        let mut on_page_boundary = false;
+
+        // Linear scan through the chunk
+        while address < chunk_end {
```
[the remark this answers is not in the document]
k-sareen: We would have to return a triple of the `address, page_is_empty, chunk_is_empty` etc. It's not too pretty -- I can try if we want

src/policy/mallocspace/metadata.rs
k-sareen: > What is the reason that you need to map N chunks both before and after the address?

Because `malloc()` is going to allocate memory _around_ the first chunk allocated. So before as well as after the first allocation.

> Is this still necessary? It seems for every malloc result, you will check if it is mapped. If not, you will invoke this method. And when you create sweep packets, you also check if chunks are mapped, and only mapped chunks will be swept. If that's the case, it seems you do not need the eager mapping here.

Well I guess the difference now is that one mutator can do it for likely the rest of the program, but we have to map the metadata for every chunk otherwise? It might introduce unnecessary slowdowns because of the synchronization via Mutex.

k-sareen: > CHUNK_METADATA.try_map_metadata_space(): This is intended to use after try_map_metadata_address_range(). Directly using this would overwrite existing mmapping.

Sorry I don't follow? My understanding was that adding the active chunk metadata to the MallocSpace metadata won't overwrite the mapping when `metadata.try_map_metadata_space()` is called.

qinsoon: OK. If malloc returns addresses that are not close to the previous ones, this also works, right? It just goes to this method a few more times and map a few more memory.

k-sareen: Yeah this works as per the `hsqldb` test for the openjdk binding -- it eagerly allocates the active chunk metadata twice (as seen from the debug messages)

qinsoon: > > CHUNK_METADATA.try_map_metadata_space(): This is intended to use after try_map_metadata_address_range(). Directly using this would overwrite existing mmapping.
> 
> Sorry I don't follow? My understanding was that adding the active chunk metadata to the MallocSpace metadata won't overwrite the mapping when `metadata.try_map_metadata_space()` is called as it is already eagerly mapped before.

I was talking about overwriting mmappings with `mmap`. I checked the code, it should be fine. Now when we map an unmapped chunk, we always use mmap with no_replace which will not overwrite existing mappings. Before the change in https://github.com/mmtk/mmtk-core/pull/334, `try_map_metadata_space()` will do a `mmap` and that may overwrite/replace existing mappings. Ignore me on this. Sorry.

src/policy/mallocspace/global.rs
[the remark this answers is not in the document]
qinsoon: I meant in the line `page_is_empty = !on_page_boundary`. I don't get the logic here. Why `page_is_empty` is the opposite of `on_page_boundary`?

src/policy/mallocspace/global.rs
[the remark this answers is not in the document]
qinsoon: I suggest giving it a try. That removes duplicate code and may possibly simplify these two methods and make them easy to read.

src/policy/mallocspace/global.rs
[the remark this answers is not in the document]
k-sareen: If an object spans a page boundary, then the next page is not empty. Hence it's the opposite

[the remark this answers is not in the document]
k-sareen: I think you're right and it should be >= for both then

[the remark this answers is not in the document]
qinsoon: > If an object spans a page boundary, then the next page is not empty. Hence it's the opposite

Oh OK. Probably rename it to `(last_)object_on_page_boundary`, or simply put some comments for it. I thought it meant if the last loop iteration crosses page boundary.

src/policy/mallocspace/global.rs
```diff
@@ -214,45 +249,180 @@ impl<VM: VMBinding> MallocSpace<VM> {
...
-            self.active_bytes.load(Ordering::Relaxed)
-        );
+        debug!("Check active chunk {:?}", chunk_start);
+        let mut chunk_is_empty = true;
+        let mut address = chunk_start;
+        let chunk_end = chunk_start + BYTES_IN_CHUNK;
+        let mut page = conversions::page_align_down(address);
+        let mut page_is_empty = true;
+        let mut on_page_boundary = false;
+        let align: usize = 128 * (1 << LOG_MIN_OBJECT_SIZE);
```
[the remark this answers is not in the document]
qinsoon: Actually you should assert the `log_min_obj_size` for alloc bit spec and mark bit spec are the same, and use the `log_min_obj_size` from one of the specs rather than the constant `LOG_MIN_OBJECT_SIZE`.

k-sareen:
Just resolved some conversations I had addressed -- will fix up everything else

src/policy/mallocspace/metadata.rs
k-sareen: Are you still happy for me to use the `CHUNK_METADATA` for eagerly mmaping the active chunk space? I understand it is using it incorrectly, but I could not find another way of doing the eager mmaping.

src/policy/mallocspace/global.rs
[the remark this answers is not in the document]
k-sareen: We need to keep track of some context with `live_bytes` (based on if it is debug or release), so a function is kinda annoying -- we can do a macro if we want?

src/policy/mallocspace/global.rs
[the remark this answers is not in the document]
k-sareen: Just realized that the accounting for the bulk loading version isn't exactly correct as we don't know if an object is on a page boundary when we've bulk loaded them and they're alive [here](https://github.com/mmtk/mmtk-core/pull/361/files#diff-612b529a28c7592c8dcab06b6364051cf6a553ec527bb40f076ec21def62b0f0R416)

k-sareen:
@qinsoon With the new object model refactor, given a `VMLocalMarkBitSpec`, how am I supposed to figure out if it is in the side or the header? Currently the `MetadataSpec` for the type is private. I've made it public in my PR so that I can access it as `.0`. Is this fine or is there a different expected method?

src/policy/mallocspace/metadata.rs
qinsoon: I guess we don't have other choices. You have to map metadata for addresses that we/mmtk haven't mapped.

qinsoon:
> @qinsoon With the new object model refactor, given a `VMLocalMarkBitSpec`, how am I supposed to figure out if it is in the side or the header? Currently the `MetadataSpec` for the type is private. I've made it public in my PR so that I can access it as `.0`. Is this fine or is there a different expected method?

@k-sareen You can dereference those types to `MetadataSpec`: https://github.com/mmtk/mmtk-core/blob/62ccbc70915ff17f8cc6f526b781d74997b63213/src/vm/object_model.rs#L275

Basically you can use those types in places where you need a `&MetadataSpace`. E.g.
https://github.com/mmtk/mmtk-core/blob/62ccbc70915ff17f8cc6f526b781d74997b63213/src/plan/gencopy/gc_work.rs#L61
or 
https://github.com/mmtk/mmtk-core/blob/62ccbc70915ff17f8cc6f526b781d74997b63213/src/plan/gencopy/global.rs#L203

k-sareen:
Should we run the performance evaluation experiments and merge this PR?

qinsoon:
> Should we run the performance evaluation experiments and merge this PR?

I thought you would submit an OpenJDK PR to support both on-side and in-header mark bit. Otherwise the on-side mark bit code is not tested in any of our tests.

We can run a performance CI, but that does not include marksweep (we are not including marksweep in performance CI). I probably should add marksweep to our script very soon, as this PR adds a lot of optimizations to it.

k-sareen:
Ah sorry -- I muddled it up. I thought I was supposed to put in the other PR after this one. I'll do it right now sorry

src/policy/mallocspace/global.rs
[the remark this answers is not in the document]
qinsoon: Sorry I don't quite understand this. Has this been fixed? If not, can you explain the issue a little bit more? @k-sareen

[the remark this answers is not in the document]
k-sareen: No it hasn't. The issue is what we only know that _something_ is alive, and not necessarily know what is alive and how many bytes is it. In order to check if an object spans a page boundary (i.e. its end has crossed a page boundary), we would have to go through the 128-bits in some fashion to figure out if one of the objects alive spans a page boundary. I don't know what the cost of this would be -- it seems like it would defeat the purpose of the bulk xor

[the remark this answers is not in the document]
qinsoon: If I understand you correctly, the issue you are talking about is that if the 128bit check does not find dead objects, we will not linear scan the region, and we do not know if there will be any live object in the region that lives across the page boundary. In that case, the next page may be incorrectly unset because we think it is empty. Is this correct?

I further checked the use of the page mark, and it seems we never actually check the page mark (`is_page_marked()` is unused). We only set and unset the mark, but never really use the mark. If that's the case, this is not going to be a problem. However, if that's the case, why do we need the page mark at all?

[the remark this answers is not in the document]
k-sareen: > If I understand you correctly, the issue you are talking about is that if the 128bit check does not find dead objects, we will not linear scan the region, and we do not know if there will be any live object in the region that lives across the page boundary. In that case, the next page may be incorrectly unset because we think it is empty. Is this correct?

Yes you are correct.

So for the time being, I am using page mark through an inefficient way, but as outlined [above](https://github.com/mmtk/mmtk-core/pull/361#issuecomment-871969511) we need to figure out a way to do the actual calculation performantly -- maybe something like a SIMD add of 16 bytes at one time.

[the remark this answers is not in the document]
qinsoon: Can you point me to the code where you access the page mark bit? `is_page_marked()` (including the unsafe one) is never used for what I searched. If the page mark is not used, then the issue we talked about doesn't matter as it affects page mark, and page mark is not used. 

I am trying to figure out what needs to be done before this can be merged.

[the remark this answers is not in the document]
k-sareen: I use page mark for my own experiments -- it's not part of this PR as it's not performant enough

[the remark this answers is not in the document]
qinsoon: That makes sense. So we can either keep the page mark metadata if you plan to submit future PRs using it (probably leave a comment for the metadata spec saying it is not used at the moment but you plan to use it later), or we can simply remove it as it is not used.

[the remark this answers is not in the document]
k-sareen: I'll add a comment -- It'll be a bit of effort to remove it and then add it again. I am basically trying to optimize the addition of the metadata as much as possible. It doesn't seem like Rust auto-vectorizes the addition when using references. Maybe we could convert the metadata to an array/`Vec` and then do an add. That might result in SIMD instructions.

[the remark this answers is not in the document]
k-sareen: Done. I'll check why the JikesRVM build fails after dinner. I haven't changed anything other than merge master into my branch.

k-sareen:
Hmm.. The failed test case is for the `PageProtect` GC. Should we just merge this now?

qinsoon:
> Hmm.. The failed test case is for the `PageProtect` GC. Should we just merge this now?

No. We will not merge a pull request if it fails the tests. If in any case it is the master branch that is broken, I will make it my priority and fix it ASAP. Wenyu has a PR today for OpenJDK based on master, and that passed the tests, so I am guessing it is not master. Probably you want to check what changes you merged into the PR.

k-sareen:
It could still be a transient bug -- it seems like the runner complained about out of memory. Maybe another benchmark was running at the same time? Can we try running the CI again?

qinsoon:
> It could still be a transient bug -- it seems like the runner complained about out of memory. Maybe another benchmark was running at the same time? Can we try running the CI again?

I started a rerun. The OOM could be caused by exceeding the limit of total mappings for the process.

qinsoon:
> It could still be a transient bug -- it seems like the runner complained about out of memory. Maybe another benchmark was running at the same time? Can we try running the CI again?

Looks like the test passed. Should I merge this PR now?

k-sareen:
Yes we should -- this will unblock the opendjk PR as well. Thank you!

PR #369, opened 2021-07-09
Refactoring side metadata offset

qinsoon:
This PR introduces `union SideMetadataOffset` that properly represents the `offset` field in `SideMetadataSpec`. `offset` is in the type `Address` for all specs in 64 bits and global specs in 32 bits, and in the type `usize` as a relative offset for local specs in 32 bits. 
* Adds `union SideMetadataOffset` and provides functions to get offset of the right type in `SideMetadataSpec`.
* Tidies up the constants in `util::metadata::side_metadata::constants`, differentiating offset constants and address constants.
* Provides easier methods `side_first()` and `side_after()` in object model specs for a binding to create side metadata specs.

Related PRs:
* OpenJDK: https://github.com/mmtk/mmtk-openjdk/pull/88
* JikesRVM: https://github.com/mmtk/mmtk-jikesrvm/pull/75

### DIFF
--- a/src/util/metadata/global.rs
+++ b/src/util/metadata/global.rs
@@ -1,9 +1,9 @@
 use crate::util::metadata::side_metadata;
-use crate::vm::ObjectModel;
-use atomic::Ordering;
 use crate::util::metadata::side_metadata::SideMetadataSpec;
 use crate::util::ObjectReference;
+use crate::vm::ObjectModel;
 use crate::vm::VMBinding;
+use atomic::Ordering;
 
 use super::header_metadata::HeaderMetadataSpec;
 
@@ -23,8 +23,6 @@ impl MetadataSpec {
     pub fn is_on_side(&self) -> bool {
         matches!(self, &MetadataSpec::OnSide(_))
     }
-
-
 }
 
 /// A function to load the specified metadata's content.
--- a/src/util/metadata/global.rs
+++ b/src/util/metadata/global.rs
@@ -1,7 +1,6 @@
 use crate::util::metadata::side_metadata;
 use crate::vm::ObjectModel;
 use atomic::Ordering;
-
 use crate::util::metadata::side_metadata::SideMetadataSpec;
 use crate::util::ObjectReference;
 use crate::vm::VMBinding;
--- a/src/util/metadata/global.rs
+++ b/src/util/metadata/global.rs
@@ -40,7 +40,7 @@ impl MetadataSpec {
 ///
 #[inline(always)]
 pub fn load_metadata<VM: VMBinding>(
-    metadata_spec: MetadataSpec,
+    metadata_spec: &MetadataSpec,
     object: ObjectReference,
     mask: Option<usize>,
     atomic_ordering: Option<Ordering>,
@@ -71,7 +71,7 @@ pub fn load_metadata<VM: VMBinding>(
 ///
 #[inline(always)]
 pub fn store_metadata<VM: VMBinding>(
-    metadata_spec: MetadataSpec,
+    metadata_spec: &MetadataSpec,
     object: ObjectReference,
     val: usize,
     mask: Option<usize>,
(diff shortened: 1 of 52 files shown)


qinsoon:
openjdk-microbm
OPENJDK_BINDING_BRANCH_REF=metadata-offset

qinsoon:
openjdk-perf-compare
OPENJDK_BINDING_BRANCH_REF=metadata-offset

qinsoon:
jikesrvm-perf-compare
JIKESRVM_BINDING_BRANCH_REF=metadata-offset

src/util/metadata/global.rs
```diff
@@ -13,16 +13,23 @@ use super::header_metadata::HeaderMetadataSpec;
...
     OnSide(SideMetadataSpec),
 }
 
 impl MetadataSpec {
-    pub fn is_on_side(&self) -> bool {
+    pub const fn is_on_side(&self) -> bool {
         matches!(self, &MetadataSpec::OnSide(_))
     }
+
+    pub const fn extract_side_spec(&self) -> &SideMetadataSpec {
```
javadamiri, line 27: As this is a public method, probably a short comment is useful

src/util/metadata/side_metadata/global.rs
```diff
@@ -16,27 +16,125 @@ use std::sync::atomic::{AtomicU16, AtomicU32, AtomicU8, AtomicUsize, Ordering};
...
+        self.is_global || cfg!(target_pointer_width = "64")
+    }
+    /// If offset for this spec relative? (chunked side metadata for local specs in 32 bits)
+    #[inline(always)]
+    pub const fn is_rel_offset(&self) -> bool {
+        !self.is_addr_offset()
+    }
+
+    #[inline(always)]
+    pub const fn get_addr_offset(&self) -> Address {
```
javadamiri, line 39: Not a strong opinion, but probably this should be `get_addr` or `get_absolute_offset` rather than `get_addr_offset`.

src/util/metadata/side_metadata/global.rs
```diff
@@ -16,27 +16,125 @@ use std::sync::atomic::{AtomicU16, AtomicU32, AtomicU8, AtomicUsize, Ordering};
...
+            self.log_min_obj_size
         ))
     }
 }
 
+/// A union of Address or relative offset (usize) used to store offset for a side metadata spec.
+/// If a spec is contiguous side metadata, it uses address. Othrewise it uses usize.
+// The fields are made private on purpose. They can only be accessed from SideMetadata which knows whether it is Address or usize.
+#[derive(Clone, Copy)]
+pub union SideMetadataOffset {
```
javadamiri, line 78: Is there any reason you chose a `union` over an `enum`?
I think an `enum` will be easier to read

javadamiri (approved):
LGTM,

Only some minor suggestions to apply if you think are good.

src/util/metadata/global.rs
qinsoon: Sure.

src/util/metadata/side_metadata/global.rs
qinsoon: Sounds good. I will change this.

src/util/metadata/side_metadata/global.rs
qinsoon: Yeah, `enum` is generally better but `enum` has a space overhead. `enum` is probably 8+1 or 8+8 bytes in this case. I just tried to do some minimal changes to clarify things without worrying about performance. 

If we would use `enum`, I would prefer doing something like this:
```rust
enum SideMetadataSpec {
  Contiguous(ContiguousSideMetadata),
  Chunked(ChunkedSideMetadata)
}
```
This would make it clearer that there are the two kinds of side metadata and internally they work differently.

qinsoon: Umm, the example may not work.

javadamiri: We do not expect a large number of specs or offsets, so I don't think we need to worry about the space overhead.

Right, making `SideMetadataSpec` an `enum` is a better idea, if you decide to do it

qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=metadata-offset

qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=metadata-offset

PR #373, opened 2021-07-12
Clear CopySpace side forwarding-status table

wenyuzhao:
Clear side forwarding status table before each GC, if CopySpace uses side metadata for forwarding bits.

This is required, otherwise, some unforwarded objects may have "forwarded" states inherited from the previous GC.

### DIFF
--- a/src/policy/copyspace.rs
+++ b/src/policy/copyspace.rs
@@ -125,6 +125,9 @@ impl<VM: VMBinding> CopySpace<VM> {
 
     pub fn prepare(&self, from_space: bool) {
         self.from_space.store(from_space, Ordering::SeqCst);
+        // Clear the metadata if we are using side forwarding status table. Otherwise
+        // objects may inherit forwarding status from the previous GC.
+        // TODO: Fix performance.
         if let MetadataSpec::OnSide(side_forwarding_status_table) =
             *<VM::VMObjectModel as ObjectModel<VM>>::LOCAL_FORWARDING_BITS_SPEC
         {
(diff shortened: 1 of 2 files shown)


src/policy/copyspace.rs
```diff
@@ -125,6 +125,15 @@ impl<VM: VMBinding> CopySpace<VM> {
 
     pub fn prepare(&self, from_space: bool) {
         self.from_space.store(from_space, Ordering::SeqCst);
+        if let MetadataSpec::OnSide(side_forwarding_status_table) =
```
wenyuzhao, line 128: This is performance-critical. We may switch to some sort of "circular" forwarding status in the future to avoid this bulk zeroing.

src/policy/copyspace.rs
```diff
@@ -125,6 +125,15 @@ impl<VM: VMBinding> CopySpace<VM> {
...
+            *<VM::VMObjectModel as ObjectModel<VM>>::LOCAL_FORWARDING_BITS_SPEC
+        {
+            side_metadata::bzero_metadata(
```
wenyuzhao, line 131: This is performance-critical. We may switch to some sort of "circular" forwarding status in the future to avoid this bulk zeroing.

qinsoon: You can put this as TODO in the code. Also probably make it clear with comments that this zeroing is only needed for side metadata, as for inheader forwarding bits, the forwarded objects are dead and we do not need to clear their metadata.

qinsoon (approved):
LGTM

qinsoon:
I just found out that the forwarding bits are cleared in `post_copy()` for both semispace and gencopy on a per-object basis. I believe that will clear the bits for both in header and side metadata. Is there any reason we need this bulk zeroing?

wenyuzhao:
> I just found out that the forwarding bits are cleared in `post_copy()` for both semispace and gencopy on a per-object basis. I believe that will clear the bits for both in header and side metadata. Is there any reason we need this bulk zeroing? @wenyuzhao

We still need it. `post_copy()` only clears metadata for the copied object, not the original object in from-space. So metadata in from-space is not cleared properly.

PR #378, opened 2021-07-15
Add large code space

wenyuzhao:
Add large code space for large code objects.

Currently this is only used by v8 binding. Other VMs can optionally use this as well, if they want to process large code differently.


qinsoon (approved):
LGTM

PR #381, opened 2021-07-16
Always enable BasePlan spaces

wenyuzhao:
This PR adds a feature to always allocate special objects into their corresponding BasePlan (or CommonPlan) spaces. V8 expects these objects to sit in correct spaces regardless of which plan MMTk is using.

To enable this feature to NoGC, another feature is added as well to make NoGC contains a CommonPlan.

---

Currently, this feature only applies to three GCs (NoGC, SS, GenCopy) that the v8 binding is going to support. Other GCs will be updated later when the v8 binding wants to use it.


qinsoon:
I noticed this in your last PR. We are not actually using those common/base spaces, which I think is a mistake. I think we do not need the feature `force_vm_spaces`, and instead we should always allocate objects into the right space. This is probably going to involve changes for the bindings as well. 

So for this PR, for `force_vm_spaces`, yes, I think we need this change, and it should not be optional (unless there are other reasons). For `nogc_common_plan`, we have to check with Steve and what his opinion is.

qinsoon:
I talked with Steve about LOS in NoGC. Steve's opinion is that he would like to maintain NoGC simple: either we add another immortal space for large objects to NoGC (so both use bump pointer) or as in this PR, have a normal LOS but conditionally compiled. 

As for the allocator mapping change, I suggest making it non optional - we always allocate objects to the right space. Let me know if you would need any help on the bindings for this change.

wenyuzhao:
> I talked with Steve about LOS in NoGC. Steve's opinion is that he would like to maintain NoGC simple: either we add another immortal space for large objects to NoGC (so both use bump pointer) or as in this PR, have a normal LOS but conditionally compiled.

I think the second option looks better. But in additional to LOS, I also need the immortal space in CommonPlan. V8 binding uses that as its "map space" (for TIB objects).

> As for the allocator mapping change, I suggest making it non optional - we always allocate objects to the right space. Let me know if you would need any help on the bindings for this change.

Sure. I'll change it.

wenyuzhao:
> > I talked with Steve about LOS in NoGC. Steve's opinion is that he would like to maintain NoGC simple: either we add another immortal space for large objects to NoGC (so both use bump pointer) or as in this PR, have a normal LOS but conditionally compiled.
> 
> I think the second option looks better. But in additional to LOS, I also need the immortal space in CommonPlan. V8 binding uses that as its "map space" (for TIB objects).
> 
> > As for the allocator mapping change, I suggest making it non optional - we always allocate objects to the right space. Let me know if you would need any help on the bindings for this change.
> 
> Sure. I'll change it.

Done. BasePlan spaces are always enabled now, except for NoGC. BasePlan spaces, LOS and immortal space is only enabled when the feature "nogc_common_plan" is enabled.

qinsoon:
> > > I talked with Steve about LOS in NoGC. Steve's opinion is that he would like to maintain NoGC simple: either we add another immortal space for large objects to NoGC (so both use bump pointer) or as in this PR, have a normal LOS but conditionally compiled.
> > 
> > 
> > I think the second option looks better. But in additional to LOS, I also need the immortal space in CommonPlan. V8 binding uses that as its "map space" (for TIB objects).
> > > As for the allocator mapping change, I suggest making it non optional - we always allocate objects to the right space. Let me know if you would need any help on the bindings for this change.
> > 
> > 
> > Sure. I'll change it.
> 
> Done. BasePlan spaces are always enabled now, except for NoGC. BasePlan spaces, LOS and immortal space is only enabled when the feature "nogc_common_plan" is enabled.

Looks good. I probably will push some changes to remove the duplicate code for each plan about creating allocator/space mapping.

wenyuzhao:
> I have done some refactoring. There is one change I made that for NoGC (with no CommonPlan), there is no allocator mapping for immortal and LOS. Using Immortal or LOS semantic will lead to a panic. I also changed JikesRVM so it uses `nogc_common_plan`. Let me know if you think this is reasonable. @wenyuzhao Otherwise, I think we can merge this PR.

Thanks for the changes. This makes the code more clear.

Yes, we can merge it after you approved the changes. Thanks:)

qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_REF=force-vm-spaces

qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=force-vm-spaces

qinsoon:
This PR seems to break V8 (possibly due to my changes). Do you need to enable `nogc_common_plan` for V8? @wenyuzhao

wenyuzhao:
> This PR seems to break V8 (possibly due to my changes). Do you need to enable `nogc_common_plan` for V8? @wenyuzhao

I found the problem. I'll create a PR for v8 with a temporary fix. This will be fully fixed after I upstream my v8-binding changes later (which enables mmtk-v8 to do real collections).

wenyuzhao:
v8-binding-test
V8_BINDING_REF=force-vm-spaces

qinsoon:
The OpenJDK tests failed due to https://github.com/mmtk/mmtk-openjdk/issues/92. I started a rerun.

qinsoon:
Since I have made some changes to the PR, it would be better if we have a third person to review this. @caizixian Can you review this PR?

qinsoon:
openjdk-microbm
OPENJDK_BINDING_BRANCH_REF=force-vm-spaces

qinsoon:
openjdk-perf-compare
OPENJDK_BINDING_BRANCH_REF=force-vm-spaces

qinsoon:
jikesrvm-perf-compare
JIKESRVM_BINDING_BRANCH_REF=force-vm-spaces

qinsoon:
I realized that this PR needs to be updated for Immix and the Gen refactoring.

caizixian (dismissed):
As discussed with @steveblackburn in a call, we don't want to have, e.g. LOS, for NoGC. Instead, we should just have multiple immortal spaces, if the VM expected to allocate some objects into dedicated spaces.

qinsoon:
> As discussed with @steveblackburn in a call, we don't want to have, e.g. LOS, for NoGC. Instead, we should just have multiple immortal spaces, if the VM expected to allocate some objects into dedicated spaces.

What about spaces in `BasePlan` like code space? We currently use `ImmortalSpace` for them, so it is not a problem for now. But they would eventually have their own space implementation (e.g. `CodeSpace` which allows execution while other spaces do not allow execution). If we also use `ImmortalSpace` for code space, that means we would have no execution permission which is wrong. If we use `CodeSpace`, how would we justify that we use `CodeSpace` for code, but use `ImmortalSpace` for large object? That would sound like an inconsistent design.

steveblackburn:
There's a fundamental difference between LOS (a connivence for the GC, with no runtime semantics) and code space (a pain for GC, with specific runtime semantics).   The runtime should never need to know about LOS, and should not be using it in general (except as an optimization which many GCs will use).    By contrast, the runtime will always need to know about code space because it exists only to reflect specific semantics desired by the runtime.

wenyuzhao:
> The runtime should never need to know about LOS, and should not be using it in general (except as an optimization which many GCs will use).

But this is not the case for V8. A large amount of V8's tests and runtime code makes strong assumptions on which space the object should be allocated to. This is the main reason for this PR to always enable those VM spaces -- to make V8 happy.

However, to satisfy V8's internal space checks, another approach would be to encode the space meta information into a side table (or in the header), at object allocation time. MMTk can put objects into whatever space the plan prefers, and use this per-object space meta info for V8's space checks. Given that the alloc-bits overhead is minor (according to Yiluo's results), this also sounds like a possible solution.

steveblackburn:
I'm not sure that that this is correct.

First, V8 _should_ not know about LOS (since it is a GC optimization, and has no runtime-visible semantics).   Of course I can believe that it _might_ want to know about LOS, and if so, this is a bug in V8, which we need to (eventually) get fixed.

Second, even if V8 wants to know about a separate LOS, since it cannot be aware of the semantics (and the semantics of LOS are neutral), it should always be correct to make the "LOS" that V8 wants to be just another instance of an immortal space.

wenyuzhao:
Hi @qinsoon @caizixian, I think this PR is ready to review again. I mainly changed three parts:

1. Limit the "always enable BasePlan spaces" to NoGC, SemiSpace and PageProtect plans. This reduces the size of this PR. The three plans are the initial plans I'm going to port to V8 binding. Changes to other plans will be done in the future when V8 binding starts to support those plans (I also have no way to test the correctness if my fork of V8 cannot support those plans).
2. Increase max bump allocators to 6 and los allocators to 2, to meet V8 binding's requirements. This brings the PRs to the other bindings: https://github.com/mmtk/mmtk-jikesrvm/pull/77 and https://github.com/mmtk/mmtk-openjdk/pull/91
3. Use immortal spaces as VM spaces for NoGC.

---

There is also another v8 binding PR to make current v8 binding work with this change: https://github.com/mmtk/mmtk-v8/pull/35

Could you please take a look again? Thanks!

qinsoon:
> Hi @qinsoon @caizixian, I think this PR is ready to review again. I mainly changed three parts:
> 
> 1. Limit the "always enable BasePlan spaces" to NoGC, SemiSpace and PageProtect plans. This reduces the size of this PR. The three plans are the initial plans I'm going to port to V8 binding. Changes to other plans will be done in the future when V8 binding starts to support those plans (I also have no way to test the correctness if my fork of V8 cannot support those plans).
> 2. Increase max bump allocators to 6 and los allocators to 2, to meet V8 binding's requirements. This brings the PRs to the other bindings: [Always enable BasePlan spaces mmtk-jikesrvm#77](https://github.com/mmtk/mmtk-jikesrvm/pull/77) and [Always enable BasePlan spaces mmtk-openjdk#91](https://github.com/mmtk/mmtk-openjdk/pull/91)
> 3. Use immortal spaces as VM spaces for NoGC.
> 
> There is also another v8 binding PR to make current v8 binding work with this change: [mmtk/mmtk-v8#35](https://github.com/mmtk/mmtk-v8/pull/35)
> 
> Could you please take a look again? Thanks!

I am concerned about (1). Though we prefer small pull requests, each pull request should include one change and the change should be complete. This PR as it is now does not sound complete to me.  It leaves our code base inconsistent among plans. With this PR, some plans (NoGC, SemiSpace, PageProtect) use the new one-to-one mapping between allocation semantics and spaces (each allocation semantic is assigned to a different space), while other plans use the old many-to-one mapping (many allocation semantics are assigned to one space). I would prefer making the same change for all the plans, unless introducing the change to other plans triggers existing bugs or issues -- in that case, for pragmatic reasons, we may have to allow this kind of inconsistency temporarily.

I also noticed that you added LOS and immortal space to NoGC unconditionally. I believe we want to keep NoGC as simple as possible, and we could allow LOS and immortal in NoGC with a feature.

wenyuzhao:
> I am concerned about (1). Though we prefer small pull requests, each pull request should include one change and the change should be complete. This PR as it is now does not sound complete to me. It leaves our code base inconsistent among plans. With this PR, some plans (NoGC, SemiSpace, PageProtect) use the new one-to-one mapping between allocation semantics and spaces (each allocation semantic is assigned to a different space), while other plans use the old many-to-one mapping (many allocation semantics are assigned to one space). I would prefer making the same change for all the plans, unless introducing the change to other plans triggers existing bugs or issues -- in that case, for pragmatic reasons, we may have to allow this kind of inconsistency temporarily.

No. This is not inconsistent. And this PR is not complete at the very beginning -- it is just one piece of a group of changes to make V8 work. Except for the NoGC problem you mentioned, this PR only adds some dead code to the codebase, which hopefully will add some extra spaces and allocator mappings to some of the plans.  This shouldn't be considered inconsistent. If so, we shouldn't provide an interface for each plan to construct/customize the spaces and mapping at the very beginning.

> I also noticed that you added LOS and immortal space to NoGC unconditionally. I believe we want to keep NoGC as simple as possible, and we could allow LOS and immortal in NoGC with a feature.

Yeah you're right. I'll fix this by adding conditional compilation features. Thanks for pointing this out.

qinsoon:
> this PR is not complete at the very beginning

So I suggested the PR to be complete before we review and merge it. 

> it is just one piece of a group of changes to make V8 work. 

We are not introducing changes to mmtk-core solely because one specific VM would need it. We introduce changes to mmtk-core when we believe the change is reasonable and general enough so other similar VMs would benefit from it as well. 

So forcing one-to-one mapping between allocation semantics and spaces is reasonable and general, and allows us to check whether an object is allocated with a certain semantic based on its space. I think this is a good change. However, only applying this change to a subset of the plans means an inconsistent change to mmtk-core. Furthermore, the fact that the choice of the subset is solely based on a specific binding compromises our goal of generality.

> Except for the NoGC problem you mentioned, this PR only adds some dead code to the codebase, which hopefully will add some extra spaces and allocator mappings to some of the plans. This shouldn't be considered inconsistent. If so, we shouldn't provide an interface for each plan to construct/customize the spaces and mapping at the very beginning.

Based on what you said, what is the reason that some plans use one-to-one mapping between allocation semantics and spaces, and other plans still use the old many-to-one mapping? How do you justify this difference among plans that would be introduced by this PR?

wenyuzhao:
@qinsoon I've resolved the nogc space mapping issue, and applied the mapping change to all the plans. Could you please take a look again? Thanks.

src/plan/mutator_context.rs
```diff
@@ -118,3 +140,155 @@ pub trait MutatorContext<VM: VMBinding>: Send + 'static {
...
+        );
+        assert!(
+            self.n_large_object as usize <= MAX_LARGE_OBJECT_ALLOCATORS,
+            "Allocator mapping declared more large object allocators than the max allowed."
+        );
+        assert!(
+            self.n_malloc as usize <= MAX_MALLOC_ALLOCATORS,
+            "Allocator mapping declared more malloc allocators than the max allowed."
+        );
+    }
```
qinsoon, line 183: Missing assertions for immix and mark compact allocator.

src/plan/nogc/mutator.rs
```diff
@@ -1,18 +1,36 @@
...
 use crate::plan::Plan;
 use crate::util::alloc::allocators::{AllocatorSelector, Allocators};
 use crate::util::{VMMutatorThread, VMWorkerThread};
 use crate::vm::VMBinding;
-use enum_map::enum_map;
-use enum_map::EnumMap;
+use enum_map::{enum_map, EnumMap};
+
+const RESERVED_ALLOCATORS: ReservedAllocators = ReservedAllocators {
+    n_bump_pointer: 3,
```
qinsoon, line 16: Without `nogc_multi_space`, this should be `1`.

docs/tutorial/code/mygc_semispace/mutator.rs
```diff
@@ -69,14 +76,11 @@ pub fn create_mygc_mutator<VM: VMBinding>(
...
         // ANCHOR: space_mapping
-        space_mapping: box vec![
-            (AllocatorSelector::BumpPointer(0), mygc.tospace()),
-            (
-                AllocatorSelector::BumpPointer(1),
-                mygc.common.get_immortal(),
-            ),
-            (AllocatorSelector::LargeObject(0), mygc.common.get_los()),
-        ],
+        space_mapping: box {
```
qinsoon, line 79: The tutorial needs to be updated for the changes. For example, [here](https://github.com/mmtk/mmtk-core/blob/4bd6ca82ad037dac97920bf3a1408e4912a6bf04/docs/tutorial/src/mygc/ss/alloc.md?plain=1#L215), we no longer set space mapping for LOS and immortal in the plan. Can you please check if there is other outdated text in the tutorial?

qinsoon (approved):
Looks good to me. I only have some minor comments.

src/plan/mutator_context.rs
wenyuzhao: Fixed.

src/plan/nogc/mutator.rs
wenyuzhao: This is not used if `nogc_multi_space` is disabled. I've renamed it to `MULTI_SPACE_RESERVED_ALLOCATORS` and wrote some comments to make it more clear.

docs/tutorial/code/mygc_semispace/mutator.rs
wenyuzhao: I updated the text. Could you please take a read again? Thanks!

src/plan/nogc/mutator.rs
qinsoon: I see. What if we just remove the feature and use `nogc_multi_space` as the default? Is there any reason we still want to keep the current (mapping all allocation semantics to bump pointer #0)?

PR #392, opened 2021-07-30
Run binding tests with extreme_assertions

qinsoon:
This PR adds workflows for running bindings with extreme_assertions.
* Add `may_trace_duplicate_edges` in plan constraints. Some plans may trace duplicate edges, and for those plan, we do not run duplicate edge check.
* Add the missing side metadata sanity check for PageProtect.
* Increase timeout for binding tests (as they now include tests with extreme_assertions).

Related PRs:
* https://github.com/mmtk/mmtk-openjdk/pull/95
* https://github.com/mmtk/mmtk-jikesrvm/pull/79

### DIFF
--- a/src/plan/pageprotect/global.rs
+++ b/src/plan/pageprotect/global.rs
@@ -170,6 +170,8 @@ impl<VM: VMBinding> PageProtect<VM> {
             ),
         };
 
+        // Use SideMetadataSanity to check if each spec is valid. This is also needed for check
+        // side metadata in extreme_assertions.
         {
             use crate::util::metadata::side_metadata::SideMetadataSanity;
             let mut side_metadata_sanity_checker = SideMetadataSanity::new();
--- a/src/plan/pageprotect/global.rs
+++ b/src/plan/pageprotect/global.rs
@@ -148,7 +148,7 @@ impl<VM: VMBinding> PageProtect<VM> {
         let mut heap = HeapMeta::new(HEAP_START, HEAP_END);
         let global_metadata_specs = SideMetadataContext::new_global_specs(&[]);
 
-        PageProtect {
+        let ret = PageProtect {
             space: LargeObjectSpace::new(
                 "los",
                 true,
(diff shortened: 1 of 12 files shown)


qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=test-extreme-assertions

qinsoon:
jikesrvm-binding-test
JIKESRVM_BINDING_TEST=test-extreme-assertions

wenyuzhao (approved):
LGTM

src/plan/pageprotect/global.rs
```diff
@@ -168,6 +168,17 @@ impl<VM: VMBinding> PageProtect<VM> {
                 &CONSTRAINTS,
                 global_metadata_specs,
             ),
+        };
+
+        {
+            use crate::util::metadata::side_metadata::SideMetadataSanity;
+            let mut side_metadata_sanity_checker = SideMetadataSanity::new();
+            ret.common
+                .verify_side_metadata_sanity(&mut side_metadata_sanity_checker);
```
caizixian, line 177: Is this part of the extreme assertions or is it a cheap check?

qinsoon: It is only called once per plan at boot time. I guess it is fine. `extreme_assertions` needs this check, but this check itself is not part of `extreme_assertions`.

caizixian: Maybe we should add some comments so that when people write new plans, they know to call `SideMetadataSanity`

qinsoon: Yeah. I will do that.

qinsoon: Sounds good. I have added some comments. But as `new()` is specific for each plan, I can only put some comments on the `Plan` trait type about a few things developers need to do in a plan constructor. I could not think of a better place to put those comments. If you think there is a better place to put those comments, please let me know.

caizixian (approved):

src/plan/pageprotect/global.rs
k-sareen: Potentially also the tutorial?

qinsoon: The tutorial actually includes the constructor code that shows the side metadata sanity checker. I have added more comments to all the plan constructors (including the tutorial plan).

PR #395, opened 2021-07-31
Fix UseThirdPartyHeap argument order in tutorial examples

osa1:
It needs to be before the program name otherwise it's considered as a
program argument instead of JVM argument.

Rest of the examples in the same file and examples in mtk-openjdk README
are correct.


caizixian (approved):
LGTM. Thanks for the fix.

PR #397, opened 2021-08-02
Dispatch mmtk-dev-env workflow

qinsoon:
This PR adds a check: for each mmtk-core commit in `master`, we will run `ci.yml` from https://github.com/mmtk/mmtk-dev-env to make sure the core can be built on the dev env.


caizixian (approved):
LGTM

PR #404, opened 2021-08-06
Fix a concurrency bug in mapping chunk side metadata for mallocspace

qinsoon:
The PR fixed the issue https://github.com/mmtk/mmtk-jikesrvm/issues/78.

### DIFF
--- a/src/policy/mallocspace/global.rs
+++ b/src/policy/mallocspace/global.rs
@@ -203,20 +203,14 @@ impl<VM: VMBinding> MallocSpace<VM> {
             let actual_size = unsafe { malloc_usable_size(raw) };
             // If the side metadata for the address has not yet been mapped, we will map all the side metadata for the address.
             if !is_meta_space_mapped(address) {
-                // In order to prevent race conditions, we synchronize on the lock
-                let _lock = CHUNK_MAP_LOCK.lock().unwrap();
-
-                // Check it once more just in case we have mapped it by other threads before we can acquire the lock.
-                if !is_meta_space_mapped(address) {
-                    let chunk_start = conversions::chunk_align_down(address);
-                    debug!(
-                        "Add malloc chunk {} to {}",
-                        chunk_start,
-                        chunk_start + BYTES_IN_CHUNK
-                    );
-                    // Map the metadata space for the associated chunk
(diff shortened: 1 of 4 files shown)


k-sareen:
Did we figure out if simply moving the `set_chunk_mark()` to the end of the function and removing the extra scope works? That would be less restrictive than what the current PR is

qinsoon:
> Did we figure out if simply moving the `set_chunk_mark()` to the end of the function and removing the extra scope works? That would be less restrictive than what the current PR is

I tried what you proposed (holding a lock in the entire scope of `map_metadata_space_for_chunk() and moving the set chunk bit to the end). That seems working fine. 

But is it possible that for a chunk, its chunk bit side metadata is mapped but the chunk bit is not set? In that case, what you propose would still map other side metadata for it (e.g. page mark bit and alloc bit)? What I am talking about is this line which will be called when a chunk is mapped but not marked. 
https://github.com/mmtk/mmtk-core/blob/c4c2dcd8ee2470f94aea5cfedafa63abf3497fcd/src/policy/mallocspace/metadata.rs#L127
In the case we have a chunk once mapped but later released in a GC (its chunk bit is still mapped but not marked), I assume its page bit and other side metadata is still mapped, and we should not map it again. But with what you proposed, I think we will map the side metadata for it again. I haven't observed this though.

Let me know if you think what I said is going to be a problem. If not, I can make the proposed change.

k-sareen:
That was the case before the chunk map PR as well -- chunks were removed from the global HashSet even their metadata re-mapped if they were mapped before. Earlier for my MS PR, I tried to do `if is_metadata_mapped(chunk) || is_chunk_marked(chunk)` (where `is_metadata_mapped(chunk)` was defined as checking if the `ALLOC_SIDE_METADATA_SPEC` for `chunk` was mapped by Mmaper) and that resulted in somehow resulted in the debug build failing at an assertion somewhere. Though it was a while ago before I incorporated other changes so it might be fine now.

qinsoon:
I have moved the lock back to `map_metadata_space_for_chunk()`. Can you please take a look? If things are fine, just approve this PR - I will get it merged. @k-sareen

k-sareen (approved):

PR #409, opened 2021-08-10
Simplify work packets

wenyuzhao:
Remove the `Generic Scheduler` <-> `MMTk Scheduler` abstraction. We don't really need to have the generic work packets abstraction. This simplifies the code a lot.


src/scheduler/worker.rs
```diff
@@ -3,30 +3,42 @@ use super::work_bucket::*;
...
 
+/// Thread-local data for each worker thread.
+///
+/// For mmtk, each gc can define their own worker-local data, to contain their required copy allocators and other stuffs.
+pub trait GCWorkerLocal {
+    fn init(&mut self, _tls: VMWorkerThread) {}
+}
+
+/// A default implementation for scheduling systems that does not require a worker-local context.
+impl GCWorkerLocal for () {}
```
qinsoon, line 20: Do we still need this?

qinsoon (approved):
Looks good. The scheduler type is used in the API. So you might need to update the bindings. I started the checks for bindings. If anything fails, please update the bindings.

wenyuzhao:
> Looks good. The scheduler type is used in the API. So you might need to update the bindings. I started the checks for bindings. If anything fails, please update the bindings.

The scheduler type visibility is `pub(crate)`. So looks like bindings do not need to be updated.

src/scheduler/worker.rs
wenyuzhao: Thanks for pointing this out. We don't need this anymore. I've removed it.

PR #413, opened 2021-08-16
Fix object logging code

### LINKED ISSUE #412: Generational barrier object logging code correctness
https://github.com/mmtk/mmtk-core/blob/master/src/plan/barriers.rs#L58

`compare_exchange_metadata` does not internally do fetch-compare-store loop. It just simply do an exclusive load and store. So the object logging code is actually incorrect.

We should do something like this https://github.com/mmtk/mmtk-core/blob/master/src/policy/immortalspace.rs#L149-L173.

@wks Not sure if this is related to your bug.

wenyuzhao:
Fixes #412 

Do object logging correctly.


qinsoon (approved):
LGTM. Probably let Kunshan take a look and see if this fixes his bug.

wenyuzhao:
openjdk-binding-test

qinsoon:
I haven't merged the Rust version upgrade PRs for the bindings. The tests will fail for now. I am waiting for the OpenJDK binding to pass the tests before I can merge.

caizixian:
@qinsoon all tests have passed. Is this PR ready for merging?

PR #416, opened 2021-08-17
Check if alloc side metadata is mapped in is_meta_space_mapped

### LINKED ISSUE #383: is_meta_space_mapped in MallocSpace may return true for non-marksweep plans
I encountered this bug when I was debugging GenCopy. What happened was that for an object that is not in any of our heap spaces, the check https://github.com/mmtk/mmtk-core/blob/f1b3ce4cf67149f724430406eea49f7a3b4c7fb2/src/policy/space.rs#L238-L250
will check whether the object is allocated by malloc.  `is_alloced_by_malloc` should return false if the object is not allocated by malloc. 
https://github.com/mmtk/mmtk-core/blob/f1b3ce4cf67149f724430406eea49f7a3b4c7fb2/src/policy/mallocspace/metadata.rs#L158-L160

However, `is_meta_space_mapped()` may incorrectly return true that indicates the chunk where the object resides is allocated by malloc, and then when we check whether there is an alloc bit for the object, another assertion in side metadata load may fail due to the the alloc bit metadata is not mapped at all. This results in a panic in debug builds. 

We expect `is_alloced_by_malloc` returns false instead of panicking.

To reproduce
---

I have pushed a branch that can reproduce this bug: https://github.com/mmtk/mmtk-core/tree/bug-is_alloced_by_malloc.
Build `slowdebug` with https://github.com/mmtk/mmtk-openjdk/commit/fd3bd974bfdb99543efd185ee29f15de17f145b7.
Run (you may need to run it several times to reproduce the bug): 
```
RUST_BACKTRACE=1 MMTK_PLAN=GenCopy build/linux-x86_64-normal-server-slowdebug/jdk/bin/java -XX:+UseThirdPartyHeap -server -XX:MetaspaceSize=100M -Xms500M -Xmx500M -jar /usr/share/benchmarks/dacapo/dacapo-2006-10-MR2.jar hsqldb
```
Panic
```
thread '<unnamed>' panicked at '0x17fd13e3f000 of size 4096 is not mapped', /home/yilin/Code/mmtk-core/src/util/memory.rs:121:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5c029265465301fe9cb3960ce2a5da6c99b8dcf2/library/std/src/panicking.rs:493:5
   1: std::panicking::begin_panic_fmt
             at /rustc/5c029265465301fe9cb3960ce2a5da6c99b8dcf2/library/std/src/panicking.rs:435:5
   2: mmtk::util::memory::panic_if_unmapped
             at /home/yilin/Code/mmtk-core/src/util/memory.rs:121:18
   3: mmtk::util::metadata::side_metadata::global::ensure_metadata_is_mapped
             at /home/yilin/Code/mmtk-core/src/util/metadata/side_metadata/global.rs:216:5
   4: mmtk::util::metadata::side_metadata::global::load_atomic
             at /home/yilin/Code/mmtk-core/src/util/metadata/side_metadata/global.rs:226:9
   5: mmtk::policy::mallocspace::metadata::is_alloced_object
             at /home/yilin/Code/mmtk-core/src/policy/mallocspace/metadata.rs:167:5
   6: mmtk::policy::mallocspace::metadata::is_alloced
             at /home/yilin/Code/mmtk-core/src/policy/mallocspace/metadata.rs:163:5
   7: mmtk::policy::mallocspace::metadata::is_alloced_by_malloc
             at /home/yilin/Code/mmtk-core/src/policy/mallocspace/metadata.rs:159:50
   8: mmtk::policy::space::SFTMap::is_in_space
             at /home/yilin/Code/mmtk-core/src/policy/space.rs:246:13
   9: mmtk::util::address::ObjectReference::is_mapped
             at /home/yilin/Code/mmtk-core/src/util/address.rs:503:9
  10: mmtk::memory_manager::is_mapped_object
             at /home/yilin/Code/mmtk-core/src/memory_manager.rs:268:5
  11: is_mapped_object
             at /home/yilin/Code/mmtk-openjdk/mmtk/src/api.rs:159:5
  12: _ZNK8MMTkHeap5is_inEPKv
  13: _ZNK8MMTkHeap14is_in_reservedEPKv
  14: _ZNK13CollectedHeap6is_oopEP7oopDesc
  15: _ZN7oopDesc6is_oopEPS_b
  16: _ZN10HandleArea15allocate_handleEP7oopDesc
  17: _ZN6HandleC1EP6ThreadP7oopDesc
  18: _ZN10JavaThread7prepareEP8_jobject14ThreadPriority
  19: JVM_StartThread
  20: <unknown>
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Aborted (core dumped)
```

k-sareen:
Closes #383. This is a simple fix wherein we check if the alloc side metadata metadata has been mapped instead of checking if the active chunk metadata has been mapped. If it hasn't then we we return false and hence don't panic on a subsequent access to mark-sweep side metadata (as in #383).


qinsoon (commented):
I am not convinced that this PR actually fixes anything (it may mitigate the bug, but I don't think it actually is the root cause of the bug). 

The old code is straightforward: check if _chunk bit_ is mapped, if so, check if it is set. However, the new code is to check if _alloc bit_ is mapped, if so, check if chunk bit is set. I don't understand why this is correct. If there is any reasoning behind this that I did not get, please let me know.

k-sareen:
I've folded `is_chunk_mapped(chunk) && is_metadata_mapped(chunk)` into one. If the latter is true, then the former is true. But when the latter is false, we have not mapped the chunk. Oddly enough I did originally try just returning `is_metadata_mapped(chunk)` since, as you rightly pointed out, checking if the chunk is marked does not make sense but that caused marksweep tests to fail (due to not all sweep packets being generated).

PR #418, opened 2021-08-23
Rename phase-based counters

caizixian:
This PR renames the column names in statistics for phase-based (rather than work-packet-based) counters.
The reason is that `.mu` and `.gc` is misleading, especially for concurrent collectors.
`PERF_COUNT_HW_CPU_CYCLES.gc` and `PERF_COUNT_HW_CPU_CYCLES.mu` do not mean the CPU cycles of the GC and those of the mutators.
Instead, we are measuring the cycles spent during stop-the-world pauses and the cycles outside stop-the-world pauses.

### DIFF
--- a/.github/workflows/micro-bm.yml
+++ b/.github/workflows/micro-bm.yml
@@ -54,7 +54,7 @@ jobs:
         with:
           repository: mmtk/ci-perf-kit
           token: ${{ secrets.GITHUB_TOKEN }}
-          ref: "0.6.4"
+          ref: "0.6.5"
           path: ci-perf-kit
           submodules: true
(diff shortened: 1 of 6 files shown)


qinsoon (approved):
Looks good. I will update our CI script for this change (we extract `time.mu` in the script I think), and I will push a commit to this PR.

qinsoon:
I have updated ci-perf-kit. It should be consistent with this change. You probably want to update `running-ng` as well. I notcied that there are occurences of `time.mu/gc` in `running-ng`. @caizixian

caizixian:
> I have updated ci-perf-kit. It should be consistent with this change. You probably want to update `running-ng` as well. I notcied that there are occurences of `time.mu/gc` in `running-ng`. @caizixian

I just updated `running-ng`

PR #422, opened 2021-08-24
make GLOBAL_SIDE_METADATA_BASE_ADDRESS public

YiluoWei:
mmtk/mmtk-openjdk#103
Need to make GLOBAL_SIDE_METADATA_BASE_ADDRESS public because we need to know it in the allocation fast path.


src/util/metadata/side_metadata/constants.rs
```diff
@@ -17,7 +17,7 @@ use super::metadata_address_range_size;
 pub(crate) const GLOBAL_SIDE_METADATA_BASE_ADDRESS: Address =
     unsafe { Address::from_usize(BYTES_IN_CHUNK) };
 #[cfg(target_pointer_width = "64")]
-pub(crate) const GLOBAL_SIDE_METADATA_BASE_ADDRESS: Address =
+pub const GLOBAL_SIDE_METADATA_BASE_ADDRESS: Address =
```
caizixian, line 20: Can you add some comments explaining why you need to make this public?

src/util/metadata/side_metadata/constants.rs
```diff
@@ -17,7 +17,7 @@ use super::metadata_address_range_size;
 pub(crate) const GLOBAL_SIDE_METADATA_BASE_ADDRESS: Address =
```
caizixian, line 17: You should probably make this public as well for consistency

caizixian (changes requested):
Please add some description to this PR ;-)

caizixian (approved):

Issue #405, opened 2021-08-06
MallocSpace maps side metadata based on cell address but set bit based on object reference


qinsoon:
`MallocSpace` maps side metadata based on the address returned from `malloc` (the cell address).
https://github.com/mmtk/mmtk-core/blob/c4c2dcd8ee2470f94aea5cfedafa63abf3497fcd/src/policy/mallocspace/global.rs#L204

But when it sets bit in `initialize_object_metadata()`, it uses the object reference.
https://github.com/mmtk/mmtk-core/blob/c4c2dcd8ee2470f94aea5cfedafa63abf3497fcd/src/policy/mallocspace/global.rs#L76-L77

This could lead to a segfault if the cell address and the object reference are not in the same chunk. E.g, malloc returns us the last few bytes in a chunk, but the object reference is the start of the next chunk.

k-sareen:
Close?

qinsoon:
Yeah. This issue has been fixed in #434.

PR #434, opened 2021-09-01
MallocSpace should map metadata for all the chunks for an allocated object

qinsoon:
This PR fixes a bug that may cause the side metadata for an object in `MallocSpace` is not properly mapped. 

The bug is triggered when an object crosses the chunk boundary. The current code only maps the chunk for the object start, but it is possible that the `ObjectReference` for the object is in the next chunk. When we try set alloc bit for the object reference, we will see the alloc bit metadata is not yet mapped. 

This PR changes the mapping related code in `MallocSpace` so it maps all the chunks for the allocated object `[start, start + size)`.

### DIFF
--- a/src/policy/mallocspace/global.rs
+++ b/src/policy/mallocspace/global.rs
@@ -200,16 +200,10 @@ impl<VM: VMBinding> MallocSpace<VM> {
 
         if !address.is_zero() {
             let actual_size = unsafe { malloc_usable_size(raw) };
-            // If the side metadata for the address has not yet been mapped, we will map all the side metadata for the address.
-            if !is_meta_space_mapped(address) {
-                let chunk_start = conversions::chunk_align_down(address);
-                debug!(
-                    "Add malloc chunk {} to {}",
-                    chunk_start,
-                    chunk_start + BYTES_IN_CHUNK
-                );
+            // If the side metadata for the address has not yet been mapped, we will map all the side metadata for the range [address, address + actual_size).
+            if !is_meta_space_mapped(address, actual_size) {
                 // Map the metadata space for the associated chunk
-                self.map_metadata_and_update_bound(chunk_start);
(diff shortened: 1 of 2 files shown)


k-sareen:
Should we try and fix #405 in this as well or make a different PR for it?

qinsoon:
> Should we try and fix #405 in this as well or make a different PR for it?

This PR should fix https://github.com/mmtk/mmtk-core/issues/405.

src/policy/mallocspace/global.rs
```diff
@@ -266,37 +260,47 @@ impl<VM: VMBinding> MallocSpace<VM> {
...
-            match self.chunk_addr_max.compare_exchange_weak(
-                max,
-                chunk_usize,
-                Ordering::AcqRel,
-                Ordering::Relaxed,
-            ) {
-                Ok(_) => break,
-                Err(x) => max = x,
+        // Update chunk_addr_max, basing on the end of the allocation: addr + size.
+        // chn
```
k-sareen, line 289: Nitpick: the comment "chn" should be deleted.

k-sareen (approved):
LGTM other than the minor nitpick. Has Eduardo been able to test this with the Julia binding to see if it fixes his issue?

src/policy/mallocspace/global.rs
qinsoon: I have removed it. Thanks.

qinsoon:
> LGTM other than the minor nitpick. Has Eduardo been able to test this with the Julia binding to see if it fixes his issue?

Yes. I tested with the Julia binding. Eduardo can build Julia's boot image with NoGC (allocation only), but failed in marksweep (allocation only) due to this bug. With the fix, we can build Julia with marksweep (allocation only). He is working on the GC part.

PR #438, opened 2021-09-07
Fix unlog bits in `ProcessModBuf` work packet

### LINKED ISSUE #412: Generational barrier object logging code correctness

wenyuzhao:
Similar to https://github.com/mmtk/mmtk-core/pull/413, this PR adds fix to the `ProcessModBuf` work packet.

Fixes #412


qinsoon (approved):
https://github.com/mmtk/mmtk-core/pull/413 fixed the issue with a compare-and-exchange loop. But here you just used a store (which does compare-and-exchange internally). Does this mean we can simply do a store as well for https://github.com/mmtk/mmtk-core/pull/413?

Can you add some comments to `compare_exchange_metadata()` and `store()` to explain how to use them correctly so we can easily avoid misuses like these in the future?

wenyuzhao:
> #413 fixed the issue with a compare-and-exchange loop. But here you just used a store (which does compare-and-exchange internally). Does this mean we can simply do a store as well for #413?

Yes, `store` does compare-and-exchange loop internally, but it never reports whether this thread is the one that success the compare-and-exchange operation. For #413, it is necessary to quit early as soon as the unlog bit is 0, so we don't need to enter the slow path every time. 

Ideally, we may need some metadata helper function like `AtomicUsize::fetch_update` in rust std library, so that we can quit the busy loop if necessary.

> Can you add some comments to `compare_exchange_metadata()` and `store()` to explain how to use them correctly so we can easily avoid misuses like these in the future?

Done

Issue #355, opened 2021-06-24
Do not trace objects in SemiSpace's to-space


wenyuzhao:
https://github.com/mmtk/mmtk-core/blob/master/src/plan/semispace/gc_work.rs#L101-L106

It is not necessary to call `Space::trace_object` for to-space objects. This makes the code confusing.

--- A feedback after meeting with Eliot Moss.

PR #440, opened 2021-09-08
Cleanup and refactor some SemiSpace and CopySpace code

k-sareen:
Some minor cleanup and refactoring of the SemiSpace and CopySpace code. Adds a few defensive (debug) asserts as a sanity check (though these should pragmatically never fail). Also closes https://github.com/mmtk/mmtk-core/issues/355.


qinsoon (approved):
LGTM

src/util/object_forwarding.rs
```diff
@@ -30,21 +26,16 @@ const FORWARDING_POINTER_MASK: usize = 0xffff_fffc;
...
-            object,
-            None,
-            Some(Ordering::SeqCst),
-        );
+        let old_value = get_forwarding_status::<VM>(object);
         if old_value != FORWARDING_NOT_TRIGGERED_YET
             || compare_exchange_metadata::<VM>(
                 &VM::VMObjectModel::LOCAL_FORWARDING_BITS_SPEC,
                 object,
                 old_value,
```
k-sareen, line 34: I just want to check that the semantics of this compare exchange haven't changed, right? I didn't see the point of the bitwise OR and I thought that if we fail, Relaxed ordering should be good enough to give a slight bit of performance.

qinsoon: I think the bit OR is not needed, given the forwarding bits are 2 bits and we do not need to preserve other bits. We used to need the bit OR as we set bits in the header (usize). I nominated @wenyuzhao to review this PR as well. He can help check this.

qinsoon: @wenyuzhao Can you also help check if we may have the same issue here as in the https://github.com/mmtk/mmtk-core/issues/412? This `compare_exchange_metadata()` may fail for object A if another object B that is close is also forwarded. Forwarding B will cause the value in the same metadata byte/word change, and A may not get forwarded at all.

If this issue also affects here, I think probably we probably want to systematically check all uses of `compare_exchange_metadata()`.

k-sareen: But isn't the compare exchange already in a loop so it will try until it can exit from the loop?

wenyuzhao (approved):
LGTM

PR #441, opened 2021-09-08
Add extreme_assertion check for edges in ProcessEdges

k-sareen:
This adds an assertion which checks that edges we get in`ProcessEdgesWork` are where we expect them to be. It is guarded by `extreme_assertions`. The assertion check is simple: it takes the `trace_object()` checks for each plan and converts them into an assertion when a new `ProcessEdgesWork` packet (for that plan) is created.


k-sareen:
Related issue: #442

k-sareen:
Very bit rotted and not useful PR anymore.

PR #443, opened 2021-09-13
Manual dispatch perf compare

qinsoon:
This PR makes some changes to our perf comparison workflow. The current perf comparison is triggered by commits in a PR with the 'PR-benchmarking' label. After the change, the perf comparison is manually triggered with a `workflow_dispatch` event with a pull request number as its input. We can either run a workflow from the Github API in `Actions`, or later send the event from our bot when detecting certain comments in a PR.

This change achieves two goals:
1. Allows triggering the perf runs manually. This avoids running perf runs for trivial commits, or building up a long work queue if there are many pushes in a short time.
2. Allows always using our access token to access our perf scripts (we used to use `GITHUB_TOKEN` for the workflows which may not have access to our internal repos).

This PR:
* moves the performance comparison jobs to a separate yml file
* changes its trigger from `pull_request` to `workflow_dispatch`
* changes the action we use to post results back to the PR. Now we also append to one same comment rather than simply adding a comment to the PR.
* disables the micro benchmark workflow, as it seems not helpful.

Note that as `workflow_dispatch` cannot be triggered from a branch (only the workflow in `master` can be triggered), I am not able to test this. I tried to test the change with a `pull_request` trigger, and it worked fine, and I reverted the trigger (see: https://github.com/mmtk/mmtk-core/pull/443/commits/ea4980d3373c0723c95ea39289578a0fa4b24758). However, it is possible that this PR does not work, and requires further fix.


qinsoon:
Running benchmarks for JikesRVM...

qinsoon:
Run finished.

qinsoon:
Running benchmarks for OpenJDK...

qinsoon:
Run finished

.github/workflows/perf-compare-ci.yml
```diff
@@ -0,0 +1,215 @@
...
+          - uses: hasura/comment-progress@v2.1.0
+            if: always()
+            with:
+                github-token: ${{ secrets.CI_ACCESS_TOKEN }}
+                repository: 'mmtk/mmtk-core'
+                number: ${{ env.PERF_PR }}
+                id: jikesrvm-perf-compare-comment
+                message: ${{ steps.cat.outputs.text }}
+                append: true
+          # Cleanup logs (this is necessary for self-hosted runners)
```
caizixian, line 107:
```suggestion
          # Cleanup logs (this is necessary for self-hosted runners)
          # [`always()`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always) ensures we clean up the logs even if some steps fail
```

.github/workflows/perf-compare-ci.yml
```diff
@@ -0,0 +1,215 @@
...
+            uses: actions/checkout@v2
+            with:
+              ref: ${{ env.MMTK_CORE_BRANCH_REF }}
+              path: mmtk-core-branch
+          # Checkout perf-kit
+          - name: Checkout Perf Kit
+            uses: actions/checkout@v2
+            with:
+              repository: mmtk/ci-perf-kit
+              token: ${{ secrets.GITHUB_TOKEN }}
```
caizixian, line 58: Aren't we supposed to use `CI_ACCESS_TOKEN` to clone private submodules?

qinsoon: Yes. The version of `ci-perf-kit` in this PR does not use private submodules, so it's fine. I can change it here though.

caizixian: Ah, right. That can be fixed in #444. No problem. Just leave it as is.

caizixian (approved):
LGTM. Please take a look at that one line of comment I added. I think it's worth clarifying why we need that conditional.

qinsoon:
> LGTM. Please take a look at that one line of comment I added. I think it's worth clarifying why we need that conditional.

Yeah. That looks fine. Thanks for the review. I will merge this PR later (possibly right before #444 ).

PR #444, opened 2021-09-13
Use new ci-perf-kit with running-ng

qinsoon:


qinsoon:
Superseded by https://github.com/mmtk/mmtk-core/pull/1005.

PR #447, opened 2021-09-17
Rename `primary` of `BasePlan` and `CommonPlan` to `full_heap`

### LINKED ISSUE #435: Misleading parameters in `prepare` and `release` in `CommonPlan` and `BasePlan`
For example, see the code below.
https://github.com/mmtk/mmtk-core/blob/8e45ad061d9327f5ec4ef3b29354070cb65dc884/src/plan/global.rs#L828-L831

We use normally use terminologies, such as full heap or major GC for this purpose. We have not used the word `primary` anywhere else in the code base.

Also see Java MMTk.

https://github.com/JikesRVM/JikesRVM/blob/38b21f5a663016dbf43771cb2d231f74db0a01c6/MMTk/src/org/mmtk/policy/LargeObjectSpace.java#L105

caizixian:
Close #435


qinsoon (approved):
LGTM

Issue #374, opened 2021-07-12
Start address passed to Space::grow_space does not necessarily need to be chunk-aligned


wenyuzhao:
Consider the following scenario:

In LargeObjectSpace, allocate one 5MB object and calls [allocate_contiguous_chunks](https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/freelistpageresource.rs#L97) during allocation.
The allocated object is guaranteed to be chunk-aligned. `highwater_mark` is updated to the start address of the 1st chunk at [here](https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/freelistpageresource.rs#L114).
The subsequent `grow_space` then receives a chunk-aligned start address.
```
     1st chunk         2nd chunk
[aaaaaaaaaaaaaaaa] [aaaaaaaa--------]
 ^
 highwater_mark
```
Then allocate another 1MB object.
highwater_mark will be updated to some _chunk-unaligned_ address within the 2nd chunk.
So under this case, **the subsequent `grow_space` will receive an unaligned start address.**
```
     1st chunk         2nd chunk
[aaaaaaaaaaaaaaaa] [aaaabbbb--------]
                        ^
                        highwater_mark
```

---

So either the [chunk alignment check](https://github.com/mmtk/mmtk-core/blob/master/src/policy/space.rs#L358) in `grow_space` is incorrect, or we should update highwater_mark to the end address of the object allocation, instead of the start address.

qinsoon:
That assertion only checks the address is chunk aligned when `new_chunk = true`. The assertion tries to make sure that we do not overwrite the SFT entry (e.g. in your example, the SFT for 2nd chunk should only be set once). A weaker assertion could be to make sure we do not overwrite SFT to a different value.

In your example, the first allocation should initialize SFT map for both chunks, and the second allocation should not return `new_chunk = true` as it is not a new chunk. Maybe we should use the end of an object as highwater_mark.

wenyuzhao:
> That assertion only checks the address is chunk aligned when `new_chunk = true`. The assertion tries to make sure that we do not overwrite the SFT entry (e.g. in your example, the SFT for 2nd chunk should only be set once). A weaker assertion could be to make sure we do not overwrite SFT to a different value.
> 
> In your example, the first allocation should initialize SFT map for both chunks, and the second allocation should not return `new_chunk = true` as it is not a new chunk. Maybe we should use the end of an object as highwater_mark.

[...]

PR #461, opened 2021-10-01
Fix OpenJDK derived root pointers

wenyuzhao:


qinsoon:
openjdk-binding-test
OPENJDK_BINDING_REF=derived-root-pointers

wenyuzhao:
@wks I could not add you as a reviewer for the mmtk-openjdk PR https://github.com/mmtk/mmtk-openjdk/pull/115. You are not on reviewers name list... But you can take a look at that as well. Thanks.

src/policy/space.rs
```diff
@@ -355,10 +355,11 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
             bytes,
             new_chunk
         );
-        debug_assert!(
-            (new_chunk && start.is_aligned_to(BYTES_IN_CHUNK)) || !new_chunk,
-            "should only grow space for new chunks at chunk-aligned start address"
-        );
+        // FIXME: This assertion is too strict. See https://github.com/mmtk/mmtk-core/issues/374
+        // debug_assert!(
```
qinsoon, line 359: I am not comfortable about just removing this assertion. As discussed in https://github.com/mmtk/mmtk-core/issues/374#issuecomment-878047774, if somehow, we have an address in the middle of a chunk and tell the function that this is a new chunk, the obvious problem is that this is NOT a new chunk and we must have already initialised the SFT map for the chunk.

By the way, why is this change included in this PR? This PR changes GC, I assume that wouldn't trigger this assertion.

wenyuzhao: When I test this PR with the new DaCapo benchmarks, it sometimes triggers this assertion failure. I feel that this assertion is too strict, which means this is actually an incorrect assertion. We shouldn't put an incorrect assertion in the code base.

Issue #428, opened 2021-08-26
Panics with File exists in byte_map_mmapper for marksweep allocation tests (examples/build.py)


qinsoon:
I observed that the marksweep allocation tests may fail due to a mapping failure. This can be reproduced on the current master (5d7ea0f50d6b289200a522c004afa821c465d021). 

Run `python examples/build.py`. This bug is random. But it only fails with the same error message on marksweep 32 bits.
```
[exec_and_redirect] ['./test_mmtk_32'] ({'MMTK_PLAN': 'MarkSweep', 'RUST_BACKTRACE': 'full', 'LD_LIBRARY_PATH': '.'})
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 17, kind: AlreadyExists, message: "File exists" }', /home/yilin/Code/mmtk-core/src/util/heap/layout/byte_map_mmapper.rs:65:77
stack backtrace:
   0: 0xf7eb2621 - std::backtrace_rs::backtrace::libunwind::trace::ha835d63cb25fe1b0
                       at /rustc/ae90dcf0207c57c3034f00b07048d63f8b2363c8/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1: 0xf7eb2621 - std::backtrace_rs::backtrace::trace_unsynchronized::h1fdb05f4a4a95c6f
                       at /rustc/ae90dcf0207c57c3034f00b07048d63f8b2363c8/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2: 0xf7eb2621 - std::sys_common::backtrace::_print_fmt::h2ca1df030c8c337b
                       at /rustc/ae90dcf0207c57c3034f00b07048d63f8b2363c8/library/std/src/sys_common/backtrace.rs:67:5
   3: 0xf7eb2621 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::haf53d6f89780a0a3
                       at /rustc/ae90dcf0207c57c3034f00b07048d63f8b2363c8/library/std/src/sys_common/backtrace.rs:46:22
   4: 0xf7e089b0 - core::fmt::write::h79c58d6b5d5e3842
                       at /rustc/ae90dcf0207c57c3034f00b07048d63f8b2363c8/library/core/src/fmt/mod.rs:1117:17
   5: 0xf7eb1c44 - std::io::Write::write_fmt::h4d8b8b53ceef04b9
[...]

PR #463, opened 2021-10-01
Don't trace objects in to-space in GenCopy

k-sareen:
Accidentally forgot to add this in #440 which adds an assert in CopySpace


k-sareen:
Failing error doesn't seem to be related to the PR -- I think it's a metadata mapping error

qinsoon:
> Failing error doesn't seem to be related to the PR -- I think it's a metadata mapping error

Yeah. https://github.com/mmtk/mmtk-core/issues/428.

qinsoon (approved):
LGTM

Issue #414, opened 2021-08-17
Tutorial issues


qinsoon:
There are some minor issues with the current tutorial:
1. The tutorial asks to modify both `plan::global` and `mygc::global`. But sometimes, it only refers to `global.rs` which can be confusing.
2. The code snippet in `create_mutator` and `create_plan` is outdated.
3. Methods like `tospace()` and `fromspace()` are used before the tutorial asks to implement them in 'Access MyGC spaces'.

I will add more to the list if there are more issues raised.

PR #469, opened 2021-10-08
Tutorial: Fix issues

rouanebannister:
See #414. 
In addition:
* Emphasised that the tutorial code may be different to the main code.
* Removed some lines in the long non-anchored code snipped in 3.1 (Create MyGC), so it won't need to be updated as often and is will take up less space.


rouanebannister:
I had forgotten I marked this as a draft, but it was because I need to do a compile with mdBook to make sure I didn't break any formatting. I'm having some issues with my computer and mdBook, but I should be able to get it fixed and mark this for review within a few hours.

PR #486, opened 2021-10-27
Precise/Non-precise stress test (merge after #482)

qinsoon:
This PR adds non-precise stress test. The current implementation of stress test which checks for stress GCs in every allocation is named as 'precise stress' in this PR. This PR further adds non-precise stress test which only checks for stress GCs in the allocation slowpaths. A new option `precise_stress` is added in this PR, which defaults to true.


src/util/alloc/bumpallocator.rs
```diff
@@ -82,7 +85,7 @@ impl<VM: VMBinding> Allocator<VM> for BumpAllocator<VM> {
     // Performs three kinds of allocations: (i) if the hard limit has been met;
```
k-sareen, line 85: Could you update the commends to say "precise" stress here as well?

k-sareen (approved):
LGTM other than minor comment. Thanks!

src/util/alloc/bumpallocator.rs
qinsoon: Done. Thanks for the review.

PR #498, opened 2021-11-10
Refactor scanned_stacks to a method increment_and_check_limit()

qinsoon:
This PR extracts the code for incrementing `scanned_stacks` to a separate method. This improves readability and also allows us to unit test the code separately.


src/util/rust_util.rs
```diff
@@ -6,3 +6,68 @@ pub const fn min_of_usize(a: usize, b: usize) -> usize {
...
+                current - limit,
+                Ordering::Release,
+                Ordering::Relaxed,
+            ) == Ok(current)
+            {
+                return true;
+            }
+        }
+    }
+    false
```
wks, line 35: This function performs memory accesses too many times.  It is a unnecessary, and not atomic. I recommend a transaction-style load-update-store loop:

```suggestion
    match u.fetch_update(Ordering::SeqCst, Ordering::Relaxed, |current| {
        if current >= limit {
            None
        } else if current + 1 < limit {
            Some(current + 1)
        } else {
            Some(0)
        }
    }) {
        Err(old) => panic!("Old vlaue already reached limit: {}", old),
        Ok(old) => old + 1 == limit
    }
```

wks (changes requested):
See the inline comments for some suggestions for the implementation of `increment_and_check_limit`.

For a work-packet-based scheduler, there should be some build-in dependency management facilities that specifies some work packets depends on some jobs being finished, and the transition to `GcProper` can be written its own work packet.  We can refactor that later.

src/util/rust_util.rs
qinsoon: I have made the suggested change with only one difference -- the update function does not return `None`. Can you take a look and see if the PR is okay now?

src/util/rust_util.rs
```diff
@@ -6,3 +6,71 @@ pub const fn min_of_usize(a: usize, b: usize) -> usize {
...
+            // If the new value is larger than the limit, this should not happen as
+            // we increment by 1 and we should have reset it to zero when the value ever reaches the limit.
+            unreachable!()
+        }
+    }) {
+        // We are not returning None from the update func, so Err is unreachable.
+        Err(_) => unreachable!(),
+        Ok(old) => old + 1 == limit,
+    }
+}
```
wks, line 39: What about letting it stick to the maximum value?
```suggestion
fn increment_and_check_limit(u: &AtomicUsize, limit: usize) -> bool {
    let old = u.fetch_add(1, Ordering::SeqCst);
    if old >= limit {
        panic!("Incremented past limit: {} -> {}, limit: {}", old, old + 1, limit);
    }
    old + 1 == limit
}
```

wks (changes requested):
Sorry, but after a second thought, I think we can change the current code so it fails faster.  We can make `increment_and_check_limit` not wrapping the counter back to 0 when reached limit, but panic immediately, so that if we ever overflow the counter, we are immediately notified for the erroneous state.  See the inline comment.

I have seen such a phenomenon when debugging https://github.com/mmtk/mmtk-openjdk/issues/93.  The counter wrapped around and continued counting, because two VMOperations are incorrectly executed concurrently and caused each stack to be scanned twice.  With this change, we can spot this error earlier.

And we should also change the following line to initialize the counter manually.  Since overflow is already detected eagerly, we no longer need to assert it is equal to 0, but just set it to 0.

```diff
- debug_assert_eq!(mmtk.plan.base().scanned_stacks.load(Ordering::SeqCst), 0);
+ mmtk.plan.base().scanned_stacks.store(0, Ordering::SeqCst);
```

src/util/rust_util.rs
qinsoon: With this change the code would be cleaner. However, one implication from the change is that we would always need some sort of global synchronization (to reset `scanned_stacks`) before running any `ScanStackRoot` packet. Currently we have no such requirement for `ScanStackRoot` -- we can just schedule them and they will work. I am not confident that the current implementation is correct though (especially whether it is correct if the number of mutators changes somehow). I will think about this a little bit more.

wks: I think it is better to have such "global synchronization".  The sole purpose of shared atomic counter is synchronization.  Stack scanning is a typical fork-join use pattern. After stopping the world, we spawn N work packets for N stacks, and wait for those N packets to complete.  I think we effectively create a "goal" of "do something after N stacks are scanned" the moment we spawn N work packets, so that's a natural point to initialize the counter, and it is why we have the counter in the first place.  It is also a kind of "global synchronization" that we "do things (shift the GC phase) **after** the counter reaches N".

Speaking of reading/writing shared variable, in Java, it is also natural to write to a variable before forking (`Thread.start`), and read from the variable in sub-threads.  Although we did not use any locks for synchronization, the Java memory model still considers the write as happening-before the read (`Thread.start` happens before the first operation of the started `Thread`), and is still correctly synchronized.

Speaking of initializing the counter, I think we can also "destroy" the counter when N stacks have been scanned.  By "destroying" I mean any subsequent increment of the counter shall result in error.  Probably we can invent our own `CountDownLatch` data structure that dictates a particular use pattern, i.e. `new()`, `set_what_to_do_when_counter_reached_limit(callback)`, `increment()`, `reset()`.  A key difference is that such `CountDownLatch` does not block a thread until the latch is released, but does something (or enables a work packet) when the count reached limit.

qinsoon: I have made changes. As with this change, the function `increment_and_check_limit()` is simple enough, and I don't think we need it to be a separate function any more. I have deleted `increment_and_check_limit()`. I put the code to `BasePlan`. You can take a look now.

wks (approved):
LGTM

PR #506, opened 2021-12-08
Add lock-free BlockPageResource for ImmixSpace

wenyuzhao:
This PR adds a lock-free block queue and uses it for Immix's clean block allocation (as `BlockPageResource`) and recyclable block allocation (as a reusable block queue).

The lock-free block queue has per-thread producer endpoints. Only GC workers can add blocks to the queue. Workers will add the blocks to their thread-local queues first, and flush them to the global pool if the queue is full.

Block allocation can only happen at block granularity. The consumers still share a single endpoint. But on block allocation fast-path, the cost is just a simple atomic increment to update the allocation cursor and pop a block.  _However, for `BlockPageResource`, considering the current design of `Space.acquire`, it still acquires a lock before calling `pr.get_new_pages`._

### Performance

_Zen 2 12/24 core:_

http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/4HVaWb

http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/JYCRMu (NEW)

Screenshot 2022-07-11 at 11 01 58 am

### TODO

- [x] Fix h2
- [x] Bug fixes
- [x] Performance on shrew (16/32 core Zen 3)


wenyuzhao:
Pushing this PR forward as this will likely solve the Immix scalability issue @clairexhuang is looking at.

src/util/heap/blockpageresource.rs
```diff
@@ -0,0 +1,417 @@
...
+pub struct BlockQueue<Block> {
+    /// First global BlockArray for fast allocation
+    head_global_freed_blocks: RwLock<Option<BlockArray<Block>>>,
+    /// A list of BlockArray that is flushed to the global pool
+    global_freed_blocks: RwLock<Vec<BlockArray<Block>>>,
+    /// Thread-local block queues
+    worker_local_freed_blocks: Vec<BlockArray<Block>>,
+    /// Total number of blocks in the whole BlockQueue
+    count: AtomicUsize,
+}
```
qinsoon, line 309: I suggest renaming this to `BlockPool`, and optionally rename `BlockArray` to `BlockQueue`. Having both `BlockArray` and `BlockQueue` is a bit confusing.

src/util/heap/blockpageresource.rs
```diff
@@ -0,0 +1,417 @@
...
+use std::fmt::Debug;
+use std::marker::PhantomData;
+use std::sync::atomic::AtomicUsize;
+use std::sync::Mutex;
+
+const UNINITIALIZED_WATER_MARK: i32 = -1;
+const LOCAL_BUFFER_SIZE: usize = 128;
+
+/// A fast PageResource for fixed-size block allocation only.
+pub struct BlockPageResource<VM: VMBinding> {
```
qinsoon, line 24: Make it clear that the current implementation only support contiguous memory. And either make it clear why discontiguous cannot be supported, or add a `TODO` for discontiguous memory.

qinsoon: I am also wondering if we can internally use `FreeListPageResource` or `MonotonePageResource` for allocating new pages. So when we allocate, we try get from the block pool first. If that fails, we get from other page resource. In that case, we do not need to worry about contiguous/discontiguous.

src/policy/immix/immixspace.rs
```diff
@@ -23,21 +23,26 @@ use crate::vm::*;
...
     MMTK,
 };
 use atomic::Ordering;
 use std::sync::{atomic::AtomicU8, Arc};
 
+// For 32-bit platforms, we still use FreeListPageResource
+#[cfg(target_pointer_width = "32")]
+use crate::util::heap::FreeListPageResource as ImmixPageResource;
+
+// BlockPageResource is for 64-bit platforms only
```
qinsoon, line 36: This does not really depend on architecture. It depends on the space being contiguous or discontiguous. However, that is not a compile-time flag. I am okay with checking pointer width, but you should make it clear that this is a hack.

Alternatively, see my comments on the `BlockPageResource`. If we can support discontiguous page resource for `BlockPageResource`, we will no longer need this conditional compilation.

qinsoon (changes requested):
I am concerned with two things:
1. Using `drop()` for executing some 'epilogue' code.
2. We may be able to implement `BlockPageResource` on top of the other page resources, so we do not need to worry about coping with contiguous/discontiguous spaces.

src/policy/immix/block.rs
```diff
@@ -275,39 +275,51 @@ impl Block {
 }
 
 /// A non-block single-linked list to store blocks.
-#[derive(Default)]
 pub struct BlockList {
```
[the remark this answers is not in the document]
wenyuzhao: I'd like to keep `BlockList` for a while. I plan to have multiple `BlockPool`s in the `BlockList`. For example two pools, mostly-empty blocks go to one pool, and mostly-full blocks go to a separate pool. I'd like to see if reusing mostly-full blocks first will improve locality.

src/util/heap/blockpageresource.rs
```diff
@@ -0,0 +1,417 @@
...
+    common: CommonPageResource,
+    /// Block granularity
+    log_pages: usize,
```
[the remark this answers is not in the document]
wenyuzhao: I think it's a bad idea to introduce more type arguments. For page-resource, page-level allocation and deallocation are already slow-path operations. I'm afraid moving `log_pages` to type arguments will not improve any performance, buy only increase code complexity.

src/scheduler/mod.rs
```diff
@@ -19,7 +19,7 @@ mod work_bucket;
 pub use work_bucket::WorkBucketStage;
 
 mod worker;
-pub use worker::GCWorker;
+pub use worker::{current_worker_ordinal, GCWorker};
```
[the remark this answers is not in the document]
wenyuzhao: Done. Changed to `pub(crate)`

src/util/heap/blockpageresource.rs
```diff
@@ -0,0 +1,417 @@
...
+    pub fn flush_all(&self) {
+        self.block_queue.flush_all()
+    }
+}
+
+/// A block list that supports fast lock-free push/pop operations
+struct BlockArray<Block> {
+    cursor: AtomicUsize,
+    data: UnsafeCell<Vec<Block>>,
+    capacity: usize,
```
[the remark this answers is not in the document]
wenyuzhao: Done

src/util/heap/blockpageresource.rs
```diff
@@ -0,0 +1,417 @@
...
+
+    /// Create an array
+    #[inline(always)]
+    fn new() -> Self {
+        let mut array = Self {
+            cursor: AtomicUsize::new(0),
+            data: UnsafeCell::new(Vec::with_capacity(Self::LOCAL_BUFFER_SIZE)),
+            capacity: Self::LOCAL_BUFFER_SIZE,
+        };
+        unsafe { array.data.get_mut().set_len(Self::LOCAL_BUFFER_SIZE) }
```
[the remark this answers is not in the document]
wenyuzhao: Fixed

src/util/heap/blockpageresource.rs
wenyuzhao: Renamed as follows:
* `BlockQueue` => `BlockPool`
* `BlockArray` => `BlockQueue`
* `BlockList` => `ReusableBlockPool`

src/util/heap/blockpageresource.rs
wenyuzhao: Switched to `FreeListPageResource`

src/policy/immix/immixspace.rs
wenyuzhao: Fixed. The conditional compilation is removed.

src/policy/immix/chunk.rs
```diff
@@ -178,6 +186,8 @@ impl Default for ChunkMap {
 struct SweepChunk<VM: VMBinding> {
     space: &'static ImmixSpace<VM>,
     chunk: Chunk,
+    /// A destructor invoked when all `SweepChunk` packets are finished.
+    _epilogue: Arc<FlushPageResource<VM>>,
```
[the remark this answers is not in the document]
wenyuzhao: Thanks. Great point. Actually, rust even never guarantees that `drop()` will always run:

> It has never been a guarantee of Rust that destructors for a type will run


https://rust-lang.github.io/rfcs/1066-safe-mem-forget.html

I'll refactor this code.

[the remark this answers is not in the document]
wenyuzhao: Fixed. I think if we're not panicking and then recovering, the `drop()` function should always be executed. But I removed `drop` anyway.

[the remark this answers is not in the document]
wenyuzhao: FYI the previously executed work packet is guaranteed to be dropped before the next `self.poll()`. Updating and opening new buckets all happen inside  `self.poll()`.  So when a new bucket opens, there should be no work packets left from the previous bucket. All of them should be dropped.

https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/worker.rs#L179-L182
```rust
loop {
    let mut work = self.poll();
    work.do_work_with_stat(self, mmtk);
}
```

src/util/heap/blockpageresource.rs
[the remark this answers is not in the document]
qinsoon: My intention was to make the code more robust. Performance was not my intention. 

Currently `BlockPool` and `BlockQueue` have a type parameter, but the type parameter does not have any bound. So you can have `BlockQueue<Address>`, but you could also have `BlockQueue<u8>`, or `BlockQueue<bool>`, which is clearly not intended. We have the trait `Region` that is suitable as type bound here. In that case, you can only have things like `BlockQueue<immix::Block>`, `BlockQueue<marksweep::Block>`, or `BlockQueue<Chunk>`.
https://github.com/mmtk/mmtk-core/blob/30771e1a1e12e3225dccec86fc330ff15e5dedc8/src/util/linear_scan.rs#L79

If we do that, `Region` also provides `LOG_BYTES`, which can be used to replace the `log_pages`.
https://github.com/mmtk/mmtk-core/blob/30771e1a1e12e3225dccec86fc330ff15e5dedc8/src/util/linear_scan.rs#L80

With these changes, we could guarantee `BlockQueue` always has a proper `Region` type, and `log_pages` matches the actual region size. Let me know if you think this makes sense.

caizixian (changes requested):
@wenyuzhao could you run the experiments again with one GC thread? @clairexhuang is seeing some results that the new implementation might be slower than the old one for some benchmarks for low thread counts.

src/util/heap/blockpageresource.rs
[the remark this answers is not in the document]
wenyuzhao: Makes sense. Fixed.

wenyuzhao:
> @wenyuzhao could you run the experiments again with one GC thread? @clairexhuang is seeing some results that the new implementation might be slower than the old one for some benchmarks for low thread counts.

Do you know which benchmarks were slowed down?

clairexhuang:
Here's the plotty: http://squirrel.anu.edu.au/plotty/claireh/barrier-performance/p/rP6fn3
I was running a normal mmtk build and a build using the blockpageresource branch on skunk (i9, 8/16 thread) and new benchmarks on 2x heap. jython, lusearch and tomcat seem to be particularly bad.

src/policy/immix/block.rs
```diff
@@ -68,6 +68,13 @@ impl BlockState {
 #[derive(Debug, Clone, Copy, PartialOrd, PartialEq)]
 pub struct Block(Address);
 
+impl Default for Block {
```
qinsoon, line 71: A block starting at address zero is not really a valid block. I am not comfortable with having an invalid default.

src/util/heap/blockpageresource.rs
```diff
@@ -0,0 +1,407 @@
...
+    worker_local_freed_blocks: Vec<BlockQueue<B>>,
+    /// Total number of blocks in the whole BlockQueue
+    count: AtomicUsize,
+}
+
+impl<B: Region> BlockPool<B> {
+    /// Create a BlockQueue
+    pub fn new(num_workers: usize) -> Self {
+        Self {
+            head_global_freed_blocks: Default::default(),
```
qinsoon, line 302: You can just use `BlockQueue::new()` here and remove the default implementations for those types.

src/util/linear_scan.rs
```diff
@@ -75,7 +76,7 @@ impl<VM: VMBinding> LinearScanObjectSize for DefaultObjectSize<VM> {
 
 /// Region represents a memory region with a properly aligned address as its start and a fixed size for the region.
 /// Region provides a set of utility methods, along with a RegionIterator that linearly scans at the step of a region.
-pub trait Region: Copy + PartialEq + PartialOrd {
+pub trait Region: Copy + PartialEq + PartialOrd + Debug {
```
qinsoon, line 79: Is `Debug` really needed? I am okay with this if it is intended. I was just reminded of this, as it changed our public API.

qinsoon (commented):
I have a few comments related to the public API changes in this PR. We should try keep our API unchanged unless it is necessary.

src/policy/immix/block.rs
wenyuzhao: I need a default block value to initialize a pre-allocated array as the thread-local block allocation/release buffer. I'll see if I can refactor this code...

src/util/address.rs
```diff
@@ -129,6 +129,14 @@ impl Shl<usize> for Address {
     }
 }
 
+/// Default constructor
+impl Default for Address {
```
qinsoon, line 133: The same issue here as the default zero block for `Block`. I suggest we remove them.

src/util/linear_scan.rs
wenyuzhao: fixed

src/util/heap/blockpageresource.rs
wenyuzhao: fixed

src/policy/immix/block.rs
wenyuzhao: fixed

src/util/address.rs
wenyuzhao: removed

wenyuzhao:
> I have a few comments related to the public API changes in this PR. We should try keep our API unchanged unless it is necessary.

The API check CI is now passed. No public API changes introduced.

qinsoon (approved):
LGTM

caizixian (approved):

Issue #185, opened 2020-12-11
Work stealing


wenyuzhao:
We need a work stealing implementation (like the one Java MMTk has) to improve load balancing for the work-packets system.

steveblackburn:
To be clear, the Java MMTk does not support work stealing.   But yes, we absolutely want support for work stealing.

https://conf.researchr.org/details/ismm-2019/ismm-2019-papers/8/Scaling-Up-Parallel-GC-Work-Stealing-in-Many-Core-Environments

PR #507, opened 2021-12-10
Work stealing

wenyuzhao:
This PR adds work-stealing support to the work-packet system.

Performance: http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/j2sUrH

Screen Shot 2022-05-23 at 2 46 49 pm

**TODO**:
- [x] Fix h2
- [x] Fix pmd

### DIFF
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -47,24 +47,23 @@ unsafe impl<VM: VMBinding> Sync for GCWorkScheduler<VM> {}
 impl<VM: VMBinding> GCWorkScheduler<VM> {
     pub fn new(num_workers: usize) -> Arc<Self> {
         let worker_monitor: Arc<(Mutex<()>, Condvar)> = Default::default();
-        let worker_group = WorkerGroup::new(num_workers);
 
         // Create work buckets for workers.
         let mut work_buckets = enum_map! {
-            WorkBucketStage::Unconstrained => WorkBucket::new(true, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::Prepare => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::Closure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::SoftRefClosure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::WeakRefClosure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::FinalRefClosure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::PhantomRefClosure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::CalculateForwarding => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::SecondRoots => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::RefForwarding => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::FinalizableForwarding => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::Compact => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::Release => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
-            WorkBucketStage::Final => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::Unconstrained => WorkBucket::new(true, worker_monitor.clone()),
+            WorkBucketStage::Prepare => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::Closure => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::SoftRefClosure => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::WeakRefClosure => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::FinalRefClosure => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::PhantomRefClosure => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::CalculateForwarding => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::SecondRoots => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::RefForwarding => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::FinalizableForwarding => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::Compact => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::Release => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::Final => WorkBucket::new(false, worker_monitor.clone()),
         };
 
         // Set the open condition of each bucket.
@@ -104,6 +103,11 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         }
 
         let coordinator_worker_shared = Arc::new(GCWorkerShared::<VM>::new());
+
+        let worker_group = WorkerGroup::new(num_workers);
+        work_buckets.values_mut().for_each(|bucket| {
+            bucket.set_group(worker_group.clone());
+        });
 
         Arc::new(Self {
             work_buckets,
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -399,7 +399,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                 worker.sender.send(CoordinatorMessage::Finish).unwrap();
             }
             // Wait
-            let _ = self.worker_monitor.1.wait(guard).unwrap();
+            let _guard = self.worker_monitor.1.wait(guard).unwrap();
             // Unpark this worker
             self.worker_group.dec_parked_workers();
         }
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -10,6 +10,7 @@ use crossbeam::deque::{self, Steal};
 use enum_map::Enum;
 use enum_map::{enum_map, EnumMap};
 use std::collections::HashMap;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::mpsc::channel;
 use std::sync::{Arc, Condvar, Mutex};
 
@@ -41,6 +42,10 @@ pub struct GCWorkScheduler<VM: VMBinding> {
     /// the `Closure` bucket multiple times to iteratively discover and process
     /// more ephemeron objects.
     closure_end: Mutex<Option<Box<dyn Send + Fn() -> bool>>>,
+    /// Counter for messages sent to the controller.
+    pub(super) issued_messages: AtomicUsize,
+    /// Counter for messages completed by the controller.
+    pub(super) completed_messages: AtomicUsize,
 }
 
 // FIXME: GCWorkScheduler should be naturally Sync, but we cannot remove this `impl` yet.
@@ -112,6 +117,8 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             coordinator_worker_shared,
             worker_monitor,
             closure_end: Mutex::new(None),
+            issued_messages: AtomicUsize::new(0),
+            completed_messages: AtomicUsize::new(0),
         })
     }
 
@@ -221,6 +228,8 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
 
     fn are_buckets_drained(&self, buckets: &[WorkBucketStage]) -> bool {
         buckets.iter().all(|&b| self.work_buckets[b].is_drained())
+            && self.issued_messages.load(Ordering::Relaxed)
+                == self.completed_messages.load(Ordering::Relaxed)
     }
 
     pub fn on_closure_end(&self, f: Box<dyn Send + Fn() -> bool>) {
@@ -287,6 +296,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
     }
 
     pub fn add_coordinator_work(&self, work: impl CoordinatorWork<VM>, worker: &GCWorker<VM>) {
+        self.issued_messages.fetch_add(1, Ordering::SeqCst);
         worker
             .sender
             .send(CoordinatorMessage::Work(Box::new(work)))
@@ -400,6 +410,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                 }
                 debug_assert!(!self.worker_group.has_designated_work());
                 // The current pause is finished if we can't open more buckets.
+                self.issued_messages.fetch_add(1, Ordering::SeqCst);
                 worker.sender.send(CoordinatorMessage::Finish).unwrap();
             }
             // Wait
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -364,8 +364,8 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
 
     #[cold]
     fn poll_slow(&self, worker: &GCWorker<VM>) -> Box<dyn GCWork<VM>> {
-        let mut guard = self.worker_monitor.0.lock().unwrap();
         loop {
+            let guard = self.worker_monitor.0.lock().unwrap();
             // Retry polling
             if let Some(work) = self.poll_schedulable_work(worker) {
                 return work;
@@ -374,6 +374,12 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             let all_parked = self.worker_group.inc_parked_workers();
             // If all workers are parked, try activate new buckets
             if all_parked {
+                // If there're any designated work, resume the workers and process them.
+                if self.worker_group.has_designated_work() {
+                    self.worker_group.dec_parked_workers();
+                    self.worker_monitor.1.notify_all();
+                    continue;
+                }
                 if self.update_buckets() {
                     // We're not going to sleep since a new bucket is just open.
                     self.worker_group.dec_parked_workers();
@@ -388,11 +394,12 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                     // Return this packet and execute it.
                     return work;
                 }
+                debug_assert!(!self.worker_group.has_designated_work());
                 // The current pause is finished if we can't open more buckets.
                 worker.sender.send(CoordinatorMessage::Finish).unwrap();
             }
             // Wait
-            guard = self.worker_monitor.1.wait(guard).unwrap();
+            let _ = self.worker_monitor.1.wait(guard).unwrap();
             // Unpark this worker
             self.worker_group.dec_parked_workers();
         }
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -9,12 +9,15 @@ use crate::vm::{GCThreadContext, VMBinding};
 use crossbeam::deque::Steal;
 use enum_map::{enum_map, EnumMap};
 use std::collections::HashMap;
-use std::sync::atomic::Ordering;
 use std::sync::mpsc::channel;
 use std::sync::{Arc, Condvar, Mutex};
 
 pub enum CoordinatorMessage<VM: VMBinding> {
+    /// Send a work-packet to the coordinator thread/
     Work(Box<dyn CoordinatorWork<VM>>),
+    /// Notify the coordinator thread that all GC tasks are finished.
+    /// When sending this message, all the work buckets should be
+    /// empty, and all the workers should be parked.
     Finish,
 }
 
@@ -96,9 +99,9 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                 open_stages.push(s);
             };
 
-            for stages in stages {
-                if stages != WorkBucketStage::Unconstrained && stages != first_stw_stage {
-                    open_next(stages);
+            for stage in stages {
+                if stage != WorkBucketStage::Unconstrained && stage != first_stw_stage {
+                    open_next(stage);
                 }
             }
         }
@@ -127,7 +130,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         // Spawn the controller thread.
         let coordinator_worker = GCWorker::new(
             mmtk,
-            0,
+            usize::MAX,
             self.clone(),
             true,
             sender.clone(),
@@ -360,10 +363,8 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
 
     #[cold]
     fn poll_slow(&self, worker: &GCWorker<VM>) -> Box<dyn GCWork<VM>> {
-        debug_assert!(!worker.shared.is_parked());
         let mut guard = self.worker_monitor.0.lock().unwrap();
         loop {
-            debug_assert!(!worker.shared.is_parked());
             // Retry polling
             if let Some(work) = self.poll_schedulable_work(worker) {
                 return work;
@@ -378,7 +379,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                     // We guarantee that we can at least fetch one packet.
                     let work = self.poll_schedulable_work(worker).unwrap();
                     // Optimize for the case that a newly opened bucket only has one packet.
-                    // We only notify_all if there're motr than one packets available.
+                    // We only notify_all if there're more than one packets available.
                     if !self.all_activated_buckets_are_empty() {
                         // Have more jobs in this buckets. Notify other workers.
                         self.worker_monitor.1.notify_all();
@@ -386,14 +387,13 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                     // Return this packet and execute it.
                     return work;
                 }
-                // The current pause is finished if if we can't open more buckets.
+                // The current pause is finished if we can't open more buckets.
                 worker.sender.send(CoordinatorMessage::Finish).unwrap();
             }
             // Wait
             guard = self.worker_monitor.1.wait(guard).unwrap();
             // Unpark this worker
             self.worker_group.dec_parked_workers();
-            worker.shared.parked.store(false, Ordering::Relaxed);
         }
     }
 
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -247,9 +247,13 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             }
             let bucket = &self.work_buckets[id];
             let bucket_opened = bucket.update(self);
-            buckets_updated |= bucket_opened;
+            buckets_updated = buckets_updated || bucket_opened;
             if bucket_opened {
-                new_packets |= !bucket.is_drained();
+                new_packets = new_packets || !bucket.is_drained();
+                // Quit the loop. There'are already new packets in the newly opened buckets.
+                if new_packets {
+                    break;
+                }
             }
         }
         buckets_updated && new_packets
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -7,6 +7,7 @@ use crate::util::opaque_pointer::*;
 use crate::vm::Collection;
 use crate::vm::{GCThreadContext, VMBinding};
 use crossbeam::deque::{self, Steal};
+use enum_map::Enum;
 use enum_map::{enum_map, EnumMap};
 use std::collections::HashMap;
 use std::sync::mpsc::channel;
@@ -74,34 +75,31 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         {
             // Unconstrained is always open. Prepare will be opened at the beginning of a GC.
             // This vec will grow for each stage we call with open_next()
-            let first_stw_stage = work_buckets.iter().nth(1).map(|(id, _)| id).unwrap();
+            let first_stw_stage = WorkBucketStage::first_stw_stage();
             let mut open_stages: Vec<WorkBucketStage> = vec![first_stw_stage];
             // The rest will open after the previous stage is done.
-            let stages = work_buckets
-                .iter()
-                .map(|(stage, _)| stage)
-                .collect::<Vec<_>>();
-            let mut open_next = |s: WorkBucketStage| {
-                let cur_stages = open_stages.clone();
-                work_buckets[s].set_open_condition(move |scheduler: &GCWorkScheduler<VM>| {
-                    let should_open = scheduler.are_buckets_drained(&cur_stages);
-                    // Additional check before the `RefClosure` bucket opens.
-                    if should_open && s == crate::scheduler::work_bucket::LAST_CLOSURE_BUCKET {
-                        if let Some(closure_end) = scheduler.closure_end.lock().unwrap().as_ref() {
-                            if closure_end() {
-                                // Don't open `LAST_CLOSURE_BUCKET` if `closure_end` added more works to `Closure`.
-                                return false;
-                            }
-                        }
-                    }
-                    should_open
-                });
-                open_stages.push(s);
-            };
-
+            let stages = (0..WorkBucketStage::LENGTH).map(WorkBucketStage::from_usize);
             for stage in stages {
                 if stage != WorkBucketStage::Unconstrained && stage != first_stw_stage {
-                    open_next(stage);
+                    let cur_stages = open_stages.clone();
+                    work_buckets[stage].set_open_condition(
+                        move |scheduler: &GCWorkScheduler<VM>| {
+                            let should_open = scheduler.are_buckets_drained(&cur_stages);
+                            // Additional check before the `RefClosure` bucket opens.
+                            if should_open && stage == LAST_CLOSURE_BUCKET {
+                                if let Some(closure_end) =
+                                    scheduler.closure_end.lock().unwrap().as_ref()
+                                {
+                                    if closure_end() {
+                                        // Don't open `LAST_CLOSURE_BUCKET` if `closure_end` added more works to `Closure`.
+                                        return false;
+                                    }
+                                }
+                            }
+                            should_open
+                        },
+                    );
+                    open_stages.push(stage);
                 }
             }
         }
@@ -242,10 +240,12 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
     fn update_buckets(&self) -> bool {
         let mut buckets_updated = false;
         let mut new_packets = false;
-        for (id, bucket) in self.work_buckets.iter() {
+        for i in 0..WorkBucketStage::LENGTH {
+            let id = WorkBucketStage::from_usize(i);
             if id == WorkBucketStage::Unconstrained {
                 continue;
             }
+            let bucket = &self.work_buckets[id];
             let bucket_opened = bucket.update(self);
             buckets_updated |= bucket_opened;
             if bucket_opened {
@@ -264,7 +264,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
     }
 
     pub fn reset_state(&self) {
-        let first_stw_stage = self.work_buckets.iter().nth(1).map(|(id, _)| id).unwrap();
+        let first_stw_stage = WorkBucketStage::first_stw_stage();
         self.work_buckets.iter().for_each(|(id, bkt)| {
             if id != WorkBucketStage::Unconstrained && id != first_stw_stage {
                 bkt.deactivate();
@@ -420,7 +420,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
 
     pub fn notify_mutators_paused(&self, mmtk: &'static MMTK<VM>) {
         mmtk.plan.base().gc_requester.clear_request();
-        let first_stw_bucket = self.work_buckets.values().nth(1).unwrap();
+        let first_stw_bucket = &self.work_buckets[WorkBucketStage::first_stw_stage()];
         debug_assert!(!first_stw_bucket.is_activated());
         first_stw_bucket.activate();
         let _guard = self.worker_monitor.0.lock().unwrap();
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -374,7 +374,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             let all_parked = self.worker_group.inc_parked_workers();
             // If all workers are parked, try activate new buckets
             if all_parked {
-                // If there're any designated work, resume the workers and process them.
+                // If there're any designated work, resume the workers and process them
                 if self.worker_group.has_designated_work() {
                     self.worker_group.dec_parked_workers();
                     self.worker_monitor.1.notify_all();
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -21,7 +21,7 @@ pub enum CoordinatorMessage<VM: VMBinding> {
 pub struct GCWorkScheduler<VM: VMBinding> {
     /// Work buckets
     pub work_buckets: EnumMap<WorkBucketStage, WorkBucket<VM>>,
-    /// workers
+    /// Workers
     pub worker_group: Arc<WorkerGroup<VM>>,
     /// The shared part of the GC worker object of the controller thread
     coordinator_worker_shared: Arc<GCWorkerShared<VM>>,
@@ -233,7 +233,12 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         self.work_buckets.values().all(|bucket| bucket.is_empty())
     }
 
-    /// Open buckets if their conditions are met
+    /// Open buckets if their conditions are met.
+    ///
+    /// This function should only be called after all the workers are parked.
+    /// No workers will be waked up by this function. The caller is responsible for that.
+    ///
+    /// Return true if there're any non-empty buckets updated.
     fn update_buckets(&self) -> bool {
         let mut buckets_updated = false;
         let mut new_packets = false;
@@ -284,6 +289,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             .unwrap();
     }
 
+    /// Check if all the work buckets are empty
     #[inline(always)]
     fn all_activated_buckets_are_empty(&self) -> bool {
         for bucket in self.work_buckets.values() {
@@ -294,38 +300,44 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         true
     }
 
+    /// Get a schedulable work packet without retry.
     #[inline(always)]
     fn pop_schedulable_work_once(&self, worker: &GCWorker<VM>) -> Steal<Box<dyn GCWork<VM>>> {
-        let mut retry = false;
+        let mut should_retry = false;
+        // Try find a packet that can be processed only by this worker.
         if let Some(w) = worker.shared.local_work.pop() {
             return Steal::Success(w);
         }
+        // Try get a packet from a work bucket.
         for work_bucket in self.work_buckets.values() {
             match work_bucket.poll(&worker.shared.local_work_buffer) {
                 Steal::Success(w) => return Steal::Success(w),
-                Steal::Retry => retry = true,
+                Steal::Retry => should_retry = true,
                 _ => {}
             }
         }
+        // Try steal some packets from any worker
         for (id, stealer) in self.worker_group.stealers.iter().enumerate() {
             if id == worker.ordinal {
                 continue;
             }
             match stealer.steal() {
                 Steal::Success(w) => return Steal::Success(w),
-                Steal::Retry => retry = true,
+                Steal::Retry => should_retry = true,
                 _ => {}
             }
         }
-        if retry {
+        if should_retry {
             Steal::Retry
         } else {
             Steal::Empty
         }
     }
 
+    /// Get a schedulable work packet.
     #[inline]
     fn pop_schedulable_work(&self, worker: &GCWorker<VM>) -> Option<Box<dyn GCWork<VM>>> {
+        // Loop until we successfully get a packet.
         loop {
             match self.pop_schedulable_work_once(worker) {
                 Steal::Success(w) => {
@@ -342,7 +354,8 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         }
     }
 
-    /// Get a schedulable work. Called by workers
+    /// Called by workers to get a schedulable work packet.
+    /// Park the worker if there're no available packets.
     #[inline]
     pub fn poll(&self, worker: &GCWorker<VM>) -> Box<dyn GCWork<VM>> {
         self.pop_schedulable_work(worker)
@@ -358,20 +371,24 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             if let Some(work) = self.pop_schedulable_work(worker) {
                 return work;
             }
-            // Park this worker
+            // Prepare to park this worker
             let all_parked = self.worker_group.inc_parked_workers();
+            // If all workers are parked, try activate new buckets
             if all_parked {
                 if self.update_buckets() {
                     self.worker_group.dec_parked_workers();
                     // We guarantee that we can at least fetch one packet.
                     let work = self.pop_schedulable_work(worker).unwrap();
                     // Optimize for the case that a newly opened bucket only has one packet.
+                    // We only notify_all if there're motr than one packets available.
                     if !self.all_activated_buckets_are_empty() {
                         // Have more jobs in this buckets. Notify other workers.
                         self.worker_monitor.1.notify_all();
                     }
+                    // Return this packet and execute it.
                     return work;
                 }
+                // The current pause is finished if if we can't open more buckets.
                 worker.sender.send(CoordinatorMessage::Finish).unwrap();
             }
             // Wait
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -47,23 +47,24 @@ unsafe impl<VM: VMBinding> Sync for GCWorkScheduler<VM> {}
 impl<VM: VMBinding> GCWorkScheduler<VM> {
     pub fn new(num_workers: usize) -> Arc<Self> {
         let worker_monitor: Arc<(Mutex<()>, Condvar)> = Default::default();
+        let worker_group = WorkerGroup::new(num_workers);
 
         // Create work buckets for workers.
         let mut work_buckets = enum_map! {
-            WorkBucketStage::Unconstrained => WorkBucket::new(true, worker_monitor.clone()),
-            WorkBucketStage::Prepare => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::Closure => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::SoftRefClosure => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::WeakRefClosure => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::FinalRefClosure => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::PhantomRefClosure => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::CalculateForwarding => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::SecondRoots => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::RefForwarding => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::FinalizableForwarding => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::Compact => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::Release => WorkBucket::new(false, worker_monitor.clone()),
-            WorkBucketStage::Final => WorkBucket::new(false, worker_monitor.clone()),
+            WorkBucketStage::Unconstrained => WorkBucket::new(true, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::Prepare => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::Closure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::SoftRefClosure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::WeakRefClosure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::FinalRefClosure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::PhantomRefClosure => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::CalculateForwarding => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::SecondRoots => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::RefForwarding => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::FinalizableForwarding => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::Compact => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::Release => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
+            WorkBucketStage::Final => WorkBucket::new(false, worker_monitor.clone(), worker_group.clone()),
         };
 
         // Set the open condition of each bucket.
@@ -103,11 +104,6 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         }
 
         let coordinator_worker_shared = Arc::new(GCWorkerShared::<VM>::new());
-
-        let worker_group = WorkerGroup::new(num_workers);
-        work_buckets.values_mut().for_each(|bucket| {
-            bucket.set_group(worker_group.clone());
-        });
 
         Arc::new(Self {
             work_buckets,
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -225,7 +225,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
 
     fn are_buckets_drained(&self, buckets: &[WorkBucketStage]) -> bool {
         buckets.iter().all(|&b| self.work_buckets[b].is_drained())
-            && self.pending_messages.load(Ordering::Relaxed) == 0
+            && self.pending_messages.load(Ordering::SeqCst) == 0
     }
 
     pub fn on_closure_end(&self, f: Box<dyn Send + Fn() -> bool>) {
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -6,7 +6,7 @@ use crate::mmtk::MMTK;
 use crate::util::opaque_pointer::*;
 use crate::vm::Collection;
 use crate::vm::{GCThreadContext, VMBinding};
-use crossbeam::deque::Steal;
+use crossbeam::deque::{self, Steal};
 use enum_map::{enum_map, EnumMap};
 use std::collections::HashMap;
 use std::sync::mpsc::channel;
@@ -106,7 +106,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             }
         }
 
-        let coordinator_worker_shared = Arc::new(GCWorkerShared::<VM>::new());
+        let coordinator_worker_shared = Arc::new(GCWorkerShared::<VM>::new(None));
 
         Arc::new(Self {
             work_buckets,
@@ -135,6 +135,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             true,
             sender.clone(),
             self.coordinator_worker_shared.clone(),
+            deque::Worker::new_fifo(),
         );
         let gc_controller = GCController::new(
             mmtk,
@@ -304,23 +305,23 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
     fn poll_schedulable_work_once(&self, worker: &GCWorker<VM>) -> Steal<Box<dyn GCWork<VM>>> {
         let mut should_retry = false;
         // Try find a packet that can be processed only by this worker.
-        if let Some(w) = worker.shared.local_work.pop() {
+        if let Some(w) = worker.shared.designated_work.pop() {
             return Steal::Success(w);
         }
         // Try get a packet from a work bucket.
         for work_bucket in self.work_buckets.values() {
-            match work_bucket.poll(&worker.shared.local_work_buffer) {
+            match work_bucket.poll(&worker.local_work_buffer) {
                 Steal::Success(w) => return Steal::Success(w),
                 Steal::Retry => should_retry = true,
                 _ => {}
             }
         }
         // Try steal some packets from any worker
-        for (id, stealer) in self.worker_group.stealers.iter().enumerate() {
+        for (id, worker_shared) in self.worker_group.workers_shared.iter().enumerate() {
             if id == worker.ordinal {
                 continue;
             }
-            match stealer.steal() {
+            match worker_shared.stealer.as_ref().unwrap().steal() {
                 Steal::Success(w) => return Steal::Success(w),
                 Steal::Retry => should_retry = true,
                 _ => {}
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -1,12 +1,12 @@
 use super::stat::SchedulerStat;
-use super::work_bucket::WorkBucketStage::*;
 use super::work_bucket::*;
-use super::worker::{GCWorker, GCWorkerShared};
+use super::worker::{GCWorker, GCWorkerShared, WorkerGroup};
 use super::*;
 use crate::mmtk::MMTK;
 use crate::util::opaque_pointer::*;
 use crate::vm::Collection;
 use crate::vm::{GCThreadContext, VMBinding};
+use crossbeam::deque::Steal;
 use enum_map::{enum_map, EnumMap};
 use std::collections::HashMap;
 use std::sync::atomic::Ordering;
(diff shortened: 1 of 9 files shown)


qinsoon:
Related issue: https://github.com/mmtk/mmtk-core/issues/185

src/scheduler/scheduler.rs
```diff
@@ -51,41 +47,46 @@ unsafe impl<VM: VMBinding> Sync for GCWorkScheduler<VM> {}
...
             // Unconstrained is always open. Prepare will be opened at the beginning of a GC.
             // This vec will grow for each stage we call with open_next()
-            let mut open_stages: Vec<WorkBucketStage> = vec![Unconstrained, Prepare];
+            let first_stw_stage = work_buckets.iter().nth(1).map(|(id, _)| id).unwrap();
+            let mut open_stages: Vec<WorkBucketStage> = vec![first_stw_stage];
             // The rest will open after the previous stage is done.
+            let stages = work_buckets
+                .iter()
+                .map(|(stage, _)| stage)
+                .collect::<Vec<_>>();
```
wks, line 80: You may use static methods from the `Enum<T>` type in enum-map 0.6.2:
```rust
let first_stw_stage: WorkBucketStage = Enum::<WorkBucketStage>::from_usize(1);

let possible_values: usize = <WorkBucketStage as Enum::<WorkBucketStage>>::POSSIBLE_VALUES;
let stages = (0..possible_values)                                                                   
        .map(|i| {                                                                                      
            let stage: WorkBucketStage = Enum::<WorkBucketStage>::from_usize(i);                        
            stage                                                                                       
        }).collect::<Vec<_>>();                                                                         
```

It is a bit awkward because the `enum-map` crate which mmtk-core currently depends on is **way too old** (0.6.2 vs the latest 2.2.0).

If we update `Cargo.toml` and bump the version to `enum-map = "=2.1.0"`, we will be able to do them much more elegantly:

```rust
let first_stw_stage = WorkBucketStage::from_usize(1);
let stages = (0..WorkBucketStage::LENGTH).map(WorkBucketStage::from_usize);
```

If you feel convenient, you can update the dependency for us in this PR, too.  Note that `2.2.0` requires rustc 1.60.0 which our moma machines currently don't have.  As a workaround, `enum-map = "=2.1.0"` will lock the version to exactly 2.1.0.  We should let our administrator update our installations when appropriate.

wks (commented):
The enum-map crate provides static methods to look up `Enum` items without concrete instances.  You may consider bumping the dependency version of enum-map, too, if that's convenient.

src/scheduler/worker.rs
```diff
@@ -19,16 +19,25 @@ pub struct GCWorkerShared<VM: VMBinding> {
     pub parked: AtomicBool,
     /// Worker-local statistics data.
     stat: AtomicRefCell<WorkerLocalStat<VM>>,
-    /// Incoming work packets to be executed by the current worker.
-    pub local_work_bucket: WorkBucket<VM>,
+    /// A queue of GCWork that can only be processed by the owned thread.
+    pub local_work: ArrayQueue<Box<dyn GCWork<VM>>>,
+    /// Local work packet queue.
+    pub local_work_buffer: Worker<Box<dyn GCWork<VM>>>,
```
qinsoon, line 25: This is not actually shared. We don't expect any other thread to access this. The `Stealer` could be shared so that other threads can steal, but the `Worker` should not be shared.

src/scheduler/worker.rs
```diff
@@ -140,22 +154,124 @@ impl<VM: VMBinding> GCWorker<VM> {
...
+    }
+
+    /// Increase the packed-workers counter.
+    /// Called before a worker is parked.
+    ///
+    /// Return true if all the workers are parked.
+    #[inline(always)]
+    pub fn inc_parked_workers(&self) -> bool {
+        let old = self.parked_workers.fetch_add(1, Ordering::Relaxed);
+        old + 1 == self.worker_count()
```
qinsoon, line 256: I suggest adding assertions to the inc/dec methods to make sure `parked_workers` is always in the range `[0, worker_count())`.

qinsoon (changes requested):

src/scheduler/worker.rs
```diff
@@ -4,13 +4,13 @@ use super::*;
 use crate::mmtk::MMTK;
 use crate::util::copy::GCWorkerCopyContext;
 use crate::util::opaque_pointer::*;
-use crate::vm::VMBinding;
+use crate::vm::{Collection, GCThreadContext, VMBinding};
 use atomic_refcell::{AtomicRef, AtomicRefCell, AtomicRefMut};
-use std::sync::atomic::{AtomicBool, Ordering};
+use crossbeam::deque::{Stealer, Worker};
```
wks, line 9: I find the presence of the `Worker` type a bit confusing in the context where we frequently deal with GC workers.  Can we rename it when importing, like `{Worker as CBWorker}` or `{Worker as CrossBeamWorker}`?  Alternatively we can always qualify it as `deque::Worker` when we use it.

src/scheduler/scheduler.rs
```diff
@@ -51,41 +47,46 @@ unsafe impl<VM: VMBinding> Sync for GCWorkScheduler<VM> {}
...
 
         // Set the open condition of each bucket.
         {
             // Unconstrained is always open. Prepare will be opened at the beginning of a GC.
             // This vec will grow for each stage we call with open_next()
-            let mut open_stages: Vec<WorkBucketStage> = vec![Unconstrained, Prepare];
+            let first_stw_stage = work_buckets.iter().nth(1).map(|(id, _)| id).unwrap();
+            let mut open_stages: Vec<WorkBucketStage> = vec![first_stw_stage];
             // The rest will open after the previous stage is done.
+            let stages = work_buckets
```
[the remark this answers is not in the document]
wks: The `enum-map` crate provides static methods to map between `usize` and the `enum` type.  See my comment below.

On the documentation of the derive macro `Enum`, the example code even asserts that the elements are mapped to 0, 1, 2, ... in order, although it never explicitly states so. https://docs.rs/enum-map/latest/enum_map/derive.Enum.html#enums-without-payload

To be safe, we may list the stages by ourselves for now, but I'll open an issue and ask the `enum-map` developers to make the guarantee in its docs.

wks (changes requested):

src/scheduler/scheduler.rs
caizixian: @wks @qinsoon I think we should leave the code as is for now. I won't be physically in Canberra so I will hold off image updates of CI machines until I get back unless it's urgent (i.e. security vulnerability).

I think this also raises a meta-question of what's our MSRV policy. Currently, our MSRV is 1.59.0, which is not that old. Any change to the MSRV could potentially affect any other projects that depend on us. Of course, we can make the decision that we aggressively update MSRV. But that's something we should at least discuss as a group.

k-sareen: Depending on way too old stuff is kinda bad, in my opinion. If we can even just bump to `v2.1.0` without disrupting the MSRV version, I think we should do that. (At some point we should do an audit that updates our dependencies)

src/scheduler/scheduler.rs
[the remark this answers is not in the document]
qinsoon: > FYI: https://gitlab.com/KonradBorowski/enum-map/-/issues/25

Got it. Thanks. However, the mapping between enum and usize is just part of the issue. The problem also includes the order of how `EnumMap` iterates its keys/values. The code in the PR assumes the first element in `work_buckets.iter()` and `work_buckets.values()` is the first stw phase. Generally, a map is an unordered collection, and the order to iterate a map is implementation specific. It just happens to work for enum map for its current implementation.

src/scheduler/scheduler.rs
```diff
@@ -350,32 +364,41 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
+                }
+                // The current pause is finished if if we can't open more buckets.
+                worker.sender.send(CoordinatorMessage::Finish).unwrap();
             }
             // Wait
             guard = self.worker_monitor.1.wait(guard).unwrap();
             // Unpark this worker
-            worker.shared.parked.store(false, Ordering::SeqCst);
+            self.worker_group.dec_parked_workers();
+            worker.shared.parked.store(false, Ordering::Relaxed);
```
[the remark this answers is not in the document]
wenyuzhao: Yes, this field is not used anymore. Removed.

src/scheduler/work_bucket.rs
```diff
@@ -1,147 +1,200 @@
...
+        }
     }
 }
 
 pub struct WorkBucket<VM: VMBinding> {
     active: AtomicBool,
-    /// A priority queue
-    queue: RwLock<BinaryHeap<PrioritizedWork<VM>>>,
+    queue: BucketQueue<VM>,
+    prioritized_queue: Option<BucketQueue<VM>>,
```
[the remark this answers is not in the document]
wenyuzhao: LXR requires the unconstrained bucket to have this prioritized_queue enabled. Yes, this is disabled for all the buckets at the moment, but it is easy to enable it, by just setting the field to a `Some(_)`. I'll add some comments explaining this and why the panic is expected if the prioritized_queue is not enabled.

Having `Option` is also necessary to the performance. We have 14 work buckets now. If we double the number of queues for each bucket, when a worker tries to get a work packet, it has to poll from 28 different queues, which doubles the cost of polling.  I have experimented with this and have seen a slow down in `time.stw` if I remove the `Option`.  This also indicates that we should try to reduce the number of buckets in the future. Currently, the number of buckets is no less than the number of synchronized phases in Java MMTk.

src/scheduler/worker.rs
```diff
@@ -140,22 +154,124 @@ impl<VM: VMBinding> GCWorker<VM> {
...
         work.do_work(self, self.mmtk);
     }
 
+    /// Entry of the worker thread.
+    /// Each worker will keep polling and executing work packets in a loop.
     pub fn run(&mut self, tls: VMWorkerThread, mmtk: &'static MMTK<VM>) {
         self.tls = tls;
         self.copy = crate::plan::create_gc_worker_context(tls, mmtk);
-        self.shared.parked.store(false, Ordering::SeqCst);
+        self.shared.parked.store(false, Ordering::Relaxed);
```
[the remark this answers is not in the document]
wenyuzhao: The field is removed

src/scheduler/worker.rs
wenyuzhao: Agreed. This is qualified to `deque::Worker`.

src/scheduler/work_bucket.rs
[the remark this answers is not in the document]
qinsoon: > LXR requires the unconstrained bucket to have this prioritized_queue enabled. Yes, this is disabled for all the buckets at the moment, but it is easy to enable it, by just setting the field to a `Some(_)`. I'll add some comments explaining this and why the panic is expected if the prioritized_queue is not enabled.
> 
> Having `Option` is also necessary to the performance. We have 14 work buckets now. If we double the number of queues for each bucket, when a worker tries to get a work packet, it has to poll from 28 different queues, which doubles the cost of polling. I have experimented with this and have seen a slow down in `time.stw` if I remove the `Option`. This also indicates that we should try to reduce the number of buckets in the future. Currently, the number of buckets is no less than the number of synchronized phases in Java MMTk.

Okay I see. 

Did you mean checking `Option` is cheaper than checking `queue.is_empty()`?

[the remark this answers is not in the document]
wenyuzhao: > Did you mean checking Option is cheaper than checking queue.is_empty()?

Generally yes. Either `queue.is_empty()` or direct `queue.pop()` has to do at least two `SeqCst` loads or fences, which is quite expensive.

src/scheduler/worker.rs
wenyuzhao: Yes. `local_work_buffer` is moved to `GCWorker`.

src/scheduler/worker.rs
```diff
@@ -140,22 +154,124 @@ impl<VM: VMBinding> GCWorker<VM> {
...
         }
     }
 }
+
+/// A worker group to manage all the GC workers (except the coordinator worker).
+pub struct WorkerGroup<VM: VMBinding> {
+    /// Shared worker data
+    pub workers_shared: Vec<Arc<GCWorkerShared<VM>>>,
+    /// Handles for stealing packets from workers
+    pub stealers: Vec<Stealer<Box<dyn GCWork<VM>>>>,
```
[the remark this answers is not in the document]
wenyuzhao: Done. Yes, it makes more sense to move to `GCWorkerShared `.

src/scheduler/scheduler.rs
wenyuzhao: I switched to `enum-map = "=2.1.0"` and removed the code that fetches the first stage by using `.iter()`.

src/scheduler/scheduler.rs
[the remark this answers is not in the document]
wenyuzhao: I switched to enum-map = "=2.1.0" and removed the ordering assumption.

src/scheduler/scheduler.rs
```diff
@@ -314,34 +285,77 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
-                    .unwrap();
+    fn poll_schedulable_work(&self, worker: &GCWorker<VM>) -> Option<Box<dyn GCWork<VM>>> {
+        // Loop until we successfully get a packet.
+        loop {
+            match self.poll_schedulable_work_once(worker) {
+                Steal::Success(w) => {
+                    return Some(w);
+                }
+                Steal::Retry => {
+                    std::thread::yield_now();
```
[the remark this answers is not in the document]
wenyuzhao: I will do a benchmark to compare this

src/scheduler/scheduler.rs
```diff
@@ -95,36 +96,18 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
-            open_next(CalculateForwarding);
-            open_next(SecondRoots);
-            open_next(RefForwarding);
-            open_next(FinalizableForwarding);
-            open_next(Compact);
-            open_next(Release);
-            open_next(Final);
+            for stages in stages {
+                if stages != WorkBucketStage::Unconstrained && stages != first_stw_stage {
+                    open_next(stages);
```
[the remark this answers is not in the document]
wenyuzhao: Done

src/scheduler/scheduler.rs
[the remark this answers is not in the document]
wenyuzhao: > I will do a benchmark to compare this

The difference is minor (< 1%). Generally looks like using `yield_now` is a little bit better. I'll leave it as is.

http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/mJ4JVC

src/scheduler/work_bucket.rs
```diff
@@ -1,147 +1,200 @@
...
+        debug_assert!(self.queue.is_empty(), "Bucket not drained before close");
+        self.active.store(false, Ordering::Relaxed);
+    }
+
+    /// Add a work packet to this bucket
+    /// Panic if this bucket cannot receive prioritized packets.
+    #[inline(always)]
+    pub fn add_prioritized(&self, work: Box<dyn GCWork<VM>>) {
+        self.prioritized_queue.as_ref().unwrap().push(work);
+        self.notify_one_worker();
```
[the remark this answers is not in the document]
wenyuzhao: I thought about this a little bit more. Looks like option 1 above is a necessary step before implementing the design proposed at https://github.com/mmtk/mmtk-core/issues/546. It builds a bottleneck for adding work packets, and the design in https://github.com/mmtk/mmtk-core/issues/546 can utilize it to either redirect each packet to the main queue or temporarily store it in a bucket.

I will implement option 1 as a separate PR (this will likely to introduce many changes in mmtk-core and bindings), and have a try on https://github.com/mmtk/mmtk-core/issues/546.

qinsoon (approved):
LGTM

qinsoon:
The fail in JikesRVM tests is a bit strange.
```console
+++ ./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/rvm -X:gc:no_reference_types=false -Xms75M -Xmx75M -jar /home/runner/work/mmtk-core/mmtk-core/mmtk-jikesrvm/.github/scripts/../../repos/jikesrvm/dacapo/dacapo-2006-10-MR2.jar bloat
===== DaCapo bloat starting =====
Error: Process completed with exit code 1.
```
There is no error message. I cannot reproduce it locally, and I haven't seen this kind of error before. I will just retry the run and see what happens.

wks (approved):
LGTM

qinsoon:
The JikesRVM tests keep failing constantly (3/3 runs failed). The JikesRVM binding and the mmtk-core are both v0.12, which passed the tests in https://github.com/mmtk/mmtk-jikesrvm/pull/110. So it is possibly either a bug in this PR, or a bug revealed by this PR. I will investigate into this as it may relate to weak reference processing (it always failed in the weak ref tests).

wks:
I run the following locally, and it has a high chance to fail with an exception.

```shell
./bin/buildit localhost RFastAdaptiveSemiSpace --use-third-party-heap=../mmtk-jikesrvm/ --use-third-party-build-configs=../mmtk-jikesrvm/jikesrvm/build/configs --use-external-source=../mmtk-jikesrvm/jikesrvm/rvm/src --m32

./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/rvm -X:gc:no_reference_types=false -Xms150M -Xmx150M -jar /path/to/dacapo-2006-10-MR2.jar eclipse
```

The log:
```
===== DaCapo eclipse starting =====
<setting up workspace...>
JikesRVM: error loading library ./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/libcore_3_1_0.so: ./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/libcore_3_1_0.so: cannot open shared object file: No such file or directory
JikesRVM: error loading library ./libcore_3_1_0.so: ./libcore_3_1_0.so: cannot open shared object file: No such file or directory
<creating projects..............................................................>
<running tests at level 0...>
<performing build tests...>
        org.apache.ant (not open) opening cleaning building
        org.junit (not open) opening cleaning building
        org.eclipse.osgi (not open) opening cleaning building
<performing type hierarchy tests...>
        Hierarchy: org.eclipse.help.internal HelpPlugin
<performing AST tests...>
        AST creation: org.eclipse.jdt.internal.compiler.parser
<performing completion tests...>
        Completion: Completion>Name>Empty
        Completion: Completion>Name>Empty>No Method
<performing search tests...>
        Searching: indexing
org.eclipse.core.runtime.InvalidRegistryObjectException: Invalid registry object
   at org.eclipse.core.internal.registry.RegistryObjectManager.basicGetObject(RegistryObjectManager.java:188)
   at org.eclipse.core.internal.registry.RegistryObjectManager.getObject(RegistryObjectManager.java:178)
   at org.eclipse.core.internal.registry.ExtensionPointHandle.getExtensionPoint(ExtensionPointHandle.java:88)
   at org.eclipse.core.internal.registry.ExtensionPointHandle.getConfigurationElements(ExtensionPointHandle.java:56)
   at org.eclipse.core.internal.registry.ExtensionRegistry.getConfigurationElementsFor(ExtensionRegistry.java:290)
   at org.eclipse.core.internal.resources.Workspace.initializeMoveDeleteHook(Workspace.java:1245)
   at org.eclipse.core.internal.resources.Workspace.getMoveDeleteHook(Workspace.java:1072)
   at org.eclipse.core.internal.resources.Resource.unprotectedDelete(Resource.java:1545)
   at org.eclipse.core.internal.resources.Resource.delete(Resource.java:664)
   at org.eclipse.core.internal.resources.Project.delete(Project.java:329)
   at dacapo.eclipse.EclipseTests.runtests(EclipseTests.java:123)
   at dacapo.eclipse.HarnessRunner.run(HarnessRunner.java:21)
   at org.eclipse.core.internal.runtime.PlatformActivator$1.run(PlatformActivator.java:226)
   at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:376)
   at dacapo.eclipse.EclipseHarness.iterate(EclipseHarness.java:35)
   at dacapo.Benchmark.run(Benchmark.java:126)
   at dacapo.TestHarness.runBenchmark(TestHarness.java:302)
   at dacapo.TestHarness.main(TestHarness.java:242)
   at Harness.main(Harness.java:5)
Digest validation failed for stderr.log, expecting 0xda39a3ee5e6b4b0d3255bfef95601890afd80709 found 0xbbeb2ed56fe68de1bb6a14e7042b2b500fb843c6
===== DaCapo eclipse FAILED =====
Validation FAILED for eclipse default

```

k-sareen:
Is that with a debug build? If not, can you check if there is any errors outputted by mmtk in a debug build?

wenyuzhao:
> I run the following locally, and it has a high chance to fail with an exception.
> 
> ```shell
> ./bin/buildit localhost RFastAdaptiveSemiSpace --use-third-party-heap=../mmtk-jikesrvm/ --use-third-party-build-configs=../mmtk-jikesrvm/jikesrvm/build/configs --use-external-source=../mmtk-jikesrvm/jikesrvm/rvm/src --m32
> 
> ./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/rvm -X:gc:no_reference_types=false -Xms150M -Xmx150M -jar /path/to/dacapo-2006-10-MR2.jar eclipse
> ```
> 
> The log:
> 
> ```
> ===== DaCapo eclipse starting =====
> <setting up workspace...>
> JikesRVM: error loading library ./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/libcore_3_1_0.so: ./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/libcore_3_1_0.so: cannot open shared object file: No such file or directory
> JikesRVM: error loading library ./libcore_3_1_0.so: ./libcore_3_1_0.so: cannot open shared object file: No such file or directory
> <creating projects..............................................................>
> <running tests at level 0...>
> <performing build tests...>
>         org.apache.ant (not open) opening cleaning building
>         org.junit (not open) opening cleaning building
>         org.eclipse.osgi (not open) opening cleaning building
> <performing type hierarchy tests...>
>         Hierarchy: org.eclipse.help.internal HelpPlugin
> <performing AST tests...>
>         AST creation: org.eclipse.jdt.internal.compiler.parser
> <performing completion tests...>
>         Completion: Completion>Name>Empty
>         Completion: Completion>Name>Empty>No Method
> <performing search tests...>
>         Searching: indexing
> org.eclipse.core.runtime.InvalidRegistryObjectException: Invalid registry object
>    at org.eclipse.core.internal.registry.RegistryObjectManager.basicGetObject(RegistryObjectManager.java:188)
>    at org.eclipse.core.internal.registry.RegistryObjectManager.getObject(RegistryObjectManager.java:178)
>    at org.eclipse.core.internal.registry.ExtensionPointHandle.getExtensionPoint(ExtensionPointHandle.java:88)
>    at org.eclipse.core.internal.registry.ExtensionPointHandle.getConfigurationElements(ExtensionPointHandle.java:56)
>    at org.eclipse.core.internal.registry.ExtensionRegistry.getConfigurationElementsFor(ExtensionRegistry.java:290)
>    at org.eclipse.core.internal.resources.Workspace.initializeMoveDeleteHook(Workspace.java:1245)
>    at org.eclipse.core.internal.resources.Workspace.getMoveDeleteHook(Workspace.java:1072)
>    at org.eclipse.core.internal.resources.Resource.unprotectedDelete(Resource.java:1545)
>    at org.eclipse.core.internal.resources.Resource.delete(Resource.java:664)
>    at org.eclipse.core.internal.resources.Project.delete(Project.java:329)
>    at dacapo.eclipse.EclipseTests.runtests(EclipseTests.java:123)
>    at dacapo.eclipse.HarnessRunner.run(HarnessRunner.java:21)
>    at org.eclipse.core.internal.runtime.PlatformActivator$1.run(PlatformActivator.java:226)
>    at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:376)
>    at dacapo.eclipse.EclipseHarness.iterate(EclipseHarness.java:35)
>    at dacapo.Benchmark.run(Benchmark.java:126)
>    at dacapo.TestHarness.runBenchmark(TestHarness.java:302)
>    at dacapo.TestHarness.main(TestHarness.java:242)
>    at Harness.main(Harness.java:5)
> Digest validation failed for stderr.log, expecting 0xda39a3ee5e6b4b0d3255bfef95601890afd80709 found 0xbbeb2ed56fe68de1bb6a14e7042b2b500fb843c6
> ===== DaCapo eclipse FAILED =====
> Validation FAILED for eclipse default
> ```

Yes I'm seeing the error as well. Looks like [this commit](https://github.com/mmtk/mmtk-core/pull/507/commits/8f7db595be9ed1b5ee3e4fa347a704d5f0ca6422) can fix it. But apparently the CI error is something I still cannot reproduce locally.

qinsoon:
> > I run the following locally, and it has a high chance to fail with an exception.
> > ```shell
> > ./bin/buildit localhost RFastAdaptiveSemiSpace --use-third-party-heap=../mmtk-jikesrvm/ --use-third-party-build-configs=../mmtk-jikesrvm/jikesrvm/build/configs --use-external-source=../mmtk-jikesrvm/jikesrvm/rvm/src --m32
> > 
> > ./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/rvm -X:gc:no_reference_types=false -Xms150M -Xmx150M -jar /path/to/dacapo-2006-10-MR2.jar eclipse
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The log:
> > ```
> > ===== DaCapo eclipse starting =====
> > <setting up workspace...>
> > JikesRVM: error loading library ./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/libcore_3_1_0.so: ./dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/libcore_3_1_0.so: cannot open shared object file: No such file or directory
> > JikesRVM: error loading library ./libcore_3_1_0.so: ./libcore_3_1_0.so: cannot open shared object file: No such file or directory
> > <creating projects..............................................................>
> > <running tests at level 0...>
> > <performing build tests...>
> >         org.apache.ant (not open) opening cleaning building
> >         org.junit (not open) opening cleaning building
> >         org.eclipse.osgi (not open) opening cleaning building
> > <performing type hierarchy tests...>
> >         Hierarchy: org.eclipse.help.internal HelpPlugin
> > <performing AST tests...>
> >         AST creation: org.eclipse.jdt.internal.compiler.parser
> > <performing completion tests...>
> >         Completion: Completion>Name>Empty
> >         Completion: Completion>Name>Empty>No Method
> > <performing search tests...>
> >         Searching: indexing
> > org.eclipse.core.runtime.InvalidRegistryObjectException: Invalid registry object
> >    at org.eclipse.core.internal.registry.RegistryObjectManager.basicGetObject(RegistryObjectManager.java:188)
> >    at org.eclipse.core.internal.registry.RegistryObjectManager.getObject(RegistryObjectManager.java:178)
> >    at org.eclipse.core.internal.registry.ExtensionPointHandle.getExtensionPoint(ExtensionPointHandle.java:88)
> >    at org.eclipse.core.internal.registry.ExtensionPointHandle.getConfigurationElements(ExtensionPointHandle.java:56)
> >    at org.eclipse.core.internal.registry.ExtensionRegistry.getConfigurationElementsFor(ExtensionRegistry.java:290)
> >    at org.eclipse.core.internal.resources.Workspace.initializeMoveDeleteHook(Workspace.java:1245)
> >    at org.eclipse.core.internal.resources.Workspace.getMoveDeleteHook(Workspace.java:1072)
> >    at org.eclipse.core.internal.resources.Resource.unprotectedDelete(Resource.java:1545)
> >    at org.eclipse.core.internal.resources.Resource.delete(Resource.java:664)
> >    at org.eclipse.core.internal.resources.Project.delete(Project.java:329)
> >    at dacapo.eclipse.EclipseTests.runtests(EclipseTests.java:123)
> >    at dacapo.eclipse.HarnessRunner.run(HarnessRunner.java:21)
> >    at org.eclipse.core.internal.runtime.PlatformActivator$1.run(PlatformActivator.java:226)
> >    at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:376)
> >    at dacapo.eclipse.EclipseHarness.iterate(EclipseHarness.java:35)
> >    at dacapo.Benchmark.run(Benchmark.java:126)
> >    at dacapo.TestHarness.runBenchmark(TestHarness.java:302)
> >    at dacapo.TestHarness.main(TestHarness.java:242)
> >    at Harness.main(Harness.java:5)
> > Digest validation failed for stderr.log, expecting 0xda39a3ee5e6b4b0d3255bfef95601890afd80709 found 0xbbeb2ed56fe68de1bb6a14e7042b2b500fb843c6
> > ===== DaCapo eclipse FAILED =====
> > Validation FAILED for eclipse default
> > ```
> 
> Yes I'm seeing the error as well. Looks like [this commit](https://github.com/mmtk/mmtk-core/pull/507/commits/8f7db595be9ed1b5ee3e4fa347a704d5f0ca6422) can fix it. But apparently the CI error is something I still cannot reproduce locally.

I have seen this error in the CI (I am debugging it in my fork): https://github.com/qinsoon/mmtk-core/runs/6619841118?check_suite_focus=true

src/scheduler/scheduler.rs
```diff
@@ -51,80 +51,64 @@ unsafe impl<VM: VMBinding> Sync for GCWorkScheduler<VM> {}
...
-                            if closure_end() {
-                                // Don't open `RefClosure` if `closure_end` added more works to `Closure`.
-                                return false;
+            let stages = (0..WorkBucketStage::LENGTH).map(WorkBucketStage::from_usize);
+            for stage in stages {
+                if stage != WorkBucketStage::Unconstrained && stage != first_stw_stage {
+                    let cur_stages = open_stages.clone();
+                    work_buckets[stage].set_open_condition(
+                        move |scheduler: &GCWorkScheduler<VM>| {
+                            let should_open = scheduler.are_buckets_drained(&cur_stages);
```
qinsoon, line 87: The previous condition for opening a bucket is `are_buckets_drained() && all_workers_parked()`. But the current condition is simply `are_buckets_drained()`. So we may open a new bucket while some workers are executing packets from previous buckets (and they may generate more work for the previous bucket). I feel we should still make sure that all work is done by the workers. 

However, I changed this to `scheduler.are_buckets_drained() && scheduler.worker_group.all_parked()`. But it didn't fix the bug.

wenyuzhao: The caller of this function assumes that all workers are parked. The `&& scheduler.worker_group.all_parked()` is redundant.

https://github.com/mmtk/mmtk-core/pull/507/files/8f7db595be9ed1b5ee3e4fa347a704d5f0ca6422#diff-561a0acf6a922ed5201f43a704c5a68bfc2431156ceb7b9bd8fa5041917f67c6R236

https://github.com/mmtk/mmtk-core/pull/507/files/8f7db595be9ed1b5ee3e4fa347a704d5f0ca6422#diff-561a0acf6a922ed5201f43a704c5a68bfc2431156ceb7b9bd8fa5041917f67c6R376-R377

qinsoon: I see.

wks:
@wenyuzhao 

> Yes I'm seeing the error as well. Looks like [this commit](https://github.com/mmtk/mmtk-core/pull/507/commits/8f7db595be9ed1b5ee3e4fa347a704d5f0ca6422) can fix it. But apparently the CI error is something I still cannot reproduce locally.

No. It doesn't.  I am running it with that commit you mentioned, and it still produces that exception.

The CI didn't print any exception, but it may be the case that the process died and the output were not flushed to the STDOUT or STDERR.

I also tried different `-Xmx` and `MMTK_THREADS` configurations.

The **smaller** the heap is, the more likely the exception will be thrown.  At 150MB heap and 16 threads, it almost guarantees to fail; at 300MB, it sometimes passes; at 600MB, it almost always passes.

The **fewer** GC threads are, the more likely the exception will be thrown. At 150MB heap, with 1 thread, it almost guarantees to fail; with up to 48 threads, there is a small chance to pass ; with 56 threads, it seems to have 50% chance to pass; with 64 threads, it almost always passes.

The relation of failure and the number of GC threads is a bit weird.  It is not like usual data races where it is more likely to go wrong with more threads.

With more GC threads and larger heap, it may also throw other exceptions than the one I printed above.  For example,

```
<performing search tests...>
        Searching: indexing
-- Stack --
(0x69376d1c 2458)   at [0x69376d1c, 0x64b15fdc] Lorg/jikesrvm/runtime/StackTrace; <init>(Lorg/jikesrvm/scheduler/RVMThread;)V at line 89 at bytecode index 15 at machine code offset 0x0000004d
(0x69376d34 2457)   at [0x69376d34, 0x64b8e527] Lorg/jikesrvm/runtime/StackTrace; <init>()V at line 65 at bytecode index 4 at machine code offset 0x00000038
(0x69376d58 1837)   at [0x69376d58, 0x645ffc8e] Ljava/lang/VMThrowable; fillInStackTrace(Ljava/lang/Throwable;)Ljava/lang/VMThrowable; at line 49 at bytecode index 31 at machine code offset 0x000000a7
(0x69376d74 8166)   at [0x69376d74, 0x64c5ee02] Ljava/lang/Throwable; fillInStackTrace()Ljava/lang/Throwable; at line 500 at bytecode index 2 at machine code offset 0x0000003e
(0x69376d98 8170)   at [0x69376d98, 0x64c5df0e] Ljava/lang/Throwable; <init>(Ljava/lang/String;)V at line 161 at bytecode index 10 at machine code offset 0x0000004a
(0x69376db0 14912)   at [0x69376db0, 0x64bf1b26] Ljava/lang/Exception; <init>(Ljava/lang/String;)V at line 78 at bytecode index 2 at machine code offset 0x00000036
(0x69376dc8 20311)   at [0x69376dc8, 0x64a525de] Ljava/lang/RuntimeException; <init>(Ljava/lang/String;)V at line 76 at bytecode index 2 at machine code offset 0x00000036
(0x69376de0 21593)   at [0x69376de0, 0x648e027b] Ljava/lang/IndexOutOfBoundsException; <init>(Ljava/lang/String;)V at line 73 at bytecode index 2 at machine code offset 0x00000036
(0x69376e30 21921)   at [0x69376e30, 0x64c9aaae] Ljava/lang/ArrayIndexOutOfBoundsException; <init>(I)V at line 85 at bytecode index 20 at machine code offset 0x00000277
(0x69376e70 4567)   at [0x69376e70, 0x6aa2b8ec] Lorg/jikesrvm/runtime/RuntimeEntrypoints; deliverHardwareException(ILorg/vmmagic/unboxed/Word;)V at line 791 at bytecode index 436 at machine code offset 0x0000040e
(0x69376e84 1)   at [0x69376e84, 0x6aa2b8ec] <hardware trap>
(0x693a9d24 31780)   at [0x693a9d24, 0x6a9ff3c2] Lorg/eclipse/jdt/internal/compiler/parser/Scanner; getNextToken()I at line 899 at bytecode index 94 at machine code offset 0x00002c40
(0x693a9d78 31588)   at [0x693a9d78, 0x6aa1cad5] Lorg/eclipse/jdt/internal/compiler/parser/Parser; parse()V at line 8460 at bytecode index 236 at machine code offset 0x0000027e
(0x693a9db8 31705)   at [0x693a9db8, 0x68bfac71] Lorg/eclipse/jdt/internal/compiler/parser/Parser; parse(Lorg/eclipse/jdt/internal/compiler/env/ICompilationUnit;Lorg/eclipse/jdt/internal/compiler/CompilationResult;II)Lorg/eclipse/jdt/internal/compiler/ast/CompilationUnitDeclaration; at line 8657 at bytecode index 119 at machine code offset 0x0000018d
(0x693a9dec 28721)   at [0x693a9dec, 0x6a9f576f] Lorg/eclipse/jdt/internal/compiler/parser/Parser; parse(Lorg/eclipse/jdt/internal/compiler/env/ICompilationUnit;Lorg/eclipse/jdt/internal/compiler/CompilationResult;)Lorg/eclipse/jdt/internal/compiler/ast/CompilationUnitDeclaration; at line 8622 at bytecode index 5 at machine code offset 0x0000005d
(0x693a9e3c 31499)   at [0x693a9e3c, 0x6aa07d48] Lorg/eclipse/jdt/internal/compiler/SourceElementParser; parseCompilationUnit(Lorg/eclipse/jdt/internal/compiler/env/ICompilationUnit;Z)Lorg/eclipse/jdt/internal/compiler/ast/CompilationUnitDeclaration; at line 1333 at bytecode index 56 at machine code offset 0x0000015b
(0x693a9e98 31635)   at [0x693a9e98, 0x6aa24443] Lorg/eclipse/jdt/internal/core/search/indexing/SourceIndexer; indexDocument()V at line 81 at bytecode index 173 at machine code offset 0x000003d8
(0x693a9ebc 31750)   at [0x693a9ebc, 0x68a67a89] Lorg/eclipse/jdt/internal/core/search/JavaSearchParticipant; indexDocument(Lorg/eclipse/jdt/core/search/SearchDocument;Lorg/eclipse/core/runtime/IPath;)V at line 72 at bytecode index 24 at machine code offset 0x000000cf
(0x693a9ef8 28269)   at [0x693a9ef8, 0x6a9d2e97] Lorg/eclipse/jdt/internal/core/search/indexing/IndexManager; indexDocument(Lorg/eclipse/jdt/core/search/SearchDocument;Lorg/eclipse/jdt/core/search/SearchParticipant;Lorg/eclipse/jdt/internal/core/index/Index;Lorg/eclipse/core/runtime/IPath;)V at line 290 at bytecode index 9 at machine code offset 0x0000006d
(0x693a9f38 31243)   at [0x693a9f38, 0x68a60550] Lorg/eclipse/jdt/internal/core/search/indexing/IndexManager$1; execute(Lorg/eclipse/core/runtime/IProgressMonitor;)Z at line 601 at bytecode index 85 at machine code offset 0x0000025b
(0x693a9f6c 26515)   at [0x693a9f6c, 0x64633bea] <unprintable normal Java frame: CompiledMethods.getCompiledMethod(26515) returned null>
(0x693a9f88 5525)   at [0x693a9f88, 0x64a4a6bf] Ljava/lang/Thread; run()V at line 745 at bytecode index 11 at machine code offset 0x00000066
(0x693a9fb8 22134)   at [0x693a9fb8, 0x64c8ff54] Lorg/jikesrvm/scheduler/RVMThread; run()V at line 2760 at bytecode index 43 at machine code offset 0x0000026b
(0x693a9ff0 22223)   at [0x693a9ff0, 0xf7f5ae21] Lorg/jikesrvm/scheduler/RVMThread; startoff()V at line 2819 at bytecode index 78 at machine code offset 0x000001b8
```

and

```
===== DaCapo eclipse starting =====
Exception in thread "MainThread" java.lang.NoClassDefFoundError: Could not find the class org.eclipse.jdt.internal.launching.StandardVMType:
        org.eclipse.jdt.internal.launching.StandardVMType
   at org.jikesrvm.classloader.TypeReference.resolveInternal(TypeReference.java:818)
   at org.jikesrvm.classloader.TypeReference.resolve(TypeReference.java:803)
   at org.jikesrvm.runtime.RuntimeEntrypoints.unresolvedNewScalar(RuntimeEntrypoints.java:268)
   at dacapo.eclipse.HarnessRunner.validJavaHome(HarnessRunner.java:38)
   at dacapo.eclipse.HarnessRunner.run(HarnessRunner.java:19)
   at org.eclipse.core.internal.runtime.PlatformActivator$1.run(PlatformActivator.java:226)
   at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:376)
   at dacapo.eclipse.EclipseHarness.iterate(EclipseHarness.java:35)
   at dacapo.Benchmark.run(Benchmark.java:126)
   at dacapo.TestHarness.runBenchmark(TestHarness.java:302)
   at dacapo.TestHarness.main(TestHarness.java:242)
   at Harness.main(Harness.java:5)
Caused by: java.lang.ClassNotFoundException: org.eclipse.jdt.internal.launching.StandardVMType
   at org.eclipse.osgi.framework.internal.core.BundleLoader.findClass(BundleLoader.java:405)
   at org.eclipse.osgi.framework.internal.core.BundleLoader.findClass(BundleLoader.java:350)
   at org.eclipse.osgi.framework.adaptor.core.AbstractClassLoader.loadClass(AbstractClassLoader.java:78)
   at java.lang.ClassLoader.loadClass(ClassLoader.java:293)
   at org.jikesrvm.classloader.TypeReference.resolveInternal(TypeReference.java:815)
   ...11 more
```

wks:
If I set `-X:gc:no_reference_types=true`, i.e. 
```
MMTK_THREADS=16 ~/projects/mmtk-github/jikesrvm/dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/rvm -X:gc:no_reference_types=true -Xms150M -Xmx150M -jar dacapo-2006-10-MR2.jar eclipse
```
there is a high chance that the process will terminate without any warning, with exit state 255.

It is possible for the process to terminate at any point. Sometimes there is no output at all.  The following is an example output.

```
===== DaCapo eclipse starting =====
<setting up workspace...>
JikesRVM: error loading library /home/wks/projects/mmtk-github/jikesrvm/dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/libcore_3_1_0.so: /home/wks/projects/mmtk-github/jikesrvm/dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/libcore_3_1_0.so: cannot open shared object file: No such file or directory
JikesRVM: error loading library ./libcore_3_1_0.so: ./libcore_3_1_0.so: cannot open shared object file: No such file or directory
<creating projects.............
```
and
```sh
$ echo $?
255
```

wenyuzhao:
Looks like the checks are passed now, after fixing the designated work issue.

@qinsoon Can you please double-check one of my fixes and ensure this is correct? https://github.com/mmtk/mmtk-core/pull/507/commits/8f7db595be9ed1b5ee3e4fa347a704d5f0ca6422 Looks like `scan_soft_refs` already calls `self.soft.scan`. Calling it twice will make the `enqueued_references` list incorrect https://github.com/mmtk/mmtk-core/blob/8f7db595be9ed1b5ee3e4fa347a704d5f0ca6422/src/util/reference_processor.rs#L371

wks:
I can still reproduce the problem locally.

Build:

```sh
./bin/buildit localhost RFastAdaptiveSemiSpace --use-third-party-heap=../mmtk-jikesrvm/ --use-third-party-build-configs=../mmtk-jikesrvm/jikesrvm/build/configs --use-external-source=../mmtk-jikesrvm/jikesrvm/rvm/src --m32
```

Run:

```sh
RUST_LOG=off MMTK_THREADS=1  ~/projects/mmtk-github/jikesrvm/dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/rvm -X:gc:no_reference_types=false -Xms150M -Xmx150M -jar dacapo-2006-10-MR2.jar eclipse
```

And it still raises `org.eclipse.core.runtime.InvalidRegistryObjectException`

wks:
I found a simple test case that can stably reproduce the problem.

Save the code below as `Main.java`:

```java
import java.lang.ref.SoftReference;
import java.util.ArrayList;

class Foo {
    int n;
    int[] m;

    public Foo(int n) {
        this.n = n;
        this.m = new int[n * 100 + 3000];
    }
}

class MySR<T> extends SoftReference<T> {
    public MySR(T object) {
        super(object);
    }
}

public class Main {
    ArrayList<Foo> hardMap = new ArrayList<Foo>();
    ArrayList<MySR<Foo>> softMap = new ArrayList<MySR<Foo>>();

    static final int NUMS = 3000;
    static final int KEEP_EVERY = 100;

    public void populateMap() {
        for (int i = 0; i < NUMS; i++) {
            Foo foo = new Foo(i);
            MySR<Foo> soft = new MySR<Foo>(foo);
            softMap.add(soft);

            if (i % KEEP_EVERY == 0) {
                hardMap.add(foo);
                System.out.println(i);
            }
        }
    }

    public void checkMap() {
        for (int i = 0; i < NUMS; i++) {
            if (i % KEEP_EVERY == 0) {
                MySR<Foo> soft = softMap.get(i);
                Foo foo = soft.get();
                if (foo == null) {
                    throw new RuntimeException("foo is null! i = " + i);
                }
            }
        }
    }

        Main main = new Main();

        System.out.println("Populating map...");
        main.populateMap();

        System.out.println("Checking map...");
        main.checkMap();

        System.out.println("Done.");
    }

}
```

Compile it with `javac` and run it with JikesRVM:

```sh
MMTK_THREADS=1  ~/projects/mmtk-github/jikesrvm/dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/rvm -X:gc:no_reference_types=false -Xms70M -Xmx70M Main
```

Run it and it will throw an exception: 
```
Checking map...
Exception in thread "MainThread" java.lang.RuntimeException: foo is null! i = 0
   at Main.checkMap(Main.java:46)
   at Main.main(Main.java:59)
```

It is a bit sensitive to the heap size and the test case size.  I think it should be small enough to trigger emergency GC, but big enough to let it finish.  You can adjust the `-Xms` and `-Xmx` cmdline arguments.

Some factors that affects the result:

-   It is necessary to use a **derived** class of `SoftReference`.  Replacing `MySR` with simple `SoftReference` will not reproduce this error.
-   Setting `MMTK_THREADS` to 1-2 should guarantee the error.  Setting it to 3 and it may crash without error (with some mysterious exit code).  But if it is too big (such as 64), it may usually pass.

wks (changes requested):
The test case above indicates that there is a bug in either this PR or somewhere in JikesRVM.  Please identify the bug before merging.

qinsoon:
> Looks like the checks are passed now, after fixing the designated work issue.
> 
> @qinsoon Can you please double-check one of my fixes and ensure this is correct? [8f7db59](https://github.com/mmtk/mmtk-core/commit/8f7db595be9ed1b5ee3e4fa347a704d5f0ca6422) Looks like `scan_soft_refs` already calls `self.soft.scan`. Calling it twice will make the `enqueued_references` list incorrect
> 
> https://github.com/mmtk/mmtk-core/blob/8f7db595be9ed1b5ee3e4fa347a704d5f0ca6422/src/util/reference_processor.rs#L371

The JikesRVM tests may still fail, though now there is a chance that the tests could pass in the CI. 

The change about weak reference is reasonable. We will only need to scan soft refs once, either after retaining the soft refs, or before scanning weak refs (the Java MMTk scans it before the weak refs). There could be a subtle difference between these two cases: after retaining soft refs, we will do a transitive closure, in which we may add more soft refs. But it should not affect correctness.

qinsoon:
> The test case above indicates that there is a bug in either this PR or somewhere in JikesRVM. Please identify the bug before merging.

Thanks for the test case. I will look into this as well.

wks:
Sorry.  The bug does not depend on having *subclass* of `SoftReference`.  In JikesRVM, a `SoftReference` registers itself via `add_soft_candidate` in its constructor, so it already covers all subclasses.

One possible cause is that the MMTk core started soft reference processing before the closure phase has completed, which may cause strongly-reachable objects erroneously identified as softly reachable.

wks:
I tried my small test case on OpenJDK, and it either passes or crash with out-of-memory.  The SoftReference is never erroneously cleared.

The following command makes the test pass on my machine, but shrinking the heap to `-Xm{s,x}20M` will make it OOM.

```sh
env MMTK_NO_REFERENCE_TYPES=false MMTK_THREADS=1 RUST_BACKTRACE=1 MMTK_PLAN=SemiSpace /home/wks/projects/mmtk-github/openjdk/build/linux-x86_64-normal-server-release/jdk/bin/java -XX:+UseThirdPartyHeap -server -XX:MetaspaceSize=100M -Xm{s,x}22M Main
```

I guess the bug is still related to clearing SoftReference, but in the VM-specific part.

wks:
It is also reproducible if I replace `SoftReference` with `WeakReference`.

wks:
Here is an even simpler test case.  This time you can use any heap size. It doesn't matter.

Just like before, it is impossible to reproduce it on OpenJDK.

```java
import java.lang.ref.WeakReference;
import java.util.ArrayList;

public class Simple {
    Object hard;
    WeakReference<Object> soft;

    public void populate() {
        Object obj = new Object();
        hard = obj;
        soft = new WeakReference(obj);
    }

    public void check() {
        Object obj = soft.get();
        if (obj == null) {
            throw new RuntimeException("obj is null!");
        }
        Object obj2 = hard;
        System.out.println(obj2);
    }

        Simple main = new Simple();

        System.out.println("Populating...");
        main.populate();

        System.out.println("Doing lots of GC...");
        for (int i = 0; i < 100; i++) {
            System.gc();
        }

        System.out.println("Checking...");
        main.check();

        System.out.println("Done.");
    }

}
```

qinsoon:
> Here is an even simpler test case. This time you can use any heap size. It doesn't matter.
> 
> Just like before, it is impossible to reproduce it on OpenJDK.
> 
> ```java
> import java.lang.ref.WeakReference;
> import java.util.ArrayList;
> 
> public class Simple {
>     Object hard;
>     WeakReference<Object> soft;
> 
>     public void populate() {
>         Object obj = new Object();
>         hard = obj;
>         soft = new WeakReference(obj);
>     }
> 
>     public void check() {
>         Object obj = soft.get();
>         if (obj == null) {
>             throw new RuntimeException("obj is null!");
>         }
>         Object obj2 = hard;
>         System.out.println(obj2);
>     }
> 
>     public static void main(String[] args) {
>         Simple main = new Simple();
> 
>         System.out.println("Populating...");
>         main.populate();
> 
>         System.out.println("Doing lots of GC...");
>         for (int i = 0; i < 100; i++) {
>             System.gc();
>         }
> 
>         System.out.println("Checking...");
>         main.check();
> 
>         System.out.println("Done.");
>     }
> 
> }
> ```

Does this test pass with `master` but fail with this PR?

wks:
@qinsoon Yes.  It always passes with mmtk-core master, no matter the number of threads.

By the way, with the `work-stealing` branch, it sometimes cannot finish all 100 GC rounds.  It may hang or crash sometimes.

wks:
There is a chance that a `WeakRefProcessing` work packet is executed before any ProcessEdgesWork is executed.  This doesn't look right.  Whenever this happens, my test case will raise exception because the WeakReference is cleared.

```
[2022-06-03T11:58:49Z INFO  mmtk::plan::global] User triggering collection
[2022-06-03T11:58:49Z WARN  mmtk::util::reference_processor] WeakRefProcessing
[2022-06-03T11:58:49Z WARN  mmtk::util::reference_processor] WeakRefProcessing End
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81289
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork End: 81289
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81291
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork End: 81291
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81292
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork End: 81292
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81293
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork End: 81293
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81299
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork End: 81299
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81300
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork End: 81300
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81302
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork End: 81302
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81303
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork End: 81303
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81306
[2022-06-03T11:58:49Z ERROR mmtk::scheduler::gc_work] ProcessEdgesWork: 81294
```

wks:
A more detailed log of a problematic GC (timestamp and type parameters are omitted):

```
BEGIN mmtk::scheduler::gc_work::ScheduleCollection
END mmtk::scheduler::gc_work::ScheduleCollection
BEGIN mmtk::scheduler::gc_work::StopMutators
END mmtk::scheduler::gc_work::StopMutators
BEGIN mmtk::scheduler::gc_work::StopMutators
BEGIN mmtk::scheduler::gc_work::Prepare
END mmtk::scheduler::gc_work::Prepare
BEGIN mmtk::scheduler::gc_work::PrepareCollector
END mmtk::scheduler::gc_work::PrepareCollector
BEGIN mmtk::scheduler::gc_work::PrepareMutator
END mmtk::scheduler::gc_work::PrepareMutator
BEGIN mmtk::scheduler::gc_work::PrepareMutator
END mmtk::scheduler::gc_work::PrepareMutator
BEGIN mmtk::scheduler::gc_work::PrepareMutator
END mmtk::scheduler::gc_work::PrepareMutator
BEGIN mmtk::scheduler::gc_work::PrepareMutator
END mmtk::scheduler::gc_work::PrepareMutator
BEGIN mmtk::scheduler::gc_work::PrepareMutator
END mmtk::scheduler::gc_work::PrepareMutator
BEGIN mmtk::scheduler::gc_work::PrepareMutator
END mmtk::scheduler::gc_work::PrepareMutator
BEGIN mmtk::scheduler::gc_work::PrepareMutator
END mmtk::scheduler::gc_work::PrepareMutator
BEGIN mmtk::scheduler::gc_work::PrepareMutator
END mmtk::scheduler::gc_work::PrepareMutator
BEGIN mmtk::scheduler::gc_work::PrepareMutator
END mmtk::scheduler::gc_work::PrepareMutator
BEGIN mmtk::scheduler::gc_work::PrepareCollector
END mmtk::scheduler::gc_work::PrepareCollector
BEGIN mmtk::util::reference_processor::SoftRefProcessing
END mmtk::util::reference_processor::SoftRefProcessing
BEGIN mmtk::util::reference_processor::WeakRefProcessing
END mmtk::util::reference_processor::WeakRefProcessing
BEGIN mmtk::scheduler::gc_work::VMProcessWeakRefs
END mmtk::scheduler::gc_work::VMProcessWeakRefs
BEGIN mmtk::util::finalizable_processor::Finalization
END mmtk::util::finalizable_processor::Finalization
BEGIN mmtk::scheduler::gc_work::PlanProcessEdges
END mmtk::scheduler::gc_work::PlanProcessEdges
BEGIN mmtk::scheduler::gc_work::PlanProcessEdges
END mmtk::scheduler::gc_work::PlanProcessEdges
...
BEGIN mmtk::scheduler::gc_work::PlanProcessEdges
END mmtk::scheduler::gc_work::PlanProcessEdges
BEGIN mmtk::scheduler::gc_work::PlanProcessEdges
END mmtk::scheduler::gc_work::PlanProcessEdges
BEGIN mmtk::scheduler::gc_work::ScanStackRoot
END mmtk::scheduler::gc_work::ScanStackRoot
BEGIN mmtk::scheduler::gc_work::ScanStackRoot
END mmtk::scheduler::gc_work::PlanProcessEdges
BEGIN mmtk::scheduler::gc_work::PlanProcessEdges
END mmtk::scheduler::gc_work::ScanStackRoot
BEGIN mmtk::scheduler::gc_work::ScanStackRoot
END mmtk::scheduler::gc_work::ScanStackRoot
BEGIN mmtk::scheduler::gc_work::ScanStackRoot
END mmtk::scheduler::gc_work::ScanStackRoot
BEGIN mmtk::scheduler::gc_work::ScanStackRoot
END mmtk::scheduler::gc_work::ScanStackRoot
BEGIN mmtk::scheduler::gc_work::ScanStackRoot
END mmtk::scheduler::gc_work::ScanStackRoot
BEGIN mmtk::scheduler::gc_work::ScanStackRoot
END mmtk::scheduler::gc_work::ScanStackRoot
BEGIN mmtk::scheduler::gc_work::ScanStackRoot
END mmtk::scheduler::gc_work::ScanStackRoot
BEGIN mmtk::scheduler::gc_work::ScanStackRoot
END mmtk::scheduler::gc_work::ScanStackRoot
BEGIN mmtk::scheduler::gc_work::ScanVMSpecificRoots
END mmtk::scheduler::gc_work::ScanVMSpecificRoots
BEGIN mmtk_jikesrvm::scan_statics::ScanStaticRoots
END mmtk_jikesrvm::scan_statics::ScanStaticRoots
END mmtk::scheduler::gc_work::PlanProcessEdges
BEGIN mmtk::scheduler::gc_work::PlanProcessEdges
BEGIN mmtk_jikesrvm::scanning::ScanGlobalRoots
END mmtk_jikesrvm::scanning::ScanGlobalRoots
BEGIN mmtk_jikesrvm::scan_statics::ScanStaticRoots
END mmtk_jikesrvm::scan_statics::ScanStaticRoots
BEGIN mmtk_jikesrvm::scanning::ScanGlobalRoots
END mmtk_jikesrvm::scanning::ScanGlobalRoots
BEGIN mmtk::scheduler::gc_work::PlanProcessEdges
END mmtk::scheduler::gc_work::PlanProcessEdges
BEGIN mmtk::scheduler::gc_work::PlanProcessEdges
END mmtk::scheduler::gc_work::PlanProcessEdges
...
```

It looks like the buckets for reference processing are opened too early.

I guess what happened is, some work packets are asynchronous.  Those mmtk-core work packets finished very quickly, but there are VM threads working in the background, and may deliver more work packets afterwards.  However, the GC workers in the core thinks all buckets are empty.  It then decide that more work packets should be opened.  The weak reference processing work packets are added by the `ScheduleCollection` work packet which is executed by the GC coordinator thread.  Therefore `WeakRefProcessing` becomes executable even before stacks are scanned.

qinsoon:
@wks Nice debugging! In ScheduleCollection, we do not schedule anything for the Closure bucket, but rather rely on the binding to generate the closure packets. Actually we do not schedule anything between scanning stacks (or stopping mutators - i can’t remember) and weak ref processing. That means there could be a bug that at some point, the schedule sees there is no work in the closure bucket and all conditions for opening a new bucket is satisfied, so it activates the weak processing bucket. This also explains why CI may pass occasionally after Wenyu fixed an issue about designated work (before the fix, the scheduler might open new buckets even if some workers have designated work). The fix probably eliminated some cases, so it reduces the chance we see the bug in CI. @wenyuzhao

wks:
I found the cause.

One difference between the JikesRVM binding and the OpenJDK binding is that in OpenJDK, any GC worker can execute the `StopMutators` work packet, while in JikesRVM, only the GC coordinator thread can execute `StopMutators`.  The background was that OpenJDK requires that the thread that stops all thread must be the same as the thread that starts the mutators. (See https://github.com/mmtk/mmtk-core/pull/450)

The `StopMutators` work packet calls `stop_all_mutator`, and then immediately notifys workers to continue. See:

```rust
// StopMutators::do_work
        trace!("stop_all_mutators start");
        mmtk.plan.base().prepare_for_stack_scanning();
        <E::VM as VMBinding>::VMCollection::stop_all_mutators::<E>(worker.tls);
        trace!("stop_all_mutators end");
        mmtk.scheduler.notify_mutators_paused(mmtk);    // THIS LINE!!!!!!!!!!!!
        if <E::VM as VMBinding>::VMScanning::SCAN_MUTATORS_IN_SAFEPOINT {
...
                for mutator in <E::VM as VMBinding>::VMActivePlan::mutators() {
                    mmtk.scheduler.work_buckets[WorkBucketStage::Prepare]
                        .add(ScanStackRoot::<E>(mutator));  // ADD WORK PACKETS
                }
        }
        mmtk.scheduler.work_buckets[WorkBucketStage::Prepare].add(ScanVMSpecificRoots::<E>::new());  // ADD WORK PACKETS
```

Note that it notifies the GC workers before adding more work packets.  This is not a problem for OpenJDK, because the `StopMutators` as a whole is executed on a GC worker.  If any GC worker is running, it does not satisfy the condition of "all mutators have parked".

However, on JikesRVM, `StopMutators` is executed on the GC coordinator thread, which is not counted as a GC worker.  If there is a slight delay after the `notify_mutators_paused` invocation, as in the following patch:

```diff
diff --git a/src/scheduler/gc_work.rs b/src/scheduler/gc_work.rs
index 615458f0..a4a2636b 100644
--- a/src/scheduler/gc_work.rs
+++ b/src/scheduler/gc_work.rs
@@ -10,6 +10,7 @@ use std::marker::PhantomData;
 use std::mem;
 use std::ops::{Deref, DerefMut};
 use std::sync::atomic::Ordering;
+use std::time::Duration;
 
 pub struct ScheduleCollection;
 
@@ -186,6 +187,9 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for StopMutators<E> {
         <E::VM as VMBinding>::VMCollection::stop_all_mutators::<E>(worker.tls);
         trace!("stop_all_mutators end");
         mmtk.scheduler.notify_mutators_paused(mmtk);
+        std::thread::sleep(Duration::from_millis(10));
         if <E::VM as VMBinding>::VMScanning::SCAN_MUTATORS_IN_SAFEPOINT {
             // Prepare mutators if necessary
             // FIXME: This test is probably redundant. JikesRVM requires to call `prepare_mutator` once after mutators are paused
```

Then GC workers will wake up, and found that all open buckets are drained.  Then the GC workers will open all subsequent work packets.  Because the delay after `notify_mutators_paused` can be arbitrary depending on operating system scheduling, arbitrarily many work packets may have been executed before the `StopMutators` work packet add `ScanStackRoot` and `ScanVMSpecificRoot` work packets, which generate all the subsequent `ProcessEdgesWork` work packets.

So the current "all GC workers have stopped" is insufficient.  It does not take the running GC coordinator thread into account.

wks:
> @wks Can you please advise which machine were you using and what is your command line args? It's weird that I tried the test case with both SemiSpace and MarkSweep multiple times, and I did not see any crashes. I'm using `fox.moma`.
> 
> My command: `dist/RFastAdaptiveMarkSweep_x86_64_m32-linux/rvm -X:gc:no_reference_types=false -Xms20M -Xmx20M Simple`

I am using my own laptop in a LXC container.  It should be reproducible on any machine.

My command line is:
```
MMTK_THREADS=2  ~/projects/mmtk-github/jikesrvm/dist/RFastAdaptiveSemiSpace_x86_64_m32-linux/rvm -X:gc:no_reference_types=false -Xm{s,x}600M Simple 1000
```

The number of threads matters.  It is more likely to reproduce it with 2 threads instead of 1.

Also note that I updated the test script in the afternoon, so that you can specify how many `System.gc()` it runs on the command line, and it will check the `WeakReference` after each GC and fail fast.

wenyuzhao:
The issue is not because the controller proceeded StopMutators packets.

It was not a problem before, because the coordinator was the one that opens new buckets. So it cannot open new buckets until the coordinator StopMutators job is finished. Now it's the last yielding GC worker to open new buckets (to reduce inter-thread messages).

I changed the open condition of the buckets so that no new buckets can open unless the coordinator finished the coordinator's work packets. Hope this can solve the problem.

wenyuzhao:
@wks Can you please confirm that the change fixes the bug on your side? I'm able to reproduce the bug now and looks like it is fixed.

src/scheduler/scheduler.rs
```diff
@@ -251,6 +225,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
 
     fn are_buckets_drained(&self, buckets: &[WorkBucketStage]) -> bool {
         buckets.iter().all(|&b| self.work_buckets[b].is_drained())
+            && self.pending_messages.load(Ordering::Relaxed) == 0
```
wks, line 228: `pending_messages` is stored with `Ordering::SeqCst` in `controller.rs`, but is loaded with `Ordering::Relaxed` here, and I can't find any fences (`std::sync::atomic::fence`).  Is there any reason to use `Relaxed` instead of `SeqCst`?  In theory, if this load has `Relaxed` order, subsequent code in this worker thread may not see the effect of store operations in the work packet (such as StopMutators) executed by the coordinator thread, due to reordering in both machine and compiler level.

src/scheduler/work_bucket.rs
```diff
@@ -1,147 +1,198 @@
...
+        if self.group.parked_workers() > 0 {
+            let _guard = self.monitor.0.lock().unwrap();
+            self.monitor.1.notify_all()
+        }
     }
+
+    #[inline(always)]
     pub fn is_activated(&self) -> bool {
-        self.active.load(Ordering::SeqCst)
+        self.active.load(Ordering::Relaxed)
```
wks, line 99: I am not sure about this.  Buckets are only updated when all workers are parked, so it is sure that `bucket.activate()` happens after all the `inc` operation of `pending_workers`.  However, there is no clear message for the waken workers to know that "if I see `bucket.is_actived() == true`, I see the effect of the execution of work packets of all previous buckets".  In fact, all workers can wake up spuriously (see my other comments).

src/scheduler/worker.rs
```diff
@@ -140,22 +145,131 @@ impl<VM: VMBinding> GCWorker<VM> {
...
+        self.workers_shared.len()
+    }
+
+    /// Increase the packed-workers counter.
+    /// Called before a worker is parked.
+    ///
+    /// Return true if all the workers are parked.
+    #[inline(always)]
+    pub fn inc_parked_workers(&self) -> bool {
+        let old = self.parked_workers.fetch_add(1, Ordering::Relaxed);
```
wks, line 244: Similar to `pending_messages`, in theory, when a worker finds that all workers are parked, it may not see the buckets being updated.

src/scheduler/scheduler.rs
```diff
@@ -261,121 +236,188 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
+                    return work;
+                }
+                debug_assert!(!self.worker_group.has_designated_work());
+                // The current pause is finished if we can't open more buckets.
+                self.pending_messages.fetch_add(1, Ordering::SeqCst);
+                worker.sender.send(CoordinatorMessage::Finish).unwrap();
             }
             // Wait
-            guard = self.worker_monitor.1.wait(guard).unwrap();
+            let _guard = self.worker_monitor.1.wait(guard).unwrap();
```
wks, line 413: This `wait` may wake up spuriously.  It may not be a problem because the condition to wake up is "there are pending work packets to do", and it is notified when new work packets are added.  However, there are memory order issues.  See my other comments regarding `WorkBucket::is_activated`.

wks (changes requested):
I see some problem with the memory orders used in various atomic operations.  They may, in theory, cause some problem.  If unsure, we can always use `SeqCst`.

wks:
@wenyuzhao I confirm that the bug is no longer reproducible on my local machine.  However, I also see some memory order issue.  See my previous comments.

src/scheduler/scheduler.rs
wenyuzhao: I'm also not 100% sure about this. We may need SeqCst so that the stores to `pending_messages` will be seen immediately. But currently, it is safe if the load does not see the new value. In this case, no buckets will open and the workers will be put to sleep. Our only use case for the coordinator worker is to execute the "StopMutators" packets. The packet will wake up the workers anyway, and the workers will decide to open more buckets again (or execute new packets).  Probably I'll just change this to SeqCst.

src/scheduler/work_bucket.rs
wenyuzhao: > if I see bucket.is_actived() == true, I see the effect of the execution of work packets of all previous buckets

This sounds like a strong assumption. To ensure this we will need to put a fence at the start and end of each packet (?).

src/scheduler/worker.rs
wenyuzhao: Yes probably either this or the update buckets code should be SeqCst, so the reordering won't happen. I'll change this.

src/scheduler/scheduler.rs
wks: It is not about this load not seeing `pending_messages` being updated or not opening buckets.  It is the opposite:

**If I see `pending_messages == 0`**, can I guarantee that I see the effect of `StopMutators`?

In other words, if

1.  The coordinator puts a work packet in the `Prepare` bucket, and 
2.  The coordinator sets `pending_messages` to 0, and
3.  A worker loads from `pending_messages`, and sees 0, and
4.  The worker looks at the `Prepare` bucket,

Is the worker guaranteed that it sees the length of the `Prepare` bucket to be at least one (seeing the packet added by the coordinator)?  It is guaranteed if (4) happens after (1).  We know that
-   (1) happens before (2) because they are in the same thread (coordinator), and
-   (3) happens before (4) because they are in the same thread (worker).

If (2) has `Release` order, and (3) has `Acquire` order and (3) sees the value written by (2), then (2) happens before (3), too.  Then we have (1) HB (2) HB (3) HB (4).

Of course `SeqCst` is stronger than both `Acquire` and `Release` and has extra guarantees.  It is easiest to just use `SeqCst`.

src/scheduler/work_bucket.rs
wks: We don't need to put explicit fences, but atomic memory operations are effectively fences.  If an "acquire" load sees the value of a "release" store, the load happens after the store.  That's enough to ensure all other loads after that "acquire" load to happen after all other stores before the "release" store.

If that's still to hard to reason about, we can use `SeqCst`: all `SeqCst` atomic memory operations performed by all threads form **a global total order.**  If every worker **increments a counter** (parked workers) after they have done all their work, and some worker **activate bucket** only after **seeing the counter maxed** (equal to num of workers), and after waking up, workers only pick up packets from a bucket after **seeing it activated**, then we have the guarantee I mentioned.  Those atomic variables (the park counter, and the activation flag) are used for synchronization.

Yes. Just use `SeqCst` if not sure, as it is the easiest to reason about.

wks (approved):

PR #508, opened 2021-12-14
SFT enum dispatch

qinsoon:
This PR replaces the current `dyn SFT` table with a `SFTDispatch` table. `SFTDispatch` allows us to statically dispatch `SFT` calls to spaces, rather than a dynamic dispatch.

We thought this would give us some improvement in terms of performance. However, in terms of using SFT for tracing objects, the performance improvement is only observable on one or two benchmarks, and the improvement is marginal.

However, though the performance improvement is not clear in tracing objects, this approach may improve other cases when using SFT. So I submit a PR for this.


qinsoon:
I am closing this PR as this does not show an observable improvement in object tracing (it might give improvement for other cases if used frequently). We may revisit this.

PR #512, opened 2022-01-03
Work packet stats is collected based on feature

qinsoon:
This PR:
* adds a feature `work_packet_stats`. By default, it is turned off (for performance reasons). Only when it is turned on, we will collect statistics for work packets. 
* adds validation functions for perf events options. 
  * When `perf_counter` is not enabled, any value for the options is considered as invalid. 
  * When `work_packet_stats` is not enabled, any value for `work_perf_events` is considered as invalid.


caizixian (approved):
LGTM

Issue #58, opened 2020-03-26
MMTk Options


qinsoon:
In GitLab by @qinsoon on Jan 30, 2020, 14:03

Currently MMTk is initialized with a set of default options. Options are mutable and the VM can use `process(name, value)` to send options to MMTk. 

According to our discussion, these are clear:
* For command line arguments (such as `-X:gc`), MMTk expects the VM to send them as a string at initialization. And those are immutable.
* MMTk also allows mutable options. VM can call certain APIs to set some options during execution, these options may change MMTk's behaviours.

For mutable options, as we are supposed to have multiple readers but one writer to the options, we could consider using lock-free data structures, such as https://github.com/jonhoo/rust-evmap.

qinsoon:
Options include plan selection, which means we need to create an MMTk instance after parsing MMTk options. This makes the MMTk instance no longer static (at the moment, all bindings create MMTk instance statically with plan selection through env vars). So besides the work of refactoring MMTk options, we have to:
* refactor bindings so MMTk instances are not static variables (we could possibly use a pointer).
* refactor mmtk-core so we no longer assume `MMTK` is always static.

PR #513, opened 2022-01-06
Add API to process options in bulk

qinsoon:
This PR solves part of https://github.com/mmtk/mmtk-core/issues/58. This PR allows bindings to set MMTk options in bulk. For example, a binding can provide a command line option for third party heap options, and use the API to forward the options to MMTk.

This PR
* adds a new API function `process_bulk()` alongside the old `process()` API for setting options. `process_bulk()` takes a string, and MMTk will parse and set options based on the string.
* adds some metadata for each option. Each option will specify whether the option can be set from environment variables, and be set from command line (through `process()`).

Related PRs:
* OpenJDK binding: https://github.com/mmtk/mmtk-openjdk/pull/132
* OpenJDK fork: https://github.com/mmtk/openjdk/pull/10


caizixian (approved):
LGTM

Issue #57, opened 2020-03-26
Uninitialized Fields


qinsoon:
In GitLab by @qinsoon on Jan 26, 2020, 11:44

A lot of structs in Rust MMTk are initialized with some uninitialized fields. We are using empty values (such as `0`, `OpaquePointer::EMPTY`), `mem::uninitialized()`, or `Option::None` for the uninitialized fields.

The reasons why those uninitialized fields exist are:
1. Rust MMTk used to create some structs as static variables, some fields are uninitialized. Rust MMTk relies on an extra `init()` step to properly set values for those fields.
1. Initialize circular references.

I suggest we do these:
1. As we are removing global states, we should remove the extra `init()` so that all the fields are initialized properly in `new()`. We should do this wherever possible.
1. If we cannot remove `init()`, we should use a wrapper to properly indicate that this field might be uninitialized such as `MaybeUninit<T>`.
2. Remove circular references if we can. 

This issue is raised in code reviews for both https://gitlab.anu.edu.au/mmtk/mmtk/merge_requests/20 and https://gitlab.anu.edu.au/mmtk/mmtk/merge_requests/22.

qinsoon:
We have gradually removed the pattern. I am not aware of any unnecessary two-step initialisation in our code base. I feel we can close this issue now. If you find anything that is related with this issue and hasn't been addressed, please reply to this issue. Otherwise, I will close this issue soon.

k-sareen:
Closed as above

PR #514, opened 2022-01-09
Cleanup GC worker init

qinsoon:
This PR resolves the discussion here: https://github.com/mmtk/mmtk-core/pull/499#discussion_r778694398.


src/scheduler/scheduler.rs
```diff
@@ -206,9 +206,9 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         buckets.iter().all(|&b| self.work_buckets[b].is_drained())
     }
 
-    pub fn initialize_worker(self: &Arc<Self>, tls: VMWorkerThread) {
+    pub fn initialize_coordinator_worker(self: &Arc<Self>, tls: VMWorkerThread) {
         let mut coordinator_worker = self.coordinator_worker.as_ref().unwrap().write().unwrap();
```
wks, line 210: This `as_ref().wrap()` is the same, but let's look at it later.

src/plan/controller_collector_context.rs
```diff
@@ -50,6 +50,15 @@ impl<VM: VMBinding> ControllerCollectorContext<VM> {
...
     pub fn run(&self, tls: VMWorkerThread) {
+        // Initialize the GC worker for coordinator. We are not using the run() method from
+        // GCWorker so we manually initialize the worker here.
+        self.scheduler
+            .read()
+            .unwrap()
+            .as_ref()
+            .unwrap()
+            .initialize_coordinator_worker(tls);
+
```
wks, line 61: The `as_ref().unwrap()` part is for getting into the `Option<_>` in `RwLock<Option<Arc<GCWorkScheduler<VM>>>>`.

This long sequence may be a consequence of MMTk using two-phase initialisation.  It is considered by Rust as an anti-pattern, but considering that GC is very low-level, there may be reasons why it has to be done this way.  Let's investigate it later.

wks (approved):
There is no problem with this refactoring.

But the recurring long sequence of `.as_ref().unwrap()` to get into `Option<_>` shows that MMTk may have two-step initialisation anti-pattern.  It is understandable because MMTk was originally developed in Java.  We may consider a more major refactoring later, if we find it problematic.

src/plan/controller_collector_context.rs
qinsoon: We have an issue tracking this: https://github.com/mmtk/mmtk-core/issues/57. In this particular case, I don't think it is because of GC being low-level.

Issue #495, opened 2021-11-04
Linear scan with alloc bit


qinsoon:
In a few places in our code, we have a pattern of linear scanning a region and use alloc bit to find objects. This pattern is used in mark sweep to find objects in active chunks/pages, and is also used in mark compact. We should refactor and extract this pattern to remove duplicate code.

qinsoon:
Note that mark compact also includes some code for linear scan with alloc bit.

Issue #485, opened 2021-10-25
Cleanup chunk-sweeping methods of `MallocSpace`


caizixian:
The two methods, `sweep_chunk_mark_on_side` and `sweep_chunk_mark_in_header`, are long and hard to understand.
The former is 200 lines long, which is really hard to read, and the two methods are quite similar structure-wise.

k-sareen:
Yeah fair -- it is very inelegant currently. Once we have a linear scan interface then we should revisit this as that would make life easier

PR #518, opened 2022-01-13
Refactor linear scan

qinsoon:
This PR closes https://github.com/mmtk/mmtk-core/issues/495 and closes https://github.com/mmtk/mmtk-core/issues/485.

This PR:
* adds `ObjectIterator` that helps iterate though objects in an address range using alloc bit, and a related trait `LinearScanObjectSize`.
* uses `ObjectIteartor` in mark sweep and mark compact
* removes current linear scan code (which is unused)
* refactors chunk sweep code for mark sweep to make it clearer (https://github.com/mmtk/mmtk-core/issues/485)

This PR has no observable performance difference for marksweep. The graph below shows the time for sweeping chunk.
WX20220113-133046


k-sareen:
Quick question -- did you benchmark the PR with both in header and on side mark bits?

qinsoon:
> Quick question -- did you benchmark the PR with both in header and on side mark bits?

No. The graph only shows the performance for in header mark bits. I can measure the other if you think it might be different.

k-sareen:
The default configuration is the bulk load from memory, right? Maybe you can run a very small (~10 invocations) benchmark result for that if possible

qinsoon:
> The default configuration is the bulk load from memory, right? Maybe you can run a very small (~10 invocations) benchmark result for that if possible

Yes, the default is bulk load with on side mark bits. That sounds good. I will do another run and probably will get the results back during the weekend.

src/util/linear_scan.rs
```diff
@@ -0,0 +1,75 @@
...
+use crate::util::ObjectReference;
+use crate::vm::ObjectModel;
+use crate::vm::VMBinding;
+use std::marker::PhantomData;
+
+/// Iterate over an address range, and find each object by alloc bit.
+/// ATOMIC_LOAD_ALLOC_BIT can be set to false if it is known that loading alloc bit
+/// non-atomically is correct (e.g. a single thread is scanning this address range, and
+/// it is the only thread that accesses alloc bit).
+pub struct ObjectIterator<VM: VMBinding, S: LinearScanObjectSize, const ATOMIC_LOAD_ALLOC_BIT: bool>
```
k-sareen, line 12: I was wondering if the type parameter `LinearScanObjectSize` is required? Would a function pointer to a `size()` function not be good enough? Or are you trying to reduce overhead of indirect function calls? (Is it even significant enough?). Either way, it's not a big deal, but was wondering about your thoughts.

k-sareen (commented):
LGTM other than the one question (will approve after discussion and results of bulk load mark bit). Nice work :)

src/util/linear_scan.rs
qinsoon: Yeah, I was considering about performance and trying to be safe, as we know using trait would have no overhead at all. Maybe using a function pointer is enough.

qinsoon:
@k-sareen The results are here.

For mark bits in the header,
ms-header

For mark bits on the side,
ms-side

k-sareen (approved):

PR #521, opened 2022-01-14
Optimize SFT access

qinsoon:


src/util/rust_util.rs
```diff
@@ -6,3 +6,56 @@ pub const fn min_of_usize(a: usize, b: usize) -> usize {
...
+        }
+    }
+
+    /// Initialize the value. This should be called before ever using the struct.
+    pub fn initialize(&self, val: T) {
+        if self
+            .initialized
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+            .is_ok()
+        {
```
wks, line 38: If another thread calls `get_ref` after the `compare_exchange` but before the `write` below, the thread will think it is already initialized.

wks (changes requested):
If accessing `SFT_MAP` is currently too slow due to `lazy_static!`, this could be a workaround.

But keep in mind that the `InitializeOnce` is inherently unsafe.  There is a *temporal dependency* between the initialization and the use.  The Rust language has been trying to prevent this using ownership and lifetime, so variables are always initialized before using.  I believe the proper solution is to make `MMTK` non-static, and refactor the creation of the `MMTK` instance.

src/util/rust_util.rs
qinsoon: I have changed the implementation of `InitializeOnce`. I think it fixes this problem. 

I am now using `Once` to guarantee the initialization only happens once and other threads will be blocked when one thread is doing the initialization. Also the initial value is now only created once.

qinsoon:
V8 test is failing somehow. I have no idea so far and I will work on it. You don't need to review this PR until the test is fixed. @wks

qinsoon:
The V8 failed build for V8 has nothing to do with the changes in this pull request. See here: https://github.com/mmtk/mmtk-v8/runs/4877688469. The V8 binding `master` builds fine before (https://github.com/mmtk/mmtk-v8/runs/4576513408), but now fails.

wks (approved):
LGTM

Issue #522, opened 2022-01-14
Inconsistent mutability of `GCWorker` across VM interface


wks:
Currently,

-   `Collection<VM>::spawn_worker_thread` provides a `ctx: Option<&GCWorker<VM>>` parameter, but
-   `memory_manager::start_worker<VM>` requires a `worker: &mut GCWorker<VM>` parameter.

The provided parameter is **immutable**, but the required parameter is **mutable**.  Since `spawn_worker_thread` is supposed to create a thread that eventually calls `start_worker`, the `ctx` argument, if `Some(worker)` should be passed to `start_worker`.  Now it is a compilation error.

In the following example, I am creating the GC threads directly in Rust, without involving VM-specific C functions.

```rust
impl Collection<Ruby> for VMCollection {
    // ...
    fn spawn_worker_thread(tls: VMThread, ctx: Option<&GCWorker<Ruby>>) {
        match ctx {
            None => {
                thread::Builder::new().name("MMTk Controller Thread".to_string()).spawn(move || {
                    memory_manager::start_control_collector(&SINGLETON, VMWorkerThread(tls))
                });
            }
            Some(worker) => { // worker is a &GCWorker<Ruby>
                thread::Builder::new().name("MMTk Worker Thread".to_string()).spawn(move || {
                    memory_manager::start_worker(VMWorkerThread(tls), worker, &SINGLETON)  // ERROR: expected &mut GCWorker<Ruby>
                });            
            }
        }
    }
    // ...
}
```

Currently, in `mmtk-openjdk`, [the `&GCWorker` is force-cast to `*mut GCWorker`](https://github.com/mmtk/mmtk-openjdk/blob/master/mmtk/src/collection.rs#L54) before passing to C++:
```rust
    fn spawn_worker_thread(tls: VMThread, ctx: Option<&GCWorker<OpenJDK>>) {
        let ctx_ptr = if let Some(r) = ctx {
            r as *const GCWorker<OpenJDK> as *mut GCWorker<OpenJDK>  // casting & to *const then to *mut
        } else {
```

## Update

After discussion, we agree that the GC worker thread should own the `GCWorker` data structure.

The scheduler needs to communicate with GC workers, so shared data structures should be isolated, and should be accessed with proper synchornisation.

The `stat` field is written into by the GC worker during GC, but read from by the scheduler during `harness_end` (in mutator).  
[...]

PR #523, opened 2022-01-17
Fix GCWorker ownership.

wks:
This PR fixes https://github.com/mmtk/mmtk-core/issues/522

We let GC worker threads own `GCWorker`, and isolate shared parts into `GCWorkerShared`.  We also used `AtomicRefCell` for `WorkerLocalStat`.

This PR does *not* fix the potential data race between the mutator (`harness_begin` and `harness_end`) and the GC.  If such a race occurs, it will panic.

I found that GC worker may access the GC scheduler (the `worker_group` field) before the GC scheduler is fully consructed.  Rust doesn't like this (and therefore there is an `Arc::get_mut_unchecked` in `scheduler.rs:106`), and that should be fixed in another PR.

### DIFF
--- a/src/scheduler/work.rs
+++ b/src/scheduler/work.rs
@@ -28,16 +28,20 @@ pub trait GCWork<VM: VMBinding>: 'static + Send {
 
         #[cfg(feature = "work_packet_stats")]
         // Start collecting statistics
-        let stat = worker
-            .stat
-            .measure_work(TypeId::of::<Self>(), type_name::<Self>(), mmtk);
+        let stat = {
+            let mut worker_stat = worker.shared.borrow_stat_mut();
+            worker_stat.measure_work(TypeId::of::<Self>(), type_name::<Self>(), mmtk)
+        };
 
         // Do the actual work
         self.do_work(worker, mmtk);
 
         #[cfg(feature = "work_packet_stats")]
         // Finish collecting statistics
-        stat.end_of_work(&mut worker.stat);
+        {
+            let mut worker_stat = worker.shared.borrow_stat_mut();
+            stat.end_of_work(&mut worker_stat);
+        }
     }
 }
 
--- a/src/scheduler/worker.rs
+++ b/src/scheduler/worker.rs
@@ -5,50 +5,78 @@ use crate::mmtk::MMTK;
 use crate::util::copy::GCWorkerCopyContext;
 use crate::util::opaque_pointer::*;
 use crate::vm::{Collection, VMBinding};
+use atomic_refcell::{AtomicRef, AtomicRefCell, AtomicRefMut};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc::Sender;
-use std::sync::{Arc, Weak};
+use std::sync::Arc;
 
 const LOCALLY_CACHED_WORKS: usize = 1;
 
+/// The part shared between a GCWorker and the scheduler.
+/// This structure is used for communication, e.g. adding new work packets.
+pub struct GCWorkerShared<VM: VMBinding> {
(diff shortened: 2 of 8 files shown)


src/scheduler/worker.rs
```diff
@@ -97,52 +109,72 @@ impl<VM: VMBinding> GCWorker<VM> {
...
     }
 }
 
 pub struct WorkerGroup<VM: VMBinding> {
-    pub workers: Vec<GCWorker<VM>>,
+    pub workers_shared: Vec<Arc<GCWorkerShared<VM>>>,
+    // Allow vec_box: GCWorker are supposed to be heap-allocated.  This Vec, however, is temporary.
+    // When spawning mutators, the boxes will be moved out of the vec, and become owned by worker threads.
+    #[allow(clippy::vec_box)]
+    workers_to_spawn: Vec<Box<GCWorker<VM>>>,
```
qinsoon, line 130: I feel it is not necessary to use `Vec<Box<T>>` here. `Vec<T>` will also put `T` to the heap. However, if you are considering giving the boxed pointer to native code, then it is okay. Probably update the comment above to state that that later we need to give the `GCWorker` pointer to native, and directly using `GCWorker` is not FFI-safe.

src/scheduler/work.rs
```diff
@@ -11,6 +11,12 @@ use std::any::{type_name, TypeId};
 /// For this case, use `WorkBucket::add_with_priority_unsync` instead.
 pub trait CoordinatorWork<VM: VMBinding>: 'static + Send + GCWork<VM> {}
 
+// Error message for borrowing `GCWorkerShared::stat`.
+#[cfg(feature = "work_packet_stats")]
+const STAT_BORROWED_MSG: &str =
```
qinsoon, line 16: Two `STAT_BORROWED_MSG` constants are declared and they look identical. You can remove one if they are the same. 

I suggest the string does not say about 'concurrent GC'. There could be multiple reasons that leads to this error. You can simply state that 'we try to borrow GCWorker.stat but it was already borrowed', and explain in the comment above this string that 'we mutably borrow GCWorker.stat when doing GC work, and immutably borrow stat in harness_begin/end. If somehow, these two happen at the same time, you will see this error'.

qinsoon (commented):
Looks good to me. Did you test this with `harness_begin/end` used?

I will approve this PR later. I will submit PRs to update V8 and JikesRVM for this PR.

src/scheduler/worker.rs
wks: It is not my intention to put `GCWorker` into a `Vec`.

Ideally, I shall create `Box<GCWorker>`, put its associated `Arc<GCWorkerShared>` into `worker_group.shared`, and then immediately spawn the GC worker thread, transferring the ownership of the `Box<GCWorker>` to the thread.

I tried that, but it didn't work.  Once a GC worker thread is spawned, it will access `Scheduler.worker_group()` immediately when started, by which time the `worker_group` has not been constructed yet.  To work around this, I have to create all `GCWorker`, put all `GCWorkerShared` into the `WorkerGroup` before spawning any GC worker threads.

So this `Vec<Box<_>>` is just a temporary location for `Box<GCWorker>` instances that has not been given to any GC worker threads.  It should not really be a permanent part of the `WorkerGroup`.

I think the presence of `workers_to_spawn` here is a bit confusing. I'll try to remove it from `WorkerGroup` and make it a return value of `WorkerGroup::new`.

src/scheduler/work.rs
wks: Will do.  They are only slightly different.

qinsoon (approved):
LGTM

Issue #520, opened 2022-01-13
Plans with multiple traces


qinsoon:
Each policy provides their `trace_object()` method and that should be used for a GC transitive closure. However, if a plan needs to do multiple traces, do we expect `trace_object()` be used for all the traces?

For example, in mark compact, the GC needs multiple traces. The `MarkCompactSpace` provides several `trace_object()` methods, and each trace will use a different one. However, in mark compact, the common policies (such as LOS) is still in use, and they only provide one `trace_object()`. For the current implementation (ad95d7797f020eaab1e608dc03b7db6a3d79839a), `trace_object()` in LOS may not keep tracing objects that have already been traced in previous traces, which leads to a bug and we do not have a full transitive closure.

There are a few things we need to figure out:
1. Do we expect `trace_object()` to be used for all the traces?
2. If not, should each policy provide a `re_trace_object()` method? What is the difference between trace and re_trace? Do we expect trace to happen before re_trace?
3. Who should provide the `re_trace_object()` for each policy? The plan or the policy itself?

qinsoon:
Discussion on Zulip: https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/Immix.20with.20Evacuation/near/267529252

qinsoon:
> There are a few things we need to figure out:
> 
> 1. Do we expect `trace_object()` to be used for all the traces?

We would expect all the traces to use `trace_object()`. Each policy should provide a `prepare()` function that can help reset some states so `trace_object()` can be used correctly.

> 2. If not, should each policy provide a `re_trace_object()` method? What is the difference between trace and re_trace? Do we expect trace to happen before re_trace?
> 3. Who should provide the `re_trace_object()` for each policy? The plan or the policy itself?

We do not expect a `re_trace_object()` method for each policy.

PR #527, opened 2022-01-20
Prepare plan before compacting in MarkCompact

qinsoon:
This PR closes https://github.com/mmtk/mmtk-core/issues/520.

This PR
1. adds a proper `max_non_los_default_alloc_bytes` for mark compact
2. adds a `prepare()` call for mark compact before the compact transitive closure
3. adds some assertions to check if the two traces in mark compact traverse the same set of objects
4. removes an assertion from LOS prepare

### DIFF
--- a/src/policy/largeobjectspace.rs
+++ b/src/policy/largeobjectspace.rs
@@ -153,7 +153,6 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
 
     pub fn prepare(&mut self, full_heap: bool) {
         if full_heap {
-            debug_assert!(self.treadmill.from_space_empty());
             self.mark_state = MARK_BIT - self.mark_state;
         }
         self.treadmill.flip(full_heap);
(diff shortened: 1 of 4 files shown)


src/policy/largeobjectspace.rs
```diff
@@ -153,7 +153,6 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
 
     pub fn prepare(&mut self, full_heap: bool) {
         if full_heap {
-            debug_assert!(self.treadmill.from_space_empty());
             self.mark_state = MARK_BIT - self.mark_state;
         }
         self.treadmill.flip(full_heap);
```
tianleq, line 158: I feel that flip the from/to space will leak dead objects. After the first trace, live objects have been copied to the to space and what left in from space are dead objects. Now after the flip, the from space will become the to space, and those dead objects will be treated as live. At this stage, sweep has not been executed yet.

qinsoon: You are right. I will work on this.

qinsoon:
https://github.com/mmtk/mmtk-core/pull/944 fixes the issue.

Issue #364, opened 2021-07-01
Remove type_refer from post_alloc in mmtk.h


qinsoon:
The parameter type_refer has been removed. However, `mmtk.h` in our repo is not updated.

k-sareen:
Closed with #530

PR #530, opened 2022-01-27
Add prefix to MMTk API

### LINKED ISSUE #477: Adding prefix to functions in examples/mmtk.h
Currently, the example header files `examples/mmtk.h` and `dummyvm/api/mmtk.h` define functions without any prefixes.  The good practice for the C language is always adding a library-specific prefix, such as `mmtk_`, to all exposed API functions.  (For C++, however, it is better to use namespaces.)

The following is an excerpt of the current header:

```c
extern MMTk_Mutator bind_mutator(void *tls);
extern void destroy_mutator(MMTk_Mutator mutator);

extern void* alloc(MMTk_Mutator mutator, size_t size, size_t align, ssize_t offset, int allocator);
extern void* alloc_slow(MMTk_Mutator mutator, size_t size, size_t align, ssize_t offset, int allocator);
extern void post_alloc(MMTk_Mutator mutator, void* refer, void* type_refer, int bytes, int allocator);

extern bool is_live_object(void* ref);
extern bool is_mapped_object(void* ref);
extern bool is_mapped_address(void* addr);
extern void modify_check(void* ref);
```

Short and generic names, like `alloc` and `modify_check`, can be defined by many programs, and is prone to conflict.  Even GC-specific names, like `is_live_objects` are subject to conflict, too.

The Ruby VM defines its own [`is_live_object` function in its source file `gc.c`](https://github.com/ruby/ruby/blob/027a3379d67922738d503511c2123989229f8d9b/gc.c#L4271).  It is defined for Ruby's own GC.  This function name the same as mmtk-core's [`is_live_object` function in `src/memory_manager.rs`](https://github.com/mmtk/mmtk-core/blob/f46b1071aefc52ac2f9c6c1a91da91dffdfec365/src/memory_manager.rs#L259), but their semantics is very different.  This name conflict has brought trouble to the MMTk Ruby binding.

(Note that Ruby API functions usually have the `rb_` prefix to avoid name conflict. The `is_live_object` is private to `gc.c`, but we need to let `gc.c` include `mmtk.h` in order to support MMTk.)

Although MMTk Core does not provide an official `mmtk.h` header, VM-binding authours usually copy that `mmtk.h` and add Rust code to implement the API functions defined in that header.  Although the binding author should define its C API in the tidiest most convenient way for the VM, in practice, they usually use the `mmtk.h` found in mmtk-core as the baseline.  Apparently the OpenJDK, V8, Ruby bindings have all done the same.

-   mmtk-openjdk:
    +   Header: https://github.com/mmtk/mmtk-openjdk/blob/c1bc366e909ace78c52e62c7faf2b0a80b54f348/openjdk/mmtk.h
    +   Rust impl: https://github.com/mmtk/mmtk-openjdk/blob/c1bc366e909ace78c52e62c7faf2b0a80b54f348/mmtk/src/api.rs
-   mmtk-v8:
    +   Header: https://github.com/mmtk/mmtk-v8/blob/310c611e8e217691af2b634a85c76dff63125c90/v8/third_party/heap/mmtk/mmtk.h
    +   Rust impl: https://github.com/mmtk/mmtk-v8/blob/310c611e8e217691af2b634a85c76dff63125c90/mmtk/src/api.rs
-   mmtk-jikesrvm:
    +   Header: https://github.com/mmtk/mmtk-jikesrvm/blob/ccf97c0b6ceefb787ebd22d88968c58913cf0d28/mmtk/api/mmtk.h
    +   Rust impl: https://github.com/mmtk/mmtk-jikesrvm/blob/ccf97c0b6ceefb787ebd22d88968c58913cf0d28/mmtk/src/api.rs
-   mmtk-ruby (previously):
    +   Header: https://github.com/wks/ruby/blob/a6b11e56687feb4f18759cc0b13114a7cde0c28a/mmtk.h
    +   Rust impl: https://github.com/wks/mmtk-ruby/blob/9035c6d22010a0b09a20ada3615aad258b0907a0/mmtk/src/api.rs

It is interesting to find that the JikesRVM binding used the prefix `jikesrvm_` for some JikesRVM-specific functions that needs to be implemented in Rust.

I suggest we add the `mmtk_` prefix to the example headers in `mmtk-core/examples/mmtk.h`, so that future binding writers know that prefixes are useful for avoiding name conflict.

I updated the [`mmtk.h` in the Ruby binding](https://github.com/wks/ruby/blob/third-party-heap-3-0/mmtk.h).  It can be used as a reference.  (p.s. I just copied the file from mmtk-core.  Some reference-processing functions may not be relevant to Ruby.)

k-sareen:
Fixes #477 and #364 

I've also added brief comments to the exposed functions in the API. My understanding is that `TraceLocal` has been superseded by work packets, hence I have completely removed them from the header. Header was formatted using `clang-format`'s Chromium option.

(Also I'm not sure why `alloc_slow` is exposed in the API -- I haven't removed it, but I think it might be an artifact of refactoring)

### DIFF
--- a/examples/allocation_benchmark.c
+++ b/examples/allocation_benchmark.c
@@ -5,11 +5,11 @@
 
 int main() {
     volatile uint64_t * tmp;
-    gc_init(1024*1024*1024);
-    MMTk_Mutator handle = bind_mutator(0);
+    mmtk_gc_init(1024*1024*1024);
+    MMTk_Mutator handle = mmtk_bind_mutator(0);
 
     for (int i=0; i<1024*1024*100; i++) {
-        tmp = alloc(handle, 8, 1, 0, 0);
+        tmp = mmtk_alloc(handle, 8, 1, 0, 0);
         #ifdef STORE
             *tmp = 42;
         #endif
--- a/examples/main.c
+++ b/examples/main.c
@@ -2,13 +2,13 @@
 #include "mmtk.h"
 
 int main(int argc, char* argv[]){
-    gc_init(1024*1024);
+    mmtk_gc_init(1024*1024);
 
-    MMTk_Mutator handle = bind_mutator(0);
-    
+    MMTk_Mutator handle = mmtk_bind_mutator(0);
+
     for (int i=0;i<4;i++){
         int arr_size = 10000;
-        int* my_arr = alloc(handle, sizeof(int)*arr_size, 8, 0, 0);
+        int* my_arr = mmtk_alloc(handle, sizeof(int)*arr_size, 8, 0, 0);
         if (!my_arr){
             printf("OOM\n");
             break;
--- a/examples/mmtk.h
+++ b/examples/mmtk.h
@@ -1,8 +1,9 @@
 // This is an example of native API for the single instance MMTk.
 
-// Note: the mmtk core does not directly provide this API. However, it provides a similar multi-instance Rust API.
-// A VM binding should write their own C header file (possibly based on this example with their own extension and modification),
-// and expose the Rust API based on their native API.
+// Note: the mmtk core does not directly provide this API. However, it provides
+// a similar multi-instance Rust API.  A VM binding should write their own C
+// header file (possibly based on this example with their own extension and
+// modification), and expose the Rust API based on their native API.
 
 #ifndef MMTK_H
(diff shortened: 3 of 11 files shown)


wks (approved):
LGTM

Issue #532, opened 2022-01-29
Refactor scheduler


wks:
## TODO list

-   [x] Introduce a context struct for the GC controller, and let it own the data it needs.
-   [x] Refactor the initialisation of `GCWorkScheduler` and eliminate the need to use excessive `Option<T>` fields.
-   [ ] Move ~~`ControllerCollectorContext`~~ `GCRequester` away from `BasePlan`.  Let global states be global.
        -   Related issue: https://github.com/mmtk/mmtk-core/issues/587

## Problem

Currently, **the `GCWorkScheduler` serves multiple purposes**.
1.  It owns the `GCWorker` for the GC controller thread.  This part should be owned by the GC controller thread.
2.  It serves as the communication hub between mutators, the controller and workers.
    -   Mutators may add work packets in write barrier.
    -   workers poll work from the scheduler
    -   the controller queries worker states, and also get coordinator work from the scheduler
    This part should be shared between threads.

On the other hand, **there is currently no "context" object for the GC controller thread**.
-   `ControllerCollectorContext` contains the `run()` method for the GC controller, but it also serves as the communication hub for triggering GC.
-   It depends on the `GCWorker` in the scheduler.  It spends most of its lifetime in `GCWorkScheduler::wait_for_completion`. Even its communication channel is owned by `GCWorkScheduler.channel.1`.

And the **`ControllerCollectorContext`** -- which apparently has nothing to do with a "plan" --** is buried deep inside `mmtk.plan.base().controller_collector_context`**.  We consider a "plan" to be a description of a GC algorithm, but `BasePlan` holds mostly the global GC states.  Understandably, the controller thread itself is one of such global state, but it is really plan-agnostic.

The **initialisation process is complicated**.  When initialising, all GC algorithms pass a `scheduler` object to the `init()` function of the `ControllerCollectorContext`. `GCWorkScheduler` also uses many `Option<T>` fields and initializes itself unsafely.

These phenomena are a result of porting MMTk from Java to Rust.  Java's programming style promotes shared states and inter-referenced objects, while Rust emphasises ownership.  This requires us to think more carefully about the ownership of data.

## Goal

The goal is to ensure each thread owns the data that it accesses exclusively, and ensure shared data are properly identified, shared and synchronised.  Threads we concern include:

-   The GC controller
-   GC workers (already [fixed](https://github.com/mmtk/mmtk-core/pull/523))

[...]

PR #533, opened 2022-01-29
Introduce a context object for GC controller thread

wks:
This is one of a series of planned changes to refactor the scheduler.  The problems are described in issue https://github.com/mmtk/mmtk-core/issues/532

This PR introduces the `GCController` struct.  Like `GCWorker`, a `GCController` shall be owned by the GC controller thread.  It owns its `GCWorker` (the `coordinator_worker`) and the receiving end of the worker-to-controller channel. It also has references to the data structures it needs to work with, including the scheduler and the `ControllerCollectorContext` which is used to trigger GC.

Because of this new struct, some related fields and operations are removed from `ControllerCollectorContext` and `GCWorkScheduler`.

There is also an API change.  `spawn_worker_thread` is renamed to `spawn_gc_thread`.  Now both GC worker threads and the GC controller thread are spawn with a context object.

This PR does not change the two-phase initialisation of `GCWorkScheduler`.  It will be done in another PR.

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -25,21 +25,6 @@ use crate::util::{Address, ObjectReference};
 use crate::vm::VMBinding;
 use std::sync::atomic::Ordering;
 
-/// Run the main loop for the GC controller thread. This method does not return.
-///
-/// Arguments:
-/// * `tls`: The thread that will be used as the GC controller.
-/// * `gc_controller`: The execution context of the GC controller threa.
-///   It is the `GCController` passed to `Collection::spawn_gc_thread`.
-/// * `mmtk`: A reference to an MMTk instance.
-pub fn start_control_collector<VM: VMBinding>(
-    tls: VMWorkerThread,
-    gc_controller: &mut GCController<VM>,
-    _mmtk: &'static MMTK<VM>,
-) {
-    gc_controller.run(tls);
-}
-
 /// Initialize an MMTk instance. A VM should call this method after creating an [MMTK](../mmtk/struct.MMTK.html)
 /// instance but before using any of the methods provided in MMTk. This method will attempt to initialize a
 /// logger. If the VM would like to use its own logger, it should initialize the logger before calling this method.
@@ -167,6 +152,21 @@ pub fn get_allocator_mapping<VM: VMBinding>(
     mmtk.plan.get_allocator_mapping()[semantics]
 }
 
+/// Run the main loop for the GC controller thread. This method does not return.
+///
+/// Arguments:
+/// * `tls`: The thread that will be used as the GC controller.
+/// * `gc_controller`: The execution context of the GC controller threa.
+///   It is the `GCController` passed to `Collection::spawn_gc_thread`.
+/// * `mmtk`: A reference to an MMTk instance.
+pub fn start_control_collector<VM: VMBinding>(
+    _mmtk: &'static MMTK<VM>,
+    tls: VMWorkerThread,
+    gc_controller: &mut GCController<VM>,
+) {
+    gc_controller.run(tls);
+}
+
 /// Run the main loop of a GC worker. This method does not return.
 ///
 /// Arguments:
@@ -175,9 +175,9 @@ pub fn get_allocator_mapping<VM: VMBinding>(
 ///   It is the `GCWorker` passed to `Collection::spawn_gc_thread`.
 /// * `mmtk`: A reference to an MMTk instance.
 pub fn start_worker<VM: VMBinding>(
+    mmtk: &'static MMTK<VM>,
     tls: VMWorkerThread,
     worker: &mut GCWorker<VM>,
-    mmtk: &'static MMTK<VM>,
 ) {
     worker.run(tls, mmtk);
 }
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -15,14 +15,13 @@ use crate::mmtk::MMTK;
 use crate::plan::AllocationSemantics;
 use crate::plan::{Mutator, MutatorContext};
 use crate::scheduler::WorkBucketStage;
-use crate::scheduler::{GCWork, GCWorker};
+use crate::scheduler::{GCController, GCWork, GCWorker};
 use crate::util::alloc::allocators::AllocatorSelector;
 use crate::util::constants::{LOG_BYTES_IN_PAGE, MIN_OBJECT_SIZE};
 use crate::util::heap::layout::vm_layout_constants::HEAP_END;
 use crate::util::heap::layout::vm_layout_constants::HEAP_START;
 use crate::util::opaque_pointer::*;
 use crate::util::{Address, ObjectReference};
-use crate::vm::Collection;
 use crate::vm::VMBinding;
 use std::sync::atomic::Ordering;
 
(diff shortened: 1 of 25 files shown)


src/memory_manager.rs
```diff
@@ -31,8 +30,12 @@ use std::sync::atomic::Ordering;
 /// Arguments:
 /// * `mmtk`: A reference to an MMTk instance.
 /// * `tls`: The thread that will be used as the GC controller.
```
KipHamiltons, line 32: These arguments have changed

wks: Yes. I need to update the comments as well.

src/plan/controller_collector_context.rs
```diff
@@ -17,7 +11,6 @@ struct RequestSync {
 pub struct ControllerCollectorContext<VM: VMBinding> {
```
qinsoon, line 11: It feels like we can change this type to `GCRequest`, as all of its fields and methods are about GC requests.

src/scheduler/controller.rs
```diff
@@ -0,0 +1,102 @@
...
+            }
+            let _guard = self.scheduler.worker_monitor.0.lock().unwrap();
+            if self.scheduler.worker_group().all_parked() && self.scheduler.all_buckets_empty() {
+                break;
+            }
+        }
+        for message in self.receiver.try_iter() {
+            if let CoordinatorMessage::Work(mut work) = message {
+                work.do_work_with_stat(worker, mmtk);
+                //self.process_coordinator_work(work);
```
qinsoon, line 88: This line could be removed. Alternatively, you can implement this method in this struct and use it instead of `work.do_work_with_stat()`. It may be helpful if we want to log stuff for coordinator work.

src/scheduler/controller.rs
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+        self.scheduler.deactivate_all();
+        // Finalization: Resume mutators, reset gc states
+        // Note: Resume-mutators must happen after all work buckets are closed.
+        //       Otherwise, for generational GCs, workers will receive and process
+        //       newly generated remembered-sets from those open buckets.
+        //       But these remsets should be preserved until next GC.
+        if let Some(mut finalizer) = self.scheduler.take_finalizer() {
```
qinsoon, line 97: Similar to the 'initializer', the 'finalizer' here is always `EndOfGC`. We could possibly remove the 'finalizer' as well (it is also a confusing name).

src/scheduler/scheduler.rs
```diff
@@ -43,7 +39,7 @@ pub struct GCWorkScheduler<VM: VMBinding> {
     /// We use this callback to process ephemeron objects. `closure_end` can re-enable
     /// the `Closure` bucket multiple times to iteratively discover and process
     /// more ephemeron objects.
-    closure_end: Mutex<Option<Box<dyn Send + Fn() -> bool>>>,
+    pub closure_end: Mutex<Option<Box<dyn Send + Fn() -> bool>>>,
```
qinsoon, line 42: It seems this does not need to be public.

src/scheduler/controller.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::MMTK;
+
+use super::{GCWorkScheduler, GCWorker};
+
+pub struct GCController<VM: VMBinding> {
+    mmtk: &'static MMTK<VM>,
+    requester: Arc<ControllerCollectorContext<VM>>,
+    scheduler: Arc<GCWorkScheduler<VM>>,
+    receiver: Receiver<CoordinatorMessage<VM>>,
+    coordinator_worker: Box<GCWorker<VM>>,
```
qinsoon, line 20: Is there any reason that this needs to be boxed?

qinsoon (commented):
Very nice refactoring. It makes the code much cleaner. As this PR is still a draft, I did not approve it. But I do not see any major issue with the PR.

src/memory_manager.rs
```diff
@@ -15,24 +15,29 @@ use crate::mmtk::MMTK;
...
 ///
 /// Arguments:
-/// * `mmtk`: A reference to an MMTk instance.
 /// * `tls`: The thread that will be used as the GC controller.
-pub fn start_control_collector<VM: VMBinding>(mmtk: &MMTK<VM>, tls: VMWorkerThread) {
-    mmtk.plan.base().control_collector_context.run(tls);
+/// * `gc_controller`: The execution context of the GC controller threa.
+///   It is the `GCController` passed to `Collection::spawn_gc_thread`.
+/// * `mmtk`: A reference to an MMTk instance.
+pub fn start_control_collector<VM: VMBinding>(
```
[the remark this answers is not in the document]
wks: Good idea.

src/scheduler/controller.rs
wks: I'll use `work.do_work_with_stata(worker, mmtk);`.  Using `self.` will borrow `self`, and Rust seems to have problem borrowing `self` mutably and `self.receiver` immutably at the same time.

src/scheduler/controller.rs
wks: That makes sense.  I think the `GCWorkScheduler` will also no longer need to schedule `EndOfGC` explicitly.

src/scheduler/controller.rs
wks: The only reason is for consistency with GC worker threads which have `Box<GCWorker>` instances.  But it doesn't have to be boxed here in `GCCoordinator`.  I'll try to remove the `Box` here.

src/plan/controller_collector_context.rs
wks: I'll call it `GCRequester` so as not to be confused by a concrete request for collection.  It is an object for making requests.

src/scheduler/controller.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::plan::controller_collector_context::ControllerCollectorContext;
+use crate::scheduler::gc_work::ScheduleCollection;
+use crate::scheduler::CoordinatorMessage;
+use crate::util::VMWorkerThread;
+use crate::vm::VMBinding;
+use crate::MMTK;
+
+use super::{GCWorkScheduler, GCWorker};
+
+pub struct GCController<VM: VMBinding> {
```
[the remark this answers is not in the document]
wks: I added comments to `GCController`.  I also added comments to `GCWorker` and the head of `controller.rs`.  Please check if my descriptions are accurate.

wks:
@qinsoon I made changes according to your suggestions.  I set this PR to be ready for review.

When I was reading the code of `GCWorker`, I think some methods of `GCWorkScheduler` can be moved to `GCWorker`, too, just like what I have done to `GCController`.  I'll do it in the next PR.

qinsoon (approved):
LGTM

qinsoon:
Binding updates for this PR:
* OpenJDK: https://github.com/mmtk/mmtk-openjdk/pull/135
* JikesRVM:
  * https://github.com/mmtk/mmtk-jikesrvm/pull/98
  * https://github.com/mmtk/jikesrvm/pull/9
* V8: https://github.com/mmtk/mmtk-v8/pull/53

PR #535, opened 2022-02-01
Update OOM semantics and expectations

### LINKED ISSUE #534: Handling Out of Memory errors
Currently MMTk's contract with the VM expects that the VM crashes or aborts immediately when MMTk calls `out_of_memory()` [1]. This is expectation is too strong and can lead to issues when trying to add proper OOM error reporting to bindings such as OpenJDK.

For more details, see this comment in the OpenJDK binding: https://github.com/mmtk/mmtk-openjdk/issues/78#issuecomment-1025287199

\[1]: https://github.com/mmtk/mmtk-core/blob/4851e7ac7e8a7bd70ce65935c4ac1ff430cab6f1/src/vm/collection.rs#L66-L73

k-sareen:
**Note: API breaking change!**

Adds `MmtkAllocationError` to differentiate between the two (main) types of OOM:
1. Critical OOM: This is the case where the OS is unable to mmap or acquire more memory. MMTk expects the VM to abort immediately if such an error is thrown.
2. Heap OOM: This is the case where the specified heap size is insufficient to execute the application. MMTk expects the binding to notify the VM about this OOM and throw an OOM error *at some point later in time*. That is to say, MMTk does not expect the VM to abort immediately.

MMTk now returns `null` in case of a Heap OOM error.

This PR also adds missing documentation for the `Allocator` trait. Please check if it makes sense.

Associated mmtk-openjdk PR: https://github.com/mmtk/mmtk-openjdk/pull/136
Associated mmtk-jikesrvm PR: https://github.com/mmtk/mmtk-jikesrvm/pull/99

Fixes #534.


src/plan/generational/copying/global.rs
```diff
@@ -138,7 +138,12 @@ impl<VM: VMBinding> Plan for GenCopy<VM> {
     /// Return the number of pages avilable for allocation. Assuming all future allocations goes to nursery.
     fn get_pages_avail(&self) -> usize {
         // super.get_pages_avail() / 2 to reserve pages for copying
-        (self.get_total_pages() - self.get_pages_reserved()) >> 1
```
k-sareen, line 141: I noticed this causing an issue with some test cases. It might be an issue for the [default implementation](https://github.com/mmtk/mmtk-core/blob/master/src/plan/global.rs#L264-L266) as well, but we've never triggered it.

k-sareen:
Not sure what the error is in the test? The strings should match. Does it not like that it's a generated string now instead of a constant string?

src/plan/generational/copying/global.rs
qinsoon: The total pages should not be greater than or equal to reserved pages. If this panics, probably there is a bug somewhere else about page accounting. I don't suggest simply using `saturating_sub`.

k-sareen:
Fixed.

Also note that originally I tried to write idiomatic Rust code, returning a `Result` type for the allocation and returning an `Err` in case we were out of memory. However, this was a real pain to work with due to `Result` types not being FFI-safe, the massive change in allocation API, as well as still being hard to implement in bindings. You can find my original attempt here: https://github.com/k-sareen/mmtk-core/commit/3284a33a1fa00d2d19ff9ded35c09ba2ca5b5212

Hence, I opted to simply return null in case of an OOM and let the binding decide how to proceed.

src/plan/generational/copying/global.rs
k-sareen: Is that true? It seems like `collection_required()` expects it to hold sometimes: https://github.com/mmtk/mmtk-core/blob/master/src/plan/global.rs#L784

qinsoon: Ah, you are right. For failed allocation (OOM), it could happen. We will increase reserved pages temporarily, and check if that exceeds total pages. If so, we will deduct the temporary reserved pages and trigger a GC.

Did you fix this because it panics here with changes in this PR?

k-sareen: I honestly didn't try with master. I just noticed GenCopy failing in debug builds due to subtraction overflow while I was testing my implementation. I will check if GenCopy fails with master.

k-sareen: It does fail on master.

qinsoon: What did you run to trigger the bug? Can you post the command line?

k-sareen:
```
RUST_BACKTRACE=1 RUST_LOG= MMTK_PLAN=GenCopy ./build/linux-x86_64-normal-server-fastdebug/jdk/bin/java -XX:+UseThirdPartyHeap -Xms5M -Xmx5M -jar /usr/share/benchmarks/dacapo/dacapo-evaluation-git-29a657f.jar zxing
--------------------------------------------------------------------------------
IMPORTANT NOTICE:  This is NOT a release build of the DaCapo suite.
Since it is not an official release of the DaCapo suite, care must be taken when
using the suite, and any use of the build must be sure to note that it is not an
offical release, and should note the relevant git hash.

Feedback is greatly appreciated.   The preferred mode of feedback is via github.
Please use our github page to create an issue or a pull request.
    https://github.com/dacapobench/dacapobench.
--------------------------------------------------------------------------------

thread '<unnamed>' panicked at 'attempt to subtract with overflow', /home/kunals/git/mmtk-core/src/plan/generational/copying/global.rs:141:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/efec545293b9263be9edfb283a7aa66350b3acbf/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/efec545293b9263be9edfb283a7aa66350b3acbf/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at /rustc/efec545293b9263be9edfb283a7aa66350b3acbf/library/core/src/panicking.rs:48:5
   3: <mmtk::plan::generational::copying::global::GenCopy<VM> as mmtk::plan::global::Plan>::get_pages_avail
             at /home/kunals/git/mmtk-core/src/plan/generational/copying/global.rs:141:9
   4: mmtk::plan::generational::global::Gen<VM>::should_next_gc_be_full_heap
             at /home/kunals/git/mmtk-core/src/plan/generational/global.rs:217:9
   5: <mmtk::plan::generational::copying::global::GenCopy<VM> as mmtk::plan::global::Plan>::release
             at /home/kunals/git/mmtk-core/src/plan/generational/copying/global.rs:127:36
   6: <mmtk::scheduler::gc_work::Release<C> as mmtk::scheduler::work::GCWork<<C as mmtk::scheduler::work::GCWorkContext>::VM>>::do_work
             at /home/kunals/git/mmtk-core/src/scheduler/gc_work.rs:114:9
   7: mmtk::scheduler::work::GCWork::do_work_with_stat
             at /home/kunals/git/mmtk-core/src/scheduler/work.rs:37:9
   8: mmtk::scheduler::worker::GCWorker<VM>::run
             at /home/kunals/git/mmtk-core/src/scheduler/worker.rs:132:13
   9: mmtk::memory_manager::start_worker
             at /home/kunals/git/mmtk-core/src/memory_manager.rs:177:5
  10: start_worker
             at /home/kunals/git/mmtk-openjdk/mmtk/src/api.rs:124:5
  11: _ZN6Thread8call_runEv
             at ./src/hotspot/share/runtime/thread.cpp:385
  12: thread_native_entry
             at ./src/hotspot/os/linux/os_linux.cpp:787
  13: start_thread
             at /build/glibc-S9d2JN/glibc-2.27/nptl/pthread_create.c:463
  14: __GI___clone
             at /build/glibc-S9d2JN/glibc-2.27/misc/../sysdeps/unix/sysv/linux/x86_64/clone.S:95
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
zsh: abort (core dumped)  RUST_BACKTRACE=1 RUST_LOG= MMTK_PLAN=GenCopy  -XX:+UseThirdPartyHeap -Xms5M
```

qinsoon: It seems there is a bug in our page accounting that after a GC, we end up with more used pages than before the GC. I will look further into this.

k-sareen: Should I remove this change then? It only hides the issue instead of fixes it.

k-sareen:
Note that mmtk-v8 does not currently override `out_of_collection()` [1]. Hence, it will panic in Rust when an OOM occurs and we will never have the case where mmtk-core has returned a `null` pointer back to the VM. I am not sure if @wenyuzhao has changed the OOM mechanism in his version of the binding, however.

src/plan/generational/copying/global.rs
qinsoon: I think this is actually possible (see: https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/copy-space-pages-used/near/270335481). I think the fix is right, and we possibly want to fix all of the occurrences of `total_pages - reserved pages`, and make it clear that it is possible that the reserved pages is larger than the total pages.

k-sareen: Right. Should I fix the other ones in this PR as well or will you create a new PR for the changes?

qinsoon: Either sounds good to me. Making it in a separate PR is cleaner, and we can quickly get it merged even before this PR.

k-sareen: I agree. I'll remove the change from this PR then.

k-sareen: Done. I think it would be best if you made the PR fixing the above since you've debugged it and understand the issue better than me. I'll mark this conversation as resolved. Thanks for looking into it!

wks:
> Also note that originally I tried to write idiomatic Rust code, returning a `Result` type for the allocation and returning an `Err` in case we were out of memory. However, this was a real pain to work with due to `Result` types not being FFI-safe, the massive change in allocation API, as well as still being hard to implement in bindings.

What about letting `memory_manager::alloc` and `Mutator::alloc` return `Result<Address, MmtkAllocationError>`, and letting the binding create its own `extern "C" fn mmtk_alloc(...) -> Address`?  This makes Rust code idiomatic, and C API functions FFI-safe.

k-sareen:
> > Also note that originally I tried to write idiomatic Rust code, returning a `Result` type for the allocation and returning an `Err` in case we were out of memory. However, this was a real pain to work with due to `Result` types not being FFI-safe, the massive change in allocation API, as well as still being hard to implement in bindings.
> 
> What about letting `memory_manager::alloc` and `Mutator::alloc` return `Result<Address, MmtkAllocationError>`, and letting the binding create its own `extern "C" fn mmtk_alloc(...) -> Address`? This makes Rust code idiomatic, and C API functions FFI-safe.

There's still a large area that that changes. Not to mention you're still going to end up returning `null` in the binding function when you have an `Err` since we need to hook into the VM's OOM error mechanism so I don't exactly see its purpose other than writing idiomatic Rust code.

qinsoon:
> > Also note that originally I tried to write idiomatic Rust code, returning a `Result` type for the allocation and returning an `Err` in case we were out of memory. However, this was a real pain to work with due to `Result` types not being FFI-safe, the massive change in allocation API, as well as still being hard to implement in bindings.
> 
> What about letting `memory_manager::alloc` and `Mutator::alloc` return `Result<Address, MmtkAllocationError>`, and letting the binding create its own `extern "C" fn mmtk_alloc(...) -> Address`? This makes Rust code idiomatic, and C API functions FFI-safe.

It is good to have idiomatic code. But the change is not necessary to be included in this PR. I think it is fine that we accept changes in this PR (allowing null pointer returned by `alloc()`), and later we can do the refactoring as suggested.

k-sareen:
I tried to make another MVP (it's not complete yet) but it's a bit messy since `Address::ZERO` is used internally in a lot of places. Instead of returning `Ok(Address::ZERO)`, I opted to make a new error called `SpaceOutOfMemory` and return that. However this results in a lot of the `?`-operator being sprayed around everywhere (as we expect `Address` to be used everywhere instead of the new `Allocation` type) which will probably result in some performance loss. Though it is all technically in the slowpath of an allocation, so potentially it doesn't matter.

src/util/alloc/allocator.rs
```diff
@@ -9,6 +9,16 @@ use crate::vm::VMBinding;
 use crate::vm::{ActivePlan, Collection};
 use downcast_rs::Downcast;
 
+#[repr(C)]
+#[derive(Debug)]
+/// A list of errors that MMTk can encounter during allocation.
+pub enum MmtkAllocationError {
```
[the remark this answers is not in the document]
qinsoon: Any opinion on this? @wks @caizixian

src/util/alloc/allocator.rs
```diff
@@ -102,32 +116,70 @@ pub fn get_maximum_aligned_size<VM: VMBinding>(
...
+    /// Note that in the case where the VM is out of memory, we return a null pointer back to the
+    /// VM and expect that the OOM error is thrown by the binding.
+    ///
+    /// Arguments:
+    /// * `size`: the allocation size in bytes.
+    /// * `align`: the required alignment in bytes.
+    /// * `offset` the required offset in bytes.
     fn alloc(&mut self, size: usize, align: usize, offset: isize) -> Address;
 
+    /// Slowpath allocation attempt. We break up the slowpath implementation into two parts
```
qinsoon, line 157: I don't know what are the two parts of our slowpath implementation.

src/util/alloc/allocator.rs
```diff
@@ -102,32 +116,70 @@ pub fn get_maximum_aligned_size<VM: VMBinding>(
...
         self.alloc_slow_inline(size, align, offset)
     }
 
+    /// Slowpath allocation attempt. This function executes the actual slowpath allocation.  A
+    /// slowpath allocation in MMTk attempts to allocate the object using the per-allocator
+    /// definition of [`alloc_slow_once`]. This function also accounts for increasing the
+    /// allocation bytes in order to support stress testing. In case precise stress testing is
+    /// being used, the [`alloc_slow_once_precise_stress`] function is used instead.
+    ///
+    /// Note that in the case where the VM is out of memory, we throw a [`HeapOutOfMemory`] error
```
qinsoon, line 175: Probably be clear that we will invoke `Collection::out_of_memory()` with `HeapOutOfMemory`. In Rust, there is no 'throw'. We probably better try not confuse readers.

src/util/alloc/allocator.rs
```diff
@@ -102,32 +116,70 @@ pub fn get_maximum_aligned_size<VM: VMBinding>(
...
+    /// and return a null pointer back to the VM with the expectation that the OOM error is thrown
```
qinsoon, line 176: I would suggest that we do not make any assumption about how a binding would implement `out_of_memory()` -- they could just panic, or return and throw their exception later. It is the binding's choice, and we do not need to make any assumption on that.

We only need to make it clear that in case of OOM we will invoke `out_of_memory()`, and then the method will return null pointer.

src/vm/collection.rs
```diff
@@ -63,13 +64,22 @@ pub trait Collection<VM: VMBinding> {
...
 
-    /// Inform the VM for an out-of-memory error. The VM can implement its own error routine for OOM.
-    /// Note the VM needs to fail in this call. We do not expect the VM to resume in any way.
+    /// Inform the VM of an out-of-memory error. The binding should hook into the VM's error
+    /// routine for OOM. Note that there are two different categories of OOM:
+    ///  * Critical OOM: This is the case where the OS is unable to mmap or acquire more memory.
+    ///    MMTk expects the VM to abort immediately if such an error is thrown.
+    ///  * Heap OOM: This is the case where the specified heap size is insufficient to execute the
+    ///    application. MMTk expects the binding to notify the VM about this OOM and throw an OOM
+    ///    error *at some point later in time*. That is to say, MMTk does not expect the VM to
```
qinsoon, line 73: We can leave the choice to the VM whether they want to return or panic.

qinsoon (approved):
LGTM. I have some minor comments mostly about documentation.

src/util/alloc/allocator.rs
k-sareen: Yes, I agree. "Parts" is misleading. I mean to say different functions since one is `#[inline(never)]` for cache considerations (to the best of my understanding).

src/util/alloc/allocator.rs
k-sareen: Yes. I agree. I defaulted to Java terminology :sweat_smile:

src/vm/collection.rs
k-sareen: Yes I will make it more explicit.

src/util/alloc/allocator.rs
```diff
@@ -26,6 +36,10 @@ pub fn align_allocation<VM: VMBinding>(
     known_alignment: usize,
     fillalignmentgap: bool,
 ) -> Address {
+    if region.is_zero() {
```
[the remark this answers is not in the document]
k-sareen: This was where one of the cases failed. Hmm I've deleted the test output files unfortunately. I will try to recreate it and see where the error occurred (from memory it was due to the `debug_assert!(!(fillalignmentgap && region.is_zero()));`).

src/util/alloc/allocator.rs
k-sareen: I agree. I thought that's essentially what I said, but "expectation" is too strong. I will update the comments.

src/util/alloc/allocator.rs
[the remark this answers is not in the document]
k-sareen: Right so it was an issue with `LargeObjectAllocator`:
```
RUST_BACKTRACE=1 RUST_LOG= MMTK_PLAN=SemiSpace ./build/linux-x86_64-normal-server-fastdebug/jdk/bin/java -XX:+UseThirdPartyHeap -Xms5M -Xmx5M OOMTest
thread '<unnamed>' panicked at 'assertion failed: !(fillalignmentgap && region.is_zero())', /home/kunals/git/mmtk-core/src/util/alloc/allocator.rs:49:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/efec545293b9263be9edfb283a7aa66350b3acbf/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/efec545293b9263be9edfb283a7aa66350b3acbf/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at /rustc/efec545293b9263be9edfb283a7aa66350b3acbf/library/core/src/panicking.rs:48:5
   3: mmtk::util::alloc::allocator::align_allocation
             at /home/kunals/git/mmtk-core/src/util/alloc/allocator.rs:49:5
   4: <mmtk::util::alloc::large_object_allocator::LargeObjectAllocator<VM> as mmtk::util::alloc::allocator::Allocator<VM>>::alloc
             at /home/kunals/git/mmtk-core/src/util/alloc/large_object_allocator.rs:35:9
   5: <mmtk::plan::mutator_context::Mutator<VM> as mmtk::plan::mutator_context::MutatorContext<VM>>::alloc
             at /home/kunals/git/mmtk-core/src/plan/mutator_context.rs:83:9
   6: mmtk::memory_manager::alloc
             at /home/kunals/git/mmtk-core/src/memory_manager.rs:132:5
   7: alloc
             at /home/kunals/git/mmtk-openjdk/mmtk/src/api.rs:86:5
   8: _ZN18MMTkMutatorContext5allocEm9Allocator
             at ./../../openjdk/mmtkMutator.cpp:21
   9: _ZNK12MemAllocator21allocate_outside_tlabERNS_10AllocationE
             at ./src/hotspot/share/gc/shared/memAllocator.cpp:272
  10: _ZNK12MemAllocator12mem_allocateERNS_10AllocationE
             at ./src/hotspot/share/gc/shared/memAllocator.cpp:370
  11: _ZNK12MemAllocator8allocateEv
             at ./src/hotspot/share/gc/shared/memAllocator.cpp:377
  12: _ZN13CollectedHeap14array_allocateEP5KlassiibP6Thread
             at ./src/hotspot/share/gc/shared/collectedHeap.cpp:457
  13: _ZN13InstanceKlass17allocate_objArrayEiiP6Thread
             at ./src/hotspot/share/oops/instanceKlass.cpp:1223
  14: _ZN10oopFactory12new_objArrayEP5KlassiP6Thread
             at ./src/hotspot/share/memory/oopFactory.cpp:87
  15: _ZN18InterpreterRuntime9anewarrayEP10JavaThreadP12ConstantPoolii
             at ./src/hotspot/share/interpreter/interpreterRuntime.cpp:254
  16: <unknown>
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

src/util/alloc/allocator.rs
k-sareen: Do you have a suggestion as to what terminology to use? I think using "functions" would work. Alternatively, I can remove the sentence entirely as they may confuse the reader (although it would be good to document optimizations if we can).

src/util/alloc/allocator.rs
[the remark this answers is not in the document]
k-sareen: I think the problem is only for the LOS Allocator so I could potentially move the null check to it's allocation function instead of here. I will have to think more about if that change will work.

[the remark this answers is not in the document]
k-sareen: Right -- it does seem like other allocators will use their cursor and the only place where this is could be an issue is the LOS allocator. I can't think of a case where cursor should ever be `null` (other than initialization of course) since we will never set the limit to be `null` (at least as far as I can see). I think the current `debug_assert!()` suffices as a check and I will move the null check to the LOS allocator.

src/util/alloc/allocator.rs
qinsoon: If you are talking about `alloc_slow()` vs `alloc_slow_inline()` vs `alloc_slow_once()`, this is my guess, after checking the Java MMTk:
* `alloc_slow()` is the slowpath function that usually will be used. It is explicitly out of line for performance. It may be overriden by some allocators and implement their own slowpath semantics (e.g. bump pointer allocator in Java MMTk overrides this method and implements card marking).
* `allow_slow_once()` is a single slowpath allocation attempt. Every allocator needs to implement it.
* `alloc_slow_inline()` has the logic of retrying slowpath allocation and is responsible of dealing with OOM. We do not expect any allocator to override it. The method exists because some allocators like the immix allocator use an intermediate path (something between fastpath and slowpath), and call this in the intermediate path. By default, `alloc_slow()` just calls `alloc_slow_inline()`.

k-sareen:
I think I've addressed most of the PR comments. Can you please review again @qinsoon?

src/util/alloc/allocator.rs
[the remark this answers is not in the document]
k-sareen: It seems like the OpenJDK binding uses `MMTk` primarily. I don't have a preference, so I can change it to `MMTk` if you feel it is more consistent.

qinsoon:
Looks good. The change in LOS looks fine to me. 

I am just wondering one thing: for mmap failure, should we actually call to the binding? Basically, the binding cannot do anything for it at all. Why don't we just panic in Rust? Alternatively, we can have a method `fatal_error()` if we want to panic in the binding. In that case, `fatal_error() -> !` won't return and we allow VM to panic in their own way. We can use it for all the cases where we panic in mmtk-core. Anyway, this is beyond the scope of this PR. We can do that change later. I think the PR is good for now.

src/util/alloc/allocator.rs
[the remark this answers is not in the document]
qinsoon: If no one advocates for any style we should use, I would suggest calliong it `MMTKAllocationError` -- this is concistent with the naming of `MMTK` and `VMBinding` in our code base.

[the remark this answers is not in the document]
k-sareen: Alternatively, we could simply call it `AllocationError` or `AllocError` in order to be consistent with everything internally (for example, we don't call it `MMTKMutator` inside mmtk core).

k-sareen:
For the mmap failure, I don't think panicking in Rust is the best idea since each VM will have a different error message or exit mechanism that we should respect in my opinion. Using MMTk vs standard GC should be as transparent as possible. I agree that a `fatal_error() -> !` function does make sense and maybe we should add that in the future.

src/util/alloc/allocator.rs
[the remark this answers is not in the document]
qinsoon: Sounds good.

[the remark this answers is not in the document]
k-sareen: Done. I've also updated the respective binding PRs for OpenJDK and JikesRVM. I'm marking this thread as resolved. Thanks!

k-sareen:
When you merge the PR can you please add the following comment to the squashed commit:
```
Note: API breaking change!

Adds `AllocationError` to differentiate between the two (main) types of OOM:
 * Critical OOM: This is the case where the OS is unable to mmap or acquire more memory.
   MMTk expects the VM to abort immediately if such an error is thrown.
 * Heap OOM: This is the case where the specified heap size is insufficient to execute the
   application. MMTk expects the binding to notify the VM about this OOM. MMTk makes no
   assumptions about whether the VM will continue executing or abort immediately.

MMTk now returns null in case of a Heap OOM error.
```

Apologies -- I should have done it when I made the first commit. Will keep it in mind for the future. Thank you!

PR #538, opened 2022-02-02
Fix the issue that available pages may be negative and cause a overflow.

qinsoon:
This PR follows the discussion here (https://github.com/mmtk/mmtk-core/pull/535#discussion_r796233194) and here(https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/copy-space-pages-used/near/270335481).

The PR fixes the issue that available pages may be negative in some cases and cause a panic on overflow arithmetic. It also renames a few methods in Plan about page accounting to make their names consistent.


src/plan/global.rs
```diff
@@ -253,30 +253,50 @@ pub trait Plan: 'static + Sync + Downcast {
...
+    /// should always be positive or 0.
+    fn get_available_pages(&self) -> usize {
+        // It is possible that the reserved pages is larger than the total pages so we are doing
+        // a saturating substraction to make sure we return a non-negative number.
+        // For example,
+        // 1. our GC trigger checks if reserved pages is more than total pages.
+        // 2. when the heap is almost full of live objects and we are doing a copying GC, it is possible
+        //    the reserved pages is larger than total pages after the copying GC (the reserved pages after a GC
+        //    may be larger than the reserved pages before a GC, as we may end up using more memory for thread local
+        //    buffers for copy allocators).
```
k-sareen, line 277: Can I suggest a minor edit to mention OOMs? Something like:
```
// 2. when the heap is almost full of live objects (such as in the case of an OOM) and we are doing a copying GC,
//    it is possible the reserved pages is larger than total pages after the copying GC (the reserved pages after
//    a GC may be larger than the reserved pages before a GC, as we may end up using more memory for thread local
//    buffers for copy allocators).
```

src/plan/global.rs
```diff
@@ -777,11 +797,11 @@ impl<VM: VMBinding> BasePlan<VM> {
...
 
         debug!(
-            "self.get_pages_reserved()={}, self.get_total_pages()={}",
-            plan.get_pages_reserved(),
+            "self.get_reserved_pages()={}, self.get_total_pages()={}",
+            plan.get_reserved_pages(),
             plan.get_total_pages()
         );
-        let heap_full = plan.get_pages_reserved() > plan.get_total_pages();
+        let heap_full = plan.get_reserved_pages() > plan.get_total_pages();
```
k-sareen, line 804: Maybe add a small comment here just to say that reserved pages > total pages in case of OOM and we have a copying collector?

src/plan/global.rs
```diff
@@ -253,30 +253,50 @@ pub trait Plan: 'static + Sync + Downcast {
...
+        //    may be larger than the reserved pages before a GC, as we may end up using more memory for thread local
+        //    buffers for copy allocators).
+        self.get_total_pages()
+            .saturating_sub(self.get_reserved_pages())
+    }
+
+    /// Get the number of pages that are reserved for collection. By default, we return 0.
+    /// For copying plans, they need to override this and calculate required pages to complete
+    /// a copying GC.
+    fn get_collection_reserved_pages(&self) -> usize {
```
k-sareen, line 285: This function seems to be overriden by most plans anyway? Do we want to just make it a stub function then?

k-sareen (commented):
LGTM other than minor changes. Thanks!

src/plan/global.rs
k-sareen: Alternatively, we should delete the overrides for non-copying plans.

src/plan/global.rs
qinsoon: Yeah. I will do that.

src/plan/global.rs
qinsoon: Sorry I don't get what you meant.

k-sareen: I meant that this case can only happen when we have an OOM right? I thought we should document it here for the future.

qinsoon: I am a bit confused when you said OOM. OOM means we run out of memory and we will tell the binding that we are out of memory, right? This is not OOM. This is just a check for GC, each slowpath allocation will go through this check.

src/plan/global.rs
qinsoon: Yeah, I will delete the overrides that basically duplicate the default implementation.

src/plan/global.rs
k-sareen: Right. Sorry I should have clarified. I am referring to this comment: https://github.com/mmtk/mmtk-core/pull/535#discussion_r796244445

qinsoon: Ah sorry, my bad 😅

k-sareen (approved):

Issue #537, opened 2022-02-02
Workers spuriously wake up


wks:
When reading the code of the scheduler, I found some problem with the current synchronisation mechanism.  This issue describes the problems I see, but I am not planning for an immediate refactoring because I don't have any concrete idea of how to solve the problems.

## Spurious wake-up

From `scheduler.rs`:

```rust
    #[cold]
    fn poll_slow(&self, worker: &GCWorker<VM>) -> Box<dyn GCWork<VM>> {
        debug_assert!(!worker.shared.is_parked());
        let mut guard = self.worker_monitor.0.lock().unwrap();
        loop {
            debug_assert!(!worker.shared.is_parked());
            if let Some((work, bucket_is_empty)) = self.pop_scheduable_work(worker) {
                if bucket_is_empty {
                    worker.sender.send(CoordinatorMessage::BucketDrained).unwrap();
                }
                return work;
            }
            // Park this worker
            worker.shared.parked.store(true, Ordering::SeqCst);
            if self.worker_group().all_parked() {
                worker.sender.send(CoordinatorMessage::AllWorkerParked).unwrap();
            }
            // Wait
            guard = self.worker_monitor.1.wait(guard).unwrap();  // ERROR1: potential spurious wake-up
            // Unpark this worker
            worker.shared.parked.store(false, Ordering::SeqCst);
        }
    }

```

The expression `self.worker_monitor.1.wait(guard)` does not follow the standard way of using condition variable.  According to the semantics of condition variable, the `wait` function can [wake up spuriously](https://en.wikipedia.org/wiki/Spurious_wakeup), and must be used in a loop:

```c
mutex.lock();
while (!conditon_is_met()) {
    condvar.wait();
}
do_something();
mutex.unlock();
```

## What's the real condition of the `wait()`?

The `poll_slow()` function attempts to get some work from local work buckets or global work buckets.  If there is no work available, it will park the current GC worker.  So **the condition it waits for could be that "there are some work packets available in any global buckets"**.  From the structure of the loop, *the current code could still be a valid use of condvar*, because it will check if "any work is available in any bucket" after waking up, although the condition is an implicit condition.

There are four places where the `worker_monitor.1` condvar is notified:
1.  In `GCWorkScheduler::update_buckets`
2.  In `GCWorkScheduler::notify_mutators_paused`
3.  In `WorkBucket::notify_all_workers` (called by `bulk_add`)
4.  In `WorkBucket::notify_one_worker` (called by `add`)

[...]

PR #539, opened 2022-02-04
Refactor scheduler and worker creation

wks:
This is part 2 of refactoring the scheduler: https://github.com/mmtk/mmtk-core/issues/532

~~**But don't merge it yet, because there is a known problem: it breaks options from command line**. See below.~~ For now we temporarily disable setting the number of GC threads from the command line.

We now construct the `GCWorkScheduler` in one step, so that we no longer need to make every field `Option<T>` and set the value during `initialize`.

To make it possible, we eliminated cyclic references by considering `MMTK`, `GCWorkScheduler` and `GCWorkerShared` as nested data, like this:
```rust
MMTK {
  ...
  sheduler: GCWorkScheduler {
    ...
    workers: vec![
      GCWorkerShared {...},
      GCWorkerShared {...},
      GCWorkerShared {...},
      ...
    ],
  }
}
```

So inner structures do not refer to outer structures.  We eliminated the `GCWorkScheduler::mmtk` field.

On the other hand, actors (threads) have references to all shared data they need.  Both `GCWorker` and `GCController` contain a reference to the `MMTK` instance, so they no longer need to navigate from the scheduler to MMTK.  And, obviously, private data of the actors (threads, including `GCWorker` and `GCController`) never refer to one another.

## Known problem

We create `GCWorkScheduler` when `MMTK` is created.  This is too early for now, because the `MMTK` instance is usually created statically by the VM.  When it is created, we still don't know how many threads we have.  The number of threads is determined by the `mmtk.options.threads` option, which can be set by either environment variable or command-line options.  This is too late for the creation of `GCWorkScheduler`.  The current PR will simply disregard any changes to the value of `mmtk.options.threads` after `GCWorkScheduler` is created.

To solve this problem, we need to make one (or both) of the following changes before applying this PR.

1.  Introducing a `MMTKBuilder`, and build the `MMTK` instance after options are parsed.
2.  Make it possible to add workers after `GCWorkScheduler` is created.
3.  Change `MMTK::scheduler` to `Option<Arc<GCWorkScheduler>>`.

We will need to do (1) anyway, because currently options are also held by an `UnsafeOptionsWrapper`, which is unsafe.

Method (2) needs to make the worker group synchronised (probably using a `RwLock`).  Adding a `RwLock` **directly** may add **performance overhead** because the work packet system frequently queries if all workers are parked.  To do it efficiently, we should redesign the synchronisation algorithm and make the lock more coarse-grained.  Issue https://github.com/mmtk/mmtk-core/issues/537 covers some of the details.

Method (2) will also enable the potential of dynamically adjusting the number of workers during execution, but let's not attempt that for now.

Method (3) simply pushes the `Option<T>` from the fields of `GCWorkScheduler` to the field of `MMTK`. It simply moves the problem to a different place.

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -196,7 +196,7 @@ pub fn initialize_collection<VM: VMBinding>(mmtk: &'static MMTK<VM>, tls: VMThre
         !mmtk.plan.is_initialized(),
         "MMTk collection has been initialized (was initialize_collection() already called before?)"
     );
-    mmtk.scheduler.initialize(*mmtk.options.threads, mmtk, tls);
+    mmtk.scheduler.spawn_gc_threads(mmtk, tls);
     mmtk.plan.base().initialized.store(true, Ordering::SeqCst);
 }
 
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -57,8 +57,15 @@ impl<VM: VMBinding> MMTK<VM> {
         // The first call will initialize SFT map. Other calls will be blocked until SFT map is initialized.
         SFT_MAP.initialize_once();
 
-        let scheduler = GCWorkScheduler::new();
         let options = Arc::new(UnsafeOptionsWrapper::new(Options::default()));
+
+        let num_workers = if cfg!(feature = "single_worker") {
+            1
+        } else {
+            *options.threads
+        };
+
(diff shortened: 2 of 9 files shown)


qinsoon (approved):
LGTM

PR #540, opened 2022-02-06
Fixes some issues with the tutorial

KipHamiltons:
I did the tutorial on a laptop and these are some of the things I noted which weren't toooooo pedantic.\
Most of these are self-explanatory, but I'll mention that the large section I removed about copying objects no longer applies to the relevant code. I am not sure if it should be replaced but it seemed to work ok when I skipped those steps :shrug:


qinsoon (approved):
LGTM. Thanks for the fix!

Issue #110, opened 2020-06-24
Move trace_object() to SFT


qinsoon:
As raised in https://github.com/mmtk/mmtk-core/pull/108#discussion_r444670594, invoking `initialize_header()` from a `Space` trait object would result in dynamic dispatching in the fast path. We can move `initialize_header()` to `SFT`, this would solve the problem.

qinsoon:
We want to do a similar refactoring for `trace_object()` as well.

qinsoon:
`initialize_header()` is in `SFT`. However, we cannot easily move `trace_object()` to `SFT` without further refactoring. `SFT` is a trait object, which means its methods need to be simple. However, `trace_object()` is a generic method (with `T: TransitiveClosure`), which is not simple. We need refactoring.

steveblackburn:
I'd like to up the priority on this.   Perhaps this can be done as part of the next bugfix/cleanup release.

qinsoon:
> `trace_object()` has two type parameters, but the types themselves are never explicitly used in the function body. So they can be replaced with `impl TransitiveClosure` and `impl CopyContext`. They are semantically equivalent, but less verbose. `impl SomeTrait` means the concrete type of the parameter can still be inferred at compile time, but we just don't mention the concrete type name. Therefore it still has the performance advantage (including opportunities to inline) of generic parameters.
> 
> ```diff
> diff --git a/src/policy/copyspace.rs b/src/policy/copyspace.rs
> index 77e75321..d6ad9154 100644
> --- a/src/policy/copyspace.rs
> +++ b/src/policy/copyspace.rs
> @@ -182,12 +182,12 @@ impl<VM: VMBinding> CopySpace<VM> {
>      }
>  
>      #[inline]
> -    pub fn trace_object<T: TransitiveClosure, C: CopyContext>(
> +    pub fn trace_object(
>          &self,
> -        trace: &mut T,
> +        trace: &mut impl TransitiveClosure,
>          object: ObjectReference,
>          semantics: AllocationSemantics,
> -        copy_context: &mut C,
> +        copy_context: &mut impl CopyContext,
>      ) -> ObjectReference {
>          trace!("copyspace.trace_object(, {:?}, {:?})", object, semantics,);
>          debug_assert!(
> ```

I think `impl Trait` is considered the same as generic type parameters in terms of object safety. That means if we would like to use a trait object, its method cannot use `<T>` as well as `a: impl Trait` as parameters.

Issue #258, opened 2021-02-24
Remove policy-specific code from plans, create space-specific work packets 


steveblackburn:
Policy-specific code should _not_ be in the Plans.     The old MMTk use the plans _only_ for dispatch to policies (eg a switch statement dispatching to various implementations of `traceObject()`.

For both software engineering and performance reasons, work packets should be homogenous w.r.t. the type of work they do.

So for something like `traceObject()`, there should be a separate work packet type for every space (_note that there needs to be one per space, rather than one per policy since spaces can maintain space-specific state_).

Then a general call to `traceObject()` will place the respective object on the correct work queue according to which space the object resides in.

The dispatch can be optimized in a variety of ways, including an address-based scheme, where certain bits within the object address denote the space ID and the dispatch requires simply indexing into an array of space-specific work queues.   This can be further improved by hiding the dispatch behind the syntactic sugar of a function call on the `ObjectReference` type.

qinsoon:
> Policy-specific code should _not_ be in the Plans. The old MMTk use the plans _only_ for dispatch to policies (eg a switch statement dispatching to various implementations of `traceObject()`.

We do have 'dispatch' code in the plans, but I am not aware of policy-specific code in the plans. Can you clarify which part of the code you meant? 

> So for something like `traceObject()`, there should be a separate work packet type for every space (_note that there needs to be one per space, rather than one per policy since spaces can maintain space-specific state_).
> 
> Then a general call to `traceObject()` will place the respective object on the correct work queue according to which space the object resides in.

Currently one work packet may include many object references and they could reside in different spaces (and policies). I don't quite get how space-specific work packets work. For example, when we trace one object and it references nodes in different spaces, currently we have a buffer to store all the nodes, and create a new work packet when the buffer is full. For space-specific packets, do we have multiple buffers and store nodes by spaces to different buffers, and create space-specific work packets when any of the buffer is full?

qinsoon:
Most of this issue is done by https://github.com/mmtk/mmtk-core/pull/542. 

There are some leftover work for mark compact and immix (including gen immix). Both mark compact and immix are still using their own trace object implementation. We should do some refactoring for them with the goal of **moving policy-specific code out of plan**. The idea is simple: **policy-specific code should not be in the Plans**. We should consider letting policies provide GC work packets and use those work packets in the plan.

Space-specific work packet is an optimisation. We possibly want to track it in another issue if the rest is done for this issue and the issue can be closed.

PR #542, opened 2022-02-07
Introduce SFTProcessEdges

qinsoon:
This PR adds a general implementation for `ProcessEdges`, called `SFTProcessEdges`. It uses `SFT.sft_trace_object()` for each policy. A plan does not need to implement their own process edges work packet if they use `SFTProcessEdges`. This PR greatly simplifies the GC work implementation for most GC plans. This PR closes https://github.com/mmtk/mmtk-core/issues/110, and it is an important step towards https://github.com/mmtk/mmtk-core/issues/258.

Currently only Immix (including GenImmix) and mark compact uses custom process edges. All the other plans use `SFTProcessEdges`.

This PR
* adds `SFT.sft_trace_object()`.
* adds `Space.set_copy_for_sft_trace()` to set copy context for each space, which is used when we invoke `sft_trace_object()`.
* adds `SFTProcessEdges`, and use it for most plans (except immix/genimmix and mark compact).

Performance
===

I measured performance for semispace and Immix. 

semispace
---

Benchmark time
ss-bm-time

STW time
ss-gc-time

~~immix~~
---

This result appears to be wrong.

<details>
  <summary>Benchmark time</summary>
immix-bm-time-new
</details>

<details>
  <summary>STW time</summary>
immix-gc-time-new
</details>


src/scheduler/gc_work.rs
```diff
@@ -476,6 +476,56 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for E {
...
+
+    #[inline]
+    fn trace_object(&mut self, object: ObjectReference) -> ObjectReference {
+        use crate::policy::space::*;
+
+        if object.is_null() {
+            return object;
+        }
+
+        // Erase <VM> type parameter
```
wks, line 504: Is it possible to add a `<VM>` type parameter to `SFTMap`, i.e. `SFTMap<VM>`?  Will that make it unnecessary to do the erasure here?

Maybe the current way of initialising `SFT_MAP` does not allow us to do this.  I am thinking about moving it into `MMTK<VM>` as a field.

src/plan/generational/copying/gc_work.rs
```diff
@@ -1,78 +1,18 @@
...
+    type ProcessEdgesWorkType = SFTProcessEdges<VM>;
 }
 
 pub(super) struct GenCopyMatureGCWorkContext<VM: VMBinding>(std::marker::PhantomData<VM>);
 impl<VM: VMBinding> crate::scheduler::GCWorkContext for GenCopyMatureGCWorkContext<VM> {
     type VM = VM;
     type PlanType = GenCopy<VM>;
-    type ProcessEdgesWorkType = GenCopyMatureProcessEdges<VM>;
+    type ProcessEdgesWorkType = SFTProcessEdges<VM>;
 }
```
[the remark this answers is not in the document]
wenyuzhao: Should we provide a default type alias `type ProcessEdgesWorkType = SFTProcessEdges<Self::VM>;`  in `crate::scheduler::GCWorkContext`? So at least we don't need to repeat this line.

src/plan/generational/global.rs
```diff
@@ -172,7 +174,7 @@ impl<VM: VMBinding> Gen<VM> {
             return self.nursery.trace_object::<T>(
                 trace,
                 object,
-                CopySemantics::PromoteMature,
+                Some(CopySemantics::PromoteMature),
```
wenyuzhao, line 177: Should this be called `PromoteNursery` or `PromoteToMature`. This confused me for a sec why we need to promote a mature object.

src/policy/markcompactspace.rs
```diff
@@ -51,6 +52,18 @@ impl<VM: VMBinding> SFT for MarkCompactSpace<VM> {
...
+    #[inline(always)]
+    fn sft_trace_object(
+        &self,
+        _trace: SFTProcessEdgesMutRef,
+        _object: ObjectReference,
+        _worker: GCWorkerMutRef,
+    ) -> ObjectReference {
+        // We should not use trace_object for markcompact space.
+        // Depending on which trace it is, we should manually call either trace_mark or trace_forward.
+        panic!("sft_trace_object() cannot be used with mark compact space")
```
wenyuzhao, line 65: It is possible to define a enum variable `trace_kind` in mark compact plan, indication which trace it is doing (should be either "mark" or "forward"). Then we can do a branch here to either `trace_mark` or `trace_forward`, based on the trace state.

wenyuzhao: This may simplify the code (?) since we don't need to define other `ProcessEdgesWork` structs in markcompact code for tracing.

src/policy/immix/immixspace.rs
```diff
@@ -77,6 +78,15 @@ impl<VM: VMBinding> SFT for ImmixSpace<VM> {
...
         crate::util::alloc_bit::set_alloc_bit(_object);
     }
+    #[inline(always)]
+    fn sft_trace_object(
+        &self,
+        _trace: SFTProcessEdgesMutRef,
+        _object: ObjectReference,
+        _worker: GCWorkerMutRef,
+    ) -> ObjectReference {
+        panic!("We do not use SFT to trace objects for Immix. sft_trace_object() cannot be used.")
```
wenyuzhao, line 88: Is this just for performance reasons?

src/plan/generational/copying/gc_work.rs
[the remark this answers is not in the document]
qinsoon: I have set the default type as Wenyu suggested, and changed `GenCopyNursery/MatureGCWorkContext` to `GenCopyGCWorkContext`.

src/policy/copyspace.rs
```diff
@@ -89,6 +102,10 @@ impl<VM: VMBinding> Space<VM> for CopySpace<VM> {
     fn release_multiple_pages(&mut self, _start: Address) {
         panic!("copyspace only releases pages enmasse")
     }
+
+    fn set_copy_for_sft_trace(&mut self, semantics: Option<CopySemantics>) {
```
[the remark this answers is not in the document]
qinsoon: We don't have to. Alternatively, we can set both as `Some(DefaultCopy)`. We could even set that in `gc_init()` rather than `prepare()`. However, as we don't copy from `tospace`, setting `tospace`'s copy semantic to `DefaultCopy` would be confusing. I feel it is more conceptually clear to set it in `prepare()` and toggle the semantics for each GC. 

Let me know what you think. I can make the change if you would prefer so.

src/scheduler/gc_work.rs
qinsoon: This is a good point. There are multiple issues here:
1. Should we have `SFTMap<VM>`? 
I think we should if we can. We do not expect different VMs using MMTk in the same process (at least at this point, that is not our goal).
2. Should `SFTMap<VM>` be put to an `MMTK<VM>` instance?
I am not sure, but possibly yes, because it includes pointers to spaces that are in one instance. However, to support multiple `MMTK` instances, we would need to know what address range is used for one instance, and what address range is still available for using. So we will need to have some kind of heap data structure that is shared across instances. Currently, these heap data structures are VM map, Mmapper, and SFT map, and they are intentionally static. But it may be possible that keep some static heap data structure, but move `SFTMap` to `MMTK` -- but I don't know how feasible that is.

Anyway, I don't think we can do this change in this PR. If you think this current hack is not acceptable, I can work on refactoring `SFTMap` first, and then let's revisit this PR.

src/policy/markcompactspace.rs
qinsoon: I gave it a try. Theoretically it could work. But we would need to set `trace_kind` 4 times: before tracing, before forwarding, before tracing finalizable objects, before forward finalizable objects. And we do not have a proper hook for the last two (we do not have a hook to run some code when a work bucket is opened) so we will probably end up having more code. 

I could work on it after this pull request, and possibly add a hook in the scheduler to call some methods in the plan when a work bucket is first opened. Let me know what you think.

Another concern is performance, as this requires us to do a branch for each `trace_object` call.

src/policy/immix/immixspace.rs
qinsoon: Yes. I got some strange results for immix. I benchmarked 3 builds: master, immix with SFTProcessEdges in the PR, immix with the original ImmixProcessEdges in the PR. Somehow, the last two showed a similar slowdown, and I couldn't figure out the reason (I expect immix with the original ImmixProcessEdges should behave the same as master). So in the end, I remove all the changes related with immix in the PR.

Maybe I should dig more into the performance for immix.

src/scheduler/gc_work.rs
wks: OK. Let's not make the change in this PR, and think about it later.

src/policy/copyspace.rs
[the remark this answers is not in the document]
wks: You are right.  It may be confusing if the semantic is `DefaultCopy` while not actually being copied from.

I also anticipate that some spaces may have different copy semantics during different kinds of GC.

Let's keep your code.

wks (commented):
It looks good for me now.  Let's wait for Wenyu's comments to be resolved.

src/policy/immix/immixspace.rs
qinsoon: I did another measurement for this PR only using `pmd` with `immix` (which I found usually has the worst slowdown). I had 200 invocations so I assume the result is robust enough. There is slowdown of 0.7% total time and 2.8% STW time. There is no difference in retired instructions, but the CPU cycles increases by 3.6% (for GC). There are also some difference in some major GC work packets as well. 

pmd-immix

I couldn't see how this PR affects immix performance.

k-sareen: This is genuinely bizarre. Are there more cache misses perchance?

qinsoon: > This is genuinely bizarre. Are there more cache misses perchance?

I didn't measure that, but I could add that the measurement.

qinsoon:
@wenyuzhao Can you take a look at this PR and the discussions and see if the comments are resolved? Let me know if there is anything I would need to do for the PR.

wenyuzhao (approved):
LGTM. I feel that the performance drop may come from other places (?), the refactoring introduced ins this PR just reveals the problem.

PR #552, opened 2022-02-22
Fix ImmixSpace liveness check

wenyuzhao:
Fix ImmixSpace's `is_live` and `get_forwarded_object` for https://github.com/mmtk/mmtk-openjdk/pull/141 .

---

The context of this PR:

OpenJDK has several object cache (or storage) that should be processed during GC. For example, `CodeCache` may contain embedded pointers. Our old approach is to treat all the objects in these storages as root pointers -- and caused memory leak. https://github.com/mmtk/mmtk-openjdk/pull/141 tries to update these storages before the end of each GC, including _removing dead objects_ and _updating forwarded objects_.

To support this. all mmtk spaces should implement `is_live` and `get_forwarded_object` correctly.

@tianleq Looks like there's some difficulty implementing `get_forwarded_object` for MarkCompactSpace? Is it because the MarkCompactSpace does not use `object_forwarding`? 

---

TODO: MarkCompactSpace needs to be fixed as well...

### DIFF
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -14,7 +14,7 @@ use crate::util::heap::HeapMeta;
 use crate::util::heap::PageResource;
 use crate::util::heap::VMRequest;
 use crate::util::metadata::side_metadata::{self, *};
-use crate::util::metadata::{self, compare_exchange_metadata, load_metadata, MetadataSpec};
+use crate::util::metadata::{self, compare_exchange_metadata, load_metadata, MetadataSpec, store_metadata};
 use crate::util::object_forwarding as ForwardingWord;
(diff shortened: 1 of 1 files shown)


tianleq:
> 
Yes, markcompact moves object through a linear scan

qinsoon:
`get_forwarded_object` is fixed in https://github.com/mmtk/mmtk-core/pull/753 for ImmixSpace. Setting mark bit in Immix `post_copy` is included in `master`. 
https://github.com/mmtk/mmtk-core/blob/5b48f8ad2e62043344942e3d88e59b6fd8d26afc/src/policy/immix/immixspace.rs#L776-L783

We can close this PR.

PR #555, opened 2022-03-06
Fix object ref SFT

qinsoon:
If a binding uses `ObjectReference` to point to an address that is not within our allocated memory, the SFT and the side metadata for the `ObjectReference` may not be initialized properly.

This PR introduces a solution by introducing a constant `ObjectModel::OBJECT_REF_OFFSET_BEYOND_CELL`. If a binding may use object reference to point to an address beyond the cell we allocate, they need to provide a `Some` value for the constant. MMTk allocators need to guarantee that the object reference for an allocation won't point to a new chunk (so its metadata is properly initialized and is the same as the allocation result).

This PR fixes https://github.com/mmtk/mmtk-jikesrvm/issues/102.

This PR:
* introduces a constant `ObjectModel::OBJECT_REF_OFFSET_BEYOND_CELL`. By default, it is `None`. If a binding may use `ObjectReference` to point to an address outside a cell, they need to set the constant to a proper `Some` value.
* Introduces a module `util::alloc::object_ref_guard`, which includes a few methods that help allocators to check and avoid breaking the invariant.
* adds a few assertions to check if our allocation addresses and object references have valid SFT entries.
* introduces a space acquire lock so we can guarantee when pages are acquired from a page resource, the SFT for the pages are always initialized.
* removes unnecessary methods in `Map32`.


qinsoon:
binding-refs
JIKESRVM_BINDING_REF=fix-object-ref-sft

src/policy/space.rs
```diff
@@ -199,38 +202,44 @@ impl<'a> SFTMap<'a> {
...
-                    self.sft.len()
-                } else {
-                    i + SPACE_PER_LINE
-                };
-                let chunks: Vec<usize> = (i..max).collect();
-                let space_names: Vec<&str> = chunks.iter().map(|&x| self.sft[x].name()).collect();
-                trace!("Chunk {}: {}", i, space_names.join(","));
-            }
+            trace!("{}", self.print_sft_map());
+        }
```
wks, line 218: The `trace!` macro itself tests if the log is enabled at the `Trace` level, therefore the `if` can be omitted.

If `print_sft_map` is only used here, and the sole purpose is to log the whole string in one atomic step, you can do this:

```rust
trace!("{}", {
    let mut res = String::new();
    for ...... {
        res.push(...);
    }
    res
});
```

src/util/alloc/object_ref_guard.rs
```diff
@@ -0,0 +1,64 @@
+//! This module includes functions to make sure an invariant: for each object we allocate (`[cell, cell + bytes)`), the metadata for
```
wks, line 1: s/make sure an invariant/make sure the following invariant always holds/

src/util/alloc/object_ref_guard.rs
```diff
@@ -0,0 +1,64 @@
...
+//! the object reference (`object_ref`) is the same as the allocated memory. Given that we always initialize metadata based on chunks,
```
wks, line 2: s/is the same as/is always in the range of/

src/util/alloc/object_ref_guard.rs
```diff
@@ -0,0 +1,64 @@
...
+//! we simply need to make sure that `object_ref` is in the same chunk as `[cell, cell + bytes)`. In other words, we avoid
+//! allocating an address for which the object reference may be in another chunk.
+//!
+//! Note that where an ObjectReference points to is defined by a binding. We only have this problem if an object reference may point
+//! to an address that is outside our allocated memory (`object_ref >= cell + bytes`). We ask a binding to specify
+//! `ObjectModel::OBJECT_REF_OFFSET_BEYOND_CELL` to be a value if their object reference may point to an address outside the allocated
```
wks, line 8: s/to be a value if their object reference may point/if an object reference may point/

src/util/heap/layout/vm_layout_constants.rs
```diff
@@ -24,6 +24,7 @@ pub const LOG_BYTES_IN_CHUNK: usize = 22;
 
 /** Coarsest unit of address space allocation. */
 pub const BYTES_IN_CHUNK: usize = 1 << LOG_BYTES_IN_CHUNK;
+pub const CHUNK_MASK: usize = (1 << LOG_BYTES_IN_CHUNK) - 1;
```
wks, line 27: There is another `CHUNK_MASK` in `src/util/metadata/side_metadata/constants.rs` that has the same definition. Should we remove that after we introduce this?

src/util/alloc/object_ref_guard.rs
```diff
@@ -0,0 +1,64 @@
...
+    );
+}
+
+/// Check if the object reference for this allocation may cross and fall into the next chunk.
+pub fn object_ref_may_cross_chunk<VM: VMBinding>(addr: Address) -> bool {
+    if VM::VMObjectModel::OBJECT_REF_OFFSET_BEYOND_CELL.is_none() {
+        return false;
+    }
+
+    (addr & CHUNK_MASK) + VM::VMObjectModel::OBJECT_REF_OFFSET_BEYOND_CELL.unwrap() > BYTES_IN_CHUNK
```
wks, line 63: Has it already crossed the chunk boundary if the sum is **equal** to, but not greater than, `BYTES_IN_CHUNK`?

src/util/alloc/malloc_allocator.rs
```diff
@@ -35,15 +36,21 @@ impl<VM: VMBinding> Allocator<VM> for MallocAllocator<VM> {
...
-            size,
-            align,
-            offset,
-            ret
-        );
-        ret
+        loop {
+            let ret = self.space.alloc(self.tls, size, align, offset);
+            if object_ref_may_cross_chunk::<VM>(ret) {
+                self.space.free(ret);
```
wks, line 42: If the malloc implementation is clever enough to reuse the memory you just `free`-ed in the next `malloc`, this loop may go on forever.  We need to think about a way to "disrupt" `malloc` a bit, for example, allocating a slightly bigger object, requiring a stricter alignment, do the second `malloc` before `free`-ing, etc.

wks (changes requested):

src/policy/space.rs
qinsoon: Yeah, you are right. I will remove the `if` statement. 

The reason why I put `print_sft_map()` to a separate function is that I used the function in a panic hook to print the SFT map upon panic in my debugging. I think it would be useful for debugging if we can print the SFT map in other places.

src/util/heap/layout/vm_layout_constants.rs
qinsoon: Thanks for pointing this out. I will remove `CHUNK_MASK` from side metadata constants.

src/util/alloc/object_ref_guard.rs
qinsoon: You are right. I will change this.

src/util/alloc/malloc_allocator.rs
qinsoon: I made changes to the code. If the first allocation does not pass the object ref check, a vec is created to store the 'bad' results, and the results are free'd when we have a 'good' result.

caizixian: Kunshan’s concern is very valid. For example, the malloc in glibc has tcache, which can be problematic.

Yi’s new approach sounds better. But it still has one potential problem. For example, if you allocate A and it’s bad, so you allocate again, you get B, which is good, and then you free A. A goes into the tcache, and for next allocation of the same size class, you will get A again, which you need to withhold, and try again. That is, once you get a bad cell and free it, it stays forever in the tcache (and it can grow the next time you hit a chunk boundary). I’m not sure whether this will manifest in real world scenarios, or whether this will introduce a performance hit (we assume that if people call into libraries for malloc, they probably don’t care too much about performance anyway)

qinsoon:
@wks I have addressed the reviews. Can you take a look again?

wks (approved):
LGTM

Issue #501, opened 2021-11-24
Debug is_live() correctness


qinsoon:
Currently we provide an API `is_live_object(object: ObjectReference) -> bool`, which is supposed to be used at any time and should return a correct result.

I quickly tested two cases in OpenJDK, and `is_live()` did not return correct results for all the plans.

Case 1: disable the allocation fastpath, assert if the object we just allocate is live after `post_alloc()`. Mark sweep failed on this.
Case 2: Based on Case 1, further use a stress factor `1`, which will trigger a GC for every allocation. Both mark sweep and immix failed on this.

qinsoon:
We need to clarify that `is_live` should only be used at certain stages during GC.

PR #560, opened 2022-03-16
Add post_copy() for ImmixCopyContext: mark object and line

qinsoon:
This PR adds `post_copy()` for `ImmixCopyContext`, which marks the new object and conditionally marks the line for the object. Without this PR, copied Immix objects were not considered as 'live' (by `is_live()`), and GC work after copying (such as reference processing) may behave incorrectly if they use `is_live()`. This PR is related to https://github.com/mmtk/mmtk-core/issues/501.


wenyuzhao (approved):
lgtm

PR #563, opened 2022-03-21
Fix object reference confusion for mark compact

qinsoon:
This pull request refactors `MarkCompactSpace`. The current implementation uses address and object reference interchangeably, which assumes bindings have no offset between object reference and address. This PR now properly uses object reference in `MarkCompactSpace`.

* `MarkCompactSpace` no longer casts an address directly to an object reference. It uses proper methods to get object references from the binding.
* `MarkCompactSpace` now stores an object reference to the new object as its forwarding pointer.
* `MarkCompactSpace` now has a few functions to access the forwarding pointer.
* A new work bucket `SecondRoots` is added. So the second round root scanning will happen before `RefForwarding`.


qinsoon:
binding-refs
OPENJDK_BINDING_REF=get-reference-when-copied-to

tianleq (approved):
LGTM

Issue #544, opened 2022-02-09
Weak reference processing


qinsoon:

qinsoon:
https://github.com/mmtk/mmtk-core/pull/564 adds a functional implementation for weak reference. Here lists the performance results for the PR: https://github.com/mmtk/mmtk-core/pull/564#issuecomment-1092328984

Remaining work based on https://github.com/mmtk/mmtk-core/pull/564:
* [ ] Should we keep Java's weak reference semantics? (soft/weak/phantom)
    * If most languages just use weak reference and it is similar to Java, maybe we should make soft/phantom as an optional feature.
    * Java's semantics require references to be enqueue'd if their referent is cleared. Maybe we should make this an optional feature as well.
* [ ] Store weak refs in buffer during object tracing. This will speed up the process edges time shown in the performance results. It cuts the [overhead from 5.4% down to 2.7%](https://squirrel.anu.edu.au/plotty/yilin/yp-2015/#0|hare-2022-04-04-Mon-122809&benchmark^build^invocation^iteration&GC^bmtime^time.other^time.stw^work.ImmixProcessEdges.time.total&|10&iteration^1^2|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^build^benchmark&). *
* [ ] Investigate GC time slowdown. We have a 10% slowdown in GC time. Process edges only have a 5% slowdown. That means we have slowdown somewhere else. It could be the weak reference work itself. Adding extra work bucket has [a slowdown of ~1%](https://squirrel.anu.edu.au/plotty/yilin/yp-2015/#0|fisher-2022-03-29-Tue-095044&benchmark^build^invocation^iteration&GC^bmtime^time.other^time.stw&|10&iteration^1^2|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^build^benchmark&). *
* [ ] Investigate batik GC time (40% slowdown in GC time). Batik only does 1 GC in the measurement.


\* Performance links are not publicly visible.

PR #564, opened 2022-03-24
Weak reference processing

qinsoon:
This PR revamps the reference processor, and schedules reference processing work. This PR is one step towards https://github.com/mmtk/mmtk-core/issues/544: it gets the old reference processor working again.

There is some more work needed for weak reference processing. So `no_reference_types` is `true` and by default, we do not process weak references.

Changes
* Rework on `ReferenceProcessor`:
  * retain soft references properly, depending on whether the collection is an emergency collection or not.
  * add a separate reference enqueue step. For references that need to be enqueue'd (by Java semantics), the reference processor will store them in a buffer, and call to the binding at the end of a GC to enqueue the references. This makes sure the references are valid during the GC.
  * reference table is now an `HashSet` rather than `Vec`. We used to assume weak references are added to MMTk once they are created, so for one reference, it will be added once. Now we allow adding weak references during GC, and a weak reference may be traced multiple times in a GC, and added multiple times to MMTk. Using `HashSet` can deduplicate the references.
  * Remove unnecessary use of `UnsafeCell` for `ReferenceProcessorSync`.
  * Remove unnecessary `unforwarded_references`: I suspect it was intended to deal with meta-circularity in JikesRVM/Java MMTk (https://github.com/JikesRVM/JikesRVM/blob/5072f19761115d987b6ee162f49a03522d36c697/MMTk/ext/vm/jikesrvm/org/jikesrvm/mm/mmtk/ReferenceProcessor.java#L96), and is no longer needed for MMTk core. For MMTk core, any Rust struct and field is 'untraced' unless we specifically 'trace' it.
  * Add a `allow_new_candidate` field to avoid adding new candidates after we finish processing references. See the comments for the field for details.
  * `process_reference()` is moved from `ReferenceGlue` (in VMBinding traits) to `ReferenceProcessor`.
  * `enqueue_references()` is added to `ReferenceGlue`.
* Add a few work packets for weak reference processing. They are simply wrapping the reference processor's methods, and are single threaded.
* Add a few more work buckets for weak reference processing: `SoftRefClosure`, `WeakRefClosure`, `PhantomRefClosure`, `FinalRefClosure`, `FinalizableForwarding`.
* Schedule the weak reference processing work properly.
* Rename `ProcessWeakRef` (which call `VMBinding::VMCollection::process_weak_refs()`) to `VMProcessWeakRef`.
* `add_soft/weak/phantom_candidate()` now no longer needs a `referent` argument.


src/util/reference_processor.rs
```diff
@@ -129,207 +162,392 @@ pub enum Semantics {
...
+        trace!(" ~> {:?} (retained)", referent.to_address());
+    }
+
+    /// Process a reference.
+    /// * If both the reference and the referent is alive, return the updated reference and update its referent properly.
+    /// * If the reference is alive, and the referent is not null but not alive, return a null pointer and the reference (with cleared referent) is enqueued.
+    /// * For other cases, return a null pointer.
+    ///
+    /// If a null pointer is returned, the reference can be removed from the reference table. Otherwise, the updated reference should be kept
+    /// in the reference table.
```
wks, line 410: It is more idiomatic in Rust to return `Option<ObjectReference>` instead.  We never store the returned `NULL` objref in any ref fields.  We only use the return value to decide if the reference should be discarded or retained.

src/util/reference_processor.rs
```diff
@@ -129,207 +162,392 @@ pub enum Semantics {
...
+    fn process_reference<E: ProcessEdgesWork>(
+        &self,
+        trace: &mut E,
+        reference: ObjectReference,
+        enqueued_references: &mut Vec<ObjectReference>,
+    ) -> ObjectReference {
```
wks, line 416:
```suggestion
    ) -> Option<ObjectReference> {
```

src/util/reference_processor.rs
```diff
@@ -129,207 +162,392 @@ pub enum Semantics {
...
+
+        trace!("Process reference: {}", reference);
+
+        // If the reference is dead, we're done with it. Let it (and
+        // possibly its referent) be garbage-collected.
+        if !reference.is_live() {
+            <E::VM as VMBinding>::VMReferenceGlue::clear_referent(reference);
+            trace!(" UNREACHABLE reference: {}", reference);
+            trace!(" (unreachable)");
+            return ObjectReference::NULL;
```
wks, line 427:
```suggestion
            return None;
```

src/util/reference_processor.rs
```diff
@@ -129,207 +162,392 @@ pub enum Semantics {
...
+        let old_referent = <E::VM as VMBinding>::VMReferenceGlue::get_referent(reference);
+        trace!(" ~> {}", old_referent);
+
+        // If the application has cleared the referent the Java spec says
+        // this does not cause the Reference object to be enqueued. We
+        // simply allow the Reference object to fall out of our
+        // waiting list.
+        if old_referent.is_null() {
+            trace!(" (null referent) ");
+            return ObjectReference::NULL;
```
wks, line 441:
```suggestion
            return None;
```

src/util/reference_processor.rs
```diff
@@ -129,207 +162,392 @@ pub enum Semantics {
...
+
+            // The reference object stays on the waiting list, and the
+            // referent is untouched. The only thing we must do is
+            // ensure that the former addresses are updated with the
+            // new forwarding addresses in case the collector is a
+            // copying collector.
+
+            // Update the referent
+            <E::VM as VMBinding>::VMReferenceGlue::set_referent(new_reference, new_referent);
+            new_reference
```
wks, line 461:
```suggestion
            Some(new_reference)
```

src/util/reference_processor.rs
```diff
@@ -129,207 +162,392 @@ pub enum Semantics {
...
+        } else {
+            // Referent is unreachable. Clear the referent and enqueue the reference object.
+            trace!(" UNREACHABLE referent: {}", old_referent);
+
+            <E::VM as VMBinding>::VMReferenceGlue::clear_referent(new_reference);
+            enqueued_references.push(new_reference);
+            ObjectReference::NULL
```
wks, line 468:
```suggestion
            None
```

src/util/reference_processor.rs
```diff
@@ -129,207 +162,392 @@ pub enum Semantics {
...
 
-        /*
-         * Reference is definitely reachable.  Retain the referent.
-         */
-        let referent = VM::VMReferenceGlue::get_referent(reference);
+        // Reference is definitely reachable.  Retain the referent.
+        let referent = <E::VM as VMBinding>::VMReferenceGlue::get_referent(reference);
         if !referent.is_null() {
-            trace.retain_referent(referent);
+            Self::keep_referent_alive(trace, referent);
```
wks, line 399: `keep_referent_alive` calls `e.trace_object`, and it may move the object, and return the forwarded reference.  But we are not calling `VMReferenceGlue::set_referent` here.  Currently the two call sites of `set_referent` are:

1.  in `forward` for MarkCompact.
2.  in `process_reference`.  But `scan` calls `process_reference` when `retain` is `false`.  But this line is executed when `retain` is `true`.

So if we use the `SemiSpace` plan, and a `SoftReference` holds the only pointer to another object, and this is the first time this edge is traced, will this `SoftReference` hold a dangling pointer after that? 

```suggestion
            let new_referent = Self::keep_referent_alive(trace, referent);
            <E::VM as VMBinding>::VMReferenceGlue::set_referent(reference, new_referent);
```

wks (changes requested):

src/util/reference_processor.rs
```diff
@@ -129,207 +162,392 @@ pub enum Semantics {
...
-        if TRACE_DETAIL {
-            trace!("{:?} Reference table is {:?}", self.semantics, references);
-            trace!(
-                "{:?} unforwardedReferences is {:?}",
-                self.semantics,
-                unforwarded_references
-            );
+
+        if !sync.enqueued_references.is_empty() {
+            debug!("enqueue: {:?}", sync.enqueued_references);
```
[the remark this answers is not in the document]
qinsoon: Good point. I will change this to `trace!()`.

src/vm/reference_glue.rs
```diff
@@ -30,16 +29,12 @@ pub trait ReferenceGlue<VM: VMBinding> {
...
-    /// Arguments:
-    /// * `trace`: A reference to a `TraceLocal` object for this reference.
-    /// * `reference`: The address of the reference. This may or may not be the address of a heap object, depending on the VM.
-    /// * `tls`: The GC thread that is processing this reference.
-    fn process_reference<T: TraceLocal>(
-        trace: &mut T,
-        reference: ObjectReference,
-        tls: VMWorkerThread,
-    ) -> ObjectReference;
+    /// For reference types, if the referent is cleared during GC, the reference
```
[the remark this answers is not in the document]
qinsoon: I feel we can use the word 'weak reference' as a general term to refer to all kinds of weak references (including soft/phantom references). Also we can use `WeakReferenceGlue`, and `WeakReferenceProcessor`.

src/util/reference_processor.rs
qinsoon: Sounds good.

src/util/reference_processor.rs
qinsoon: That is not necessary. However, the code is a bit confusing. What actually happens is this:
1. In `SoftRefClosure`:
    * if emergency collection { retain soft refs (no update) }
3. In `WeakRefClosure`:
    * scan soft refs (with update)
    * scan weak refs (with update)

I will change this to:
1. In `SoftRefClosure`:
    * if emergency collection { retain soft refs (no update) }
    * scan soft refs (with update)
4. In `WeakRefClosure`:
    * scan weak refs (with update)

src/util/reference_processor.rs
qinsoon: With this change, we are able to use a more concise way to iterate on the reference table `references.iter().filter_map(|reff| self.process_reference(...))`.

qinsoon:
binding-refs
OPENJDK_BINDING_REF=weak-ref
JIKESRVM_BINDING_REF=weak-ref
V8_BINDING_REF=update-pr-564

qinsoon:
Performance results for the PR (Immix)

Benchmark time: 1% slowdown with weak ref processing
weak-ref-bmtime

STW time: 10% slowdown with weak ref
weak-ref-gctime
 
Immix process edges time: 5% slowdown with weak ref
weak-ref-process-edges

<details>
  <summary>Data link (not publicly visible)</summary>
  <a href="https://squirrel.anu.edu.au/plotty/yilin/yp-2015/#0|hare-2022-04-04-Mon-122809&benchmark^build^invocation^iteration&GC^bmtime^time.other^time.stw^work.ImmixProcessEdges.time.total&|10&iteration^1^2&build^2^jdk-mmtk-weak-ref-buffer&benchmark^2^h2|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^build^benchmark&">link</a>
</details>

wks:
The following program crashes on OpenJDK with MMTk, but runs fine without `-XX:+UseThirdPartyHeap`:
```java
import java.lang.ref.SoftReference;

class Foo {
    public SoftReference<Foo> sr;
    int num;
}

public class SoftStrongSoft {
    public static SoftReference<Foo> setUp() {
        Foo foo1 = new Foo();
        foo1.num = 30;

        Foo foo2 = new Foo();
        foo2.num = 40;

        SoftReference<Foo> sr1 = new SoftReference<Foo>(foo1);
        SoftReference<Foo> sr2 = new SoftReference<Foo>(foo2);

        foo1.sr = sr2;

        return sr1;
    }

    public static void fillUpTheHeap() {
        String[] ar = new String[1000000];
        for (int i = 1; i < 1000000; i++) {
            String str = String.valueOf(i);
            if (i % 100 == 0) {
                ar[i] = str;
            }
        }
    }

        SoftReference<Foo> sr1 = setUp();

        fillUpTheHeap();
        //System.gc();
        //Runtime.getRuntime().gc(); // Just in case System.gc doesn't trigger GC.

        Foo foo1 = sr1.get();
        Foo foo2 = foo1.sr.get();

        System.out.println(foo1.num);
        System.out.println(foo2.num);
    }
}
```
Run with:
```shell
javac SoftStrongSoft.java -source 11 -target 11
export MMTK_NO_REFERENCE_TYPES=false
/path/to/java -XX:+UseThirdPartyHeap -Xmx64M SoftStrongSoft
```
Result:
```
[2022-04-08T04:37:18Z INFO  mmtk::memory_manager] Initialized MMTk with SemiSpace
[2022-04-08T04:37:19Z INFO  mmtk::plan::global]   [POLL] copyspace0: Triggering collection
[2022-04-08T04:37:19Z INFO  mmtk::scheduler::gc_work] End of GC
[2022-04-08T04:37:19Z INFO  mmtk::plan::global]   [POLL] copyspace1: Triggering collection
[2022-04-08T04:37:19Z INFO  mmtk::scheduler::gc_work] End of GC
30
Exception in thread "main" java.lang.NullPointerException
        at SoftStrongSoft.main(SoftStrongSoft.java:45)
```

The first `SoftReference`, i.e. `sr1` is retained, but the second `sr2` is not.  I confirmed that emergency collection did not happen.

I think we should repeat the soft reference processing until no further objects can be reached by traversing soft-strong-soft-srong-soft-...-soft-strong references.  FinalizerReference will need it, too.  It should not be needed for weak/phantom references because weak/phantom references never keep their referents alive.

qinsoon:
> The following program crashes on OpenJDK with MMTk, but runs fine without `-XX:+UseThirdPartyHeap`:
> 
> ```java
> import java.lang.ref.SoftReference;
> 
> class Foo {
>     public SoftReference<Foo> sr;
>     int num;
> }
> 
> public class SoftStrongSoft {
>     public static SoftReference<Foo> setUp() {
>         Foo foo1 = new Foo();
>         foo1.num = 30;
> 
>         Foo foo2 = new Foo();
>         foo2.num = 40;
> 
>         SoftReference<Foo> sr1 = new SoftReference<Foo>(foo1);
>         SoftReference<Foo> sr2 = new SoftReference<Foo>(foo2);
> 
>         foo1.sr = sr2;
> 
>         return sr1;
>     }
> 
>     public static void fillUpTheHeap() {
>         String[] ar = new String[1000000];
>         for (int i = 1; i < 1000000; i++) {
>             String str = String.valueOf(i);
>             if (i % 100 == 0) {
>                 ar[i] = str;
>             }
>         }
>     }
> 
>     public static void main(String[] args) {
>         SoftReference<Foo> sr1 = setUp();
> 
>         fillUpTheHeap();
>         //System.gc();
>         //Runtime.getRuntime().gc(); // Just in case System.gc doesn't trigger GC.
> 
>         Foo foo1 = sr1.get();
>         Foo foo2 = foo1.sr.get();
> 
>         System.out.println(foo1.num);
>         System.out.println(foo2.num);
>     }
> }
> ```
> 
> Run with:
> 
> ```shell
> javac SoftStrongSoft.java -source 11 -target 11
> export MMTK_NO_REFERENCE_TYPES=false
> /path/to/java -XX:+UseThirdPartyHeap -Xmx64M SoftStrongSoft
> ```
> 
> Result:
> 
> ```
> [2022-04-08T04:37:18Z INFO  mmtk::memory_manager] Initialized MMTk with SemiSpace
> [2022-04-08T04:37:19Z INFO  mmtk::plan::global]   [POLL] copyspace0: Triggering collection
> [2022-04-08T04:37:19Z INFO  mmtk::scheduler::gc_work] End of GC
> [2022-04-08T04:37:19Z INFO  mmtk::plan::global]   [POLL] copyspace1: Triggering collection
> [2022-04-08T04:37:19Z INFO  mmtk::scheduler::gc_work] End of GC
> 30
> Exception in thread "main" java.lang.NullPointerException
>         at SoftStrongSoft.main(SoftStrongSoft.java:45)
> ```
> 
> The first `SoftReference`, i.e. `sr1` is retained, but the second `sr2` is not. I confirmed that emergency collection did not happen.
> 
> I think we should repeat the soft reference processing until no further objects can be reached by traversing soft-strong-soft-srong-soft-...-soft-strong references. FinalizerReference will need it, too. It should not be needed for weak/phantom references because weak/phantom references never keep their referents alive.

On second thought, I feel there is no guarantee that soft references in this example will be retained, and it is not a correctness bug if the soft references are not retained. It probably will depend on the GC implementation.

wks (approved):
I observed the same behaviour in JikesRVM.  The Java API documentation seems to allow this behaviour for SoftReference.

This PR looks good now.

qinsoon:
This PR introduces weak reference, but weak reference processing is disabled by default. The following results show that with disabled weak reference processing, the GC/STW time is unchanged, in comparison with `master` (benchmarked with OpenJDK). The performance issues with weak reference processing will be further addressed in future PRs.

I will merge this PR soon.

weak-ref-disable

Issue #6, opened 2020-03-26
Rust nightly features audit


qinsoon:
In GitLab by @pzakopaylo01 on Dec 16, 2017, 19:29

This issue tracks dependencies on nightly rust for the mmtk project. Please add any nightly features that are used to the list below.

* inline assembly (`asm`). Required for upcalls into JikesRVM due to different calling convention.

* compile-time evaluated functions (`const_fn`). Required because we have a global synchronised counter. Since this counter is explicitly written to be lockless, we would incur an unacceptable performance overhead by wrapping it in a `lazy_static!`.

PR #565, opened 2022-03-29
Use stable Rust toolchain

qinsoon:
This PR removes our use of Rust nightly features. It closes https://github.com/mmtk/mmtk-core/issues/6.

Code changes in the PR is driven by removing nightly features:
* `associated_type_defaults`: We can no longer use `SFTProcessEdges` as the default `ProcessEdgesWorkType` for `GCWorkContext`. We specify `SFTProcessEdges` in each trait impl.
* `box_syntax`: We now use `Box::new(X)` rather than `box X`.
* `const_fn_trait_bound`: We cannot use trait bound `<T: X>` with const functions.
  * `InitializeOnce` now is created without an initialization function. The function is provided with `initialize_once()`.
  * `Fixture::new` is not a const function any more.
* `adt_const_params`: We can only use types like integers and bool as const type parameters for a type (we cannot use enum). So for Immix related work packet types, we use `u8` to represent the trace kind instead of enum.
* `step_trait`: I refactor this heavily. Immix `Line`/`Block`/`Chunk` implements step, so we can iterate on them. Though `Step` provides both `forward` and `backward`, we only use `forward`. I introduce `trait Region` and `struct RegionIterator<R>`, and line/block/chunk all implements `Region` and can use `RegionIterator`.


wks (approved):
LGTM

Issue #559, opened 2022-03-14
Remove TransitiveClosure param from Space::trace_object and Scanning::scan_object


wks:
TL;DR: The type parameter of `XxxxxSpace::trace_object<T: TransitiveClosure>` and the `trace: &mut T` parameter can be removed, and the information should be carried by return value.

TL;DR2: The `trace: &mut T` parameter of `Scanning::scan_object` should also be removed and replaced by a lambda. (See [comment below](https://github.com/mmtk/mmtk-core/issues/559#issuecomment-1067683479))

TODO list:
-   [ ] Remove the `<T: TransitiveClosure>` parameter from `XxxxxSpace::trace_object`
-   [x] Remove the `<T: TransitiveClosure>` parameter from `Scanning::scan_object`
-   [ ] Remove the `TransitiveClosure` trait completely

# What?

Each space ([like this one](https://github.com/mmtk/mmtk-core/blob/afa2718d7e73ff2b18c3e92826939c525b5d09f1/src/policy/copyspace.rs#L204)) has a `trace_object<T: TransitiveClosure>` method that takes a `trace: &mut T` parameter.

We mentioned the removal of the `<T: TransitiveClosure>` type parameter in https://github.com/mmtk/mmtk-core/issues/110.  In that issue, @qinsoon mentioned that once we remove plan-specific ProcessEdgesWork, we can remove it.

But I think we can remove it earlier than that, and *plan-specific edge-processing is not even the core of the problem*.  The `trace` parameter is more like an "out parameter" in some languages.

# What does `trace: &mut T` do?

It only serves one purpose for any Space: **to enqueue the newly marked object**.

`Space::trace_object` does two things:

1.  Mark an object if it is not already marked.  When an object is marked (in GC terms, when a "grey" object becomes "black"), its edges should be enqueued (i.e. adjacent object shall turn "grey").
2.  Forward an object if it is not already forwarded.  If the current object reference points to a tombstone, find the address of the forwarded object.

So the result of the function should also consist of two parts:
1.  If the object is newly marked, and
2.  if the object is forwarded and, if it is, what is the new address.

The current `trace_object` signature is
```rust
    pub fn trace_object<T: TransitiveClosure>(
        &self,
        trace: &mut T,
        object: ObjectReference,
        semantics: Option<CopySemantics>,
        worker: &mut GCWorker<VM>,
    ) -> ObjectReference {
```

[...]

PR #566, opened 2022-03-30
Split the TransitiveClosure trait

wks:
This is the first step of removing the `TransitiveClosure` trait: https://github.com/mmtk/mmtk-core/issues/559

Moved the `process_edge` function of the `TransitiveClosure` trait into
a dedicated trait `EdgeVisitor` for visiting edges of an object (or
stack roots).  As a result,

1.  VM bindings only require the `EdgeVisitor` trait for object
    scanning.  VM bindings no longer use the `TransitiveClosure` in any
    way.

2.  Now `TransitiveClosure` only serves one purpose in `mmtk-core`, i.e.
    as a super-trait of ProcessEdgesWork for the `process_node` method.
    It is only called by `trace_object` of different spaces.

With this change, it will be easy to refactor `trace_object` and remove
`TransitiveClosure` completely.


wks:
binding-refs
OPENJDK_BINDING_REF=scanning-no-transitive-closure
JIKESRVM_BINDING_REPO=wks/mmtk-jikesrvm
JIKESRVM_BINDING_REF=scanning-no-transitive-closure
V8_BINDING_REPO=wks/mmtk-v8
V8_BINDING_REF=scanning-no-transitive-closure

src/vm/scanning.rs
```diff
@@ -20,11 +27,11 @@ pub trait Scanning<VM: VMBinding> {
     /// in favor of bulk scanning `scan_objects`.
     ///
     /// Arguments:
-    /// * `trace`: The `TransitiveClosure` to use for scanning.
+    /// * `edge_visitor`: Called back for each edge.
     /// * `object`: The object to be scanned.
     /// * `tls`: The GC worker thread that is doing this tracing.
-    fn scan_object<T: TransitiveClosure>(
-        trace: &mut T,
+    fn scan_object<EV: EdgeVisitor>(
```
qinsoon, line 33: I would suggest we change and clean up for `scan_objects()` as well. I would think the signature for `scan_objects()` should be the same as `scan_object()`, except that we use `&[ObjectReference]` instead of `ObjectReference`. 

There are multiple issues with `scan_objects()` (some are related with the changes in this PR, some are not):
* `scan_objects()` should take `EdgeVisitor` as type parameter rather than `ProcessEdgesWork`. Actually in OpenJDK's `scan_objects()` implementation, it simply creates its own `ObjectsClosure` and call `scan_object()` (which expects `EdgeVisitor`).
* OpenJDK does not need to create its own `ObjectsClosure`, and it can use the `ObjectsClosure` from mmtk-core. Actually OpenJDK first introduced `ObjectsClosure`, then `ObjectsClosure` was added to mmtk-core as V8 would use it as well. However, we never change that in OpenJDK. I am mentioning this as you also changed that part of the code in the OpenJDK binding.
* The parameter `tls` in `scan_object()` is inconsistent with `worker` in `scan_objects()`. I think we can get `tls` from `worker` but not the other way around. We probably can change `tls` in `scan_object()` to `worker` as well. I think @udesou has this change in his mmtk-core fork, and would like to upstream this change at some point as well.

As you are changing code around this, I would suggest trying to fix these issues as well. Maybe try include these changes if you find them related or easy enough to do.

qinsoon: Can you remove the comments that says `scan_object()` may be removed in the future? I think I wrote that, and it is incorrect. We probably will keep both. Generally, we would use `scan_objects()`. But for some policies like Immix, we would also need to do some extra work for each scanned object (like marking lines), we would have to use `scan_object()`.

qinsoon (commented):
LGTM. I will change this to an approval when we are about to merge this (probably after v0.11).

src/vm/scanning.rs
wks: >   I would suggest we change and clean up for scan_objects() as well. I would think the signature for scan_objects() should be the same as scan_object(), except that we use &[ObjectReference] instead of ObjectReference.

Yes. I agree. I also think all functions that can only be called by GC worker threads should take `&mut GCWorker` as the first parameter. `GCWorker` has a `tls: VMWorkerThread` field, and `VMWorkerThread` is only meaningful when calling C functions.

wks: > There are multiple issues with scan_objects() (some are related with the changes in this PR, some are not):
>
>    scan_objects() should take EdgeVisitor as type parameter rather than ProcessEdgesWork. Actually in OpenJDK's scan_objects() implementation, it simply creates its own ObjectsClosure and call scan_object() (which expects EdgeVisitor).
>    OpenJDK does not need to create its own ObjectsClosure, and it can use the ObjectsClosure from mmtk-core. Actually OpenJDK first introduced ObjectsClosure, then ObjectsClosure was added to mmtk-core as V8 would use it as well. However, we never change that in OpenJDK. I am mentioning this as you also changed that part of the code in the OpenJDK binding.

I am surprised that the OpenJDK binding did not fork the work packet into multiple work packets.  I would assume that if `objects.size()` is too large, it will split it into multiple work packets.  After a second thought, it can't, because the openjdk binding is not privileged to create the private `ScanObjects` work packet.

I think it is worth refactoring `scan_objects` so that it is the `mmtk-core`'s responsibility to create the `ProcessEdgesWork` work packet.

qinsoon: > I am surprised that the OpenJDK binding did not fork the work packet into multiple work packets. I would assume that if `objects.size()` is too large, it will split it into multiple work packets. After a second thought, it can't, because the openjdk binding is not privileged to create the private `ScanObjects` work packet.

I guess we will not see an `ObjectReference` slice larger than `ProcessEdgesWork::CAPACITY`. So it should be fine. Plus creating more packets has the overhead of dispatching them.

> I think it is worth refactoring `scan_objects` so that it is the `mmtk-core`'s responsibility to create the `ProcessEdgesWork` work packet.

I am not exactly sure what you were referring to. But generally I find it confusing that some work is created from the binding, and it is not obvious if we only look at the mmtk-core code and try figure out where the packet is created. So it is probably a good idea if we move some packet creation back to mmtk-core. However, we would still allow a binding to create work if they want to.

wks: I made changes according to your suggestions.  However, I couldn't change `tls: VMWorkerThread` to `&mut GCWorker<VM>` because of some borrowing issue. It will break the following simple code:
```rust
let mut closure = ObjectsClosure::<E>::new(worker);  // NOTE: first mutable borrow here
<E::VM as VMBinding>::VMScanning::scan_objects(worker, &self.buffer, &mut closure);  // ERROR: worker was mutably borrowed.
```

Rust doesn't like a reference to be borrowed in two places. That is actually right, because if it were borrowed twice, both `ObjectClosure` and `scan_objects` could modify `worker` and it is easy to make mistakes like "changing a collection while iterating through it".

The ideal way to do this kind of callback in Rust is to add a `&mut GCWorker` parameter to `EdgeVisitor::visit_edge`, too, and let `scan_objects` pass the `&mut Worker` back.  But I think it is overly complicated.  In fact, neither OpenJDK nor JikesRVM used the `tls: VMWorkerThread` when scanning an object (JikesRVM uses it for debug purpose).  It doesn't need the Rust-level `GCWorker` structure, either.

So I made both `scan_object` and `scan_objects` take `tls: VMWorkerThread` as parameter for the following reasons:
1.  mmtk-core does not need to access any data in `GCWorekr` to support scanning object. It only uses `GCWorker` to handle the *result* of scanning objects (i.e. the edges).
2.  If any VM really needs the TLS to scan objects, we just give the TLS to the VM.

src/policy/immix/immixspace.rs
```diff
@@ -592,15 +592,12 @@ impl<Edges: ProcessEdgesWork> ScanObjectsAndMarkLines<Edges> {
 }
 
 impl<E: ProcessEdgesWork> GCWork<E::VM> for ScanObjectsAndMarkLines<E> {
-    fn do_work(&mut self, worker: &mut GCWorker<E::VM>, mmtk: &'static MMTK<E::VM>) {
+    fn do_work(&mut self, worker: &mut GCWorker<E::VM>, _mmtk: &'static MMTK<E::VM>) {
```
qinsoon, line 595: As `GCWorker` has a field that points to `MMTK`, it seems we do not need this `mmtk` argument in `do_work()`. Right? No change is needed for the PR.

qinsoon (approved):
LGTM

src/policy/immix/immixspace.rs
wks: That's right.  We don't need the `&MMTK<VM>` argument to `do_work`, because `GCWorker` already refers to all shared and owned data structures it needs for executing `GCWork`.  Not only that, but `GCController` also has an `&MMTK<VM>` reference, because it is a thread-local context, too, and `mmtk` is something it needs for doing its work.

I made `GCWorker::mmtk` public. I think it should be accessible to `GCWork`, because `GCWork`, by definition, is a piece of work `GCWorker` performs, and should have access to `GCWorker` fields.

caizixian:
@wks , as discussed during the meeting, perhaps we can think of better names for some types and traits. In particular, the `TransitiveClosure` type now only has `process_node`. Considering that we might eventually remove that trait, maybe it doesn't matter whether we rename it now.

wks:
@caizixian That's right.  I'll remove it in two more PRs.  One of them will remove the only use of `TransitiveClosure::process_node`, namely in `trace_object`; the other will remove the `TransitiveClosure` trait itself, as planned in https://github.com/mmtk/mmtk-core/issues/559

PR #567, opened 2022-03-31
Minimal supported rust version

qinsoon:
This PR adds minimal supported Rust version, and adds test matrix for our CI tests.

* adds `rust-version` in `Cargo.toml` to specify MSRV. MSRV 1.57.0
* adds `cargo-msrv.yml` which verify MSRV, and if it fails, it figures out MSRV (but won't attempt to update it in `Cargo.toml`).
* updates `rust-toolchain` to `1.59.0`, which we use for test.
* adds test matrix. Currently we test [`x86_64`, `i686`] x [`MSRV`, `rust-toolchain`]
  * as we have a separate 32 bit test now, tests for 32 bit is removed from our test scripts. The test script will be run on both `x86_64` and `i686`.
  * fix a few issues related to tests in `i686`.


caizixian (approved):
LGTM

PR #568, opened 2022-03-31
Fix allocation bytes increment for precise stress test

k-sareen:
This change fixes two related bugs:

  1. When a GC occurs, there will be multiple nested allocation calls
     that will try to increase the allocation bytes resulting in double
     counting of an allocation. We mitigate this by only allowing the
     outermost scope to increase the allocation bytes

  2. ImmixAllocator missed updates to the allocation bytes counter after
     the first GC. This was due to the ImmixAllocator reusing lines
     which implicitly updated the cursor and limit. The updated cursor
     and limit didn't fail the test for fast path and hence, the
     allocator missed updates to the allocation bytes. We fix this by
     directly calling the slow path (instead of going through the fast
     path) after obtaining a recycled line in the case when stress test
     is active

### DIFF
--- a/src/util/alloc/allocator.rs
+++ b/src/util/alloc/allocator.rs
@@ -118,20 +118,24 @@ pub trait Allocator<VM: VMBinding>: Downcast {
     /// Return the [`VMThread`] associated with this allocator instance.
     fn get_tls(&self) -> VMThread;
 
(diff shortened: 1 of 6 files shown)


k-sareen:
I also updated some of the doc comments where I noticed the links were broken.

k-sareen:
Closed in favour of #571

PR #571, opened 2022-04-11
Fix allocation bytes increment for precise stress test

k-sareen:
This change fixes two related bugs:

  1. Currently, when a thread-local block is exhausted, there will be
     multiple nested allocation calls (specifically to
     `alloc_slow_inline`) that will try to increase the allocation bytes
     resulting in double counting of an allocation. We mitigate this by
     ensuring that calls to `alloc_slow_inline` are not re-entrant

  2. ImmixAllocator missed updates to the allocation bytes counter after
     the first GC. This was due to the ImmixAllocator reusing lines
     which implicitly updated the cursor and limit. The updated cursor
     and limit didn't fail the test for fast path and hence, the
     allocator missed updates to the allocation bytes. We fix this by
     directly calling the slow path (instead of going through the fast
     path) after obtaining a recycled line in the case when stress test
     is active

### DIFF
--- a/src/plan/global.rs
+++ b/src/plan/global.rs
@@ -373,7 +373,7 @@ pub struct BasePlan<VM: VMBinding> {
     stacks_prepared: AtomicBool,
     pub mutator_iterator_lock: Mutex<()>,
     // A counter that keeps tracks of the number of bytes allocated since last stress test
-    allocation_bytes: AtomicUsize,
+    pub allocation_bytes: AtomicUsize,
     // Wrapper around analysis counters
     #[cfg(feature = "analysis")]
     pub analysis_manager: AnalysisManager<VM>,
--- a/src/util/alloc/immix_allocator.rs
+++ b/src/util/alloc/immix_allocator.rs
@@ -7,6 +7,7 @@ use crate::policy::space::Space;
 use crate::util::alloc::Allocator;
 use crate::util::linear_scan::Region;
 use crate::util::opaque_pointer::VMThread;
+use crate::util::rust_util::unlikely;
 use crate::util::Address;
 use crate::vm::*;
 
@@ -34,14 +35,6 @@ pub struct ImmixAllocator<VM: VMBinding> {
     request_for_large: bool,
     /// Hole-searching cursor
     line: Option<Line>,
-    /// Are we doing alloc slow when stress test is turned on. This is only set to true,
-    /// during the allow_slow_once_stress_test() call. In the call, we will restore the correct
-    /// limit for bump allocation, and call alloc() to try resolve the allocation request with
-    /// the thread local buffer. If we cannot do the allocation from the thread local buffer,
-    /// we will eventually call allow_slow_once_stress_test(). With this flag set to true, we know
-    /// we are resolving an allocation request and have failed the thread local allocation. In
-    /// this case, we will acquire new block from the space.
-    alloc_slow_for_stress: bool,
 }
 
 impl<VM: VMBinding> ImmixAllocator<VM> {
@@ -147,29 +140,24 @@ impl<VM: VMBinding> Allocator<VM> for ImmixAllocator<VM> {
             return ret;
         }
 
-        // We are not yet required to do a stress GC. We will try to allocate from thread local buffer if possible.
-        // Restore the fake limit to the normal limit so we can do thread local alloaction normally.
+        // We are not yet required to do a stress GC. We will try to allocate from thread local
+        // buffer if possible.  Restore the fake limit to the normal limit so we can do thread
+        // local allocation normally. Check if we have exhausted our current thread local block,
+        // and if so, then directly acquire a new one
         self.restore_limit_for_stress();
-        let ret = if self.alloc_slow_for_stress {
-            // If we are already doing allow_slow for stress test, and reach here, it means we have failed the
-            // thread local allocation, and we have to get a new block from the space.
+        let ret = if self.require_new_block(size, align, offset) {
+            // We don't have enough space in thread local block to service the allocation request,
+            // hence allocate a new block
             trace!(
                 "{:?}: alloc_slow_once_precise_stress - acquire new block",
                 self.tls
             );
             self.acquire_clean_block(size, align, offset)
         } else {
-            // Indicate that we are doing alloc slow for stress test. If the alloc() cannot allocate from
-            // thread local buffer, we will reach this method again. In that case, we will need to poll, rather
-            // than attempting to alloc() again.
-            self.alloc_slow_for_stress = true;
-            // Try allocate. The allocator will try allocate from thread local buffer, if that fails, it will
-            // get a clean block.
-            trace!("{:?}: alloc_slow_once_precise_stress - alloc()", self.tls);
-            let ret = self.alloc(size, align, offset);
-            // Indicate that we finish the alloc slow for stress test.
-            self.alloc_slow_for_stress = false;
-            ret
+            // This `alloc()` call should always succeed given the if-branch checks if we are out
+            // of thread local block space
+            trace!("{:?}: alloc_slow_once_precise_stress - alloc()", self.tls,);
+            self.alloc(size, align, offset)
         };
         // Set fake limits
         self.set_limit_for_stress();
@@ -200,7 +188,6 @@ impl<VM: VMBinding> ImmixAllocator<VM> {
             large_limit: Address::ZERO,
             request_for_large: false,
             line: None,
-            alloc_slow_for_stress: false,
         }
     }
 
@@ -231,7 +218,23 @@ impl<VM: VMBinding> ImmixAllocator<VM> {
     fn alloc_slow_hot(&mut self, size: usize, align: usize, offset: isize) -> Address {
         trace!("{:?}: alloc_slow_hot", self.tls);
         if self.acquire_recyclable_lines(size, align, offset) {
-            self.alloc(size, align, offset)
+            // If stress test is active, then we need to go to the slow path instead of directly
+            // calling `alloc()`. This is because the `acquire_recyclable_lines()` function
+            // manipulates the cursor and limit if a line can be recycled and if we directly call
+            // `alloc()` after recyling a line, then we will miss updating the `allocation_bytes`
+            // as the newly recycled line will service the allocation request. If we set the stress
+            // factor limit directly in `acquire_recyclable_lines()`, then we risk running into an
+            // loop of failing the fastpath (i.e. `alloc()`) and then trying to allocate from a
+            // recyclable line.  Hence, we bring the "if we're in stress test" check up a level and
+            // directly call `alloc_slow_inline()` which will properly account for the allocation
+            // request as well as allocate from the newly recycled line
+            let stress_test = self.plan.base().is_stress_test_gc_enabled();
+            let precise_stress = self.plan.base().is_precise_stress();
+            if unlikely(stress_test && precise_stress) {
+                self.alloc_slow_inline(size, align, offset)
+            } else {
+                self.alloc(size, align, offset)
+            }
         } else {
             self.alloc_slow_inline(size, align, offset)
         }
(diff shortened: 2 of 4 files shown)


k-sareen:
I believe that `alloc_slow_hot` is not considered fast path. Let me know if I should run performance analysis on standard Immix, though.

src/util/alloc/immix_allocator.rs
```diff
@@ -126,48 +120,47 @@ impl<VM: VMBinding> Allocator<VM> for ImmixAllocator<VM> {
...
-            self.alloc_slow_for_stress = true;
             // Try allocate. The allocator will try allocate from thread local buffer, if that fails, it will
             // get a clean block.
-            trace!("{:?}: alloc_slow_once_precise_stress - alloc()", self.tls);
-            let ret = self.alloc(size, align, offset);
-            // Indicate that we finish the alloc slow for stress test.
-            self.alloc_slow_for_stress = false;
-            ret
+            trace!("{:?}: alloc_slow_once_precise_stress - alloc()", self.tls,);
+            self.alloc(size, align, offset)
```
qinsoon, line 163: Is this `alloc()` call guaranteed to succeed? If so, the comments above it need an update. If not, does that mean `alloc_slow_once_precise_stress()` is still reentrant (and we may have the issue of double counting some allocations)?

src/util/alloc/immix_allocator.rs
```diff
@@ -229,7 +221,19 @@ impl<VM: VMBinding> ImmixAllocator<VM> {
...
-            self.alloc(size, align, offset)
+            // If stress test is active, then we need to go to the slow path instead of directly
+            // calling alloc(). This is because the `acquire_recyclable_lines()` function
+            // manipulates the cursor and limit; if we directly call `alloc()` then we will miss
+            // updating the allocation as the TLAB will service the allocation request. If we set
+            // the limit in `acquire_recyclable_lines()`, then we risk running into an infinite
+            // loop of failing the fastpath and then trying to allocate from a recyclable line.
+            let stress_test = self.plan.base().is_stress_test_gc_enabled();
+            let precise_stress = self.plan.base().is_precise_stress();
+            if stress_test && precise_stress {
```
qinsoon, line 232: This seems to be the only change that may affect performance. It would be better to have it measured.

qinsoon (commented):
LGTM. It is much cleaner than the previous PR. And thanks a lot for fixing/adding comments in the code. It would be better if we have performance results for it before we merge it.

src/util/alloc/immix_allocator.rs
k-sareen: Alright -- will measure it :+1:

src/util/alloc/immix_allocator.rs
k-sareen: Hmm. I am fairly certain it should succeed barring any OOM cases as the `require_new_block()` function basically acts like an `alloc()` call (note the hierarchy of checks in that function) without actually allocating anything. So if the function returns `false`, then we almost certainly have space for the current allocation request. Let me know if I should test this further.

Good catch regarding the comments. I will update them

k-sareen:
image
image
image

<details>
<summary> plotty link (not public) </summary>
[http://squirrel.anu.edu.au/plotty/kunals/bench-test/p/UVvbyE](http://squirrel.anu.edu.au/plotty/kunals/bench-test/p/UVvbyE)
</details>

Seems like it does have a minor performance hit for a few benchmarks. The run was 20 invocations with 5 iterations each.

qinsoon:
> image image image
> 
> plotty link (not public)
> Seems like it does have a minor performance hit for a few benchmarks. The run was 20 invocations with 5 iterations each. For some reason h2 and graphchi have 1 outlier which overwhelms the entire graph. The performance trend might be more visible if you hide h2 as well.

I feel the results look fine. I have seen weird results for `h2` as well. But it would be better if you can run it again with like 40 invocations.

k-sareen:
I've removed the formatting and documentation changes -- will submit a new PR for those.

k-sareen:
If the above solution is not agreeable, then I propose another solution where we return an `enum` for `acquire_recyclable_lines()` which can essentially be one of three cases: we have successfully recycled a line, we have successfully recycled a line but we're in stress test, and we have not been able to recycle a line. Only the first case will actually call `alloc()` the other two will call `alloc_slow_inline()` automatically and hence we can remove the extra check in the current solution (though we will then have to insert a new check in `acquire_recyclable_lines()` anyway).

We need the three values, in my eyes at least, since we use `acquire_recyclable_lines()` in `require_new_block()` which checks if we are totally out of space in our current thread local buffer. Hence, there needs to be a distinction between "we are totally out of space" and "we've recycled a line but we're in stress test". This solution could be considered cleaner, but unknown performance overhead (which I can measure if required).

k-sareen:
@qinsoon what is the status of this? Should I do more performance evaluation/try the above approach as well or can we merge this PR?

qinsoon:
> @qinsoon what is the status of this? Should I do more performance evaluation/try the above approach as well or can we merge this PR?

I don't think we will merge this PR at this state, as we know it has slowdown. Besides, the result for `tomcat` looks noisy and showed an improvement which we don't expect. But it may further cover the actual overhead on average.

You could simply try use `likely` to tell the compiler, stress test is an unlikely case. In that case, we would just do a check but no jump. This may or may not improve performance.

You could try what you proposed earlier. It may help. 

The last resort could be just making `precise_stress` a feature. So its related code always is conditionally compiled.

k-sareen:
Alright -- I'll try the `likely` and my above solution :+1:

k-sareen:
It's actually not that bad, if any performance regression at all: http://squirrel.anu.edu.au/plotty/kunals/bench-test/p/yBDS7E

Everything is under ~0.05% with all error bars overlapping. This is 30 invocations and 5 iterations. The only outlier is tomcat, but that can be explained by the fact that it just happened to trigger more GCs and hence the STW time jumped up a bit (the number of GCs seems pretty volatile for this benchmark -- maybe controlling number of GCs with non-precise stress would have been a better experiment).

Build "Change in alloc_slow_hot" is this PR but with `unlikely()` and build "Change in acquire_recyclable_lines" is my above mentioned solution + `unlikely()`.

k-sareen:
Inlining images for posterity:

<details> <summary> Number of GCs </summary>
image
</details>
<details> <summary> Mutator Time </summary>
image
</details>
<details> <summary> STW Time </summary>
image
</details>
<details> <summary> Total Time </summary>
image
</details>

qinsoon:
> It's actually not that bad, if any performance regression at all: http://squirrel.anu.edu.au/plotty/kunals/bench-test/p/yBDS7E The `h2` benchmark is too noisy (even on `master`), so if you remove it you can see that there is no noticeable performance overhead involved with the change.
> 
> Everything is under ~0.05-1% with all error bars overlapping. This is 30 invocations and 5 iterations. The only outlier is `tomcat`, but that can be explained by the fact that it just happened to trigger more GCs and hence the STW time jumped up a bit (the number of GCs seems pretty volatile for this benchmark -- maybe controlling number of GCs with non-precise stress would have been a better experiment).
> 
> Build "Change in alloc_slow_hot" is this PR but with `unlikely()` and build "Change in acquire_recyclable_lines" is my above mentioned solution + `unlikely()`.
> 
> Note this is only on one microarchitecture (Intel i7-6700K, Skylake). If required, then I can run the same experiment on a different microarchitecture.

The result looks good. Can you add the changes to the PR? Either is fine.

src/plan/global.rs
```diff
@@ -441,6 +442,9 @@ impl<VM: VMBinding> BasePlan<VM> {
         global_side_metadata_specs: Vec<SideMetadataSpec>,
     ) -> BasePlan<VM> {
         let stats = Stats::new(&options);
+        use crate::util::constants::DEFAULT_STRESS_FACTOR;
+        let is_stress_test_gc_enabled = *options.stress_factor != DEFAULT_STRESS_FACTOR
```
qinsoon, line 446: I am guessing you were trying to optimize this. But this happens when we create `MMTK`, and options can be set after creating `MMTK`. If options are set after creating `MMTK`, this value could be incorrect.

There are a few things you can do to fix this:
1. revert the change, and always check the option where it is used. 
2. do not allow those options to be set by command line arguments (we do have a few options that cannot be set through command line args - but this is not preferred, and will be fixed at some point). 
https://github.com/mmtk/mmtk-core/blob/3dbdd7ae1076fb022f483b8b91cf37425dc6f0ee/src/util/options.rs#L341-L344
3. add another option, something like `enable_stress_gc`. The two factor options will only be used when `enable_stress_gc` is true.

k-sareen: Ah sorry I didn't realize that would be an issue. I noticed that the check is potentially expensive since it is three loads + 1 comparison. Its performance does not matter in `alloc_slow_inline()`, which is the only other place it is used. However, for the intermediate path, its performance overhead was noticeable. Hence, I optimized it to be 1 load instead.

Hmm... I haven't looked into the option setting code yet, but would it not be easier/better to just update the new boolean field if and when the commandline stress factor option is processed?

qinsoon: > would it not be easier/better to just update the new boolean field if and when the commandline stress factor option is processed?

I feel that would be worse. That would mean the options need to know that it is being cached in the plan (so it can update the value in the plan), and it introduces unnecessary dependency. You could just add a new option `enable_stress_gc`.

k-sareen: Right I see. In that case, what is the semantics of `enable_stress_gc`? If `enable_stress_gc=false` (I guess which is the default?) but `stress_factor=8192` then will we stress test or not?

qinsoon: I assume no. If we do not enable stress gc, we do not further check the factors (it would be good to give a warning if stress/analysis factors are set but stress gc is disabled).

qinsoon: All what I am saying above is under the assumption that the old code of checking actually has an overhead. If it does not have overhead, we could just check against the factor options. You could simply revert this change first, and run one or two benchmarks against the build here: https://github.com/mmtk/mmtk-core/pull/571#issuecomment-1104629725, and see if there is any overhead at all.

k-sareen: Alright. Did you mean this build (i.e. the one with `unlikely()`): https://github.com/mmtk/mmtk-core/pull/571#issuecomment-1120199298?

qinsoon: Umm, using the later build with `unlikely` seems better. You just need to figure out the cost of 'cached boolean' vs 'accessing option directly', with or without `unlikely()` (but it would be better with unlikely, as you said). You could just pick one or two benchmarks that seemed to have a slowdown in https://github.com/mmtk/mmtk-core/pull/571#issuecomment-1104629725.

k-sareen: https://squirrel.anu.edu.au/plotty/kunals/bench-test/#0|2022-06-03-immix-stress-fix-shrew-2022-06-03-Fri-155155&benchmark^build^invocation^iteration&time^time.other^time.stw&|10&iteration^1^4&benchmark^2^cassandra&benchmark^2^h2|20&1^invocation|70&build^immix-stress-pr|30&1&benchmark&build;jdk11-master-is_stress-core_28e1af1d-jdk_040da09-vm_ca90b43f0f5|41&Histogram%20(with%20CI)^build^benchmark&

Both builds have minimal, if any, impact on the mutator time. I'll stick with the no cached boolean that is currently on master then.

src/util/rust_util.rs
```diff
@@ -7,6 +7,31 @@ pub const fn min_of_usize(a: usize, b: usize) -> usize {
...
 
+// likely() and unlikely() compiler hints in stable Rust
+// [1]: https://github.com/rust-lang/hashbrown/blob/a41bd76de0a53838725b997c6085e024c47a0455/src/raw/mod.rs#L48-L70
+// [2]: https://users.rust-lang.org/t/compiler-hint-for-unlikely-likely-for-if-branches/62102/3
+#[cfg(not(feature = "nightly"))]
+#[inline]
+#[cold]
+fn cold() {}
+
+#[cfg(not(feature = "nightly"))]
```
qinsoon, line 18: I think you missed these lines:
```rust
#[cfg(feature = "nightly")]
use core::intrinsics::{likely, unlikely};
```
https://github.com/rust-lang/hashbrown/blob/a41bd76de0a53838725b997c6085e024c47a0455/src/raw/mod.rs#L45

Otherwise we cannot compile with a nightly compiler.

k-sareen: Yep my bad. Done

qinsoon:
binding-refs
OPENJDK_BINDING_REF=fix/immix-stress

k-sareen:
Hm? Why is MarkCompact failing?

qinsoon:
> Hm? Why is MarkCompact failing?

I started a rerun. I started the CI tests before adding the comment so it wasn't executing with your OpenJDK PR.

qinsoon:
The tests passed now. The JikesRVM test failed once (https://github.com/mmtk/mmtk-core/actions/runs/2445005252/attempts/3), but I think it is https://github.com/mmtk/mmtk-jikesrvm/issues/108 and unrelated with this PR.

qinsoon (approved):
LGTM

PR #572, opened 2022-04-19
Manual malloc space

qinsoon:


qinsoon:
I will redo this pull request. There will be some changes to this draft PR:
* We will provide `malloc`/`free` API, rather than using `alloc(..., AllocationSemantic::MALLOC)`.
* The `malloc` API should have the same signature as the library `malloc()`.
* A binding could call `malloc` in a non-application thread (which does not have a thread local `Mutator`).
* We could forward the malloc call to a third party library, or use our mimalloc implementation. For mimalloc, we need to maintain some thread-local metadata for mimalloc and that is not a part of `Mutator`. 
* If we forward the malloc call to third party malloc, we should store no metadata for it. We should not have any extra overhead.
* For our APIs such as `is_in_mmtk_spaces()`, we may return different results depending on our malloc implementation. For example, when we use thirdparty malloc, the returned address is not in our space. When we use our native malloc, the address is in our space.
* The `malloc` API has nothing to do with the (malloc) mark sweep -- no reuse code or anything.
* We could provide options on 1. whether the malloc'd memory is counted in our heap size, 2. whether we may trigger a GC in malloc (I am not sure if this is a good idea -- we could just count the malloc'd size, and GC in the next normal allocation).

PR #575, opened 2022-04-27
Use macro to generate process edges for plans

qinsoon:
This PR adds `PlanProcessEdges`, and a few macros/traits that are used in `PlanProcessEdges`. This PR removes policy-specific code from plan, and use macros to generate trace object work for each plan. This PR closes https://github.com/mmtk/mmtk-core/issues/258, and closes https://github.com/mmtk/mmtk-core/issues/576.

* add a trait `PolicyTraceObject`. Each policy provides an implementation of this. With this trait, a plan no longer include any policy-specific code in trace object.
* add a trait `PlanTraceObject` and related macros. With the macro, plan implementer can declaratively specify trace object behavior with attributes, and the implementation of `PlanTraceObject` will be generated by the macro.
* add a type `PlanProcessEdges`. This uses `PlanTraceObject` to provide a general implementation of `ProcessEdgesWork`. We no longer need any plan-specific process edges work.


src/policy/immix/immixspace.rs
```diff
@@ -113,6 +117,46 @@ impl<VM: VMBinding> Space<VM> for ImmixSpace<VM> {
...
+
+    #[inline(always)]
+    fn create_scan_work<E: ProcessEdgesWork<VM = VM>>(
+        &'static self,
+        nodes: Vec<ObjectReference>,
+    ) -> Box<dyn GCWork<VM>> {
+        Box::new(crate::policy::immix::ScanObjectsAndMarkLines::<E>::new(
+            nodes, false, self,
+        ))
+    }
```
wks, line 146: I have some crazy ideas about removing ScanObjectsAndMarkLines completely.

https://github.com/mmtk/mmtk-core/issues/576

You can try it in this PR if it can make your work less laborious, or we can do it in another PR.

macros/trace_object/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+    } else {
+        // Otherwise, just panic
+        quote! {
+            fn create_scan_work<E: crate::scheduler::gc_work::ProcessEdgesWork<VM = VM>>(&'static self, nodes: Vec<crate::util::ObjectReference>) -> Box<dyn crate::scheduler::GCWork<VM>> {
+                panic!("Unable to create a scan work packet for the plan (the plan does not name a #[scan_work] field, or a #[trace] field")
+            }
+        }
+    }
+}
```
wks, line 160: Ditto. If https://github.com/mmtk/mmtk-core/issues/576 is applicable, this entire function can go away, too, like `PolicyTraceObject::create_scan_work` itself.

macros/trace_object/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+extern crate proc_macro_error;
+extern crate quote;
+
+use proc_macro::TokenStream;
+use proc_macro_error::proc_macro_error;
+use syn::{parse_macro_input};
+use proc_macro_error::abort_call_site;
+use quote::quote;
+use syn::{DeriveInput, Field, TypeGenerics};
+use syn::__private::TokenStream2;
```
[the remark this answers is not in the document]
qinsoon: Yeah. That sounds good. I was puzzled by the types. I will fix this.

Cargo.toml
```diff
@@ -18,6 +18,8 @@ crate-type = ["rlib"]
 doctest = false
 
 [dependencies]
+macro-trace-object = { path = "macros/trace_object"}
```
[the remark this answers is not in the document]
qinsoon: Sure. I will add a prefix. But I would prefer putting each macro to a seperate crate so that each crate will be straightforward and simple to understand.

qinsoon:
Performance result for this PR. This PR only changes GC. These results show the stop-the-world time. More results can be found in the links at the end. The results are based on https://github.com/mmtk/mmtk-core/pull/575/commits/d09c58af475560da5cd827d3bc3c19e77a56209e (with the `scan_object()` change).

semispace
macro-ss-stw

gen immix
macro-genix-stw

immix
macro-genix-stw


Perf links (not publicly visible)
<details> <summary> semispace </summary>
https://squirrel.anu.edu.au/plotty/yilin/yp-2015/#0|bear-2022-05-05-Thu-100348-ss-macro-process-edges-new&benchmark^build^invocation^iteration&bmtime^time.other^time.stw&|10&iteration^1^2|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^build^benchmark&
</details>
<details> <summary> gen immix </summary>
https://squirrel.anu.edu.au/plotty/yilin/yp-2015/#0|bear-2022-05-02-Mon-095232-genix-macro-process-edges-new&benchmark^build^invocation^iteration&bmtime^time.other^time.stw&|10&iteration^1^2|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^build^benchmark&
</details>
<details> <summary> immix </summary>
https://squirrel.anu.edu.au/plotty/yilin/yp-2015/#0|bear-2022-04-28-Thu-152509-immix-macro-process-edges&benchmark^build^invocation^iteration&bmtime^time.other^time.stw&|10&iteration^1^2|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^build^benchmark&
</details>

src/policy/immix/immixspace.rs
qinsoon: This change is done in the PR.

src/plan/mod.rs
```diff
@@ -37,7 +37,7 @@ pub use plan_constraints::DEFAULT_PLAN_CONSTRAINTS;
 mod tracelocal;
 pub use tracelocal::TraceLocal;
 
-mod transitive_closure;
+pub(crate) mod transitive_closure;
```
wks, line 40: See comments in `transitive_closure.rs`.  If we move the code to a different module, we don't need to make this change.

src/scheduler/gc_work.rs
```diff
@@ -421,31 +421,40 @@ pub trait ProcessEdgesWork:
         // So maximum 1 `ScanObjects` work can be created from `nodes` buffer
     }
 
-    /// Create a new scan work packet. If SCAN_OBJECTS_IMMEDIATELY, the work packet will be executed immediately, in this method.
+    /// Start the a scan work packet. If SCAN_OBJECTS_IMMEDIATELY, the work packet will be executed immediately, in this method.
```
wks, line 424: Actually "Start" is not accurate. If not `SCAN_OBJECTS_IMMEDIATELY`, the work packet doesn't start until it is scheduled on a GC worker.

"Create a new" is more accurate in this context, at least it doesn't imply when it is executed.

wks (changes requested):
Mostly OK, except some minor spelling, documentation and structural problems.

src/plan/transitive_closure.rs
```diff
@@ -71,3 +71,163 @@ impl<'a, E: ProcessEdgesWork> Drop for ObjectsClosure<'a, E> {
...
+use crate::mmtk::MMTK;
+use crate::plan::Plan;
+use crate::scheduler::gc_work::ProcessEdgesBase;
+use std::ops::{Deref, DerefMut};
+
+/// This provides an implementation of [`ProcessEdgesWork`](scheduler/gc_work/ProcessEdgesWork). A plan that implements
+/// `PlanTraceObject` can use this work packet for tracing objects.
+pub struct PlanProcessEdges<
+    VM: VMBinding,
+    P: 'static + Plan<VM = VM> + PlanTraceObject<VM> + Sync,
```
wenyuzhao, line 118: `Plan` is already `'static  + Sync`

src/plan/transitive_closure.rs
```diff
@@ -71,3 +71,163 @@ impl<'a, E: ProcessEdgesWork> Drop for ObjectsClosure<'a, E> {
...
+    const KIND: TraceKind,
+> {
+    plan: &'static P,
+    base: ProcessEdgesBase<VM>,
+}
+
+impl<
+        VM: VMBinding,
+        P: 'static + PlanTraceObject<VM> + Plan<VM = VM> + Sync,
```
wenyuzhao, line 127: `Plan` is already `'static  + Sync`

src/plan/transitive_closure.rs
```diff
@@ -71,3 +71,163 @@ impl<'a, E: ProcessEdgesWork> Drop for ObjectsClosure<'a, E> {
...
+impl<E: ProcessEdgesWork, P: 'static + Plan<VM = E::VM> + PlanTraceObject<E::VM> + Sync>
+    GCWork<E::VM> for PlanScanObjects<E, P>
+{
+    fn do_work(&mut self, worker: &mut GCWorker<E::VM>, _mmtk: &'static MMTK<E::VM>) {
+        trace!("PlanScanObjects");
+        {
+            let tls = worker.tls;
+            let mut closure = ObjectsClosure::<E>::new(worker);
+            for object in &self.buffer {
+                use crate::vm::Scanning;
```
wenyuzhao, line 226: Move this to the top?

macros/trace_object/src/util.rs
```diff
@@ -0,0 +1,45 @@
...
+    fields: &'f FieldsNamed,
+    attr_name: &str,
+) -> Option<&'f Field> {
+    let mut result = None;
+
+    'each_field: for field in fields.named.iter() {
+        if let Some(attr) = get_field_attribute(field, attr_name) {
+            if result.is_none() {
+                result = Some(field);
+                continue 'each_field;
```
wenyuzhao, line 36: There're not nested loops. I guess a simple `continue` is enough?

wenyuzhao (commented):
Great PR. It helps refactor a large part of the codebase and makes the code cleaner and shorter. It also removes the dynamic dispatch introduced in `SFTProcessEdges`.

Here're some of my high-level concerns:
1. The macro and the unification of `trace_object` forces all the policies to have the same `trace_object` signature. But different policies will have different implementations of `trace_object` with different behaviours and parameters, or even different return values.  One immediate drawback of this PR is that if a policy wants to pass one more argument, it has to change the unified interface, which will also change the macro definition code and all the other policies to adapt to the interface change. This adds a cost to future development.
2. `#[derive(PlanTraceObject)]` implies that each plan or policy can only be bind to one `trace_object` implementation. For GC like Immix which requires multiple implementations, it has to introduce another generic args `TraceKind`. I'm afraid this is probably not a nice and clean approach. `TraceKind` was introduced _only_ because ImmixSpace has two different `ProcessEdges`, and they share a large amount of common code. But this is not the case for other policies.  For policies with multiple different `ProcessEdges` and `trace_object`, introducing `TraceKind` again in `PlanTraceObject` will only make things more complicated.
   * In addition, different policies defines different `TraceKind` values, this makes the meaning of `TraceKind` become different at different places. This also makes the compiler (or human) harder to check if all the `TraceKind` value matching and comparison code are written correctly.

qinsoon:
Thanks for the reviews. Before going into each comment, I will just address your concerns first.

> Here're some of my high-level concerns:
> 
> 1. The macro and the unification of `trace_object` forces all the policies to have the same `trace_object` signature. But different policies will have different implementations of `trace_object` with different behaviours and parameters, or even different return values.  One immediate drawback of this PR is that if a policy wants to pass one more argument, it has to change the unified interface, which will also change the macro definition code and all the other policies to adapt to the interface change. This adds a cost to future development.

Do you have an example in your mind about extra parameters in `trace_object()`? If we need extra parameters, there are two things we can do:
1. if the extra parameter is general and other policy may need it as well, we can change `trace_object()` and change the macro. 
2. if the extra parameter is not general and only specific to a policy (first we need a good reason to explain why it is special), one can implement their own `trace_object()` method, and not use `PlanTraceObject`/`PolicyTraceObject`.

> 2. `#[derive(PlanTraceObject)]` implies that each plan or policy can only be bind to one `trace_object` implementation. For GC like Immix which requires multiple implementations, it has to introduce another generic args `TraceKind`. I'm afraid this is probably not a nice and clean approach. `TraceKind` was introduced _only_ because ImmixSpace has two different `ProcessEdges`, and they share a large amount of common code. But this is not the case for other policies.  For policies with multiple different `ProcessEdges` and `trace_object`, introducing `TraceKind` again in `PlanTraceObject` will only make things more complicated.
>    
>    * In addition, different policies defines different `TraceKind` values, this makes the meaning of `TraceKind` become different at different places. This also makes the compiler (or human) harder to check if all the `TraceKind` value matching and comparison code are written correctly.

Depending on `TraceKind`, `trace_object<const K: TraceKind>()` are actually specialized into different implementations. Both immix and mark compact have different traces. In my opinion, for example, in immix, calling `fast_trace_object()` and `trace_objecct()` has no difference than `trace_object<TRACE_KIND_FAST>()` and `trace_object<TRACE_KIND_DEFRAG>()`.

For your concern about using `u8` for `TraceKind`, it is less ideal. But it can be mitigated that each policy will only check their known trace kind, and panic for unknown traces.

src/plan/transitive_closure.rs
```diff
@@ -71,3 +71,163 @@ impl<'a, E: ProcessEdgesWork> Drop for ObjectsClosure<'a, E> {
...
+pub trait PlanTraceObject<VM: VMBinding> {
+    /// Trace objects in the plan. Generally one needs to figure out
+    /// which space an object resides in, and invokes the corresponding policy
+    /// trace object method.
+    fn trace_object<T: TransitiveClosure, const KIND: TraceKind>(
+        &self,
+        trace: &mut T,
+        object: ObjectReference,
+        worker: &mut GCWorker<VM>,
+    ) -> ObjectReference;
```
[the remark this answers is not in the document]
qinsoon: Yeah. I will add comments to this method.

macros/trace_object/src/derive_impl.rs
```diff
@@ -0,0 +1,125 @@
...
+            if self.#f_ident.in_space(__mmtk_objref) {
+                use crate::policy::gc_work::PolicyTraceObject;
+                <#f_ty as PolicyTraceObject #ty_generics>::post_scan_object(&self.#f_ident, __mmtk_objref);
+                return;
+            }
+        }
+    });
+
+    quote! {
+        #[inline(always)]
```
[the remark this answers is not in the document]
qinsoon: I will make the change.

macros/trace_object/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+                #[inline(always)]
+                #may_move_objects_function
+            }
+        }
+    } else {
+        abort_call_site!("`#[derive(PlanTraceObject)]` only supports structs with named fields.")
+    };
+
+    // Debug the output
+    // println!("{}", output.to_token_stream());
```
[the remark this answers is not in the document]
qinsoon: I would prefer adding the import here, and still keep this as comments. If I remember correctly, `cargo expand` does not work if there is any mistake in the generated code and the code cannot compile. In that case, we need this debug line to print out the generated code.

src/plan/transitive_closure.rs
```diff
@@ -71,3 +71,163 @@ impl<'a, E: ProcessEdgesWork> Drop for ObjectsClosure<'a, E> {
         self.flush();
     }
 }
+
```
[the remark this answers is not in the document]
qinsoon: It makes sense to move `PlanProcessEdges` to `scheduler/gc_work.rs`. What do you suggest about `PlanTraceObject`?

Splitting the `scheduler::gc_work` module should be a separate PR. I would prefer not doing it in this PR.

src/plan/transitive_closure.rs
```diff
@@ -71,3 +71,163 @@ impl<'a, E: ProcessEdgesWork> Drop for ObjectsClosure<'a, E> {
...
+    /// Post-scan objects in the plan. Each object is scanned by `VM::VMScanning::scan_object()`, and this function
+    /// will be called after the `VM::VMScanning::scan_object()` as a hook to invoke possible policy post scan method.
+    /// If a plan does not have any policy that needs post scan, this method can be implemented as empty.
+    /// If a plan has a policy that has some policy specific behaviors for scanning (e.g. mark lines in Immix),
+    /// this method should also invoke those policy specific methods for objects in that space.
+    fn post_scan_object(&self, object: ObjectReference);
+
+    /// Whether objects in this plan may move. If any of the spaces used by the plan may move objects, this should
+    /// return true.
```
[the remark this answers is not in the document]
qinsoon: The comments for the trait include a sentence saying that the methods should be inlined.

src/scheduler/gc_work.rs
qinsoon: The method below this method is called `create_scan_work()`. We could call this `start_or_dispatch_scan_work()`, or simply `dispatch_scan_work()`.

src/policy/gc_work.rs
```diff
@@ -0,0 +1,40 @@
...
+
+/// This trait defines policy-specific behavior for tracing objects.
+/// The procedural macro #[derive(PlanTraceObject)] will generate code
+/// that uses this trait. We expect any policy to implement this trait.
+/// For the sake of performance, the implementation
+/// of this trait should mark methods as `[inline(always)]`.
+pub trait PolicyTraceObject<VM: VMBinding> {
+    /// Trace object in the policy. If the policy copies objects, we should
+    /// expect `copy` to be a `Some` value.
+    fn trace_object<T: TransitiveClosure, const KIND: TraceKind>(
```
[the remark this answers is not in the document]
qinsoon: > `TraceKind` is only for immix. I did not intend to make it general. It was used specifically for immix to reduce code duplication instead of creating two ProcessEdges packets.

Using `TraceKind` can also reduce code duplication for mark compact, which does mark trace and forward trace.

> Also, different policies or plans may use different variations of `trace_object`, with different parameters and behaviours. Unifying the interface and using TraceKind will add a cost for a policy to add a different `trace_object` -- it has to modify this interface and change other policies' code.

I am hoping that this `trace_object()` could work for future policies. But if it does not work, we can expand the `trace_object()` in this trait to support it. Alternatively, if it is really necessary, a policy can choose not to implement this trait -- the only issue is that a plan with that policy cannot use `PlanTraceObject` and `PlanProcessEdges`.

> I suggest either not generalizing trace_object, or adding a way for different ProcessEdges to be able to call different trace_object functions.

A plan can implements `ProcessEdgesWork` themselves.

macros/trace_object/src/derive_impl.rs
```diff
@@ -0,0 +1,125 @@
+use quote::quote;
+use syn::{Field, TypeGenerics};
+use proc_macro2::TokenStream as TokenStream2;
+
+use crate::util;
+
+pub fn generate_trace_object<'a>(
```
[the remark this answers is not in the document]
qinsoon: Fixed.

macros/trace_object/src/derive_impl.rs
```diff
@@ -0,0 +1,125 @@
...
+    ty_generics: &TypeGenerics,
+) -> TokenStream2 {
+    // If any space or the parent may move objects, the plan may move objects
+    let space_handlers = space_fields.iter().map(|f| {
+        let ref f_ty = f.ty;
+
+        quote! {
+            || <#f_ty as PolicyTraceObject #ty_generics>::may_move_objects::<KIND>()
+        }
+    });
```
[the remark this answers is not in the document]
qinsoon: Mostly right. But `Gen` is the fallback for generational plans. As the nursery copy space is copying, we will always move objects. For example, for non-defrag generational immix, we still move objects. This code was intended to deal with that.

src/plan/transitive_closure.rs
```diff
@@ -71,3 +71,163 @@ impl<'a, E: ProcessEdgesWork> Drop for ObjectsClosure<'a, E> {
...
+use crate::plan::Plan;
+use crate::scheduler::gc_work::ProcessEdgesBase;
+use std::ops::{Deref, DerefMut};
+
+/// This provides an implementation of [`ProcessEdgesWork`](scheduler/gc_work/ProcessEdgesWork). A plan that implements
+/// `PlanTraceObject` can use this work packet for tracing objects.
+pub struct PlanProcessEdges<
+    VM: VMBinding,
+    P: 'static + Plan<VM = VM> + PlanTraceObject<VM> + Sync,
+    const KIND: TraceKind,
```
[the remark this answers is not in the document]
qinsoon: A plan can implement its own `ProcessEdgesWork`.

The current `TraceKind` assumes a plan will only use one policy that has special trace kinds. For example, we cannot use mark compact space and immix space in the same plan. I think that is unlikely for now. But if we need to cope with that case, we could use each bit in the `TraceKind` `u8` to represent a trace.

src/plan/transitive_closure.rs
```diff
@@ -71,3 +71,163 @@ impl<'a, E: ProcessEdgesWork> Drop for ObjectsClosure<'a, E> {
...
+        Box::new(PlanScanObjects::<Self, P>::new(self.plan, nodes, false))
+    }
+
+    #[inline(always)]
+    fn trace_object(&mut self, object: ObjectReference) -> ObjectReference {
+        if object.is_null() {
+            return object;
+        }
+        self.plan
+            .trace_object::<Self, KIND>(self, object, self.worker())
```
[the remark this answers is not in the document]
qinsoon: We have to discuss this case by case. Either we generalize that, or we use custom `ProcessEdgesWork`.

src/policy/immix/immixspace.rs
```diff
@@ -113,6 +116,44 @@ impl<VM: VMBinding> Space<VM> for ImmixSpace<VM> {
...
+    #[inline(always)]
+    fn post_scan_object(&self, object: ObjectReference) {
+        if super::MARK_LINE_AT_SCAN_TIME && !super::BLOCK_ONLY {
+            debug_assert!(self.in_space(object));
+            self.mark_lines(object);
+        }
+    }
+
+    #[inline(always)]
+    fn may_move_objects<const KIND: crate::policy::gc_work::TraceKind>() -> bool {
```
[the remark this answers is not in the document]
qinsoon: Will fix this.

src/policy/immix/immixspace.rs
```diff
@@ -113,6 +116,44 @@ impl<VM: VMBinding> Space<VM> for ImmixSpace<VM> {
     }
 }
 
+impl<VM: VMBinding> crate::policy::gc_work::PolicyTraceObject<VM> for ImmixSpace<VM> {
+    #[inline(always)]
+    fn trace_object<T: TransitiveClosure, const KIND: crate::policy::gc_work::TraceKind>(
```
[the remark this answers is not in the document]
qinsoon: Will fix this.

src/plan/immix/global.rs
```diff
@@ -26,8 +26,14 @@ use std::sync::Arc;
 use atomic::Ordering;
 use enum_map::EnumMap;
 
+use mmtk_macro_trace_object::PlanTraceObject;
+
+#[derive(PlanTraceObject)]
 pub struct Immix<VM: VMBinding> {
+    #[post_scan]
```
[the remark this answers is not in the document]
qinsoon: Yeah. They should be inlined. But we need to check which space the object is in, and invoke `post_scan()` on the space. With `#[post_scan]`, we only check for spaces that are explicitly marked with the tag.

src/policy/immix/immixspace.rs
```diff
@@ -113,6 +116,44 @@ impl<VM: VMBinding> Space<VM> for ImmixSpace<VM> {
...
+        }
+    }
+
+    #[inline(always)]
+    fn post_scan_object(&self, object: ObjectReference) {
+        if super::MARK_LINE_AT_SCAN_TIME && !super::BLOCK_ONLY {
+            debug_assert!(self.in_space(object));
+            self.mark_lines(object);
+        }
+    }
```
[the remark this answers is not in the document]
qinsoon: > I feel that this is not generalized enough. This only adds a hook after scanning each object, so immix can mark spanning lines. However, for example LXR needs to do extra stuff when scanning the object.

This hook assumes we always call `VMScanning::scan_object()`, then allows a policy to do some policy-specific stuff. Do you have an example?

> Even for a similar task (e.g. scanning objects or processing edges), I think it's still better to define different work packets for different requirements. You can't enumerate all the special needs for all the policies in a unified interface. This is also the power of the work packets design: If you need to do something different, just define a new packet.

Correct. But this PR addresses a separate issue: If you need to do something *similar*, do not duplicate code.

src/plan/generational/copying/gc_work.rs
```diff
@@ -1,10 +1,20 @@
...
 use crate::vm::*;
 
+use crate::plan::transitive_closure::PlanProcessEdges;
+use crate::policy::gc_work::DEFAULT_TRACE;
+
+pub struct GenCopyNurseryGCWorkContext<VM: VMBinding>(std::marker::PhantomData<VM>);
+impl<VM: VMBinding> crate::scheduler::GCWorkContext for GenCopyNurseryGCWorkContext<VM> {
+    type VM = VM;
+    type PlanType = GenCopy<VM>;
+    type ProcessEdgesWorkType = GenNurseryProcessEdges<Self::VM>;
```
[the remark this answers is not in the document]
qinsoon: `GenNurseryProcessEdges` is manually implemented, it does not use `PlanProcessEdges`/`PlanTraceObject`. I will make it clear that the attributes only affect `PlanTraceObject` and `PlanProcessEdges`.

src/plan/transitive_closure.rs
qinsoon: I have removed this. Thanks.

src/plan/transitive_closure.rs
qinsoon: I have removed this. Thanks.

src/plan/transitive_closure.rs
qinsoon: Do you mean moving this to the top of the file? It is only used in this scope.

macros/trace_object/src/util.rs
qinsoon: Yeah. I have removed the label.

macros/trace_object/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "mmtk-macro-trace-object"
```
wks, line 2: When the next version is published, this crate will be visible on https://crates.io/crates/mmtk-macro-trace-object

We'd better not litter crates.io with too many crates.

macros/trace_object/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
```
wks, line 3: We should give it the same version as mmtk-core's

wks (changes requested):
We should probably have only one macro crate called `mmtk-macros` rather than multiple macro crate for different features, such as `mmtk-macro-trace-object`.

We may have more macros in the future.  Then,
1.  Different macros may share common utilities, such as handling attributes.
2.  The sub crates that contain macros will also appear on crates.io

The second point matters.  If we have multiple crates, then there will be multiple `mmtk-macro-xxx` crates on crates.io.  They will be visible to the world as:
-   https://crates.io/crates/mmtk-macro-trace-object
-   https://crates.io/crates/mmtk-macro-foo
-   https://crates.io/crates/mmtk-macro-bar
-   https://crates.io/crates/mmtk-macro-baz
-   ...

And they need to be versioned, too.  If we add more features, we will introduce even more crates.  Although having multiple crates doesn't seem to violate crates.io's policy (They do, however, disallow "using an automated tool to claim ownership of a large number of package names".), it is much harder to maintain so many crates.

macros/trace_object/src/lib.rs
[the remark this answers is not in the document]
wks: What about guarding those lines with `cfg!(feature="debug_output")`, or using `log::trace!(...)`, or even `if false`?  At least we still get syntax checking.

src/plan/transitive_closure.rs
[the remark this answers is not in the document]
wks: @qinsoon `PlanTraceObject` can be in the same place as `trait Plan` itself, i.e. `src/plan/global.rs`.  It represents something that a `Plan` does, although mainly used by the `PlanProcessEdges` work packet (but it can be used elsewhere as well).  I suppose other methods, such as `get_used_pages`, can also be derived.

macros/trace_object/Cargo.toml
qinsoon: Does this macro crate get published automatically when mmtk-core is published? Do we need to publish this macro crate separately?

wks: I tried `cargo publish --dry-run` in `mmtk-core`.  It shows that I must push `mmtk-macro-trace-object` to crates-io before pushing `mmtk-core`, because `mmtk-core` depends on `mmtk-macro-trace-object`.

However, it is okay to `cargo publish --dry-run` in `mmtk-core/macros/trace_object`, because it has no unresolved dependencies.

There is an open issue about this topic: https://github.com/rust-lang/cargo/issues/1169

qinsoon: I see. I probably need to update our github action for publishing.

macros/trace_object/src/lib.rs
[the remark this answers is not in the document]
qinsoon: Yeah. Sounds good. I have added a constant for debug print.

src/plan/transitive_closure.rs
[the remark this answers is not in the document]
qinsoon: I have moved `PlanTraceObject` to `plan/global.rs`, and moved `PlanProcessEdges` to `scheduler/gc_work.rs`.

src/plan/mod.rs
qinsoon: Done. I have reverted the change.

macros/trace_object/Cargo.toml
qinsoon: Done.

macros/trace_object/Cargo.toml
qinsoon: Done.

src/plan/transitive_closure.rs
qinsoon: As Kunshan suggested, this code is moved to `scheduler::gc_work`. We no longer need the `use` in the scope.

wks (commented):
It looks good to me now.  Please make sure Wenyu's comments are also addressed.

wks (approved):

wenyuzhao (approved):
LGTM

PR #578, opened 2022-04-29
Get finalizable object before object dies

wks:
Add an API to get a reference to an object registered for finalization even before the object actual dies.  This is useful for VMs which calls all finalizers before VM exits, such as Ruby.

Draft: **This has not been tested, yet.**  Currently, the `Finalization` work packet depends on `ProcessEdgesWork`, but Ruby can only process objects at a time, and cannot enqueue edges.  This prevents me from testing it on Ruby.

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -490,13 +490,31 @@ pub fn add_finalizer<VM: VMBinding>(mmtk: &'static MMTK<VM>, object: ObjectRefer
 /// * `mmtk`: A reference to an MMTk instance.
 pub fn get_finalized_object<VM: VMBinding>(mmtk: &'static MMTK<VM>) -> Option<ObjectReference> {
     if *mmtk.options.no_finalizer {
-        warn!("get_object_for_finalization() is called when no_finalizer = true");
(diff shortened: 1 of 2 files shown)


qinsoon:
I am wondering if we should just have the API to get all the objects rather than getting them one by one. When a binding wants to do this kind of 'finalization befere exit', they would almost certainly do it for all the finalizable objects. So it makes sense that we just give them all the objects.

And we probably want to check with @udesou and see if this would work fine for him as well.

wks:
@qinsoon Agreed.  It doesn't make sense to get a candidate first and then get another ready-to-finalize object.  I think getting all candidates and all ready-to-finalize objects at once should be better.

An alternative design is to set a "VM-exiting" flag to the finalization processor.  When set to true, all existing candidates will be dumped to the `ready_to_finalize` vector, and any subsequent calls to `add_finalize` will go straight into `ready_to_finalize`.  The risk could be that one finalizer may create another finalizer and this can go on forever.  That's bad VM design anyway.

qinsoon:
I am closing this PR, as the changes are included in https://github.com/mmtk/mmtk-core/pull/609.

PR #582, opened 2022-04-30
Allow longer timeout for OpenJDK tests

qinsoon:


qinsoon:
It seems jobs on github-hosted runners can only run up to 6 hours. I will have to fix this in the OpenJDK binding. I am closing this PR.

PR #590, opened 2022-05-11
fix thread state transition in harness

tianleq:


qinsoon (changes requested):
I don't think it is necessary to make this change in mmtk-core. The change in https://github.com/mmtk/mmtk-openjdk/pull/157 should be limited to the binding.

tianleq:
> I don't think it is necessary to make this change in mmtk-core. The change in [mmtk/mmtk-openjdk#157](https://github.com/mmtk/mmtk-openjdk/pull/157) should be limited to the binding.
Yes, this is not necessary. I just want to make them consistent.

qinsoon:
> > I don't think it is necessary to make this change in mmtk-core. The change in [mmtk/mmtk-openjdk#157](https://github.com/mmtk/mmtk-openjdk/pull/157) should be limited to the binding.
> > Yes, this is not necessary. I just want to make them consistent.

The change makes those methods inconsistent with other methods in the header file. My opinion is that we provide the 'standard' header file as an example, and each binding can start from this, and customize it to their needs. Exposing `harness_begin()` as `mmtk_harness_begin_impl()` is the choice of the OpenJDK binding, and it does not affect the header file we provide in mmtk-core.

tianleq:
> > > I don't think it is necessary to make this change in mmtk-core. The change in [mmtk/mmtk-openjdk#157](https://github.com/mmtk/mmtk-openjdk/pull/157) should be limited to the binding.
> > > Yes, this is not necessary. I just want to make them consistent.
> 
> The change makes those methods inconsistent with other methods in the header file. My opinion is that we provide the 'standard' header file as an example, and each binding can start from this, and customize it to their needs. Exposing `harness_begin()` as `mmtk_harness_begin_impl()` is the choice of the OpenJDK binding, and it does not affect the header file we provide in mmtk-core.

okay. I see

PR #593, opened 2022-05-16
Allow bindings to handle unknown objects in trace_object()

qinsoon:
This PR adds a new method to `ActivePlan` to allow bindings to handle unknown objects encountered in object tracing.


src/policy/space.rs
```diff
@@ -181,8 +181,9 @@ impl SFT for EmptySpaceSFT {
         _object: ObjectReference,
         _worker: GCWorkerMutRef,
     ) -> ObjectReference {
+        // We do not have the `VM` type parameter here, so we cannot forward the call to the VM.
         panic!(
-            "Call trace_object() on {} (chunk {}), which maps to an empty space",
+            "Call trace_object() on {} (chunk {}), which maps to an empty space. SFTProcessEdges does not support the fallback to vm_trace_object().",
             _object,
```
wks, line 187: This `_object` has the same problem.  Although it is not related to the bindings, it is nice to fix it, too.

wks (changes requested):
The logic is OK.

There is only a small problem related to the naming of used/unused parameters.

qinsoon:
I have removed the underscore prefix for those variables.

wks (approved):
LGTM

PR #597, opened 2022-05-18
Remove ProcessEdgesBase::worker

wks:
The `ProcessEdgesBase::worker` field is a `*mut GCWorker<VM>`.  This is unsafe.

```rust
pub struct ProcessEdgesBase<VM: VMBinding> {
    pub edges: Vec<Address>,
    pub nodes: Vec<ObjectReference>,
    mmtk: &'static MMTK<VM>,
    worker: *mut GCWorker<VM>,  // This is unsafe. Remove it
    pub roots: bool,
} 
```

`ProcessEdgesWork` needs to access the worker instance because

1.  `XxxxxxSpace::trace_object` needs a `GCWorker<VM>`  reference to get the `CopyContext`, and
2.  `flush()` needs a `GCWorker<VM>` to submit or execute `ScanObjects` work packets.

Currently, `ProcessEdgesWork` attempts to give itself access to `GCWorker` by holding a `*mut GCWorker<VM>` in the `ProcessEdgesBase::worker` field.  *This is unsafe*, because the work packet is not associated to a `GCWorker` until it is executed by a `GCWorker`.  The access to the `*mut GCWorker<VM>` pointer is invalid before `do_work` starts, and after `do_work` finishes.

In idiomatic Rust, if the `GCWorker` is only valid to access during the execution of `gc_work`, it should be passed to `gc_work` as a `&mut GCWorker<VM>` reference.  Rust's borrowing semantics will ensure that it is only borrowed during the execution of `gc_work`.

Actually `GCWork::do_work` already has a `worker: &mut GCWorker<VM>` parameter.  So we can pass it through levels of function calls to give them access to `GCWorker`.

This PR attempts to do this.

**DRAFT**: I added `&mut GCWorker` parameter to too many functions, but only a few functions actually use it.  This may indicate that there is still some room for refactoring.  One possibility is to introduce a struct that has a lifetime parameter `'w` and contains a `&'w mut GCWorker`, and use that struct as the `self`, but only during the execution of `do_work`.

### DIFF
--- a/src/plan/generational/gc_work.rs
+++ b/src/plan/generational/gc_work.rs
@@ -1,6 +1,6 @@
 use crate::plan::generational::global::Gen;
 use crate::policy::space::Space;
-use crate::scheduler::gc_work::*;
+use crate::scheduler::{gc_work::*, GCWorker};
 use crate::util::{Address, ObjectReference};
 use crate::vm::*;
 use crate::MMTK;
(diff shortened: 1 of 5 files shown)


wks:
https://github.com/mmtk/mmtk-core/pull/1468 removed `ProcessEdgesWork` and its associated `ProcessEdgesBase`.  The new architecture no longer captures `&mut Worker` when creating the work packet.  Instead, the work packet simply uses the parameter `worker: &mut Worker` of the `GCWork::do_work` method.  This PR is no longer needed.

PR #598, opened 2022-05-19
Decouple root scanning from ProcessEdgesWork.

wks:
This **draft** PR is a proposal for changing the root-scanning interface.

When scanning roots, the VM should only report what roots it has, not deciding what work packets to create.  But currently, `scan_thread_root`, `scan_thread_roots` and `scan_vm_specific_root` take a generic type parameter `<E: ProcessEdgesWork>` which binds them to a `ProcessEdgesWork`.  This may not be general enough, in two ways:

1.  Some VMs cannot present roots as edges.  They can only present roots as nodes because they cannot update the root edges.  One example is the Ruby VM.  It uses conservative stack scanning, and cannot update roots.  It gives the GC a list of **values** on the stack, and some of them are roots (can be filtered using the `is_mmtk_object` function in the core).
2.  Some GC algorithms do root scanning significantly differently than our current `ProcessEdgesWork`.  Although `ProcessEdgesWork` is still a trait, it is too tied to some kinds of GC.  For example, it contains the logic of tracing objects and  scanning objects (via `ScanObjects`).  The real culprit is the `ProcessEdgesWork`.

However, from the code in this PR, we observe that we need to wrap `ProcessEdgesWork` in such a complicated manner just to get a chance to call the plan-specific `trace_object` function.  Probably it is easier to do this refactoring if we eliminate (or weaken) the `ProcessEdgesWork` trait first.

### DIFF
--- a/src/scheduler/gc_work.rs
+++ b/src/scheduler/gc_work.rs
@@ -263,6 +263,60 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for VMProcessWeakRefs<E> {
     }
 }
 
+struct NonMovingProcessEdgesWork<E: ProcessEdgesWork> {
+    delegate: E,
+    nodes: Vec<ObjectReference>,
+    phantom: PhantomData<E>,
+}
+
+impl<E: ProcessEdgesWork> NonMovingProcessEdgesWork<E> {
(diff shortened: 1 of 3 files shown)


wks:
Superseded by https://github.com/mmtk/mmtk-core/pull/611

PR #602, opened 2022-05-31
Immix should declare that it will use forwarding specs in VMObjectModel.

qinsoon:
Immix uses forwarding bits and pointers. However, it does not add them to its side metadata context. So if a binding uses side metadata for those specs, Immix won't initialise the side metadata for those specs.

`CopySpace` properly deals with this:
https://github.com/mmtk/mmtk-core/blob/e0273429f5239ab03407b4a93b52b21f45930d88/src/policy/copyspace.rs#L141-L144

### DIFF
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -166,6 +166,8 @@ impl<VM: VMBinding> ImmixSpace<VM> {
                 MetadataSpec::OnSide(Block::MARK_TABLE),
                 MetadataSpec::OnSide(ChunkMap::ALLOC_TABLE),
                 *VM::VMObjectModel::LOCAL_MARK_BIT_SPEC,
+                *VM::VMObjectModel::LOCAL_FORWARDING_BITS_SPEC,
+                *VM::VMObjectModel::LOCAL_FORWARDING_POINTER_SPEC,
             ]
         } else {
             vec![
@@ -174,6 +176,8 @@ impl<VM: VMBinding> ImmixSpace<VM> {
                 MetadataSpec::OnSide(Block::MARK_TABLE),
                 MetadataSpec::OnSide(ChunkMap::ALLOC_TABLE),
                 *VM::VMObjectModel::LOCAL_MARK_BIT_SPEC,
+                *VM::VMObjectModel::LOCAL_FORWARDING_BITS_SPEC,
+                *VM::VMObjectModel::LOCAL_FORWARDING_POINTER_SPEC,
             ]
         })
     }


wks:
>   TracingDelegate looks quite similar to PlanTraceObject. We could just rename PlanTraceObject to TracingDelegate. Reusing code is also an important part of software engineering.

`TracingDelegate` can also be implemented for SFT.  For example, `SFTTracingDelegate::trace_object` could call `SFT::sft_trace_object`, like `SFTProcessEdges` does.

`TracingDelegate` intends to extract "good parts" from `SFTProcessEdges` and `PlanProcessEdges` into `SFTTracingDelegate` and `PlanTracingDelegate`, and make them reusable.  My plan is to reuse `TracingDelegate` in other work packets as well, such as:

-   **`struct TracingProcessEdges<D: TracingDelegate>`**: Replacing `trait ProcessEdgesWork`.
-   **`struct TracingProcessEdges<D: TracingDelegate>`**: Replacing `struct ScanObjects` and `struct PlanScanObjects`.  It scans object like `ScanObjects` does, but can optionally process its fields (edges), too, if the VM (Ruby) does not support edge enqueuing for some objects.

Neither of them can have subclasses, but plans can customise them by provide different `TracingDelegate` instances.

However, if we plan to commit to `PlanTraceObject` and dismiss SFT, we won't need `TracingDelegate`.  Instead, we just embed a `plan: &'static p` where `P: Plan<VM = VM> + PlanTraceObject<VM>`, and those work packets can call into the plan directly.

>   I don't think we need a special type for reference processor. Reference processor could use any type that we use for tracing (currently ProcessEdgesWork, or any type that will supersede ProcessEdgesWork).

If the "special type" means `SoftRefProcessing` (and its weak/phantom counterparts), then we do need it.  The logic of reference processing is still a bit different from ordinary edge processing.  Edge processing does the following:
1.  Make an object queue `q`
2.  Load objref from the edge
3.  Call `trace_object(&mut q, objref)`
4.  Store the new objref back to the edge (if moved).
5.  Repeat 2-4 until all edges are processed
6.  If `q` is not empty, create an object-scanning work packet with all elements in `q`
7.  Execute the object-scanning work packet, or submit the work packet to the scheduler.

`SoftRefProcessing` is different in the following aspects:
-   Step 2 and 4 are different, because it accesses SoftReference objects via special VM API.
-   If it is weak reference processing, or it decides not to retain soft references, it will only do step 2-4 for reachable Soft/WeakReference.

`TracingDelegate` supports step 3, so the reference processor can still call it.

`TracingDelegate` indirectly supports step 7.  It provides `post_scan_object` which is currently the only difference between different `XxxxxScanObjects` work packets.  And the code for step 7 (creating a work packet) is trivial (just a few lines of code).  If code repetition is a problem, we can still abstract it out in a function.

Here is an in-progress work for `TracingProcessEdges::gc_work` which does what `ProcessEdgesWork::gc_work` does with `TracingDelegate`: https://github.com/wks/mmtk-core/blob/8217c09480451e4ed7b43a8cea3b2aead8e1913e/src/scheduler/gc_work.rs#L872

qinsoon:
We need to make sure that Immix will properly clear forward bits and pointers if they are in the side metadata.

qinsoon:
Closing this PR. https://github.com/mmtk/mmtk-core/pull/753 includes this chagne.

wks:
I didn't include `VMObjectModel::LOCAL_FORWARDING_POINTER_SPEC` in https://github.com/mmtk/mmtk-core/pull/753.  I should include that for consistency even though the overhead may be too large if we put the forwarding pointer on the side.

qinsoon:
> I didn't include `VMObjectModel::LOCAL_FORWARDING_POINTER_SPEC` in #753. I should include that for consistency even though the overhead may be too large if we put the forwarding pointer on the side.

We once discussed this. I think the conclusion was that forwarding pointer should always be in the dead object (destroying the object). As the object is dead, neither the runtime or the application should access a dead object. Thus it is fine for us to just destroy the object and put a forwarding pointer there. Thus forwarding pointer should not be put to the side, and it does not need to be a `MetadataSpec` that the binding can config.

However, with the new requirement in Ruby, this conclusion may not be true any more. Anyway, just bear in mind that we were not sure if we should allow the binding to specify the location of forwarding pointer.

PR #603, opened 2022-05-31
Replace custom_derive with strum

k-sareen:
The custom_derive crate has been inactive since 2016. This commit
replaces its use with the strum crate which is more recent and in active
development.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,14 +21,12 @@ doctest = false
 # MMTk macros
 mmtk-macros = { version = "0.12.0", path = "macros/" }
 
-custom_derive = "0.1"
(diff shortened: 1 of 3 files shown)


k-sareen:
I was just doing a check of what dependencies we use and found that the custom_derive crate [1] had not been in active development since 2016 (so 6 years ago). It's used to convert the MMTK_PLAN and MMTK_NURSERY_ZEROING environment variables from a String to an Enum. I've replaced its use with the more recent and in active development strum crate [2].

[1]: https://github.com/DanielKeep/rust-custom-derive
[2]: https://github.com/Peternator7/strum

Issue #573, opened 2022-04-22
Generalise edges


wks:
TL;DR: An edge may not be just `*mut ObjectReference`.

-   With CompressedOops, it can be `*mut u32`;
-   In Julia, it can be `{pointer: *mut ObjectPointer, offset: usize}`;
-   In Ruby, the VM calls the GC with the field value, and the GC returns the new field value, so an edge is virtual.

We generalise edges into a `EdgeShape` trait with two methods: `load` and `store`.  `mmtk-core` only depends on these two abstract operations, and the VM defines the concrete representation of edges.

# Problem

An **edge** is a **reference field**.  An edge holds a reference to an object (can be null).  Edges can be in an object, on the stack, or in other locations.

Currently, MMTk Core assumes an edge is a memory slot of the size of `ObjectReference` (word sized).  For example, the following snippet is from `mmtk-core`:
```rust
    #[inline]
    fn process_edge(&mut self, slot: Address) {
        let object = unsafe { slot.load::<ObjectReference>() };
        let new_object = self.trace_object(object);
        if Self::OVERWRITE_REFERENCE {
            unsafe { slot.store(new_object) };
        }
    }
```
Here `slot` is just an address. When loading, it assumes the slot holds a word-sized data, and is directly interpreted as an `ObjectReference`.  When storing, the new `ObjectReference` is written into the slot.

This is not general enough.  In VMs, an edge may not be simply a memory slot that holds an `ObjectReference`.

## Compressed oops

OpenJDK supports "compressed oops".  On 64-bit systems, the JVM can limit the heap size to 32GB and require 8-byte alignment so that a pointer to an object only has 32 effective bits (See: https://www.baeldung.com/jvm-compressed-oops).  By doing this, each object reference only occupies 32 bits of space instead of 64 bits.  The VM becomes more space-efficient at the cost of some computation overhead.

-   Loading an object reference from a slot that holds a compressed pointer needs bit operation:  `objref = ((*compressedOop) << 3)`.
-   Storing an object reference into such a slot also needs bit operation: `*compressedOop = objref >> 3` (assuming bits in `objref[63..35]` are all zero).

## Julia: pointer with offset

Sometimes a Julia array may hold a pointer to the middle of its underlying buffer, and record the offset from the beginning of the buffer to the pointer.  This is convenient for the "soft deletion" of elements in the beginning of the array (See https://github.com/JuliaLang/julia/issues/43056).

A simplified version of it is shown below:
```cpp
[...]

PR #606, opened 2022-06-02
Edge trait

wks:
We make the representation of edges VM-specific.  The MMTk core only
depends on the abstract Edge trait.  This allows VM bindings to custom
the behaviour of loading/storing of edges, and support features like
compressed OOPs.

Some example `Edge` implementations are in `vmbindings/dummyvm/src/edges.rs`.  There is a test case `vmbindings/dummyvm/src/tests/edges_test.rs` to show what to expect.

This PR involves API change.  Migration guide is in the commit message.

Binding PRs:
-   OpenJDK: https://github.com/mmtk/mmtk-openjdk/pull/160
-   JikesRVM: https://github.com/mmtk/mmtk-jikesrvm/pull/121
-   V8: https://github.com/mmtk/mmtk-v8/pull/66
-   Ruby: https://github.com/mmtk/mmtk-ruby/pull/9

Closes: https://github.com/mmtk/mmtk-core/issues/573

### DIFF
--- a/src/vm/edge_shape.rs
+++ b/src/vm/edge_shape.rs
@@ -60,6 +60,7 @@ pub trait Edge: Copy + Send + Debug + PartialEq + Eq + Hash {
 /// A simple edge implementation that represents a word-sized slot where an ObjectReference value
 /// is stored as is.  It is the default edge type, and should be suitable for most VMs.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[repr(transparent)]
 pub struct SimpleEdge {
     slot_addr: *mut Atomic<ObjectReference>,
 }
@@ -99,9 +100,16 @@ impl Edge for SimpleEdge {
     }
 }
 
-/// For backword compatibility.
-/// We let Address implement Edge so existing bindings that use `Address` to represent an edge can
-/// continue to work.
+/// For backword compatibility, we let `Address` implement `Edge` so that existing bindings that
+/// use `Address` to represent an edge can continue to work.
+///
+/// However, we should use `SimpleEdge` directly instead of using `Address`.  The purpose of the
+/// `Address` type is to represent an address in memory.  It is not directly related to fields
+/// that hold references to other objects.  Calling `load()` and `store()` on an `Address` does
+/// not indicate how many bytes to load or store, or how to interpret those bytes.  On the other
+/// hand, `SimpleEdge` is all about how to access a field that holds a reference represented
+/// simply as an `ObjectReference`.  The intention and the semantics are clearer with
+/// `SimpleEdge`.
 impl Edge for Address {
     #[inline(always)]
     fn load(&self) -> ObjectReference {
@@ -112,4 +120,12 @@ impl Edge for Address {
     fn store(&self, object: ObjectReference) {
         unsafe { Address::store(*self, object) }
     }
+}
+
+#[test]
+fn a_simple_edge_should_have_the_same_size_as_a_pointer() {
+    assert_eq!(
+        std::mem::size_of::<SimpleEdge>(),
+        std::mem::size_of::<*mut libc::c_void>()
+    );
 }
--- a/src/vm/edge_shape.rs
+++ b/src/vm/edge_shape.rs
@@ -1,0 +1,115 @@
+use std::fmt::Debug;
+use std::hash::Hash;
+
+use atomic::Atomic;
+
+use crate::util::{Address, ObjectReference};
+
+/// An abstract edge.  An edge holds an object reference.  When we load from it, we get an
+/// ObjectReference; we can also store an ObjectReference into it.
+///
+/// This intends to abstract out the differences of reference field representation among different
+/// VMs.  If the VM represent a reference field as a word that holds the pointer to the object, it
+/// can use the default `SimpleEdge` we provide.  In some cases, the VM need to implement its own
+/// `Edge` instances.
+///
+/// For example:
+/// -   The VM uses compressed pointer (Compressed OOP in OpenJDK's terminology), where the heap
+///     size is limited, and a 64-bit pointer is stored in a 32-bit slot.
+/// -   The VM uses tagged pointer, where some bits of a word are used as metadata while the rest
+///     are used as pointer.
+/// -   A field holds a pointer to the middle of an object (an object field, or an array element,
+///     or some arbitrary offset) for some reasons.
+///
+/// When loading, `Edge::load` shall decode its internal representation to a "regular"
+/// `ObjectReference` which is applicable to `ObjectModel::object_start_ref`.  The implementation
+/// can do this with any appropriate operations, usually shifting and masking bits or subtracting
+/// offset from the address.  By doing this conversion, MMTk can implement GC algorithms in a
+/// VM-neutral way, knowing only `ObjectReference`.
+///
+/// When GC moves object, `Edge::store` shall convert the updated `ObjectReference` back to the
+/// edge-specific representation.  Compressed pointers remain compressed; tagged pointers preserve
+/// their tag bits; and offsetted pointers keep their offsets.
+///
+/// The methods of this trait are called on hot paths.  Please ensure they have high performance.
+/// Use inlining when appropriate.
+///
+/// Note: this trait only concerns the representation (i.e. the shape) of the edge, not its
+/// semantics, such as whether it holds strong or weak references.  If a VM holds a weak reference
+/// in a word as a pointer, it can also use `SimpleEdge` for weak reference fields.
+pub trait Edge: Copy + Send + Debug + PartialEq + Eq + Hash {
+    /// Load object reference from the edge.
+    fn load(&self) -> ObjectReference;
+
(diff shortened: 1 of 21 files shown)


src/vm/edge_shape.rs
```diff
@@ -0,0 +1,115 @@
...
+    #[inline(always)]
+    fn store(&self, object: ObjectReference) {
+        unsafe { (*self.slot_addr).store(object, atomic::Ordering::Relaxed) }
+    }
+}
+
+/// For backword compatibility.
+/// We let Address implement Edge so existing bindings that use `Address` to represent an edge can
+/// continue to work.
+impl Edge for Address {
```
qinsoon, line 105: Maybe it is worth documenting that we suggest using `SimpleEdge` instead of `Address`, and why.

vmbindings/dummyvm/src/edges.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn store(&self, object: ObjectReference) {
+        let begin = object.to_address();
+        let middle = begin + self.offset;
+        unsafe { (*self.slot_addr).store(middle, atomic::Ordering::Relaxed) }
+    }
+}
+
+/// This edge presents the object reference itself to mmtk-core.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct ValueEdge {
```
qinsoon, line 133: Will we use a type like this when a binding wants to do node enqueuing? If so, we probably should just call it `Node` or at least in the comment, state this is also known as node.

qinsoon (approved):
LGTM

qinsoon:
I suggest we also get a review from @wenyuzhao on this PR.

src/vm/edge_shape.rs
wks: Yes.   I'll add more comments about why we should move away from using Address directly as an edge.

vmbindings/dummyvm/src/edges.rs
wks: I thought this `ValueEdge` could help node-enqueuing, but I later found it doesn't help.  I already removed all comments that mentioned "edge enqueuing" in this PR, but I want to keep this example, just to demonstrate that the developers can think out of the box when implementing `Edge`.  I'll add some comments to clarify the purpose of this example.

vmbindings/dummyvm/src/edges.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+impl Edge for ValueEdge {
+    fn load(&self) -> ObjectReference {
+        self.value
+    }
+
+    fn store(&self, _object: ObjectReference) {
+        // No-op.  Value edges are immutable.
```
qinsoon, line 155: I don't think we can call this an edge. We call `store()` to update the edge, but this implementation cannot update. The `store()` impl should be at least `unreachable!()` or `panic!()`, as the impl cannot do what it is asked. However, that will cause MMTk crash, as we will call `store`. I think we should just remove this type -- it is confusing.

wks: OK. I think I'll just remove the `ValueEdge` type because it is confusing.

src/vm/edge_shape.rs
```diff
@@ -0,0 +1,115 @@
...
+    #[inline(always)]
+    fn prefetch_store(&self) {
+        // no-op by default
+    }
+}
+
+/// A simple edge implementation that represents a word-sized slot where an ObjectReference value
+/// is stored as is.  It is the default edge type, and should be suitable for most VMs.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct SimpleEdge {
```
wenyuzhao, line 63: `#[repr(transparent)]` so that the size of the struct is equal to the pointer size.

src/vm/mod.rs
```diff
@@ -45,6 +46,9 @@ where
     type VMActivePlan: ActivePlan<Self>;
     type VMReferenceGlue: ReferenceGlue<Self>;
 
+    /// The type of edges in this VM.
+    type VMEdge: edge_shape::Edge;
```
wenyuzhao, line 50: Is it possible to do a runtime switch? For example, OpenJDK may need to support compressed pointers as an optional run-time feature, and cannot be activated at build-time.

wenyuzhao: Also probably this can be moved to VMObjectModel?

src/vm/edge_shape.rs
wks: It doesn't have to.  What matters is its member `slot_addr: *mut Atomic<ObjectReference>,` **points** to a location that has pointer-sized thing.

Edge instances are passed from the VM to mmtk-core in callbacks like `scan_object` and edge-scanning functions.  More often than not, `Edge` instances will *not* have the same size as the pointer.  The `OffsetEdge` type in the [vmbindings/dummyvm/src/edges.rs](https://github.com/mmtk/mmtk-core/pull/606/files/24b7e47fc1647efec38f3d41e9419de619f70d99#diff-e1acc241ac12bd754b4502e45a5b174b3db0360ac50f11fed7f2c79d0e5e4a03) is an example.  An `Edge` instance also holds *information* about an edge, such as offsets, or tag bits, in addition to the address of the field.

src/vm/mod.rs
wks: If the "feature" is a Cargo feature, we can define it like this:
```rust
enum OpenJDKEdge {
    Uncompressed(*mut AtomicUsize),
    #[cfg(features = "compressed_oop")]
    Compressed(*mut AtomicU32),
}
```

But if we want it enabled/disabled by a command-line flag, the easiest way (may affect performance, but need to be confirmed with tests) is simply not use the `OpenJDKEdge::Compressed` case.  As an optimisation, maybe we can parameterise `OpenJDK` to `OpenJDK<EdgeType>` so that depending on the command line options, we can instantiate MMTk with different (parameterised) VMBinding implementation.

qinsoon: Maybe we could check the performance impact of having `enum Edge` (but never using the other case). We do hope that is minimal.

qinsoon: > Also probably this can be moved to VMObjectModel?

I don't think `ObjectModel` is the right place to put it. I think it is fine to put the type in `VMBinding` (as it is now). If we plan to move it, I think `Scanning` fits well. All related changes are in the `Scanning` trait. This is an indication that the type is related with `Scanning`.

wks: > Also probably this can be moved to VMObjectModel?

In theory, we can.  But then instead of `VM::VMEdge`, we will need to write `<<VM as VMBinding>::VMObjectModel as ObjectModel<VM>>::VMEdge` everywhere.

A refactoring may be possible to let VMBinding *require* VMObjectModel, instead of including a VMObjectModel as a member.  I have seen similar problems when I tried to refactor ReferenceProcessor.  I tried to add a function in `VMCollection` that depends on a type in `VMReferenceGlue`, and I have to navigate from VMCollection to VMBinding, and from there to VMReferenceGlue.  Sometimes the compiler is not smart enough to figure out the VM type passed to the function to VMCollection is the same VM as the VM in `VMReferenceGlue<VM>`.

I'll elaborate the refactoring in a different issue.

wenyuzhao: Using `enum OpenJDKEdge` will (1) introduce an extra branch for each edge load and store (2) For a mark-queue `Vec<Edge>`, there will be one word (100%) overhead for each entry. With compressed pointers on, the overhead will be 300%. Probably I'll parameterize `OpenJDK`.

wenyuzhao: > In theory, we can. But then instead of VM::VMEdge, we will need to write <<VM as VMBinding>::VMObjectModel as ObjectModel<VM>>::VMEdge everywhere.

I see. Yes I agree it should stay in VMBinding.

wenyuzhao: I think I don't really need another edge type for compressed pointers. The size of an edge **address** is always one word, no matter whether it's a `*mut AtomicU32` or `*mut AtomicUsize `. The only thing I need to replace is the handler of the load/store operations.

src/vm/edge_shape.rs
wks: Well, I'll just add `#[repr(transparent)]` because some VMs may want to pass field address from C to Rust and want Rust to treat it as an edge.

src/vm/mod.rs
wks: If space overhead is a problem, the OpenJDK binding can implement `Edge` with a tagged pointer.  It saves space, but still needs a test on bits.

Other bindings may not be able to use this trick.  Ruby already uses tagged pointers, and Julia has offsetted edges where the offset has to be recorded.

wks:
I updated the PR.
1.  I replaced the `ValueEdge` example with `TaggedEdge`, because we do have some VMs using tagged pointers.
2.  I added comments about why we should prefer `SimpleEdge` over `Address`.
3.  I added `#[repr(transparent)]` to `SimpleEdge`.

Issue #574, opened 2022-04-27
Only update reference if the object is actually moved?


qinsoon:
In the line in `ProcessEdgesWork`,
https://github.com/mmtk/mmtk-core/blob/0babba20290d3c4e4cdb2a83284aa7204c9a23cc/src/scheduler/gc_work.rs#L449-L452

we check a constant `OVERWRITE_REFERENCE` and write back the reference returned by `trace_object()`. `OVERWRITE_REFERENCE`
 is defined per `ProcessEdgesWork`, which is currently per plan. In a copying plan, `OVERWRITE_REFERENCE` is `true`, and we always update references. 

However, this means there are a few cases where we update references but we do not have to:
* an object is in a non-copying space: e.g. objects in immortal or large object space in a copying plan
* an object is in a copying space but in a particular trace, the policy is not moving objects: e.g. objects in an immix space in a non-defrag GC (non-moving), or objects in the mature immix space in generational immix for a non-defrag GC (nursery is still copying, but objects in mature space is non-moving for that GC).

We may want to check if this introduces any measurable overhead.

There could be different solutions if this brings any overhead:
* Per policy per trace work packet. So `OVERWRITE_REFERENCE` is defined per policy per trace.
* Check if `new_object` is the same as `object` before write. (conditional vs memory write).

qinsoon:
Alternative solutions:
* Let policy's `trace_object()` returns whether they copied object or not (e.g. returns `Option<ObjectReference>`).
* Pass the slot address to policy's `trace_object()`. If the policy copies objects, they are responsible to update the slot.

wks:
One way to "let policy's `trace_object()` returns whether they copied object or not" is mentioned here: https://github.com/mmtk/mmtk-core/issues/559

Example:
```rust
pub struct TraceObjectResult {
    pub is_newly_marked: bool,  // true if the object is newly marked
    pub forwarded: Option<ObjectReference>,  // If an object is forwarded, return Some(newobj), otherwise None.
}

#[inline]
fn process_edge(&mut self, slot: Address) {
    let object = unsafe { slot.load::<ObjectReference>() };
    let TraceObjectResult(newly_marked, forwarded) = self.trace_object(object);
    if newly_marked {
        self.process_node(forwarded.unwrap_or(object));  // enqueue IFF newly marked
    }
    if let Some(new_object) = forwarded {
        unsafe { slot.store(new_object) };  // store IFF forwarded.
    }
}
```

[...]

PR #607, opened 2022-06-06
Replace TransitiveClosure with an ObjectQueue trait

wks:
As we discussed in https://github.com/mmtk/mmtk-core/issues/559, `TransitiveClosure` should be split into two traits, one for object scanning, and the other for enqueuing objects.  This PR is the latter: it replaces `TransitiveClosure` with an object-enqueuing trait `ObjectQueue`, and remove `TransitiveClosure` completely.

With this change, the `trace_object` method now takes an `&mut impl ObjectQueue` parameter.  For `ProcessEdgesWork`, it should pass `&mut self.base().nodes` instead of `self` to `trace_object`.

There is an [alternative solution](https://github.com/mmtk/mmtk-core/compare/master...wks:trace-object-return2) which uses the return value to indicates which object to enqueue.  I think this PR is a more direct approach, because "enqueuing" is part of the semantics of `trace_object`, and it is better to just expose a queue to `trace_object` so it can queue, instead of letting the caller enqueue objects.  Consequently, unlike the [alternative solution](https://github.com/mmtk/mmtk-core/compare/master...wks:trace-object-return2), this PR cannot address the issue of "only updating edges when the object is actually moved" (https://github.com/mmtk/mmtk-core/issues/574).  It shall be addressed in a separate PR.

Closes: https://github.com/mmtk/mmtk-core/issues/559


wks:
I think the code is ready, but I am waiting for the benchmark result.

wks:
I ran Dacapo Chopin on bobcat.moma (Alder Lake), comparing the master and this branch. 20 iterations, 3 invocations each, 4.8x min heap size.

Statistics show that the effect on STW time is negligible. (Outliers (zscore >= 3) are removed.)

Geomean of relative STW time across different benchmarks:
-   SemiSpace: 1.0038804314826324
-   Immix: 0.998644089075811
-   GenCopy: 1.0000574290637219
-   GenImmix: 0.9942469216074474

image

FYI: Here are the scattered points for all executions (with outliers removed), together with box plots and violin plots, just in case you don't trust the means and standard deviations. Note that the plots for some plan-benchmark combinations are missing.  For some combinations (such as SemiSpace-batik), GC did not happen during the execution; For some combinations, (such as luindex and pmd with generational collectors) it crashed. (Running lusearch and pmd with generational collectors will result in crashes for both master and the branch. We should investigate later.)

image

k-sareen:
> For some combinations, (such as luindex and pmd with generational collectors) it crashed. (Running lusearch and pmd with generational collectors will result in crashes for both master and the branch. We should investigate later.)

Yes, I found this earlier when I was doing a performance evaluation of MMTk and brought it up in a meeting. I recall @wenyuzhao mentioning that it might be related to the barriers. I believe he will be updating the barrier code due to his LXR PR, so it might be resolved with it.

src/plan/generational/global.rs
```diff
@@ -178,42 +178,42 @@ impl<VM: VMBinding> Gen<VM> {
     }
 
     /// Trace objects for spaces in generational and common plans for a full heap GC.
-    pub fn trace_object_full_heap<T: TransitiveClosure>(
+    pub fn trace_object_full_heap<Q: ObjectQueue>(
```
wenyuzhao, line 181: Shall we mark this as `#[inline(always)]`? Also the `trace_object_nursery` function below as well

src/plan/global.rs
```diff
@@ -601,33 +601,33 @@ impl<VM: VMBinding> BasePlan<VM> {
         pages
     }
 
-    pub fn trace_object<T: TransitiveClosure>(
+    pub fn trace_object<Q: ObjectQueue>(
```
wenyuzhao, line 604: Not sure if it is helpful to mark this as `#[cold]` or `#[inline(never)]`. This may reduce the inlined code size of the hot trace_object calls.

wenyuzhao (approved):
LGTM. Just two minor questions regarding `#[inline]`.

One thing that may be helpful is configuring [codegen-units](https://doc.rust-lang.org/cargo/reference/profiles.html#codegen-units) in the bindings and setting it to a low value (like `1`) may improve codegen quality. So functions in mmtk-core are more likely to be inlined.

src/plan/generational/global.rs
wks: I tried.  However, if I mark this as `#[inline(always)]`, the Rust compiler will decide not to inline another call inside this function.  Then I marked another function as `#[inline(always)]`, and then the Rust compiler will again decide not to inline yet another function.  And this went on and on.  So I decided to fix the inlining problem later, not in this PR, because the performance of generational GC doesn't seem to be affected by this PR.

src/plan/global.rs
wks: Yes.  It may reduce the inline code size.  But I am not sure if we should mark it as `#[cold]` or `#[inline(never)]`.

1.  The VM space (if exists) may not be cold.
2.  If `in_space` is cheap and simple, inlining it should help skipping objects that are not in those spaces.

PR #608, opened 2022-06-07
Malloc API

qinsoon:
This PR adds a set of malloc related API, and another set of malloc API that will count allocation size into MMTk heap under the feature `malloc_counted_size`. This PR adds things described in https://github.com/mmtk/mmtk-core/pull/572#issuecomment-1125613524.

The changes in this PR:
* tidy up the malloc library import
* provide a set of standard malloc functions (`malloc`, `calloc`, `realloc` and `free`)
* provide a set of malloc functions that will count the malloc size into MMTk heap under the feature `malloc_counted_size`
* provide `gc_poll()` that can be used to check GC when a binding uses counted malloc functions. Small refactoring to allow `poll()` without a space reference.


vmbindings/dummyvm/src/tests/fixtures/mod.rs
```diff
@@ -14,15 +17,27 @@ pub trait FixtureContent {
...
+            serial_lock: None,
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn new_serial_tests() -> Self {
+        Self {
+            content: AtomicRefCell::new(None),
+            once: Once::new(),
+            serial_lock: Some(Mutex::default()),
```
wks, line 40: If a fixture is serial, then neither `AtomicRefCell` nor `Once` are necessary. It is easier to create a dedicated `SerialFixture` type, and wrap both `Option<Box<T>>` and an `initialized` flag into a big `Mutex`.

wks (commented):
Mostly OK.  I only have one question and one suggestion.

wks (approved):
LGTM

Issue #479, opened 2021-10-19
Make the unit of work packet stats consistent


caizixian:
Currently, the work packet stats report times in nanoseconds while the phase-based stats report times in milliseconds. This is confusing and causes issues when trying to calculate derived values in plotty.

cc @steveblackburn @elise-palethorpe

PR #613, opened 2022-06-22
Fix updating of work packet statistics

k-sareen:
Work packets can have different `TypeID`s but the same `work_name()`
since work names strip type parameters whereas `TypeID`s don't. Hence,
work packet statistics used to overwrite previously existing values
leading to incorrect values of the `work.*.{count,max,min,total}`
fields.

This commit fixes this bug by correctly updating the values. This commit
also closes https://github.com/mmtk/mmtk-core/issues/479 by converting
the work packet times to milliseconds.

### DIFF
--- a/src/scheduler/stat.rs
+++ b/src/scheduler/stat.rs
@@ -45,15 +45,20 @@ impl SchedulerStat {
     /// Used during statistics printing at [`crate::memory_manager::harness_end`]
     pub fn harness_stat(&self) -> HashMap<String, String> {
         let mut stat = HashMap::new();
+        let mut counts = HashMap::<String, usize>::new();
+        let mut times = HashMap::<String, f64>::new();
         // Work counts
         let mut total_count = 0;
         for (t, c) in &self.work_counts {
             total_count += c;
             let n = self.work_id_name_map[t];
-            stat.insert(
-                format!("work.{}.count", self.work_name(n)),
-                format!("{}", c),
-            );
+            // We can have the same work names for different TypeIDs since work names strip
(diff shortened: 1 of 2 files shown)


k-sareen:
Should I close #479 in this PR as well? I should just need to divide by `10e6`

qinsoon:
> Should I close #479 in this PR as well? I should just need to divide by `10e6`

Yeah, sure. Just make sure you try it out with plotty after your change.

k-sareen:
Some work packets execute faster than microseconds so they get truncated to 0.000ms. I think this is fine, but if not then let me know: http://squirrel.anu.edu.au/plotty/kunals/bench-test/p/qXJueZ

qinsoon:
LGTM. Thanks for the fix. As Kunal raised [here](https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/Work.20packet.20stats/near/287012781), the total work time does not match the stop-the-world time (even after considering the number of GC threads). This issue may not be related with this PR. But I would suggest we merge this PR after figuring out the issue.

k-sareen:
Apologies -- I can't do exponent notation :sweat_smile: Meant to do `1e6` for 10^6 but did `10e6` instead so values were off by a factor of 10. So single threaded GC numbers make sense now. Multi-threaded GC numbers still have `n_gc_threads * time.stw` >> `total.work.time.total`, but I imagine part of it is due to scheduling overheads and maybe not all threads evenly participate in GC work?

qinsoon (approved):
LGTM

Issue #610, opened 2022-06-17
Performance regression on mutator time for semispace in PR#555


qinsoon:
@wenyuzhao reported performance regression on semispace between 519e8f3c(20211206) and cd6d8984(20220322). The most noticeable issue is that the mutator time for `lusearch` is slowed down by 60-70%. The regression is only measurable with many mutator threads.
> 8 cores: http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/xDFxUB Mar-2022 build is 7% faster on time, 3% slower on time.other
16 cores: http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/CsGQdJ Mar-2022 build is 3% faster on time, 8% slower on time.other
32 cores: http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/jCt85v Mar-2022 build is 16% slower on time, 77% slower on time.other

The complete discussion is here: https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/Performance.20regression

After some measurement, the problematic commit is this one: https://github.com/mmtk/mmtk-core/pull/555, which added a lock in the allocation slowpath. That is possibly the issue. I will further look into it.
![ss-regression](https://user-images.githubusercontent.com/1664709/174212125-e9c4d1af-037c-4ac8-96b3-4c9a98a7fcbc.png)

PR #614, opened 2022-06-24
Minimize the scope of acquire_lock

qinsoon:
This PR changes the scope of `Space.acquire_lock`. The lock was introduced in https://github.com/mmtk/mmtk-core/pull/555, and it caused bad allocation performance when we have many threads (e.g. more than 24 threads, mutator threads or GC threads with copying allocators). This pull request reduces the scope of this `acquire_lock`, and the overhead from it is mostly neglectable. 

This closes https://github.com/mmtk/mmtk-core/issues/610.

Mutator performance after this pull request
ss-regression-mutator


wks (approved):
LGTM

PR #619, opened 2022-06-30
Notify workers when finishing all coord work

wks:
The coordinator thread did not notify GC workers when it finishes
executing all coordinator work packets.  That caused deadlock when all
GC workers are parked before the coordinator finishes executing all
coordinator work packets.

We let the coordinator thread notify one GC worker after draining all
pending coordinator work packets.  One GC worker is enough to identify
that the condition to open more buckets is met.

### DIFF
--- a/src/scheduler/controller.rs
+++ b/src/scheduler/controller.rs
@@ -74,15 +74,20 @@ impl<VM: VMBinding> GCController<VM> {
         match message {
             CoordinatorMessage::Work(mut work) => {
                 work.do_work_with_stat(worker, mmtk);
-                self.scheduler
-                    .pending_messages
+                let old_count = self
+                    .scheduler
+                    .pending_coordinator_packets
                     .fetch_sub(1, Ordering::SeqCst);
+                if old_count == 1 {
+                    // When the coordinator finishes executing all coordinator work packets,
+                    // it is a chance to open more work buckets.
+                    // Notify one worker so it can open buckets.
+                    let _guard = self.scheduler.worker_monitor.0.lock().unwrap();
+                    self.scheduler.worker_monitor.1.notify_one();
+                }
                 false
             }
             CoordinatorMessage::Finish => {
-                self.scheduler
-                    .pending_messages
-                    .fetch_sub(1, Ordering::SeqCst);
                 // Quit only if all the buckets are empty.
                 // For concurrent GCs, the coordinator thread may receive this message when
                 // some buckets are still not empty. Under such case, the coordinator
@@ -109,9 +114,6 @@ impl<VM: VMBinding> GCController<VM> {
         debug_assert!(!self.scheduler.worker_group.has_designated_work());
         // Sometimes multiple finish messages will be sent. Skip them.
         for message in self.receiver.try_iter() {
-            self.scheduler
-                .pending_messages
-                .fetch_sub(1, Ordering::SeqCst);
             match message {
                 CoordinatorMessage::Work(_) => unreachable!(),
                 CoordinatorMessage::Finish => {}
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -43,7 +43,7 @@ pub struct GCWorkScheduler<VM: VMBinding> {
     /// more ephemeron objects.
     closure_end: Mutex<Option<Box<dyn Send + Fn() -> bool>>>,
     /// Counter for pending coordinator messages.
-    pub(super) pending_messages: AtomicUsize,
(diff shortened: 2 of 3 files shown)


src/scheduler/controller.rs
```diff
@@ -74,15 +74,20 @@ impl<VM: VMBinding> GCController<VM> {
...
+                let old_count = self
+                    .scheduler
+                    .pending_coordinator_packets
                     .fetch_sub(1, Ordering::SeqCst);
+                if old_count == 1 {
+                    // When the coordinator finishes executing all coordinator work packets,
+                    // it is a chance to open more work buckets.
+                    // Notify one worker so it can open buckets.
+                    let _guard = self.scheduler.worker_monitor.0.lock().unwrap();
+                    self.scheduler.worker_monitor.1.notify_one();
```
wenyuzhao, line 86: What confuses me is that although there is a queue for all the pending messages, currently we only send one work packet to the coordinator to stop all the mutators. For this stop mutators packet, it will call `notify_mutators_paused` to wake up all the workers [1]. So in theory there should be no deadlock issue?

Waking up workers here is a more general solution. Does that mean we can remove the wake-up code here [1]?

[1] https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L445

src/scheduler/scheduler.rs
```diff
@@ -386,31 +391,38 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             if all_parked {
                 // If there're any designated work, resume the workers and process them
                 if self.worker_group.has_designated_work() {
-                    self.worker_group.dec_parked_workers();
+                    assert!(
```
qinsoon, line 394: Is this assertion correct? Between the check `poll_schedulable_work()` and this line, there could be work pushed to its `designated_work`.

qinsoon (commented):
Looks good to me. But I am not quite familiar with the code, and not familiar with the issue. You probably will need an approval from @wenyuzhao on this.

src/scheduler/controller.rs
wks: `notify_mutators_paused` will wake up all workers **so that workers can scan mutator stacks**.  However, after scanning all stacks, the GC workers will sleep again.  Once all root-scanning work packets are done, and all GC workers are asleep, there will be nothing to wake up the worker **again**.

The lines I added here are for waking up the workers *the second time*, i.e. after all roots are scanned and all workers are sleeping because there are pending coordinator work packets.

This and `notify_mutator_paused` are independent, and cannot replace each other.  The code [1] represents one necessary condition (i.e. all mutators have stopped) to open "the first STW bucket" (i.e. Prepare).  The code I added fixes a problem that makes it fail to open the *second* STW bucket (i.e. Closure).

wenyuzhao: I see

wenyuzhao (approved):

src/scheduler/scheduler.rs
wks: Currently, yes; but in general, maybe no.

Currently, only ordinary (non-coordinate) work packets can add designated work.  Because the entire `poll_slow` function is protected by a lock, if a GC worker reaches this point, all other GC workers must have already been blocked at the `wait(guard)` call below in this function (therefore cannot push any designated work), and the current GC worker must have checked if itself has any pending designated work at the `self.poll_schedulable_work` call above in this function.  So the current GC worker must not have any designated work at this point.

If we add another coordinator work packet (in addition to `StopMutators` which is the only work packet executed on the coordinator at this moment) in the future that can send designated work to GC workers, we will need to remove this assertion, and probably update the control flow here, too.

qinsoon: Does it mean we can push work to designated work from a normal worker (and normal work packets), but we cannot push work to designated work from the coordinator (and coordinator work packets)? If this is the case, it sounds fine -- we just need to make it clear that a coordinator work packet should not push work to `designated_work`.

wks: That's right.  If we push work to designated work from coordinator, this assertion will fail. But a coordinator work packet has no problem adding more coordinator work packets, because it just pushes to the coordinator's message queue.

qinsoon: Can you add some comments for this? For example, in `designated_work`, make it clear that we should only push to work to it from normal workers. For this assertion, make it clear that the assertion failure may be due to pushing work to `designated_work` from the coordinator.

wks: Yes.  I updated the comment for `designated_work` and mentioned that designated work cannot be added from the GC controller thread.

Issue #623, opened 2022-07-13
Allow the plan to be set as a command-line flag


chrisseaton:
We'd like to be able to set the plan via a flag on the command line.

https://github.com/mmtk/mmtk-ruby/issues/1

I can see the comment about why this is not possible at the moment. How deep does this design assumption go? Should we alleviate it sooner rather than later?

```
// The plan to use. This needs to be initialized before creating an MMTk instance (currently by setting env vars)
```

Slightly higher priority for me is at least having a more sensible default plan - either a different default hard-coded into MMTk itself, or being able to set a different default in our VM (which would then mean we can also implement the flag.)

We'd like people to be able to just run our VM as normal, without having to set an environment variable.

Or, as a workaround is it safe to set the environment variable ourselves, within the VM, before we first use MMTk? Is that sensible?

CC @qinsoon

qinsoon:
> Or, as a workaround is it safe to set the environment variable ourselves, within the VM, before we first use MMTk? Is that sensible?

Yeah, that is possible. However, you would have to set the environment variables before creating a `MMTK` instance (not before first using MMTk). That may not be easy unless you create the MMTk instance lazily.

Ideally the issue will go away with https://github.com/mmtk/mmtk-core/issues/541. But that would be a much significant refactoring. I will think about if there is a workaround.

qinsoon:
I did some refactoring in https://github.com/mmtk/mmtk-core/pull/625. Now plan (and all the other options) can be set from command line. Setting options through env vars still works though.

Issue #541, opened 2022-02-07
Refactoring MMTK instantiation


wks:
# Problem

Currently, the `MMTK` instance is usually created statically.

The practice of using static singletons is https://github.com/mmtk/mmtk-core/issues/58brought from Java JikesRVM, but this is not idiomatic in Rust.  Such practice forced two-phase initialisation and forced bad and unsafe practices, including:

-   [X] Using `Option<T>` for late-initialised fields, and force casting `&` to `&mut` to initialise shared data.
    -   This has been addressed in issue https://github.com/mmtk/mmtk-core/issues/532, and fixed in https://github.com/mmtk/mmtk-core/pull/539
-   [X] `MMTK::options` is an `UnsafeOptionsWrapper`, allowing unsafe update to the shared "supposedly-immutable" `MMTK` singleton.
    -   Fixed in https://github.com/mmtk/mmtk-core/pull/625 by introducing a mutable `MMTKBuilder` that creates an `MMTK` instance with immutable `options` field.
-   [X] There are objects that need to be created/initialised before the MMTK instance is created, such as `SFTMap`, `VMMap` and `Mmapper`.  Making MMTK a static singleton makes such dependencies hard to manage.
    -   Now those objects are still global singletons implemented as `static` of `lazy_static!`, but `MMTK` is created after them.  Therefore the dependencies are clearer.

Related issues:

*   [x] https://github.com/mmtk/mmtk-core/issues/57
*   [X] https://github.com/mmtk/mmtk-core/issues/58
*   [ ] https://github.com/mmtk/mmtk-core/issues/100

# Proposal

Introduce a `MMTKBuilder` type that builds the `MMTK` instance.

Following the convention of some builder types in the Rust standard library, such as `std::thread::Builder` and `std::fs:DirBuilder`, the `MMTKBuilder` should be used like the following:

```rust
let mmtk = MMTKBuilder::new()
    .heap_size(256 * 1024 * 1024)
    .threads(8)
    .create();
```

The `MMTKBuilder` should contain mutable fields so that the user can specify options gradually.  But once the `MMTK` instance is created, the options will be read-only.  This means `UnsafeOptionWrapper` will no longer be necessary.

By the time `MMTKBuilder.create()` is called, the options should contain all relevant information to create and initialise plans.  Because of this, the `gc_init` methods of many types (such as `Plan`, `Space`, etc.) may no longer be necessary, and can be replaced by a `new` method.

# Q/A

[...]

PR #625, opened 2022-07-15
Add MMTKBuilder and allow all options to be set from command line

qinsoon:
This PR adds `MMTKBuilder`, and changes APIs around initializing MMTk. The change allows all the options to be set from command line, including heap size, plan, GC threads, etc.

This PR closes https://github.com/mmtk/mmtk-core/issues/623. This PR is also an initial step for https://github.com/mmtk/mmtk-core/issues/541. For https://github.com/mmtk/mmtk-core/issues/541, we may need further changes.

* Add `MMTKBuilder`
* `gc_init()` now takes a reference to `MMTKBuilder` as its argument, and returns a boxed pointer to the new MMTk instance created.
* `gc_init()` is renamed to `mmtk_init()`.
* `gc_init()` no longer requires a `heap_size` argument. `heap_size` now is an option.
* allow options for `plan`, `threads`, `vm_space_size` to be set by command line
* move `examples/mmtk.h` to `docs/header/mmtk.h`. Building C code in `examples` no longer uses this header file.

Related PRs:
* https://github.com/mmtk/mmtk-openjdk/pull/166
* https://github.com/mmtk/openjdk/pull/12
* https://github.com/mmtk/mmtk-jikesrvm/pull/118
* https://github.com/mmtk/mmtk-v8/pull/64

### DIFF
--- a/examples/allocation_benchmark.c
+++ b/examples/allocation_benchmark.c
@@ -5,8 +5,7 @@
 
 int main() {
     volatile uint64_t * tmp;
-    mmtk_set_heap_size(1024*1024*1024);
-    mmtk_gc_init();
+    mmtk_init24*1024*1024);
     MMTk_Mutator handle = mmtk_bind_mutator(0);
 
     for (int i=0; i<1024*1024*100; i++) {
--- a/examples/allocation_benchmark.c
+++ b/examples/allocation_benchmark.c
@@ -5,7 +5,7 @@
 
 int main() {
     volatile uint64_t * tmp;
-    mmtk_init24*1024*1024);
+    mmtk_init(24*1024*1024);
     MMTk_Mutator handle = mmtk_bind_mutator(0);
 
     for (int i=0; i<1024*1024*100; i++) {
--- a/examples/allocation_benchmark.c
+++ b/examples/allocation_benchmark.c
@@ -5,7 +5,7 @@
 
 int main() {
     volatile uint64_t * tmp;
-    mmtk_init(24*1024*1024);
+    mmtk_init(1024*1024*1024);
     MMTk_Mutator handle = mmtk_bind_mutator(0);
 
     for (int i=0; i<1024*1024*100; i++) {
--- a/examples/allocation_benchmark.c
+++ b/examples/allocation_benchmark.c
@@ -5,7 +5,8 @@
 
 int main() {
     volatile uint64_t * tmp;
-    mmtk_gc_init(1024*1024*1024);
+    mmtk_set_heap_size(1024*1024*1024);
+    mmtk_gc_init();
     MMTk_Mutator handle = mmtk_bind_mutator(0);
 
     for (int i=0; i<1024*1024*100; i++) {
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -27,8 +27,19 @@ use crate::vm::VMBinding;
 use std::sync::atomic::Ordering;
 
 /// Initialize an MMTk instance. A VM should call this method after creating an [MMTK](../mmtk/struct.MMTK.html)
-/// instance but before using any of the methods provided in MMTk. This method will attempt to initialize a
-/// logger. If the VM would like to use its own logger, it should initialize the logger before calling this method.
+/// instance but before using any of the methods provided in MMTk (except `process()` and `process_bulk()`).
+///
+/// We expect a binding to ininitialize MMTk in the following steps:
+///
+/// 1. Create an [MMTK](../mmtk/struct.MMTK.html) instance. Usually this is done statically, i.e. a binding holds one singleton as a static variable for MMTk.
+/// 2. Set command line options for MMTk by [process()](./fn.process.html) or [process_bulk()](./fn.process_bulk.html). At this point,
+///    other APIs should not be used, as MMTk is not yet initialized.
+/// 3. Initialize MMTk by calling this function, `gc_init()`. After this call, most MMTk APIs are avialable for use. But MMTk no longer expects
+///    new command line options. If a binding calls `process()` after this point, it may have no effect.
+/// 4. Enable garbage collection in MMTk by [enable_collection()](./fn.enable_collection.html). A binding should only call this once its
+///    thread system is ready. MMTk will not trigger garbage collection before this call.
+///
+/// Not that This method will attempt to initialize a logger. If the VM would like to use its own logger, it should initialize the logger before calling this method.
 /// Note that, to allow MMTk to do GC properly, `initialize_collection()` needs to be called after this call when
 /// the VM's thread system is ready to spawn GC workers.
 ///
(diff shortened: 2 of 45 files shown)


qinsoon:
binding-refs
OPENJDK_BINDING_REF=mmtk-builder
JIKESRVM_BINDING_REF=mmtk-builder
V8_BINDING_REF=mmtk-builder

examples/allocation_benchmark.c
```diff
@@ -5,7 +5,8 @@
 
 int main() {
     volatile uint64_t * tmp;
-    mmtk_gc_init(1024*1024*1024);
+    mmtk_set_heap_size(1024*1024*1024);
+    mmtk_gc_init();
```
wks, line 9: The `mmtk_set_heap_size` in the example may give the reader a false impression that the heap size can be changed during execution, which is not true.  Without a concrete implementation as in dummyvm, the reader may not realise that it sets an option in the builder before creating an MMTk instance.

For the purpose of example, I think it is OK to keep the old code `mmtk_gc_init(1024 * 1024 * 1024);` because the C API doesn't have to correspond one-to-one with what `memory_manager` provides.  Conceptually, a C function `mmtk_gc_init` could both set the option and create the instance.  We can also change the name to `mmtk_init` or `init_mmtk` to avoid confusion.

wks (changes requested):
Having `MMTKBuilder` to enable safe instantiation of `MMTK` is a significant improvement.

The main problem of this PR, I think, is the handling of Options.  We previously used `UnsafeCell` because we update the options *after* `MMTK` is instantiated.  With the builder introduced, it *is* safe to update options *before* instantiating `MMTK`.  It is no longer necessary to use `UnsafeCell`.

vmbindings/dummyvm/src/lib.rs
```diff
@@ -26,14 +27,24 @@ impl VMBinding for DummyVM {
...
+/// This is used to ensure we initialize MMTk at a specified timing.
+pub static MMTK_INITIALIZED: AtomicBool = AtomicBool::new(false);
+
 lazy_static! {
-    pub static ref SINGLETON: MMTK<DummyVM> = MMTK::new();
+    pub static ref BUILDER: MMTKBuilder = MMTKBuilder::new();
+    pub static ref SINGLETON: MMTK<DummyVM> = {
+        debug_assert!(!MMTK_INITIALIZED.load(Ordering::Relaxed));
+        let ret = mmtk::memory_manager::gc_init(&BUILDER);
+        MMTK_INITIALIZED.store(true, std::sync::atomic::Ordering::Relaxed);
```
wks, line 47: If this is only for assertion purpose, `Relaxed` should be fine.

But if we use it for actual inter-thread synchronisation (like in the OpenJDK binding), this store should have at least `Release` order, and loads of `MMTK_INITIALIZED` should be at least `Acquired`.  Or we can use `SeqCst` for both load and store for even stronger order.

I suggest we either add a comment here and mention `Relaxed` is not enough for serious inter-thread synchronisation.  Alternatively, we can use `Acquire` and `Release` (or `SeqCst` for both) here just in case someone would just copy this part of the code into their binding.

src/mmtk.rs
```diff
@@ -39,28 +39,77 @@ use crate::util::rust_util::InitializeOnce;
 // A global space function table that allows efficient dispatch space specific code for addresses in our heap.
 pub static SFT_MAP: InitializeOnce<SFTMap<'static>> = InitializeOnce::new();
 
+// MMTk builder. This is used to set options before actually creating an MMTk instance.
+pub struct MMTKBuilder {
+    /// The options for this instance.
+    options: Arc<UnsafeOptionsWrapper>,
```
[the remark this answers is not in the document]
qinsoon: I removed `UnsafeOptionsWrapper`.

src/memory_manager.rs
```diff
@@ -324,8 +341,8 @@ pub fn disable_collection<VM: VMBinding>(mmtk: &'static MMTK<VM>) {
 /// * `mmtk`: A reference to an MMTk instance.
 /// * `name`: The name of the option.
 /// * `value`: The value of the option (as a string).
-pub fn process<VM: VMBinding>(mmtk: &'static MMTK<VM>, name: &str, value: &str) -> bool {
-    unsafe { mmtk.options.process(name, value) }
+pub fn process(builder: &MMTKBuilder, name: &str, value: &str) -> bool {
+    unsafe { builder.set_option(name, value) }
```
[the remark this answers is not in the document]
qinsoon: I have made this change as well. I am using `Mutex<MMTKBuilder>`. As we still use lazy static variables for `MMTK` and we need the builder to initialize `MMTK`, we cannot have a non-static `MMTKBuilder`. We may need to find a better pattern for this.

examples/allocation_benchmark.c
qinsoon: Done. I changed it to `mmtk_init(heap_size)` (`gc_init()` is renamed to `mmtk_init()` -- I think it is a good idea, as it avoids the confusion with `initialize_collection()`). As for the C API, in this PR, I moved `examples/mmtk.h` to `docs/header/mmtk.h`. This is meant to be the canonical header, and the C signatures in the header tries to be one-to-one mapped from the Rust API. The DummyVM binding uses a header in `vmbindings/dummyvm/api/mmtk.h`, which is different from the canonical version.

vmbindings/dummyvm/src/api.rs
```diff
@@ -2,24 +2,27 @@
...
-    let singleton_mut = unsafe { &mut *(&*SINGLETON as *const MMTK<DummyVM> as *mut MMTK<DummyVM>) };
-    memory_manager::gc_init(singleton_mut, heap_size)
+pub extern "C" fn mmtk_gc_init() {
+    assert!(!crate::MMTK_INITIALIZED.load(Ordering::Relaxed));
+    lazy_static::initialize(&SINGLETON);
+}
+
+#[no_mangle]
+pub extern "C" fn mmtk_set_heap_size(size: usize) {
+    memory_manager::process(&BUILDER, "heap_size", size.to_string().as_str());
```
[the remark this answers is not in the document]
qinsoon: Done. I added a `set()` method to `MMTKOption`.

vmbindings/dummyvm/src/lib.rs
qinsoon: I changed the ordering to `SeqCst`.

examples/allocation_benchmark.c
```diff
@@ -5,7 +5,7 @@
 
 int main() {
     volatile uint64_t * tmp;
-    mmtk_gc_init(1024*1024*1024);
+    mmtk_init24*1024*1024);
```
wks, line 8: This looks like a typo

vmbindings/dummyvm/src/api.rs
```diff
@@ -2,24 +2,30 @@
...
-    // # Safety
-    // Casting `SINGLETON` as mutable is safe because `gc_init` will only be executed once by a single thread during startup.
-    #[allow(clippy::cast_ref_to_mut)]
-    let singleton_mut = unsafe { &mut *(&*SINGLETON as *const MMTK<DummyVM> as *mut MMTK<DummyVM>) };
-    memory_manager::gc_init(singleton_mut, heap_size)
+pub extern "C" fn mmtk_init(heap_size: usize) {
+    // set heap size first
+    {
+        let mut builder = BUILDER.lock().unwrap();
+        assert!(builder.options.heap_size.set(heap_size));
```
wks, line 22: Although Rust always executes code in `assert!`, it is generally not considered a good practice to put code with important side effect inside assertions.

wks (changes requested):
Mostly good. Only some small problems.

examples/allocation_benchmark.c
```diff
@@ -5,7 +5,7 @@
 
 int main() {
     volatile uint64_t * tmp;
-    mmtk_gc_init(1024*1024*1024);
+    mmtk_init(24*1024*1024);
```
wks, line 8: This is 24 (twenty four) megabytes, but the loop below attempted to allocate 800 megabytes of objects.  Is this intentional? Did you mean `1024*1024*1024`?

wks (approved):
LGTM

PR #630, opened 2022-07-26
Introduce Bounded and Fixed nurseries

k-sareen:
Introduce Bounded and Fixed nurseries

This commit adds Bounded and Fixed nursery types and changes how the
nursery size is set. A Bounded nursery has a lower bound of 2 MB but a
variable upper bound (set to be 1 TB on 64-bit by default), whereas a
Fixed nursery controls both the upper and lower bound of the nursery and
sets them to be the same value. By default, MMTk uses a Bounded nursery.
The nursery size and type can be set via command line arguments or
environment variables, for example, setting MMTK_NURSERY="Fixed:8192"
will create a Fixed nursery of size 8192 bytes.

This commit also changes how minor and major GCs are triggered to be
more in line with the Java MMTk.

**Note**: VM bindings may want to change the
`ObjectModel::VM_WORST_CASE_COPY_EXPANSION` constant depending on the
worst case expansion that can occur due to object sizes changing when
copied.

### DIFF
--- a/src/plan/generational/global.rs
+++ b/src/plan/generational/global.rs
@@ -14,12 +14,13 @@ use crate::util::heap::VMRequest;
 use crate::util::metadata::side_metadata::SideMetadataSanity;
 use crate::util::metadata::side_metadata::SideMetadataSpec;
 use crate::util::options::Options;
+use crate::util::statistics::counter::EventCounter;
 use crate::util::ObjectReference;
 use crate::util::VMWorkerThread;
-use crate::vm::VMBinding;
+use crate::vm::{ObjectModel, VMBinding};
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use mmtk_macros::PlanTraceObject;
 
@@ -37,6 +38,7 @@ pub struct Gen<VM: VMBinding> {
     pub gc_full_heap: AtomicBool,
     /// Is next GC full heap?
     pub next_gc_full_heap: AtomicBool,
+    pub full_heap_gc_count: Arc<Mutex<EventCounter>>,
 }
 
 impl<VM: VMBinding> Gen<VM> {
@@ -48,27 +50,33 @@ impl<VM: VMBinding> Gen<VM> {
         mmapper: &'static Mmapper,
         options: Arc<Options>,
     ) -> Self {
+        let nursery = CopySpace::new(
+            "nursery",
+            false,
+            true,
+            VMRequest::fixed_extent(options.get_max_nursery(), false),
+            global_metadata_specs.clone(),
+            vm_map,
+            mmapper,
+            &mut heap,
+        );
+        let common = CommonPlan::new(
+            vm_map,
+            mmapper,
+            options,
+            heap,
+            constraints,
+            global_metadata_specs,
+        );
+
+        let full_heap_gc_count = common.base.stats.new_event_counter("majorGC", true, true);
+
         Gen {
-            nursery: CopySpace::new(
-                "nursery",
-                false,
-                true,
-                VMRequest::fixed_extent(crate::util::options::NURSERY_SIZE, false),
-                global_metadata_specs.clone(),
-                vm_map,
-                mmapper,
-                &mut heap,
-            ),
-            common: CommonPlan::new(
-                vm_map,
-                mmapper,
-                options,
-                heap,
-                constraints,
-                global_metadata_specs,
-            ),
+            nursery,
+            common,
             gc_full_heap: AtomicBool::default(),
             next_gc_full_heap: AtomicBool::new(false),
+            full_heap_gc_count,
         }
     }
 
@@ -88,6 +96,9 @@ impl<VM: VMBinding> Gen<VM> {
     /// Prepare Gen. This should be called by a single thread in GC prepare work.
     pub fn prepare(&mut self, tls: VMWorkerThread) {
         let full_heap = !self.is_current_gc_nursery();
+        if full_heap {
+            self.full_heap_gc_count.lock().unwrap().inc();
+        }
         self.common.prepare(tls, full_heap);
         self.nursery.prepare(true);
         self.nursery
@@ -101,6 +112,18 @@ impl<VM: VMBinding> Gen<VM> {
         self.nursery.release();
     }
 
+    /// Independent of how many pages remain in the page budget (a function of heap size), we must
+    /// ensure we never exhaust virtual memory. Therefore we must never let the nursery grow to the
+    /// extent that it can't be copied into the mature space.
+    ///
+    /// Returns `true` if the nursery has grown to the extent that it may not be able to be copied
+    /// into the mature space.
+    fn virtual_memory_exhausted<P: Plan>(&self, plan: &P) -> bool {
+        ((plan.get_collection_reserved_pages() as f64
+            * VM::VMObjectModel::VM_WORST_CASE_COPY_EXPANSION) as usize)
+            > plan.get_mature_physical_pages_available()
+    }
+
     /// Check if we need a GC based on the nursery space usage. This method may mark
     /// the following GC as a full heap GC.
     pub fn collection_required<P: Plan>(
@@ -110,8 +133,13 @@ impl<VM: VMBinding> Gen<VM> {
         space: Option<&dyn Space<VM>>,
     ) -> bool {
         let nursery_full = self.nursery.reserved_pages()
-            >= (conversions::bytes_to_pages_up(*self.common.base.options.max_nursery));
+            >= (conversions::bytes_to_pages_up(self.common.base.options.get_max_nursery()));
+
         if nursery_full {
+            return true;
+        }
+
+        if self.virtual_memory_exhausted(plan) {
             return true;
         }
 
@@ -139,7 +167,7 @@ impl<VM: VMBinding> Gen<VM> {
 
     /// Check if we should do a full heap GC. It returns true if we should have a full heap GC.
     /// It also sets gc_full_heap based on the result.
-    pub fn request_full_heap_collection(&self, total_pages: usize, reserved_pages: usize) -> bool {
+    pub fn requires_full_heap_collection<P: Plan>(&self, plan: &P) -> bool {
         // Allow the same 'true' block for if-else.
         // The conditions are complex, and it is easier to read if we put them to separate if blocks.
         #[allow(clippy::if_same_then_else)]
@@ -165,8 +193,10 @@ impl<VM: VMBinding> Gen<VM> {
         {
             // Forces full heap collection
             true
+        } else if self.virtual_memory_exhausted(plan) {
+            true
         } else {
-            total_pages <= reserved_pages
+            plan.get_total_pages() <= plan.get_reserved_pages()
         };
 
         self.gc_full_heap.store(is_full_heap, Ordering::SeqCst);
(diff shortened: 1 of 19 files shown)


k-sareen:
I tried adding a timer which only timed the major GCs but it was either still counting some mutator time or it was panicking due to `None` in the statistics code. Currently I've only added a counter for the number of major GCs

k-sareen:
If you think the commit is too big, then let me know and I'll break it up into:
1. `PageResource`/`HeapLayout` changes; and
2.  Bounded and Fixed nursery changes.

src/util/heap/layout/map.rs
```diff
@@ -32,6 +32,10 @@ pub trait Map: Sized {
 
     fn get_contiguous_region_size(&self, start: Address) -> usize;
 
+    fn get_available_discontiguous_chunks(&self) -> usize;
```
qinsoon, line 35: Can you add comments for these methods? https://github.com/JikesRVM/JikesRVM/blob/0b6002e7d746a829d56c90acfc4bb5c560faf634/MMTk/src/org/mmtk/utility/heap/layout/Map.java#L141

src/util/heap/layout/map.rs
```diff
@@ -32,6 +32,10 @@ pub trait Map: Sized {
...
+
+    fn get_chunk_consumer_count(&self) -> usize;
```
qinsoon, line 37: Comments: https://github.com/JikesRVM/JikesRVM/blob/0b6002e7d746a829d56c90acfc4bb5c560faf634/MMTk/src/org/mmtk/utility/heap/layout/Map.java#L149

src/util/heap/layout/map64.rs
```diff
@@ -152,6 +152,16 @@ impl Map for Map64 {
         unreachable!()
     }
 
+    fn get_available_discontiguous_chunks(&self) -> usize {
+        // We don't use discontiguous chunks for 64-bit
+        0
```
qinsoon, line 157: This should panic instead of returning 0.

src/util/heap/layout/map64.rs
```diff
@@ -152,6 +152,16 @@ impl Map for Map64 {
...
+    }
+
+    fn get_chunk_consumer_count(&self) -> usize {
+        // We don't use discontiguous chunks for 64-bit
+        0
```
qinsoon, line 162: This should panic instead of returning 0.

src/util/heap/monotonepageresource.rs
```diff
@@ -28,9 +28,9 @@ pub struct MonotonePageResource<VM: VMBinding> {
 
 struct MonotonePageResourceSync {
     /** Pointer to the next block to be allocated. */
-    cursor: Address,
+    pub cursor: Address,
```
qinsoon, line 31: These fields do not need to be public. From what I can see, you are not using them outside this module.

src/util/options.rs
```diff
@@ -239,6 +254,67 @@ macro_rules! options {
...
 }
 
+#[derive(Copy, Clone, EnumString, Debug)]
+pub enum NurseryKind {
+    Bounded,
+    Fixed,
+}
+
+#[derive(Copy, Clone, Debug)]
+pub struct NurserySize {
```
qinsoon, line 264: Need some comments for these types (`NurseryKind` and `NurserySize`).

src/util/options.rs
```diff
@@ -259,11 +335,11 @@ options! {
     // Should we eagerly finish sweeping at the start of a collection? (not supported)
     eager_complete_sweep:  bool                 [env_var: true, command_line: true]  [always_valid] = false,
     // Should we ignore GCs requested by the user (e.g. java.lang.System.gc)?
-    ignore_system_g_c:     bool                 [env_var: true, command_line: true]  [always_valid] = false,
-    // The upper bound of nursery size.
-    max_nursery:           usize                [env_var: true, command_line: true]  [|v: &usize| *v > 0 ] = DEFAULT_MAX_NURSERY,
-    // The lower bound of nusery size.
-    min_nursery:           usize                [env_var: true, command_line: true]  [|v: &usize| *v > 0 ] = DEFAULT_MIN_NURSERY,
+    ignore_system_gc:      bool                 [env_var: true, command_line: true]  [always_valid] = false,
```
qinsoon, line 338: Our CI script currently uses this option -- I will update the script, and update the `ci-perf-kit` revision in the PR.

src/util/options.rs
```diff
@@ -259,11 +335,11 @@ options! {
...
+    // FIXME: This is not a good way to have conflicting options -- we should refactor this
+    // The nursery size for generational plans. It can be one of Bounded or Fixed. The nursery size
+    // can be set like "Fixed:8192", for example, to have a Fixed nursery size of 8192 bytes
+    nursery:               NurserySize          [env_var: true, command_line: true]  [always_valid] = NurserySize { kind: NurseryKind::Bounded, min: DEFAULT_MIN_NURSERY, max: DEFAULT_MAX_NURSERY },
```
qinsoon, line 342: What if the specified size for bounded nursery is smaller than `DEFAULT_MIN_NURSERY`? Is that valid? I feel the validation should be: `min > 0 and max > 0 and max >= min`.

src/util/options.rs
```diff
@@ -259,11 +335,11 @@ options! {
...
     // Should we ignore GCs requested by the user (e.g. java.lang.System.gc)?
-    ignore_system_g_c:     bool                 [env_var: true, command_line: true]  [always_valid] = false,
-    // The upper bound of nursery size.
-    max_nursery:           usize                [env_var: true, command_line: true]  [|v: &usize| *v > 0 ] = DEFAULT_MAX_NURSERY,
-    // The lower bound of nusery size.
-    min_nursery:           usize                [env_var: true, command_line: true]  [|v: &usize| *v > 0 ] = DEFAULT_MIN_NURSERY,
+    ignore_system_gc:      bool                 [env_var: true, command_line: true]  [always_valid] = false,
+    // FIXME: This is not a good way to have conflicting options -- we should refactor this
+    // The nursery size for generational plans. It can be one of Bounded or Fixed. The nursery size
+    // can be set like "Fixed:8192", for example, to have a Fixed nursery size of 8192 bytes
```
qinsoon, line 341: Probably make it clear that for `Bounded`, the specified size is the upper bound.

qinsoon (commented):
LGTM. The ported new methods need comments. And there are some minor issues.

I will need to update CI scripts for the `ignore_system_gc` option.

src/plan/generational/immix/global.rs
```diff
@@ -167,6 +167,9 @@ impl<VM: VMBinding> Plan for GenImmix<VM> {
         }
         self.last_gc_was_full_heap
             .store(full_heap, Ordering::Relaxed);
+
+        self.gen
```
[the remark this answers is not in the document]
k-sareen: Yeah I had previously tried it but Rust was angry at me. I'll try and see if I can refactor something to get it to work

src/util/options.rs
k-sareen: Yes sorry. I was making the option more consistent with the others (like `full_heap_system_gc`).

src/util/options.rs
k-sareen: Ooh that's a good point. I'll validate the sizes in the `new()` function for `NurserySize`

qinsoon: You can change the `[always_valid]` to `|v: &NurserySize| v.min > 0 && v.max > 0 && v.max >= v.min`.

src/plan/generational/global.rs
```diff
@@ -14,15 +14,18 @@ use crate::util::heap::VMRequest;
...
 use crate::util::VMWorkerThread;
 use crate::vm::VMBinding;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use mmtk_macros::PlanTraceObject;
 
+const WORST_CASE_COPY_EXPANSION: f64 = 1.5;
```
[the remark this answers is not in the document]
k-sareen: Hmm... Yes I think you're right. In that case is 1.5x correct for OpenJDK and our other VMs? I can add it to `vm::ObjectModel`.

k-sareen:
Yes sorry -- will document code further :+1:

src/plan/generational/global.rs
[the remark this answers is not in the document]
qinsoon: Yeah. 1.5 should fine. For OpenJDK, 1 is probably fine (I assume we do not grow object size during copying for OpenJDK). I don't know how much this would affect performance. If it is significant, we may want to make 1 as the default value, and change the value for JikesRVM to 1.5.

src/plan/generational/immix/global.rs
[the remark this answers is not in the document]
qinsoon: If it needs some major refactoring, you can leave it as it is and add `TODO` comments for it. If it is trivial, probably just change it in the PR.

src/plan/generational/global.rs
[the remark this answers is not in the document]
k-sareen: I've moved it to `ObjectModel`, but have kept the value to still be 1.5. I think OpenJDK technically doesn't change the size of copied objects, but I'm not certain so I'm being conservative currently.

k-sareen:
Done. Please don't merge until I've done a performance evaluation after rebasing. I only have stale results.

qinsoon:
> Done. Please don't merge until I've done a performance evaluation after rebasing. I only have stale results.

Yeah. I won't merge it right now. I will update the CI script version in this PR as well.

qinsoon (approved):

k-sareen:
Performance results for [`GenImmix`](https://squirrel.anu.edu.au/plotty/kunals/bench-test/#0|2022-07-27-nursery-pr-performance-boar-2022-07-27-Wed-153152&benchmark^build^buildstring^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&mmtk_gc^1^GenImmix|20&1^invocation|70&build^nursery-pr|30&2&benchmark|41&Histogram%20(with%20CI)^build^benchmark&)
Performance results for [`GenCopy`](https://squirrel.anu.edu.au/plotty/kunals/bench-test/#0|2022-07-27-nursery-pr-performance-boar-2022-07-27-Wed-153152&benchmark^build^buildstring^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&mmtk_gc^1^GenCopy|20&1^invocation|70&build^nursery-pr|30&2&benchmark|41&Histogram%20(with%20CI)^build^benchmark&)

Note I used a very generous/roomy heap factor of 4x min heap. I thought it would showcase the difference more starkly. Mostly the PR sees a performance improvement but `GenCopy` slows down for a few benchmarks which I believe is due to an increased number of GCs from the changes to how minor/major GCs are triggered

qinsoon:
That looks good. I have also updated the CI scripts. We can merge this PR.

qinsoon:
@k-sareen Can you take a look at the failed test? Just check if it is reproducible, and if it is related with the changes in this pull request.

k-sareen:
`luindex` is known to be broken for the generational plans (at least the new `luindex`). Likely to do with barriers. See: https://github.com/mmtk/mmtk-openjdk/issues/156. I'll investigate and see why this one failed.

qinsoon:
> `luindex` is known to be broken for the generational plans (at least the new `luindex`). Likely to do with barriers. See: [mmtk/mmtk-openjdk#156](https://github.com/mmtk/mmtk-openjdk/issues/156). I'll investigate and see why this one failed.

It is possible that this is a bug revealed by this PR. I don't think this PR changes any code that may cause segfault. But it changes the GC pattern and that may reveal new bugs.

Issue #458, opened 2021-09-30
Side metadata issues


qinsoon:
This issue tracks problems of our side metadata implementation.

* [ ] We assume we only use a certain amount of address space and calculate how many bytes we would need for a side metadata spec, so we can layout another side metadata spec immediately after this spec. However, if somehow we use extra address space (for example, we assume we only use address space within our heap range, but malloc returns an arbitrary address), we may end up calculating a side metadata address that is actually out of bound for this spec, and belongs to the spec after it. We have no check for this (Malloc space implements its own check routine tough). 
* [ ] For 64 bits, global side metadata base address is `0x600_0000_0000`. For 64 bits, the range for the first space is `[0x200_0000_0000, 0x400_0000_0000)`, for the second space `[0x400_0000_0000, 0x600_0000_0000)`, and for the third space `[0x600_0000_0000, 0x800_0000_0000)`. This means our global side metadata clashes with the third space we have.
* [ ] When we unmap side metadata (at the end of tests), we do not tell the mmapper about the unmapping. The mmapper would still think we have mapped those chunks.
* [ ] When calculating the side metadata address and size to map (e.g. `try_mmap_contiguous_metadata_space()`), we are not considering overflow. If we try reserve side metadata for the whole address range (e.g. 32 bits), we will have an arithmetic overflow.

PR #633, opened 2022-08-02
Feature to turn Immix into non-moving GC

wks:
1.  [X] We should rebase it after https://github.com/mmtk/mmtk-core/pull/629 is merged.
2.  [X] Ruby crashes with segmentation fault when using the non-moving Immix.  The stack trace indicates there are some issues with the alloc bit when using Immix.
        -   The problem was caused by the metadata overlapping with the large object space.  I added a workaround to temporarily solve this problem.
        -   Related issue: https://github.com/mmtk/mmtk-core/issues/458

Added some features to disable defragmentation and make Immix a
non-moving GC.  It is useful during developing new VM bindings so that
the VM can use Immix before the VM fully supports object moving.

The `immix_no_defrag` feature disables defragmentation in ImmixSpace.
The Immix Plan now reports itself as a non-moving collector
(IMMIX_CONSTRAINTS.moves_objects = false) if policy::immix::DEFRAG is
false.

The `immix_smaller_block` feature reduces the size of blocks of
ImmixSpace.  This should mitigate fragmentation when defragmentation is
disabled.

Also worked around metadata and space address clash by moving metadata to a higher address.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,13 @@ global_alloc_bit = []
 # conservative garbage collection support
 is_mmtk_object = ["global_alloc_bit"]
 
+# The following two features are useful for using Immix for VMs that do not support moving GC.
+
+# Disable defragmentation for ImmixSpace.  This makes Immix a non-moving plan.
+immix_no_defrag = []
+# Reduce block size for ImmixSpace.  This mitigates fragmentation when defrag is disabled.
+immix_smaller_block = []
+
 # Run sanity GC
 sanity = []
 # Run analysis
--- a/src/plan/immix/global.rs
+++ b/src/plan/immix/global.rs
@@ -38,7 +38,7 @@ pub struct Immix<VM: VMBinding> {
 }
 
 pub const IMMIX_CONSTRAINTS: PlanConstraints = PlanConstraints {
-    moves_objects: true,
+    moves_objects: crate::policy::immix::DEFRAG,
     gc_header_bits: 2,
     gc_header_words: 0,
     num_specialized_scans: 1,
--- a/src/policy/immix/block.rs
+++ b/src/policy/immix/block.rs
@@ -84,7 +84,10 @@ impl From<Block> for Address {
 }
 
 impl Region for Block {
+    #[cfg(not(feature = "immix_smaller_block"))]
     const LOG_BYTES: usize = 15;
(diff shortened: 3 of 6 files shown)


wks:
I tested with OpenJDK and compared the effect of different block sizes.  The command is:
```
MMTK_THREADS=16 RUST_LOG=info RUST_BACKTRACE=1 MMTK_PLAN=Immix /path/to/linux-x86_64-normal-server-slowdebug/jdk/bin/java -XX:+UseThirdPartyHeap -Xm{s,x}100M -jar /path/to/dacapo-9.12-MR1-bach.jar lusearch
```
I ran it with block sizes of 2^13 and 2^15.  My expected result is, because non-moving Immix cannot reuse partially empty blocks, the heap will become more fragmented if the block size is large, and it will trigger more GC because the heap will become full more often.

But the reality is just the opposite. When running with smaller block size, GC becomes more frequent.

|block size|number of GC|
|---|---|
|2^13|21|
|2^15|15|

How to explain this phenomenon?

qinsoon:
I started CI tests for bindings -- we need to check if the metadata address changes break any binding (they may hard-code the value).

Also I also requested a review from @udesou  to make sure this change also works fine for the Julia binding.

qinsoon (approved):
LGTM

udesou:
I'm running the Julia correctness tests at the moment, will approve as soon as they are done.

qinsoon:
The V8 test failed. But that seems to be a problem from previous PRs. We should fix that in the binding.
The API check failed. I am not sure if it is because the PR is from a fork. I will need to check that.

wks:
The failure in Public API Check failed in the "Documenting mmtk" step.  Some comments contain cross-references to code which the tool does not recognise.  They are not related to this PR.

qinsoon:
> The failure in Public API Check failed in the "Documenting mmtk" step. Some comments contain cross-references to code which the tool does not recognise. They are not related to this PR.

I think it invokes `cargo doc` to generate an API list in JSON format. It succeeded for master, but failed when it tried to checkout code for the pull request (possibly because this PR is from a fork). It is not related to the PR. I will work on it separately.

udesou:
The Julia tests succeeded. All good with merging from that perspective.

udesou (approved):

Issue #348, opened 2021-06-15
Mimalloc Port (Free List Allocator and Mark Sweep Space)


paigereeves:
The goal is to have an implementation of malloc/free based on [mimalloc](https://www.microsoft.com/en-us/research/uploads/prod/2019/06/mimalloc-tr-v1.pdf).

**Free List Allocator**
- Thread-local lists of owned 64kB blocks. Separated (possibly implicitly) into available, exhausted, and unswept blocks.
- Segregated size classes.
- Eager or lazy sweeping.
- A method to sweep a marked block by constructing a free list.
- (Lazy) allocate to unswept blocks by constructing free lists as required and moving swept blocks to the available list.
- (Lazy) allocation is attempted first from a thread's available list, then its unswept list, then other threads' unswept lists, then fresh memory from the associated space.
- Reconstruct thread-local block lists after marking.
- A global list of all active 64kB blocks in that space, used for stealing unswept blocks from other threads.

**Mark Sweep Space**
- Referencing [the old MMTk's MarkSweepSpace policy](https://github.com/JikesRVM/JikesRVM/blob/master/MMTk/src/org/mmtk/policy/MarkSweepSpace.java), but excluding any functionality including free lists and size classes, for example, which are considered utility and exist within the allocator.
- Send a signal to allocator threads when marking is completed.
- Should be able to replace this with a (yet unimplemented) reference counting policy without any changes from the allocator's side.

Issue #480, opened 2021-10-19
Assert VMBinding implementation in MMTk core


qinsoon:
MMTk core requires the VM binding to implement the `VMBinding` trait. However, if they do not implement the traits correctly, or sometimes miss a few methods (and return invalid results instead), MMTk core would just fail strangely, and does not give any useful warning or debug message.

For example, I encountered a case where the `get_next_mutator()` is not implemented properly. Because of this, the preparation for each mutator is not called at all, and the mutator does not have its thread local state reset for the GC, which leads to a strange error when the mutator tries to allocate after GC. We should have some sort of assertions for this, like check the number of mutators in GC preparation, and file a warning if the number is 0.

I think we would like to go through the `VMBinding` trait, and try have assertions for each method if possible. This could help the binding implementer greatly.

qinsoon:
> For example, I encountered a case where the `get_next_mutator()` is not implemented properly. Because of this, the preparation for each mutator is not called at all, and the mutator does not have its thread local state reset for the GC, which leads to a strange error when the mutator tries to allocate after GC. We should have some sort of assertions for this, like check the number of mutators in GC preparation, and file a warning if the number is 0.

We should find a way to assert if a mutator is properly reset in the GC. We would have the same issue if we miss any mutator.

PR #643, opened 2022-08-17
Mimalloc Allocator

paigereeves:
This pull request introduces an allocator based on Microsoft's mimalloc (https://www.microsoft.com/en-us/research/uploads/prod/2019/06/mimalloc-tr-v1.pdf) used in a MarkSweep GC. This closes https://github.com/mmtk/mmtk-core/issues/348.

This PR has a few leftover issues: https://github.com/mmtk/mmtk-core/issues/688

General changes in this PR (if requested, I can separate these changes to different PRs):
* `destroy_mutator()`:
  * now takes `&mut Mutator` instead of `Box<Mutator>`. The reclamation of the `Mutator` memory is up to the binding, and allows the binding to copy the `Mutator` value to their thread local data structure, in which case, Rust cannot reclaim the memory as if it is a `Box`.
  * now calls each allocator for their allocator-specific `on_destroy()` behavior.
* Extract `Chunk` and `ChunkMap` from Immix, and make it general for other policies to use.
* Changes in `VMBinding` constants:
  * `LOG_MIN_ALIGNMENT` is removed. We still provide `MIN_ALIGNMENT`. This avoids confusion that a binding may override one constant but not the other.
  * `MAX_ALIGNMENT_SHIFT` is removed. We provide `MAX_ALIGNMENT` for the same reason as above.
  * Add `USE_ALLOCATION_OFFSET: bool`. If a binding never use allocation offset (i.e. `offset === 0`), they can set this to `false`, and MMTk core could use this for some optimizations.
* Changes in `ObjectModel`:
  * Add `UNIFIED_OBJECT_REFERENCE_ADDRESS: bool` to allow a binding to tell us if the object reference uses the same address as its object start and its to_address.
  * Add `OBJECT_REF_OFFSET_LOWER_BOUND` to allow binding to tell us roughly where the object reference points to (with respect of the allocation address).
  * Changes to related methods to cope with this change.

Mark sweep changes in this PR:
* add two features for marksweep
  * `eager_sweeping`: sweep unused memory eagerly in each GC. Without this feature, unused memory is swept when we attempt to allocate from them.
  * `malloc_mark_sweep`: the same as our previous `malloc_mark_sweep`. When this is used, the mark sweep plan uses `MallocSpace` and `MallocAllocator`.
* Move the current `policy::mallocspace` to `policy::marksweepspace::malloc_ms`
* Add `policy::marksweepspace::native_ms` for the mark sweep space with mimalloc.
* Add `util::alloc::free_list_allocator`.


wks:
Yes. It needs to be merged/rebased with the `master` version because the architecture and the API have changed a lot.

I also see `#[cfg(feature="malloc")]` macros everywhere.  Essentially, a space using our own free list implementation and a space that uses external `malloc` libraries are two different spaces.  A plan that uses those different spaces should also be considered different plans.  But as both use the mark-sweep algorithm, there should be some code that can be shared between those two plans/spaces.

src/util/heap/chunk_map.rs
```diff
@@ -114,7 +88,7 @@ impl ChunkMap {
         unsafe { Self::ALLOC_TABLE.store::<u8>(chunk.start(), state as u8) };
         // If this is a newly allcoated chunk, then expand the chunk range.
         if state == ChunkState::Allocated {
-            debug_assert!(!chunk.start().is_zero());
+            assert!(!chunk.start().is_zero());
```
qinsoon, line 91: I should revert this to `debug_assert!`

qinsoon:
binding-refs
OPENJDK_BINDING_REF=mimalloc-ms-support
JIKESRVM_BINDING_REF=mimalloc-ms-support
V8_BINDING_REF=update-pr-643

qinsoon:
@wks @wenyuzhao I have requested reviews from you two. Before you do a comprehensive review, could you let me know if I should get this PR split into a few smaller PRs?

src/plan/marksweep/global.rs
```diff
@@ -95,47 +100,50 @@ impl<VM: VMBinding> Plan for MarkSweep<VM> {
...
-        #[cfg(not(feature = "global_alloc_bit"))]
-        let global_metadata_specs = SideMetadataContext::new_global_specs(&[
-            ALLOC_SIDE_METADATA_SPEC,
-            ACTIVE_CHUNK_METADATA_SPEC,
-        ]);
-
-        let res = MarkSweep {
-            ms: MallocSpace::new(global_metadata_specs.clone()),
-            common: CommonPlan::new(
+    #[allow(unused_variables)] // scheduler only used by marksweepspace
```
wks, line 103: I removed this line, but didn't see any "unused variables" warnings.  Is it still relevant?

qinsoon: I think it can be removed.

src/util/alloc/immix_allocator.rs
```diff
@@ -304,12 +306,14 @@ impl<VM: VMBinding> ImmixAllocator<VM> {
                     block.start(),
                     block.end()
                 );
+                let (cursor, limit) =
+                    adjust_thread_local_buffer_range::<VM>(block.start(), block.end());
```
wenyuzhao, line 310: Is there any reason to adjust block start/end addresses? Looks like Java Immix does not need this?

https://github.com/JikesRVM/JikesRVM/blob/master/MMTk/src/org/mmtk/utility/alloc/ImmixAllocator.java#L192-L198

src/util/alloc/object_ref_guard.rs
```diff
@@ -20,46 +20,72 @@ use crate::util::Address;
 use crate::vm::ObjectModel;
 use crate::vm::VMBinding;
 
-/// Adjust limit for thread local buffer to make sure that we will not allocate objects whose object reference may
+/// Adjust start and end for thread local buffer to make sure that we will not allocate objects whose object reference may
```
wenyuzhao, line 23: > we will not allocate objects whose object reference may be in another chunk.

It is totally ok to have objects like this. JikesRVM always does that. If you'd like to get the chunk owning the object, you should use `VMObjectModel::ref_to_address`. `object.to_address()` is incorrect for this case.

src/util/alloc/large_object_allocator.rs
```diff
@@ -36,21 +36,20 @@ impl<VM: VMBinding> Allocator<VM> for LargeObjectAllocator<VM> {
 
     fn alloc(&mut self, size: usize, align: usize, offset: isize) -> Address {
         #[cfg(debug_assertions)]
-        crate::util::alloc::object_ref_guard::assert_object_ref_in_cell::<VM>(size);
+        crate::util::alloc::object_ref_guard::assert_object_ref_in_allocation::<VM>(size);
```
wenyuzhao, line 39: To my understanding, this assertion should always be true and is redundant. The only case for JikesRVM to have an object pointer pointing outside of the object is zero-sized objects. For these zero-sized objects, (1) they should not go to LOS (2) even if they do, they will never be placed at the end of a chunk or page, since LOS allocate memory at page granularity.

vmbindings/dummyvm/src/api.rs
```diff
@@ -36,7 +36,7 @@ pub extern "C" fn mmtk_bind_mutator(tls: VMMutatorThread) -> *mut Mutator<DummyV
 
 #[no_mangle]
 pub extern "C" fn mmtk_destroy_mutator(mutator: *mut Mutator<DummyVM>) {
-    memory_manager::destroy_mutator(unsafe { Box::from_raw(mutator) })
+    memory_manager::destroy_mutator(unsafe { &mut *mutator })
```
wenyuzhao, line 39: Is there any memory leak issue by doing so?

src/util/metadata/mark_bit.rs
```diff
@@ -0,0 +1,16 @@
+use crate::util::ObjectReference;
+use crate::vm::VMBinding;
+use crate::vm::VMLocalMarkBitSpec;
+use std::sync::atomic::Ordering;
+
+impl VMLocalMarkBitSpec {
+    /// Set the mark bit for the object to 1
+    pub fn set<VM: VMBinding>(&self, object: ObjectReference, ordering: Ordering) {
```
wenyuzhao, line 8: it is probably better to rename the methods to `mark`/`is_marked`.

src/util/linear_scan.rs
```diff
@@ -116,6 +116,11 @@ pub trait Region: Copy + PartialEq + PartialOrd + From<Address> + Into<Address>
     fn containing<VM: VMBinding>(object: ObjectReference) -> Self {
         Self::from(VM::VMObjectModel::ref_to_address(object).align_down(Self::BYTES))
     }
+    /// Check if the given address is in the region.
+    #[inline(always)]
+    fn includes_address(&self, addr: Address) -> bool {
+        addr >= self.start() && addr < self.end()
```
wenyuzhao, line 122: Slightly faster implementation: `Self::align(addr) == self.0`

src/util/alloc_bit.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) const ALLOC_SIDE_METADATA_SPEC: SideMetadataSpec =
 pub const ALLOC_SIDE_METADATA_ADDR: Address = ALLOC_SIDE_METADATA_SPEC.get_absolute_offset();
 
 pub fn map_meta_space_for_chunk(metadata: &SideMetadataContext, chunk_start: Address) {
+    eprintln!("map meta space for chunk {}", chunk_start);
```
wenyuzhao, line 16: Is this a debug leftover?

src/util/alloc/bumpallocator.rs
```diff
@@ -29,9 +29,10 @@ pub struct BumpAllocator<VM: VMBinding> {
 }
 
 impl<VM: VMBinding> BumpAllocator<VM> {
-    pub fn set_limit(&mut self, cursor: Address, limit: Address) {
+    pub fn set_limit(&mut self, start: Address, limit: Address) {
+        let (cursor, limit) = adjust_thread_local_buffer_range::<VM>(start, limit);
```
wenyuzhao, line 33: Same as ImmixAllocator. Probably this is unnecessary.

src/memory_manager.rs
```diff
@@ -103,12 +106,14 @@ pub fn bind_mutator<VM: VMBinding>(
...
+/// Report to MMTk that a mutator that is no longer needed. A binding should not attempt
+/// to use the mutator after this call. MMTk will not attempt to reclaim the memory for the
+/// mutator, so a binding should properly reclaim the memory for the mutator after this call.
 ///
 /// Arguments:
 /// * `mutator`: A reference to the mutator to be destroyed.
-pub fn destroy_mutator<VM: VMBinding>(mutator: Box<Mutator<VM>>) {
-    drop(mutator);
+pub fn destroy_mutator<VM: VMBinding>(mutator: &mut Mutator<VM>) {
+    mutator.on_destroy();
```
wenyuzhao, line 116: Does this imply that bindings should be responsible for releasing `Box<Mutator<VM>>`?

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,447 @@
...
+        }
+
+        self.store_free_list(last);
+    }
+
+    /// This is a naive implementation that is inefficient but should be correct.
+    /// In this implementation, we simply go through each possible object
+    /// reference and see if it has the mark bit set. If we find mark bit, that means the cell is alive. If we didn't find
+    /// the mark bit in the entire cell, it means the cell is dead.
+    fn naive_brute_force_sweep<VM: VMBinding>(&self) {
```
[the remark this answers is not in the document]
wks: Or we can work out the in-cell offset given any address inside that object.
```rust
let any_address = ObjectModel::ref_to_cell(objref);
let in_cell_offset = (any_address - block_start) % cell_size;
```
Then we know the start address of an allocation (which is equal to the start of the cell in the freelist allocator) by
```rust
let start_of_cell = any_address - in_cell_offset;
assert_eq!(start_of_cell, ObjectModel::object_start_ref(objref));
```
In this way, we only need to do arithmetic (and avoid accessing object headers) to know the cell start.  Then we can reuse the per-word mark bit side-metadata.

src/util/alloc/free_list_allocator.rs
```diff
@@ -0,0 +1,530 @@
...
+    // This will usually be the first block on the available list. If all available blocks are found
+    // to be full, other lists are searched
+    // This function allows different available block lists -- normal allocation uses self.avaialble_blocks, and precise stress test uses self.avialable_blocks_stress.
+    #[inline(always)]
+    fn find_free_block_with(
+        available_blocks: &mut BlockLists,
+        consumed_blocks: &mut BlockLists,
+        size: usize,
+        align: usize,
+    ) -> Block {
```
wks, line 222: I think we still need to take offset into consideration when searching for free block.

For example, if the VM requests an object of 64 bytes, it can be allocated into the 64-byte size class; but if the VM requests that the object must be aligned to 16 bytes at +4 byte offset, then the object has to be at least 64 + (16 - 4) = 76 bytes large, and it has to be allocated into the next size class, i.e. 80 bytes.

```suggestion
    fn find_free_block_with(
        available_blocks: &mut BlockLists,
        consumed_blocks: &mut BlockLists,
        size: usize,
        align: usize,
        offset: usize,
    ) -> Block {
```

src/util/alloc/free_list_allocator.rs
```diff
@@ -0,0 +1,530 @@
...
+            self.available_blocks[bin].push(block);
+        }
+    }
+
+    /// Tries to recycle local blocks if there is any. This is a no-op for eager sweeping mark sweep.
+    #[cfg(not(feature = "eager_sweeping"))]
+    #[inline]
+    fn recycle_local_blocks(&mut self, size: usize, align: usize, _stress_test: bool) -> Block {
+        // attempt to sweep
+        loop {
```
wks, line 264: We could just move the `#[cfg(...)]` macro inside the function.
```suggestion
    #[inline]
    fn recycle_local_blocks(&mut self, size: usize, align: usize, _stress_test: bool) -> Block {
        if cfg!(not(feature = "eager_sweeping")) {
            return Block::ZERO_BLOCK; // We should really make Block non-zero and return None here.
        }    
        // attempt to sweep
        loop {
```

src/util/alloc/free_list_allocator.rs
```diff
@@ -0,0 +1,530 @@
...
+                self.consumed_blocks.get_mut(bin).unwrap().push(block);
+            }
+        }
+        Block::ZERO_BLOCK
+    }
+
+    /// Tries to recycle local blocks if there is any. This is a no-op for eager sweeping mark sweep.
+    #[cfg(feature = "eager_sweeping")]
+    #[inline]
+    fn recycle_local_blocks(&mut self, _size: usize, _align: usize, _stress_test: bool) -> Block {
```
wks, line 293: With `cfg!` inside the function, this function definition can be removed.

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -0,0 +1,369 @@
...
+use crate::util::VMThread;
+use crate::vm::ObjectModel;
+use std::sync::Mutex;
+
+/// The result for `MarkSweepSpace.acquire_block()`. `MarkSweepSpace` will attempt
+/// to allocate from abandoned blocks first. If none found, it will get a new block
+/// from the page resource.
+pub enum BlockAcquireResult {
+    /// A new block we just acquired from the page resource
+    Fresh(Block),
```
wks, line 40: Consider adding a special case for space exhaustion, because `acquire_block` may fail.
```suggestion
    /// No block is available.
    Exhausted,
    /// A new block we just acquired from the page resource
    Fresh(Block),
```

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -0,0 +1,369 @@
...
+                    let block = Block::from(abandoned_unswept[bin].pop().start());
+                    return BlockAcquireResult::AbandonedUnswept(block);
+                }
+            }
+        }
+
+        BlockAcquireResult::Fresh(Block::from(
+            self.acquire(tls, Block::BYTES >> LOG_BYTES_IN_PAGE),
+        ))
+    }
```
wks, line 328: `self.acquire` may return 0 and that indicates GC is required. I suggest adding a special result `BlockAcquireResult::Exhausted` for that case as I mentioned above.
```suggestion
        let addr = self.acquire(tls, Block::BYTES >> LOG_BYTES_IN_PAGE);
        if addr == Block::ZERO_BLOCK { // We should make Block non-zero and check for None/Some.
            BlockAcquireResult::Exhausted
        } else {
            BlockAcquireResult::Fresh(Block::from(addr))
        }
    }
```

wks (changes requested):
The biggest problem so far is still that `Block` may be `ZERO_BLOCK`.  This effectively makes every `Block` nullable.  It is OK for C, but in Rust we should use `Option<Block>` when a block may not exist.

There is also a global problem that doc comments before functions use two slashes (`//`) instead of three (`///`).

src/util/heap/chunk_map.rs
```diff
@@ -33,48 +28,25 @@ impl From<Chunk> for Address {
...
 }
 
 impl Chunk {
     /// Chunk constant with zero address
-    const ZERO: Self = Self(Address::ZERO);
-    /// Log blocks in chunk
-    pub const LOG_BLOCKS: usize = Self::LOG_BYTES - Block::LOG_BYTES;
-    /// Blocks in chunk
-    pub const BLOCKS: usize = 1 << Self::LOG_BLOCKS;
+    pub const ZERO: Self = Self(Address::ZERO);
```
wks, line 36: Like `struct Block(Address)`, we may want `Chunk` to be non-zero, and use `Option<Chunk>`, too, for the same reason.  But we may postpone the refactoring to another PR because `Chunk` is not introduced in this PR.

src/util/metadata/side_metadata/spec_defs.rs
```diff
@@ -75,8 +75,29 @@ define_side_metadata_specs!(
     IX_BLOCK_DEFRAG = (global: false, log_num_of_bits: 3, log_bytes_in_region: crate::policy::immix::block::Block::LOG_BYTES),
     // Mark blocks by immix
     IX_BLOCK_MARK   = (global: false, log_num_of_bits: 3, log_bytes_in_region: crate::policy::immix::block::Block::LOG_BYTES),
-    // Mark chunks by immix
-    IX_CHUNK_MARK   = (global: false, log_num_of_bits: 3, log_bytes_in_region: crate::policy::immix::chunk::Chunk::LOG_BYTES),
+    // Mark chunks (any plan that uses the chunk map should include this spec in their local sidemetadata specs)
+    CHUNK_MARK   = (global: false, log_num_of_bits: 3, log_bytes_in_region: crate::util::heap::chunk_map::Chunk::LOG_BYTES),
+    // Mark chunks by (native mimalloc) marksweep
+    // MS_CHUNK_MARK   = (global: false, log_num_of_bits: 3, log_bytes_in_region: crate::policy::marksweepspace::chunk::Chunk::LOG_BYTES),
```
wks, line 81: If this is no longer needed, we can remove it.

vmbindings/dummyvm/src/tests/allocate_align_offset.rs
```diff
@@ -0,0 +1,45 @@
...
+pub fn allocate_offset() {
+    MUTATOR.with_fixture(|fixture| {
+        const OFFSET: isize = 4;
+        let min = DummyVM::MIN_ALIGNMENT;
+        let max = DummyVM::MAX_ALIGNMENT;
+        info!("Allowed alignment between {} and {}", min, max);
+        let mut align = min;
+        while align <= max {
+            info!("Test allocation with alignment {} and offset {}", align, OFFSET);
+            let addr = api::mmtk_alloc(fixture.mutator, 8, align, OFFSET, AllocationSemantics::Default);
```
wks, line 40: As I mentioned in a previous comment, adding an offset may require the object to be allocated in a larger size class than the size requested.  In this case, the 8-byte allocation should be allocated in a 16-byte cell because of `OFFSET`, but I suspect it currently isn't.  I recommend add an assertion that the allocated region (from `addr` to `addr + size`) lies within the cell.

There is no public API that exposes that information.  We may add a `debug_assert!` inside the freelist allocator instead of this test case.

Alternatively, we can optionally (e.g. via a feature, maybe named "testing") expose some functions for performing white-box testing, only.

src/util/alloc/immix_allocator.rs
qinsoon: Is there a fix that I can do for this in this PR?

src/plan/mutator_context.rs
```diff
@@ -129,6 +129,27 @@ impl<VM: VMBinding> MutatorContext<VM> for Mutator<VM> {
...
 
+impl<VM: VMBinding> Mutator<VM> {
+    /// Get all the valid allocator selector (no duplicate)
+    fn get_all_allocator_selectors(&self) -> Vec<AllocatorSelector> {
+        use itertools::Itertools;
+        self.config
+            .allocator_mapping
+            .iter()
+            .map(|(_, selector)| *selector)
+            .dedup()
```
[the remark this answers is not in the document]
qinsoon: Thanks. I will fix this.

src/policy/marksweepspace/native_ms/block.rs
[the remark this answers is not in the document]
qinsoon: > I wonder if we can optimise this if `ObjectModel` provides a function that is guaranteed to return any address within an object given an object reference (currently `ObjectModel::ref_to_address`). If each cell has the same size, then given any address within an object, we can workout the cell index by dividing the in-block offset by the cell size. In this way, we don't need a per-object mark bit, but a per-cell mark bit local to each block.

What you said would work. It is conceptually simpler. But it has some drawbacks:
1. We already have an object mark bit (in header or on side), and we do not want to introduce other metadata for the cells.
2. If we do not use object mark bit, and use the cell bit instead, it might be slower than what this code is doing. Cell size for each block is stored in side metadata. To get the cell address, we need to load the cell size, and then do computation. If this happens for any query on the object liveness, it could be slow as well.
3. This 'naive' method is the worst case we could have (which has ~10-15% mutator overhead), but the simplest method (`simple_sweep()`) which works for OpenJDK, there is no overhead. Between those two cases, we still have chance to improve.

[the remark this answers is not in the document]
qinsoon: > Or we can work out the in-cell offset given any address inside that object.
> 
> ```rust
> let any_address = ObjectModel::ref_to_cell(objref);
> let in_cell_offset = (any_address - block_start) % cell_size;
> ```
> 
> Then we know the start address of an allocation (which is equal to the start of the cell in the freelist allocator) by
> 
> ```rust
> let start_of_cell = any_address - in_cell_offset;
> assert_eq!(start_of_cell, ObjectModel::object_start_ref(objref));
> ```
> 
> In this way, we only need to do arithmetic (and avoid accessing object headers) to know the cell start. Then we can reuse the per-word mark bit side-metadata.

The address we return from `alloc()` is based on the alignment and offset for that allocation. The `in_cell_offset` is not a constant. We can only optimize if we know some invariants, like the conditional checks for using `simple_sweep()`.

src/policy/marksweepspace/native_ms/block_list.rs
```diff
@@ -0,0 +1,318 @@
...
+                self.last = ZERO_BLOCK;
+            } else {
+                next.store_prev_block(ZERO_BLOCK);
+                self.first = next;
+                next.store_block_list(self);
+            }
+        } else {
+            if next.is_zero() {
+                prev.store_next_block(next);
+                prev.store_next_block(ZERO_BLOCK);
```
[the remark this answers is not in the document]
qinsoon: Yeah, this is mistake. I will fix it.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,447 @@
...
+        block.0
+    }
+}
+
+impl Region for Block {
+    const LOG_BYTES: usize = 16;
+}
+
+impl Block {
+    pub const ZERO_BLOCK: Self = Self(Address::ZERO);
```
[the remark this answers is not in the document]
qinsoon: I have done the refactoring, and I am using `Block(NonZerousize)` now.

src/policy/marksweepspace/native_ms/block_list.rs
```diff
@@ -0,0 +1,318 @@
...
+                next.store_prev_block(prev);
+            }
+        }
+    }
+
+    // Pop the first block in the list
+    pub fn pop(&mut self) -> Block {
+        let rtn = self.first;
+        if rtn.is_zero() {
+            return rtn;
```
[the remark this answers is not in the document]
qinsoon: Fixed.

src/policy/marksweepspace/native_ms/block_list.rs
```diff
@@ -0,0 +1,318 @@
...
+            next.store_prev_block(ZERO_BLOCK);
+            self.first.store_block_list(self);
+        }
+        rtn.store_next_block(ZERO_BLOCK);
+        rtn.store_prev_block(ZERO_BLOCK);
+        rtn
+    }
+
+    // Push block to the front of the list
+    pub fn push(&mut self, block: Block) {
```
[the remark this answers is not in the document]
qinsoon: Fixed.

src/memory_manager.rs
qinsoon: Yes, the binding should be responsible for releasing. That is made clear in the documentation in this PR.

This change is necessary. Our current API (which requires `Box<Mutator>`) does not really work at all in some cases. For example, a binding copies the mutator content to their thread local storage. When they call `destroy_mutator()`, they pass their thread local storage pointer as `Box<Mutator>`, but Rust is not able to drop the `Box` as it is not the original memory Rust allocates.

qinsoon: > I have a different concern. If the API is symmetric, their behaviours should be symmetric, too. If `create_mutator` allocates a box and returns a `Box<Mutator<VM>>`, then `destroy_mutator` should take a `Box<Mutator<VM>>` and destroy the box. That's intuitive. Given that we want the mutator to be created and destroyed differently, I suggest we change the API so the creation and the destruction look different.

If you look at `bind_mutator` and `destroy_mutator`, yes, I think you are right. It is intuitive that 'destroy' would undo whatever you have done in 'bind'. However, we currently have `bind_mutator() -> Box<Mutator>`, `destroy_mutator(&mut Mutator)` and `flush_mutator(&mut Mutator)`. I feel it is more or less okay -- you create the mutator first, then do different things with it. The documentation makes it clear what each method expects, and how a binding can manage the `Box`'s lifetime. I feel it is okay. Let me know what you think.

> Creation (`bind_mutator`) can remain like this, or change it to a method of `MMTK`: `mmtk.create_mutator(tls)`.
> 
> I suggest expose `Mutator::on_destroy(&mut self)` (or rename it to simply `destroy()`) as a public method. When the VM calls this method on the `Mutator` in its own data structure, it is clear that it is not the same as the creation API.
> 
> p.s. This is a chance to gradually move away from the `memory_manager` module. See #658

`Mutator::on_destroy()` is public.

src/util/alloc/free_list_allocator.rs
qinsoon: I think `offset` does not matter here. If I understand correct, `offset` means our allocation result `res` should have `(res + offset).is_aligned_to(align)`, and the allocation range is `[res, res + size)`.

`offset` is not passed to this method, as it is not required by `get_maximum_aligned_size()`.
https://github.com/mmtk/mmtk-core/blob/06144875903b36b432587eb7462e0272750b71d9/src/util/alloc/allocator.rs#L94-L98

src/memory_manager.rs
```diff
@@ -103,12 +106,14 @@ pub fn bind_mutator<VM: VMBinding>(
     mutator
 }
 
-/// Reclaim a mutator that is no longer needed.
+/// Report to MMTk that a mutator that is no longer needed. A binding should not attempt
```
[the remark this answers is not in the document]
qinsoon: Fixed.

src/policy/marksweepspace/native_ms/block_list.rs
```diff
@@ -0,0 +1,318 @@
...
+            self.first.store_prev_block(block);
+            block.store_prev_block(ZERO_BLOCK);
+            self.first = block;
+        }
+        block.store_block_list(self);
+    }
+
+    // Append one block list to another
+    // The second block list left empty
+    pub fn append(&mut self, list: &mut BlockList) {
```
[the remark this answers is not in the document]
qinsoon: Fixed.

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -0,0 +1,369 @@
...
+
+    pub fn block_clear_metadata(&self, block: Block) {
+        let clear_metadata = |spec: &SideMetadataSpec| match spec.log_num_of_bits {
+            0..=3 => spec.store_atomic::<u8>(block.start(), 0, Ordering::SeqCst),
+            4 => spec.store_atomic::<u16>(block.start(), 0, Ordering::SeqCst),
+            5 => spec.store_atomic::<u32>(block.start(), 0, Ordering::SeqCst),
+            6 => spec.store_atomic::<u64>(block.start(), 0, Ordering::SeqCst),
+            _ => unreachable!(),
+        };
```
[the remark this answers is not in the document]
qinsoon: I added `set_zero()` and `set_zero_atomic()` in `SideMetadataSpec`. It is quite trivial to add it.

src/util/alloc/free_list_allocator.rs
```diff
@@ -0,0 +1,530 @@
...
+        Block::ZERO_BLOCK
+    }
+
+    /// Get a block from the space.
+    fn acquire_fresh_block(&mut self, size: usize, align: usize, stress_test: bool) -> Block {
```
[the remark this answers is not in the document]
qinsoon: The function name is not from the original mimalloc. I changed it to `acquire_global_block()`.

src/util/alloc/free_list_allocator.rs
qinsoon: Done

src/policy/marksweepspace/native_ms/global.rs
qinsoon: Done

src/policy/marksweepspace/native_ms/global.rs
qinsoon: Done

src/util/alloc/mod.rs
```diff
@@ -29,9 +28,15 @@ pub use malloc_allocator::MallocAllocator;
...
+pub use free_list_allocator::FreeListAllocator;
+
 /// Mark compact allocator (actually a bump pointer allocator with an extra heade word)
 mod markcompact_allocator;
 pub use markcompact_allocator::MarkCompactAllocator;
 
 /// Embedded metadata pages
 pub(crate) mod embedded_meta_data;
+
+pub use crate::policy::marksweepspace::malloc_ms::is_alloced_by_malloc;
```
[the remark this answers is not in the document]
qinsoon: Yeah, I removed this line.

vmbindings/dummyvm/src/tests/allocate_align_offset.rs
qinsoon: I added an assertion in the allocator to check if the allocation is within the cell. It seems working fine with these tests.

qinsoon:
@wks I should have addressed your comments. I will work on Wenyu's comments.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,462 @@
...
+    pub fn load_tls(&self) -> VMThread {
+        let tls = Block::TLS_TABLE.load_atomic::<usize>(self.start(), Ordering::SeqCst);
+        VMThread(OpaquePointer::from_address(unsafe {
+            Address::from_usize(tls)
+        }))
+    }
+
+    pub fn is_zero(&self) -> bool {
+        self.start().is_zero()
+    }
```
wks, line 219: Now a `Block` cannot be zero.

wks, line 219: Now a block cannot be zero.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,462 @@
...
+
+    pub fn has_free_cells(&self) -> bool {
+        debug_assert!(!self.is_zero());
```
wks, line 222: And this assertion is no longer necessary.

src/util/alloc/free_list_allocator.rs
```diff
@@ -0,0 +1,530 @@
...
+            unswept_blocks: new_empty_block_lists(),
+            consumed_blocks: new_empty_block_lists(),
+        }
+    }
+
+    // Find a free cell within a given block
+    fn block_alloc(&mut self, block: Block) -> Address {
+        if block.is_zero() {
+            return unsafe { Address::zero() };
+        }
```
wks, line 148: This can be removed because `Block` cannot be zero.
```suggestion
```

wks (changes requested):

src/memory_manager.rs
wks: Yes.  It is OK to "create it first and then do different things with it".  I think eventually VMs will call `mutator.destroy()` and `mutator.flush()` directly and not call top-level functions in `memory_manager`, but we can leave that refactoring to another day.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,462 @@
...
+pub struct Block(NonZeroUsize);
+
+impl From<Address> for Block {
+    #[inline(always)]
+    fn from(address: Address) -> Block {
+        debug_assert!(address.is_aligned_to(Self::BYTES));
+        debug_assert!(!address.is_zero());
+        Self(unsafe { NonZeroUsize::new_unchecked(address.as_usize()) })
+    }
+}
```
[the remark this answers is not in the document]
qinsoon: We expect to turn an aligned address to a region:
https://github.com/mmtk/mmtk-core/blob/06144875903b36b432587eb7462e0272750b71d9/src/util/linear_scan.rs#L79

We do expect it to always succeed, and we do not expect to handle the error case in any circumstance. I feel it is unnecessary to use `TryFrom/Into`, which is designed for 'conversions that may fail in a controlled way under some circumstances'. I don't think that is our case. For us, if the assertion fails, that means we have a bug in the code that we should fix.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,462 @@
...
+use crate::{
+    util::{metadata::side_metadata::SideMetadataSpec, Address, OpaquePointer, VMThread},
+    vm::VMBinding,
+};
+
+use std::num::NonZeroUsize;
+
+/// A 64KB region for MiMalloc.
+/// This is also known as MiMalloc page. We try to avoid getting confused with the OS 4K page. So we call it block.
+/// This is a non-zero block, and `Option<Block>` can still a word size value.
```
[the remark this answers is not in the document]
qinsoon: Done

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,462 @@
...
+// TODO: What if we actually use the first block?
```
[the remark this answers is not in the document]
qinsoon: I think it is unlikely. But just in case, we may have to use `Block(Address)`. Then we can distinguish `Block(Address::ZERO)` and `None`. I changed the comments a little bit.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,462 @@
...
+#[derive(Debug, Clone, Copy, PartialOrd, PartialEq)]
+#[repr(C)]
```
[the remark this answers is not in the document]
qinsoon: Done. Thanks.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,462 @@
...
+            .is_ok()
+    }
+
+    pub fn load_prev_block(&self) -> Option<Block> {
+        let prev = unsafe { Block::PREV_BLOCK_TABLE.load::<usize>(self.start()) };
+        if prev == 0 {
+            None
+        } else {
+            Some(Block(unsafe { NonZeroUsize::new_unchecked(prev) }))
+        }
```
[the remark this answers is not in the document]
qinsoon: Done. I am a bit surprised that you could this. Is `Block` used as a function name here?

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    pub fn load_next_block(&self) -> Option<Block> {
+        let next = unsafe { Block::NEXT_BLOCK_TABLE.load::<usize>(self.start()) };
+        if next == 0 {
+            None
+        } else {
+            Some(Block(unsafe { NonZeroUsize::new_unchecked(next) }))
+        }
```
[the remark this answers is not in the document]
qinsoon: Done

src/policy/marksweepspace/native_ms/block.rs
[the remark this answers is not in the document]
wks: I also have some thoughts of introducing `NonZeroAddress` which wraps `NonZeroUsize` and similar things for `Chunk` and other things.  That's for another refactoring, and I'll write it up in an issue soon.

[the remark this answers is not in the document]
qinsoon: I thought about that when I refactored the code based on your opinion. It may not be a good idea though. In the Rust Immix paper, we wrote this:
```quote
A safer alternative in Rust is to use Option<Address> initialized as None
to indicate that there is no valid value. However, this adds an additional conditional
and a few run-time checks to extract the actual address value in the
performance-critical path of allocation, which adds around 4% performance overhead.
We deem this tradeoff not to be worthwhile given the paramount importance of the allocation fast path...
```

src/policy/marksweepspace/native_ms/block.rs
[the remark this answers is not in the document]
wks: Yes.  The name of a tuple-like struct can be used as a function.

See: https://doc.rust-lang.org/reference/expressions/struct-expr.html#tuple-struct-expression

src/policy/marksweepspace/native_ms/block.rs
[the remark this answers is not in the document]
wks: If the purpose is to use `Block` as a `Region` and that requires `From<Address>`, we can keep this `From<Address>` implementation as is for now.

But I think there is a deeper problem in `Region`, because if a bug-free program only ever converts aligned addresses to `Region`, then `Region::is_aligned` should always return true (an it should be unnecessary).  It is similar to the `Block::is_zero` function and the unnecessary assertions.

I don't fully agree with the use of `From` instead of `TryFrom`.  In the standard library, for example, `i64` implements `From<i32>`, but `i32` doesn't implement `From<i64>`.  If a piece of code converts `i64` to `i32`, but "the i64 here should always fit into i32 unless there is a bug", it still doesn't make it a valid use case of `impl From<i64> for i32`.  It needs to use `try_from` because *the Rust language* cannot protect the programmer from supplying an out-of-range value.

Likewise, in our program, an `Address` can express any address, aligned or not, zero or not.  The Rust language cannot prevent the programmer from converting an unaligned or zero `Address` to `Block`.  I think `Block::try_from(xxx).unwrap()` and `Block::try_from(xxx).expect("message")` also mean "if this fails, it is a bug, and we should fix it".

src/policy/marksweepspace/native_ms/block.rs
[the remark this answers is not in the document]
wks: The unsigned and signed NonZeroXxxx types were only introduced in 1.28.0 and 1.34.0, respectively, in the year 2018 and 2019, respectively.  I think that explains why there was a performance penalty when that paper was published in 2016, because the compiler was not able to optimise for `Option<X>` where X is just an integer.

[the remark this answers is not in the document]
wks: By the way, the penalty may be worse if a function returns `Option<Address>` because that, according to the calling convention, requires the value to be returned on the stack (or in two registers on some architectures), and that could be a real performance hit.

src/policy/marksweepspace/native_ms/block.rs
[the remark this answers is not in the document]
qinsoon: I think the key point of `TryFrom/TryInto` is error handling. We do not need error handling in this case. We could simply hard fail with a panic if the input address is invalid, and fix the code that causes the issue. 

We could require `Region` to implement both: `trait Region: From<Address> + TryFrom<Address>`. But we would always use `Block::from(Block::align(addr))`, rather than `Block::try_from(addr).unwrap_or_else(|| Block::from(Block::align(addr)))`. In that case, why should we implement `TryFrom`?

src/policy/marksweepspace/native_ms/block.rs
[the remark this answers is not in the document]
qinsoon: That could be one part of the reason. Another big reason is that for bump pointer allocator, we initialize `cursor` and `limit` to `Address::ZERO`. We could simply check if `cursor + size < limit` (which works for both initial values and valid values. If we use `Option<Address>` (it doesn't matter `Address` is `usize` or `NonZeroUsize`), we will need an extra check to see if `cursor/limit` is `Some`.

src/policy/marksweepspace/native_ms/block.rs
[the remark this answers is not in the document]
wks: What about adding a method `unsafe fn from_address_unchecked(addr: Address) -> Self` to the `Region` trait, and not require `From<Address>` or `TryFrom<Address>`?  We can document it so that it is a contract that the caller must ensure `addr` is aligned and non-null.  This can distinguish it from `From::from` which "never fails" and `TryFrom::try_from` which "may fail in a controlled way".

src/policy/marksweepspace/native_ms/block_list.rs
```diff
@@ -0,0 +1,315 @@
...
+                .lock
+                .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+                .is_ok();
+        }
+    }
+
+    /// Unlock list
+    pub fn unlock(&mut self) {
+        self.lock.store(false, Ordering::SeqCst);
+    }
```
wks, line 148: I think this function needs more comment, and probably renamed to reflect its true purpose.  It literally took me hours to figure out what this lock is for.  I looked at where each of the methods (`push`, `pop`, `append`, `remove`) are called, but most of the time they are only mutating private data structures of a mutator/allocator, and don't need the `BlockList` to be locked.

```suggestion
    /// Lock the list for parallel sweeping.
    ///
    /// During the Release phase, multiple GC worker threads can sweep chunks and release mutators
    /// at the same time, and the same `BlockList` can be reached by traversing blocks in a chunk,
    /// and also by traversing blocks held by a mutator.  This lock is necessary to prevent
    /// multiple GC workers from mutating the same `BlockList` instance.
    ///
    /// This lock should only be used for that purpose.  There should be no race outside the
    /// Release phase.
    pub fn lock_for_sweep(&mut self) {
        let mut success = false;
        while !success {
            success = self
                .sweep_lock
                .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
                .is_ok();
        }
    }

    /// Unlock the list for parallel sweeping.
    pub fn unlock_for_sweep(&mut self) {
        self.sweep_lock.store(false, Ordering::SeqCst);
    }
```

src/policy/marksweepspace/native_ms/block.rs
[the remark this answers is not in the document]
qinsoon: Yeah. That sounds reasonable. Rust says `From` should never fail, and that means we should not have `From<Address>` for `Region` at all.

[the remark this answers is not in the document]
qinsoon: I submitted a PR for this: https://github.com/mmtk/mmtk-core/pull/695.

src/policy/marksweepspace/native_ms/block_list.rs
qinsoon: I added comments as suggested. But I didn't change the method name.

vmbindings/dummyvm/src/api.rs
qinsoon: Yes. Thanks for pointing this out. I will add a line after `destroy_mutator()` to turn the pointer back to a box, and let Rust to reclaim it.

src/util/metadata/mark_bit.rs
qinsoon: Sounds good to me. I was avoiding using `mark`, as we may use cyclic mark in the future. In that case, setting to `1` does not mean marking. But we could leave that to future refactoring.

src/util/linear_scan.rs
qinsoon: Done.

src/util/alloc_bit.rs
qinsoon: Yes. I will remove it.

qinsoon:
@wenyuzhao I have addressed most of your comments, except for those related to the object reference. I plan to fix them in a separate PR (which should get merged before this PR). We can come back to those comments later.

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -0,0 +1,366 @@
...
+                // Block is live. Increment the allocated block count.
+                allocated_blocks += 1;
+            }
+        }
+        // Set this chunk as free if there is not live blocks.
+        if allocated_blocks == 0 {
+            self.space.chunk_map.set(self.chunk, ChunkState::Free)
+        }
+    }
+}
```
wks, line 366: I have some radical thoughts about this.  Maybe this `SweepChunk` work packet is completely unnecessary.

If each `BlockList` is owned by exactly one mutator (or the global `MarkSweepSpace`), and each `Block` is owned by exactly one `BlockList`, then we can traverse all existing blocks simply by traversing all `BlockList` instances.  That means visiting all `BlockList` in mutators and the lists in `MarkSweepSpace::abandoned_lists`.  Doing so allows blocks to be swept while the current thread has exclusive access to its `BlockList`, therefore removing the need to have `lock()` in `BlockList` completely.

But I prefer doing this under the protection of Rust's ownership model.  From the current code, it looks like there is no way for `Block` instances to be created anywhere else (other than `MarkSweepSpace::acquire_block`), held anywhere else, or discarded before they are swept.  Rust's ownership model can enforce that, but currently our `Block` type implements `Copy`, has pointer semantics, and bypasses the ownership model.  I elaborated my thoughts here: https://github.com/mmtk/mmtk-core/issues/696.

src/util/alloc/immix_allocator.rs
qinsoon: This code is removed in https://github.com/mmtk/mmtk-core/pull/699.

src/util/alloc/object_ref_guard.rs
qinsoon: This is removed in https://github.com/mmtk/mmtk-core/pull/699.

src/util/alloc/large_object_allocator.rs
qinsoon: This is removed in https://github.com/mmtk/mmtk-core/pull/699.

qinsoon:
I believe I have addressed the issues you pointed out. Can you review the PR again and let me know if there are any further changes required? @wks @wenyuzhao

src/util/address.rs
```diff
@@ -501,7 +501,9 @@ impl ObjectReference {
     #[inline(always)]
     pub fn to_address<VM: VMBinding>(self) -> Address {
         use crate::vm::ObjectModel;
-        VM::VMObjectModel::ref_to_address(self)
+        let to_address = VM::VMObjectModel::ref_to_address(self);
+        debug_assert!(!VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS || to_address == self.to_raw_address(), "The binding claims unified object reference address, but for object reference {}, ref_to_address() returns {}", self, to_address);
```
wks, line 505: Hmmm... I think `ObjectReference::to_address` should be a synonym of `VM::VMObjectModel::ref_to_address`.  But since it is the VM that implements `VM::VMObjectModel::ref_to_address`, we can't offload this assertion to the binding.  This makes `ObjectReference::to_address` execute one extra assertion (when debugging, of course), and it feels odd.

I don't know if any kind of code instrumenting tool (aspect-oriented programming, design by contract, post conditions, etc.) can help inserting this assertion after every implementation of `VM::VMObjectModel::ref_to_address`.  If there is no such tool, I am OK with this.

src/vm/object_model.rs
```diff
@@ -379,6 +379,24 @@ pub trait ObjectModel<VM: VMBinding> {
     /// mature space for generational plans.
     const VM_WORST_CASE_COPY_EXPANSION: f64 = 1.5;
 
+    /// If this is true, the binding guarantees that an object reference's raw address equals to the return value of the ref_to_address method,
+    /// and equals to the return value of the ref_to_object_start method. This is a very strong guarantee, but it is also helpful for MMTk to
```
wks, line 383: It is "A equals B" or "A is equal to B".

```suggestion
    /// If this is true, the binding guarantees that an object reference's raw address is always equal to the return value of the `ref_to_address` method
    /// and the return value of the `ref_to_object_start` method. This is a very strong guarantee, but it is also helpful for MMTk to
```

wks (changes requested):
Mostly OK. Only some small problems remain.

src/util/address.rs
qinsoon: Thanks for pointing this out. It is a general goal for us to assert that a binding implements those traits correctly (see https://github.com/mmtk/mmtk-core/issues/480). However, I don't know if there is a better way to achieve this, other than just asserting at where we use those traits. If there is any good instrumentation tool for Rust, it is worth a try, as it can help us solve this general issue.

However, I wouldn't think that (debug) assertions change the semantics or the behavior of the method. They just help ensure some invariants (which are already part of the semantics) are not broken. Thus I don't think adding assertions changes the fact that `ObjectReference::to_address` is a synonym of `VM::VMObjectModel::ref_to_address`.

src/vm/object_model.rs
qinsoon: Done.

wks (approved):

wks:
It looks good to me now.

wenyuzhao (approved):

wks:
I managed to get the mmtk-ruby binding working.  However, if I switch from malloc MarkSweep to native MatkSweep, it will crash in `post_alloc` when allocating new objects after GC because the VO-bit is already set.  (FYI, Ruby uses the "is_mmtk_object" feature which turns on the "global_alloc_bit" feature.)

I am still investigating.

wks:
I think the implementation of `bzero_metadata` is unsound when we use it to clear a bit range, not whole pages.  Log shows that sometimes `bzero_metadata` will clear 0 bytes when clearing the VO-bits for a 56-byte cell.

```
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff1, end: 0xc0804015ff2, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fca8
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fcb0, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff2, end: 0xc0804015ff3, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fce0
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fce8, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff3, end: 0xc0804015ff4, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fd18
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fd20, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff4, end: 0xc0804015ff5, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fd50
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fd58, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff5, end: 0xc0804015ff6, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fd88
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fd90, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff6, end: 0xc0804015ff7, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fdc0
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fdc8, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff7, end: 0xc0804015ff8, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fdf8
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fe00, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff8, end: 0xc0804015ff8, len: 0
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fe30
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fe38, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff8, end: 0xc0804015ff9, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fe68
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fe70, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ff9, end: 0xc0804015ffa, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fea0
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fea8, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ffa, end: 0xc0804015ffb, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057fed8
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057fee0, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ffb, end: 0xc0804015ffc, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057ff10
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057ff18, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ffc, end: 0xc0804015ffd, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057ff48
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057ff50, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ffd, end: 0xc0804015ffe, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057ff80
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057ff88, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015ffe, end: 0xc0804015fff, len: 1
[2022-11-30T09:25:55Z INFO  mmtk::policy::marksweepspace::native_ms::block] clearing alloc bits. cell: 0x2010057ffb8
[2022-11-30T09:25:55Z INFO  mmtk::util::alloc_bit] bzero_alloc_bit start: 0x2010057ffc0, size: 56
[2022-11-30T09:25:55Z INFO  mmtk::util::metadata::side_metadata::global] zeroing: start: 0xc0804015fff, end: 0xc0804015fff, len: 0
```

`bzero_metadata` uses `memset` to set **byte ranges**.  The granularity of VO-bit metadata is one *bit* per 8 bytes.  Then a 56-byte cell corresponds to 7 bits in the VO-bit metadata.  As a result, if we use `bzero_metadata` to clear VO-bits when doing `naive_brute_force_sweep`, for every 8 consecutive cells, 7 of them will have the start of the VO-bits in one byte, and the end of the VO-bits in another, while 1 of them have both the start and the end of the VO-bits in the same byte.  That explains the log.

(FYI, for Ruby, `obj.to_raw_address() == ObjectModel::ref_to_address(obj) == ObjectModel::ref_to_object_start(obj) + 8`)

I think it is unsound to clear 0 or 1 bytes.  It should only clear the bit range the object occupies, which may span multiple bytes.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,450 @@
...
+                if cursor >= cell + cell_size {
+                    // We now stepped to the next cell. This means we did not find mark bit in the current cell, and we can add this cell to free list.
+                    debug!(
+                        "{:?} Free cell: {}, last cell in freelist is {}",
+                        self, cell, last
+                    );
+
+                    // Clear alloc bit: we don't know where the object reference actually is, so we bulk zero the possible area.
+                    #[cfg(feature = "global_alloc_bit")]
+                    crate::util::alloc_bit::ALLOC_SIDE_METADATA_SPEC.bzero_metadata(
```
wks, line 370: The current implementation of `bzero_metadata` can only clear whole bytes in the metadata, but the VO-bit range of a cell may not fill whole bytes, and may span multiple bytes.  This will cause this `bzero_metadata` call to zero more or less bits than actually needed.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,450 @@
...
+                        cell + VM::VMObjectModel::OBJECT_REF_OFFSET_LOWER_BOUND,
+                        cell_size,
```
wks, line 372: If we know that an object cannot span multiple cells, we don't need to clear this many bytes.  We only need to clear the metadata bits for all bytes from `cell + VM::VMObjectModel::OBJECT_REF_OFFSET_LOWER_BOUND` to the end of this cell.

src/policy/marksweepspace/native_ms/block.rs
wks: And it is preferrable to call `crate::util::alloc_bit::bzero_alloc_bit` to ensure all accesses to the VO-bit metadata are encapsulated.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,450 @@
...
+        let mut cell = self.start();
+        let mut last = unsafe { Address::zero() };
+        while cell + cell_size <= self.start() + Block::BYTES {
+            // We know the cell and the object reference is the same.
+            if !VM::VMObjectModel::LOCAL_MARK_BIT_SPEC
+                .is_marked::<VM>(ObjectReference::from_raw_address(cell), Ordering::SeqCst)
+            {
+                // clear alloc bit if it is ever set.
+                #[cfg(feature = "global_alloc_bit")]
+                crate::util::alloc_bit::ALLOC_SIDE_METADATA_SPEC.store_atomic::<u8>(
```
wks, line 301: It is preferable to call `crate::util::alloc_bit::bzero_alloc_bit` to ensure all accesses to the VO-bit metadata are encapsulated.

src/policy/marksweepspace/native_ms/block.rs
```diff
@@ -0,0 +1,450 @@
...
+
+                if cursor >= cell + cell_size {
+                    // We now stepped to the next cell. This means we did not find mark bit in the current cell, and we can add this cell to free list.
+                    debug!(
+                        "{:?} Free cell: {}, last cell in freelist is {}",
+                        self, cell, last
+                    );
+
+                    // Clear alloc bit: we don't know where the object reference actually is, so we bulk zero the possible area.
+                    #[cfg(feature = "global_alloc_bit")]
```
wks, line 369: If native MarkSweep relies on the VO-bit to identify allocated cells, we should clear the VO-bit regardless of whether the user enabled the `global_alloc_bit` feature.

```suggestion
```

wks (changes requested):
There is a problem related to clearing the VO-bit metadata.  It may affect JikesRVM as well.

src/policy/marksweepspace/native_ms/block.rs
qinsoon: Feels like this is a bug in `bzero_metadata`. I can fix it in a separate PR. Is there a way to work around the bug in this PR?

src/policy/marksweepspace/native_ms/block.rs
qinsoon: An object will not span multiple cells. But I don't quite understand what you said. This code uses bulk zero to clear one cell, as we do not know where the object is, and where the alloc bit is set.

src/policy/marksweepspace/native_ms/block.rs
wks: One possible workaround is not using `bzero_metadata`. Instead, use `crate::util::alloc_bit::ALLOC_SIDE_METADATA_SPEC.set_zero` to clear one bit at a time.  The granularity of `ALLOC_SIDE_METADATA_SPEC` is one bit per word (8 bytes).  For the largest size class in mimalloc, it is 8192 bits for 8192 words.

src/policy/marksweepspace/native_ms/block.rs
wks: For example, if the cell size is 128 bytes, and `OBJECT_REF_OFFSET_LOWER_BOUND` is equal to 24, then we only need to clear the metadata bits for the interval of `[cell + 24, cell + 128)` instead of `[cell + 24, cell + 24 + 128)`.  The VO-bit cannot be set in the range of `[cell + 128, cell + 128 + 24)` because that is in the next cell.

Concretely,
```
crate::util::alloc_bit::ALLOC_SIDE_METADATA_SPEC.bzero_metadata(
                        cell + VM::VMObjectModel::OBJECT_REF_OFFSET_LOWER_BOUND,
                        cell_size - VM::VMObjectModel::OBJECT_REF_OFFSET_LOWER_BOUND,
```

src/policy/marksweepspace/native_ms/block.rs
qinsoon: I submitted a PR to fix the issue: https://github.com/mmtk/mmtk-core/pull/707

qinsoon:
binding-refs
OPENJDK_BINDING_REF=mimalloc-ms-support
JIKESRVM_BINDING_REF=mimalloc-ms-support
V8_BINDING_REF=update-pr-643

src/policy/marksweepspace/native_ms/block.rs
qinsoon: I also changed this to use `bzero_alloc_bit`.

src/policy/marksweepspace/native_ms/block.rs
qinsoon: I changed to zero from `cell` for `cell_size` bytes. As we now guarantee we mark object metadata within its allocation (i.e. a cell here), we do not really need `OBJECT_REF_OFFSET_LOWER_BOUND`.

src/policy/marksweepspace/native_ms/block.rs
qinsoon: Done

src/policy/marksweepspace/native_ms/block.rs
qinsoon: Native mark sweep does not use VO-bit in the algorithm. We will have to keep this `cfg`.

wks (approved):
LGTM

Issue #612, opened 2022-06-21
Barrier fastpath in Rust


qinsoon:
We should provide a reference implementation of barriers' fastpath in Rust, just like what we have for the allocators. They may be directly called by a binding if a binding does not want to implement a fastpath. They also serve as a reference implementation if a binding wants to implement the fastpath in the binding side.

PR #644, opened 2022-08-18
Add a full impl in Rust for write barrier

qinsoon:
This pull request adds a write barrier fastpath implementation in Rust, and adds a barrier API. It closes https://github.com/mmtk/mmtk-core/issues/612.

Changes:
* rename the current `Barrier::post_write_barrier()` to `Barrier::post_write_barrier_slow()`.
* add `Barrier::post_write_barrier()` which provides a full implementation of write barrier.
* add `memory_manager::post_write_barrier()`.
* rename `WriteTarget` to `BarrierWriteTarget` as it will be exposed.
* remove `MutatorContext::record_modified_node()`.
* modify the slowpath in `ObjectRememberingBarrier` -- it attempts to do CAS first, and only check the bit after the CAS fails.

Related PRs:
* https://github.com/mmtk/mmtk-openjdk/pull/176

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -122,6 +122,7 @@ pub fn flush_mutator<VM: VMBinding>(mutator: &mut Mutator<VM>) {
 /// * `align`: Required alignment for the object.
 /// * `offset`: Offset associated with the alignment.
 /// * `semantics`: The allocation semantic required for the allocation.
+#[inline(always)]
 pub fn alloc<VM: VMBinding>(
     mutator: &mut Mutator<VM>,
     size: usize,
@@ -149,6 +150,7 @@ pub fn alloc<VM: VMBinding>(
 /// * `refer`: The newly allocated object.
 /// * `bytes`: The size of the space allocated for the object (in bytes).
 /// * `semantics`: The allocation semantics used for the allocation.
+#[inline(always)]
 pub fn post_alloc<VM: VMBinding>(
     mutator: &mut Mutator<VM>,
     refer: ObjectReference,
@@ -167,6 +169,7 @@ pub fn post_alloc<VM: VMBinding>(
 /// Arguments:
 /// * `mutator`: The mutator for the current thread.
 /// * `target`: The target for the write operation.
+#[inline(always)]
 pub fn post_write_barrier<VM: VMBinding>(mutator: &mut Mutator<VM>, target: BarrierWriteTarget) {
     mutator.barrier().post_write_barrier(target)
 }
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -167,10 +167,7 @@ pub fn post_alloc<VM: VMBinding>(
 /// Arguments:
 /// * `mutator`: The mutator for the current thread.
 /// * `target`: The target for the write operation.
-pub fn post_write_barrier<VM: VMBinding>(
-    mutator: &mut Mutator<VM>,
-    target: BarrierWriteTarget,
-) {
+pub fn post_write_barrier<VM: VMBinding>(mutator: &mut Mutator<VM>, target: BarrierWriteTarget) {
     mutator.barrier().post_write_barrier(target)
 }
 
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -14,6 +14,7 @@
 use crate::mmtk::MMTKBuilder;
 use crate::mmtk::MMTK;
 use crate::plan::AllocationSemantics;
+use crate::plan::BarrierWriteTarget;
 use crate::plan::{Mutator, MutatorContext};
 use crate::scheduler::WorkBucketStage;
 use crate::scheduler::{GCController, GCWork, GCWorker};
@@ -155,6 +156,22 @@ pub fn post_alloc<VM: VMBinding>(
     semantics: AllocationSemantics,
 ) {
     mutator.post_alloc(refer, bytes, semantics);
+}
+
+/// The write barrier by MMTk. This is a *post* write barrier, which we expect a binding to call
+/// *after* they modify an object. For performance reasons, a VM should implement the write barrier
+/// fast-path on their side rather than just calling this function.
+///
+/// TODO: We plan to replace this API with a subsuming barrier API.
+///
+/// Arguments:
+/// * `mutator`: The mutator for the current thread.
+/// * `target`: The target for the write operation.
+pub fn post_write_barrier<VM: VMBinding>(
+    mutator: &mut Mutator<VM>,
+    target: BarrierWriteTarget,
+) {
+    mutator.barrier().post_write_barrier(target)
 }
 
 /// Return an AllocatorSelector for the given allocation semantic. This method is provided
--- a/src/plan/barriers.rs
+++ b/src/plan/barriers.rs
@@ -24,21 +24,23 @@ impl BarrierSelector {
 }
 
 /// For field writes in HotSpot, we cannot always get the source object pointer and the field address
-pub enum WriteTarget {
+pub enum BarrierWriteTarget {
     Object(ObjectReference),
     Slot(Address),
 }
 
 pub trait Barrier: 'static + Send {
     fn flush(&mut self);
-    fn post_write_barrier(&mut self, target: WriteTarget);
+    fn post_write_barrier(&mut self, target: BarrierWriteTarget);
+    fn post_write_barrier_slow(&mut self, target: BarrierWriteTarget);
 }
 
 pub struct NoBarrier;
 
 impl Barrier for NoBarrier {
     fn flush(&mut self) {}
-    fn post_write_barrier(&mut self, _target: WriteTarget) {}
+    fn post_write_barrier(&mut self, _target: BarrierWriteTarget) {}
+    fn post_write_barrier_slow(&mut self, _target: BarrierWriteTarget) {}
 }
 
 pub struct ObjectRememberingBarrier<E: ProcessEdgesWork> {
@@ -64,11 +66,9 @@ impl<E: ProcessEdgesWork> ObjectRememberingBarrier<E> {
     #[inline(always)]
     fn log_object(&self, object: ObjectReference) -> bool {
         loop {
-            let old_value =
-                load_metadata::<E::VM>(&self.meta, object, None, Some(Ordering::SeqCst));
-            if old_value == 0 {
-                return false;
-            }
+            // Try set the bit from 1 to 0 (log object). This may fail, if
+            // 1. the bit is cleared by others, or
+            // 2. other bits in the same byte may get modified if we use side metadata
             if compare_exchange_metadata::<E::VM>(
                 &self.meta,
                 object,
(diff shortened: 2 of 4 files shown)


src/plan/barriers.rs
```diff
@@ -112,9 +133,17 @@ impl<E: ProcessEdgesWork> Barrier for ObjectRememberingBarrier<E> {
...
+            _ => unreachable!(),
+        }
+    }
+
+    #[inline(always)]
+    fn post_write_barrier_slow(&mut self, target: BarrierWriteTarget) {
         match target {
-            WriteTarget::Object(obj) => {
+            BarrierWriteTarget::Object(obj) => {
                 self.enqueue_node(obj);
```
wks, line 147: Should we just call `self.barrier_slow(obj)` to reduce code duplication?  The only difference is inlining.
```suggestion
                self.barrier_slow(obj);
```

wks (changes requested):
Only a small problem about code duplication.

src/plan/barriers.rs
qinsoon: Right, the difference is inlining. I was trying to achieve these: 1. make it clear in the code that for the write barrier implementation, which is fastpath (inlined), and which is slowpath (never inlined), and 2. make sure that we do not unnecessarily make another call in Rust when a binding calls the slowpath (from JIT'd code into Rust). Having one `inline(always)` method, and another `inline(never)` method that calls the inlined method can achieve this.

src/memory_manager.rs
```diff
@@ -157,6 +158,19 @@ pub fn post_alloc<VM: VMBinding>(
...
+/// The write barrier by MMTk. This is a *post* write barrier, which we expect a binding to call
+/// *after* they modify an object. For performance reasons, a VM should implement the write barrier
+/// fast-path on their side rather than just calling this function.
+///
+/// TODO: We plan to replace this API with a subsuming barrier API.
+///
+/// Arguments:
+/// * `mutator`: The mutator for the current thread.
+/// * `target`: The target for the write operation.
+pub fn post_write_barrier<VM: VMBinding>(mutator: &mut Mutator<VM>, target: BarrierWriteTarget) {
```
wks, line 170: Should we expose both `post_writer_barrier` and `post_write_barrier_slow` in `memory_manager`, or neither of them?  The OpenJDK binding defined the `extern "C"` version of both, but they call the methods on `mutator.barriers()` directly, bypassing `memory_manager::post_write_barrier`.

src/plan/barriers.rs
wks: Yes.  That makes sense if the intention is better documentation.

To reduce a level of function call (if that's important), however, we may label all those functions as `#[inline(always)]`, including `memory_manager::post_write_barrier` (we may or may not want to add `memory_manager::post_write_barrier`. See the other comment near  `memory_manager::post_write_barrier`). The only functions that are not inlined are the `extern "C"` functions defined in the bindings.

qinsoon: There are some differences here: 1. `post_write_barrier` is the full barrier implementation in Rust. For an efficient binding, they probably won't call it anyway. But `post_write_barrier_slow` is the slowpath, a binding that aims performance will call it. 2. With no inline directive, a compiler may choose to inline the function. I was trying to avoid `inline(never)`, for which case a compiler is explicitly told not to inline.

But anyway, I can add `#[inline(always)]` as you suggested.

src/memory_manager.rs
qinsoon: I thought about this. I am not quite sure. In my view, `memory_manager` is a collection of common APIs that are useful, more friendly and easy to find for the bindings. For most methods in `memory_manager`, there are some 'inner' methods in other modules that can be used by a binding and does the same thing or almost the same thing. 

So for the arguments above, we should include `post_write_barrier()` in `memory_manager`. We do expect a binding to use it and be able to find it easily. As for `post_write_barrier_slow()`, I am not quite sure. What I was thinking about is that if a binding would like to call the slowpath, they will have to look at the barrier implementation, implement the fastpath and they will find the slowpath method anyway. So it is not necessary to include it in `memory_manager`. Besides, `post_write_barrier_slow` does not have complete semantics of what it does -- it is just a part of a write barrier. Putting it to `memory_manager` and exposing it to every user may be confusing.

This is just what I was thinking about. Let me know your suggestion.

src/plan/barriers.rs
```diff
@@ -78,7 +78,15 @@ impl<E: ProcessEdgesWork> ObjectRememberingBarrier<E> {
                 Ordering::SeqCst,
                 Ordering::SeqCst,
             ) {
+                // We just logged the object
                 return true;
+            } else {
+                let old_value =
+                    load_metadata::<E::VM>(&self.meta, object, None, Some(Ordering::SeqCst));
```
wenyuzhao, line 85: Doing a test when CAS fails here does not look like a normal way people would write CAS loops. For example this is how jikesrvm does the compare and swap: https://github.com/JikesRVM/JikesRVM/blob/5072f19761/MMTk/src/org/mmtk/policy/LargeObjectSpace.java#L242-L247

I think this is due to the API design problem of the side metadata. `compare_exchange_metadata` (L72 above) operates on bits not bytes, and the inner implementation already performs an extra load. But after the load, it does not check if the loaded value is already in the correct state:
https://github.com/mmtk/mmtk-core/blob/42262c296e164b5b1d05103e527ce5a9aee7cfd2/src/util/metadata/side_metadata/global.rs#L471-L484

This is the fundamental problem to force us to do an extra load here. With this problem, either the old approach (do the load and test before CAS) or the new approach (do the load after CAS) will always cause extra loads. And also notice that the code for metadata address computation will be repeated as well, because for this case `self.meta` is not a compile-time constant, and the metadata code can hardly be inlined and optimized.

The ideal solution would be either (1) removing the load in  `compare_exchange_metadata`, or (2) doing the check inside `compare_exchange_metadata`. Alternatively, (3) we can add a rust-std-style [`fetch_update`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_update) interface to simplify the barrier CAS loop:

```rust
side_metadata::fetch_update(&self.meta, Ordering::SeqCst, , Ordering::SeqCst, |old_bit: usize| {
    if old_bit == 0 {
        None
    } else {
        Some(1)
    }
});
```

This `fetch_update` function will also ensure that the metadata address will only be compiled once, regardless of how the compiler does its optimization.

I actually have the 2nd and the 3rd option implemented in my lxr branch:
[1] https://github.com/wenyuzhao/mmtk-core/blob/lxr-merge/src/util/metadata/side_metadata/global.rs#L590-L592
[2] https://github.com/wenyuzhao/mmtk-core/blob/lxr-merge/src/util/metadata/side_metadata/global.rs#L734-L803

qinsoon: Thanks for the comments. Both 2 and 3 look good. I would suggest that I will do that change after this pull request. Also I found that for `bits_num_log == 6`, our side metadata uses `AtomicUsize` rather than `AtomicU64` -- that should be a bug. I will do a PR to fix both issues.

Can you let me know for this PR, should I revert this change, or just keep it? My intention is to turn the unconditional load into a conditional load (only load after CAS fails).

src/memory_manager.rs
wks: Yes. That makes sense.  Providing a binding for the full barrier `post_write_barrier` makes it easier for VM to begin with (not using fast paths), but the semantics of `post_write_barrier_slow` is specific to each barrier.  What's more, the JIT compiler in the VM probably want to generate a function call that calls the slow path of the specific barrier, and probably they can create their `extern "C"` function specific to each barrier.  I am not sure about this, but it's a possibility.

So I think it is OK to leave it like this.

src/plan/barriers.rs
wks: I prefer 3.  Because the metadata may be finer than a byte, any attempt to provide a plain `compare_exchange_metadata` API inevitably have to involve an extra load inside the function, so as not to modify surrounding bits, unless the function also takes the old byte as the parameter, which is ugly.  Designing a `fetch_update` API will also allow us to use the underlying `AtomicU8.fetch_update` in the Rust standard library.  In theory, this allows the compiler to implement it with load-link store-conditional (LL-SC) instead of compare-exchange on some architectures (almost anything except x86). I don't know if the Rust compiler already does this, but it is possible.

Another alternative is to provide `fetch_and_metadata`, `fetch_or_metadata`, `fetch_xor_metadata`.  Some architectures provide special instructions for atomic bit operations that never fails, and therefore never need to be used in a loop.  In Rust, there is [`AtomicU8.fetch_and`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU8.html#method.fetch_and), and such methods are [available in the `atomic_traits` crate](https://docs.rs/atomic-traits/latest/atomic_traits/fetch/trait.And.html), too.

wenyuzhao: Yes it makes sense to do metadata API refactoring in another PR.

> Can you let me know for this PR, should I revert this change, or just keep it?

Please keep it. This looks more correct. To me, it doesn't make sense for the old implementation to load the bit once for a quick test and load it again for CAS.

src/plan/barriers.rs
qinsoon: I added `#[inline(always)]` to `post_write_barrier()` in `memory_manager`. To be consistent, I also added `#[inline(always)]` to `alloc` and `post_alloc` in `memory_manager`.

src/plan/barriers.rs
qinsoon: I actually attempted 3. The `metadata` field in the object remembering barrier is actually `Metadata` rather than `SideMetadata` (because the unlogged bit could be in the object header). So it means if we add `fetch_update`, we need to add it to the `ObjectModel` trait and allow bindings to implement it for in-header metadata. For that reason, I thought it was better to put it to a separate pull request.

I will submit a pull request for that.

wks: I tried compiling `fetch_and` and `fetch_update` to AArch64 and RISCV64.  `fetch_and` always results in simpler code without loop.  `fetch_update` is implemented as a CAS loop on AArch64, but implemented as an LR/SC loop on RISCV64.

So the message is very clear.  If we only do bit-set and bit-clear, use `fetch_or` and `fetch_and`;  Otherwise, prefer `fetch_update` over a hand-written CAS loop.

src/plan/barriers.rs
wks: What about the code duplication here? Should we call `self.barrier_slow(obj);` or keep it like this?

qinsoon: `barrier_slow()` is forced out of line. If we call `barrier_slow()`, there will be always an extra call when the binding calls the slowpath.

wks: One solution is to let both `barrier_slow` and `post_write_barrier_slow` call a common function, such as `barrier_slow_impl`, which has `#[inline(always)]`.  But if the post write barrier itself is going to be replaced by a subsuming barrier, I am okay with leaving it as is.

qinsoon: > One solution is to let both `barrier_slow` and `post_write_barrier_slow` call a common function, such as `barrier_slow_impl`, which has `#[inline(always)]`.

`enqueue_node()` is the common function with `#[inline(always)]`.

wks: OK.  That makes sense, at least for the object remembering barrier.

wks (approved):

wks:
I don't have further objections.

qinsoon:
binding-refs
OPENJDK_BINDING_REF=update-pr-644

PR #645, opened 2022-08-22
CopySpace should not clear alloc bit if there is no allocation into the space

qinsoon:
This PR fixes a bug in `CopySpace::release()`, that we should not clear alloc bit if we haven't allocated into the space yet. This bug can be reproduced in this scenario: use `GenCopy` and force every GC to be full heap. In this case, the first GC is a full heap GC, and we copy objects from nursery to `tospace`. At the end of the GC, `fromspace` is empty, and `CopySpace::release()` will be called on the empty space. As we never allocated in this `fromspace`, its side metadata is unmapped, and this bzero call will cause segfault.


wks (approved):
LGTM

PR #646, opened 2022-08-23
Move to Ubuntu-22.04 in CI tests

qinsoon:


caizixian (approved):
LGTM. The binding tests are probably still broken but we will fix them in separate PRs.

Issue #313, opened 2021-05-13
Duplicate edges in Marksweep


javadamiri:
I was trying to test PR #310 (mainly `src/util/side_metadata/sanity.rs`) with marksweep when I detected random panics due to duplicate edges in both JikesRVM and OpenJDK bindings.
This is the assertion which fails:

https://github.com/mmtk/mmtk-core/blob/1fc73b360e128e3ce072538f213b4013cae91b8b/src/util/edge_logger.rs#L26

Note:
After solving this issue, we need to add CI checks for the side metadata sanity checker (under `extreme_assertions` and with marksweep plan)

javadamiri:
I found a source of duplicate edges in marksweep. In the following method in `MallocSpace`:
```
pub fn trace_object<T: TransitiveClosure>(
    &self,
    trace: &mut T,
    object: ObjectReference,
) -> ObjectReference {
    if object.is_null() {
        return object;
    }
    let address = object.to_address();
    assert!(
        self.in_space(object),
        "Cannot mark an object {} that was not alloced by malloc.",
        address,
    );
    if !is_marked::<VM>(object) {
        set_mark_bit::<VM>(object);
        trace.process_node(object);
    }
    object
}
```
multiple workers may run `if !is_marked::<VM>(object) {` and stop just before the `set_mark_bit::<VM>(object);` call.
In that case, all of them will mark the object and call `process_node` which will ultimately create duplicate edges.

To avoid this, we should replace the `if !is_marked` with something like the following pseudo-code:
```
loop {
    if is_marked(object) {
        break;
    } else if cmp_exchange_mark_bit(object, old: 0, new: 1) {
        trace.process_node(object);
        break;
    }
}
```
However, I'm not sure this is really what we want here.

In the current implementation, I see no reason for using atomic operations in `is_marked` and `set_mark_bit`.
Changing these to none atomic operations will still produce duplicate edges, but I don't think it can lead to concurrency bugs, because it only means we may mark some objects multiple times.

In conclusion, it seems to me that duplicate edges are probably Ok here and we can even change the atomic ops to none-atomic ones.
However, I'm not 100% sure about this.
@qinsoon @steveblackburn , what do you think?

steveblackburn:
Duplicate edges are fine in a normal non-copying tracing mark.

[...]

PR #647, opened 2022-08-23
Typed access to metadata

qinsoon:
This PR refactors metadata, and makes all its access associated with a num type. This PR also greatly reduces code duplication in the metadata module, and adds tests for both header metadata and side metadata access.

Metadata in the following context refers to both side metadata and header metadata.
* All metadata access functions are moved to `HeaderMetadata`/`SideMetadata` as methods.
* Introduce the `MetadataValue` trait so we can handle `u8/16/32/64/size` uniformly.
* All metadata access methods needs a number type for the get/set value type. It needs to match what is defined in the spec, otherwise it will panic.
* Fixe the issue for `u64` metadata on 32 bits. It should work properly now.
* Add unit tests.
* Header metadata methods in `ObjectModel` now have a default implementation that uses our header metadata module.

Related PRs:
* https://github.com/mmtk/mmtk-openjdk/pull/178
* https://github.com/mmtk/mmtk-jikesrvm/pull/122
* https://github.com/mmtk/mmtk-v8/pull/67


src/util/metadata/mod.rs
```diff
@@ -225,3 +225,17 @@ pub mod side_metadata;
 pub(crate) mod log_bit;
 
 pub use global::*;
+
+// We currently only support metadata that has a length equal to or smaller than the architecture's word size.
+// That means we do not support metadata of more than 8 bytes, and we do not support side metadata of 8 bytes on 32 bits architectures.
+// This macro is used to check word size. It makes sure that the given block is only executed on 64 bits architectures, and panics on other architectures.
+macro_rules! only_available_on_64bits {
+    ($b: block) => {
+        if cfg!(target_pointer_width = "64")
```
wks, line 234: Actually `AtomicU64` may be available even if the pointer width is 32 bits.  For example, on ARMv7, the `LDRD` and `STRD` instructions can load/store a 64-bit aligned region atomically.  `LDREXD` paired with `STREXD` can implement atomic update (including compare-exchange( at 64-bit granularity.

I suggest we do not add `only_available_on_64bits` now.  We assume `AtomicU64` is always available.  By doing this, when we ever compile mmtk-core on a target that doesn't support `AtomicU64`, we will get a compilation error, and we can fix that later.

A more precise feature testing macro is `target_has_atomic_load_store` for load and store, and `target_has_atomic` for atomic read-modify-write operations.  At this moment, both of them are unstable features, but may be stablized soon.

wks (changes requested):

src/util/metadata/header_metadata.rs
```diff
@@ -6,16 +6,16 @@ use std::sync::atomic::AtomicUsize;
...
+use crate::util::metadata::only_available_on_64bits;
 use crate::util::ObjectReference;
 
 const LOG_BITS_IN_U16: usize = 4;
-const BIT_IN_U16: usize = 1 << LOG_BITS_IN_U16;
+const BITS_IN_U16: usize = 1 << LOG_BITS_IN_U16;
 const LOG_BITS_IN_U32: usize = 5;
 const BITS_IN_U32: usize = 1 << LOG_BITS_IN_U32;
+const LOG_BITS_IN_U64: usize = 6;
+const BITS_IN_U64: usize = 1 << LOG_BITS_IN_U64;
```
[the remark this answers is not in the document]
qinsoon: Maybe it doesn't matter too much. We won't get these constants wrong, and they won't change.

src/util/metadata/header_metadata.rs
```diff
@@ -52,7 +52,7 @@ pub fn load_metadata(
     debug_assert!(optional_mask.is_none() || metadata_spec.num_of_bits >= 8,"optional_mask is only supported for 8X-bits in-header metadata. Problematic MetadataSpec: ({:?})", metadata_spec);
 
     // metadata smaller than 8-bits is special in that more than one metadata value may be included in one AtomicU8 operation, and extra shift and mask is required
-    let res: usize = if metadata_spec.num_of_bits < BITS_IN_BYTE {
+    let res: usize = if metadata_spec.num_of_bits < 8 {
```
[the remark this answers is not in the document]
qinsoon: Right. This change looks strange. I made the change as in other methods with the same pattern, they use literals rather than the constants. I was trying to make it consistent. I will change this (and other methods) as well.

src/util/metadata/mod.rs
qinsoon: Yeah. As Kunshan and I discussed, we will make 64bits metadata usable on 32bits architecture where it is possible. This requires us changing the metadata-related API (we cannot use `usize` as arguments and return values if we want to use 64bits values on 32bits architectures)

qinsoon: I did some refactoring on this. Basically I will introduce a new trait `MetadataValue`: when storing and loading from metadata, this type is used to describe the value. `u8, u16, u32, u64` implement this trait.

The old code looks like this (will be referred as *untyped*):
https://github.com/mmtk/mmtk-core/blob/d8bcf90991224cfe06153689b14c3e1eea0e025e/src/util/metadata/side_metadata/global.rs#L671-L707

This gets compiled to (in a use case of 1-bit alloc bit):
```asm
mmtk::util::alloc_bit::is_alloced_object_unsafe_untyped:
 mov     rcx, rdi
 mov     rax, rdi
 shr     rax, 6
 shr     cl, 3
 and     cl, 7
 movabs  rdx, 6597069766656
 movzx   eax, byte, ptr, [rax, +, rdx]
 mov     edx, 1
 shl     edx, cl
 and     edx, eax
 shr     rdx, cl
 cmp     edx, 1
 sete    al
 ret
```

After introducing the `MetadataValue` type, and extracting some common code, the new code looks like this (will be referred as *typed*):
```rust
#[inline(always)]
pub unsafe fn typed_load<T: MetadataValue>(metadata_spec: &SideMetadataSpec, data_addr: Address) -> T {
    side_metadata_access::<T, _, _, _>(metadata_spec, data_addr, || {
        let meta_addr = address_to_meta_address(metadata_spec, data_addr);
        let bits_num_log = metadata_spec.log_num_of_bits;
        if bits_num_log < 3 {
            let lshift = meta_byte_lshift(metadata_spec, data_addr);
            let mask = meta_byte_mask(metadata_spec) << lshift;
            let byte_val = meta_addr.load::<u8>();

            num_traits::FromPrimitive::from_u8((byte_val & mask) >> lshift).unwrap()
        } else {
            meta_addr.load::<T>()
        }
    }, |v| {
        #[cfg(feature = "extreme_assertions")]
        sanity::typed_verify_load(metadata_spec, data_addr, v);
    })
}

fn assert_int_type<T: Bits>(spec: &SideMetadataSpec) {
    let log_b = spec.log_num_of_bits;
    match log_b {
        _ if log_b < 3 => assert_eq!(T::LOG2, 3),
        3..=6 => assert_eq!(T::LOG2, log_b as u32),
        _ => unreachable!("side metadata > {}-bits is not supported", 1 << log_b),
    }
}

#[inline(always)]
fn side_metadata_access<T: MetadataValue, R: Copy, F: Fn() -> R, V: Fn(R)>(spec: &SideMetadataSpec, data_addr: Address, access_func: F, verify_func: V) -> R {
    #[cfg(feature = "extreme_assertions")]
    let _lock = sanity::SANITY_LOCK.lock().unwrap();

    if cfg!(debug_assertions) {
        assert_int_type::<T>(spec);
        ensure_metadata_is_mapped(spec, data_addr);
    }

    let ret = access_func();

    verify_func(ret);

    return ret;
}
```

This gets compiled to (in the same use case as above):
```asm
mmtk::util::alloc_bit::is_alloced_object_unsafe_typed:
 mov     rcx, rdi
 mov     rax, rdi
 shr     cl, 3
 and     cl, 7
 mov     dl, 1
 shl     dl, cl
 shr     rax, 6
 movabs  rsi, 6597069766656
 and     dl, byte, ptr, [rax, +, rsi]
 shr     dl, cl
 cmp     dl, 1
 sete    al
 ret
```

I think performance-wise, the new code should be mostly the same as before. And the new code should be much clearer and has less duplicated code. I plan to finish the refactoring in this PR.

qinsoon:
I pushed some draft code here. It is not yet ready for review.

qinsoon:
The following is the benchmark results for the PR (running Immix with 2x min heap). This is no measurable performance difference for this refactoring.

### STW time
Screen Shot 2022-09-08 at 11 46 04 am

### Benchmark time
Screen Shot 2022-09-08 at 11 47 36 am

<details><summary>Plotty link (not public)</summary>
https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|shrew-2022-09-07-Wed-155555&benchmark^build^invocation^iteration&bmtime^time.other^time.stw&|10&iteration^1^2|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^build^benchmark&
</details>

qinsoon:
binding-refs
OPENJDK_BINDING_REF=update-pr-647
JIKESRVM_BINDING_REF=update-pr-647
V8_BINDING_REF=update-pr-647

qinsoon:
@wks The PR is ready for review again. The focus of the PR changed and there were some major changes since it was reviewed last time.

src/plan/barriers.rs
```diff
@@ -104,7 +103,10 @@ impl<E: ProcessEdgesWork> ObjectRememberingBarrier<E> {
 
     #[inline(always)]
     fn barrier(&mut self, obj: ObjectReference) {
-        if load_metadata::<E::VM>(&self.meta, obj, None, None) == 0 {
+        // Perform a quick non-atomic load for performance.
+        // It is okay if this check fails occasionally and
+        // the execution goes to the slowpath, we can take care of that in the slowpath.
+        if unsafe { self.meta.load::<E::VM, u8>(obj, None) == 0 } {
```
wks, line 109:
```suggestion
        if self.meta.load_atomic::<E::VM, u8>(obj, None, Ordering::Relaxed) == 0 {
```
The metadata may still be concurrently modified by another thread.  According to C++11 memory model, if this non-atomic load does not have a happens-before relationship with a store, it is still considered a data race, which has undefined behaviour.  So we should still use `load_atomic` if there is any possibility to have a concurrent (no happens-before) store.

If performance is important, we can use the `Ordering::Relaxed` order.  On most platforms, a "relaxed load uses the same instruction as non-atomic load.  But because it uses the Rust-level atomic load, Rust compiler can no longer assume there is no concurrent modifications (because otherwise it is undefined behaviour) and "optimise out" this load.

wks (changes requested):

src/plan/barriers.rs
qinsoon: I have made the change as suggested.

src/util/metadata/global.rs
```diff
@@ -31,171 +30,308 @@ impl MetadataSpec {
...
-                unsafe { side_metadata::load(metadata_spec, object.to_address()) }
+    /// A function to non-atomically load the specified metadata's content.
+    ///
+    /// # Arguments:
+    ///
+    /// * `metadata_spec`: is one of the const `MetadataSpec` instances from the ObjectModel trait, for the target metadata. Whether the metadata is in-header or on-side is a VM-specific choice.
+    /// * `object`: is a reference to the target object.
+    /// * `mask`: is an optional mask value for the metadata. This value is used in cases like the forwarding pointer metadata, where some of the bits are reused by other metadata such as the forwarding bits.
+    ///
+    /// # Returns the metadata value.
```
wks, line 42: The doc is written in the Markdown format, so this line means a title whose text is "Return the metadata value".  A more natural way to do it should be

```markdown
/// # Return value
///
/// Returns the metadata value
```

Or simply remove the title
```markdown
/// Returns the metadata value.
```

It looks like we have been misusing this in many places.  We don't have to fix it now.  We can tidy up the docs in a separate PR.

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+            3..=6 => assert_eq!(T::LOG2, log_b as u32),
+            _ => unreachable!("side metadata > {}-bits is not supported", 1 << log_b),
+        }
+    }
+
+    /// Bulk-zero a specific metadata for a chunk.
+    ///
+    /// # Arguments
+    ///
+    /// * `metadata_spec` - The specification of the target side metadata.
```
wks, line 133:
```suggestion
```
This is `self` now.

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+    /// * `chunk_start` - The starting address of the chunk whose metadata is being zeroed.
```
wks, line 134: Argument names changed.

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+
+    /// This is a wrapper method for implementing side metadata access. It does nothing other than
+    /// calling the access function with no overhead, but in debug builds,
+    /// it includes multiple checks to make sure the access is sane.
+    /// * check whether the given value type matches the number of bits for the side metadata.
+    /// * check if the side metadata memory is mapped.
+    /// * check if the side metadata content is correct based on a sanity map (only for extreme assertions).
+    #[inline(always)]
+    #[allow(unused_variables)] // data_addr is not used in release build
+    fn side_metadata_access<T: MetadataValue, R: Copy, F: FnMut() -> R, V: FnMut(R)>(
```
wks, line 200:
```suggestion
    fn side_metadata_access<T: MetadataValue, R: Copy, F: FnOnce() -> R, V: FnOnce(R)>(
```
Both `access_func` and `verify_func` are called only once, so it is OK to use `FnOnce` which has less restrictions.

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+        self.side_metadata_access::<T, _, _, _>(
+            data_addr,
+            || {
+                let meta_addr = address_to_meta_address(self, data_addr);
+                let bits_num_log = self.log_num_of_bits;
+                if bits_num_log < 3 {
+                    let lshift = meta_byte_lshift(self, data_addr);
+                    let mask = meta_byte_mask(self) << lshift;
+                    let old_val = meta_addr.load::<u8>();
+                    let new_val = (old_val & !mask) | (metadata.to_u8().unwrap() << lshift);
```
wks, line 281: `metadata` is not masked here.  It is an error if the caller passes a `metadata` value that has more than `bits_num_log` bits.  I recommend adding a `debug_assert!` to ensure `metadata` does not have more bits than necessary.

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+            data_addr,
+            || {
+                let meta_addr = address_to_meta_address(self, data_addr);
+                let bits_num_log = self.log_num_of_bits;
+                if bits_num_log < 3 {
+                    let lshift = meta_byte_lshift(self, data_addr);
+                    let mask = meta_byte_mask(self) << lshift;
+                    let metadata_u8 = metadata.to_u8().unwrap();
+                    let mut old_val = unsafe { meta_addr.load::<u8>() };
+                    let mut new_val = (old_val & !mask) | (metadata_u8 << lshift);
```
wks, line 330: Same here.  I recommend `debug_assert!` that `metadata_u8` does not have more than `bits_num_log` significant bits.

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+
+                    while unsafe {
+                        meta_addr
+                            .compare_exchange::<AtomicU8>(old_val, new_val, order, order)
```
wks, line 334: We can just use `fetch_update` instead of our hand-written cmpxchg loop.

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+        update: F,
+    ) -> u8 {
+        let lshift = meta_byte_lshift(self, data_addr);
+        let mask = meta_byte_mask(self) << lshift;
+        let mut old_val = unsafe { meta_addr.load::<u8>() };
+        let mut new_sub_val = update((old_val & mask) >> lshift) & (mask >> lshift);
+        let mut new_val = (old_val & !mask) | (new_sub_val << lshift);
+
+        while unsafe {
+            meta_addr
```
wks, line 423: Similar to `atomic_store`, we can use `fetch_update` instead of hand-written cmpxchg loop.

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+            || {
+                let meta_addr = address_to_meta_address(self, data_addr);
+                if self.log_num_of_bits < 3 {
+                    FromPrimitive::from_u8(self.fetch_ops_on_bits(
+                        data_addr,
+                        meta_addr,
+                        order,
+                        order,
+                        |x: u8| x & val.to_u8().unwrap(),
+                    ))
```
wks, line 519: `fetch_and` doesn't need cmpxchg.  Just set irrelevant bits to 1 and they will not be affected.

```suggestion
let mask = meta_byte_mask(self) << lshift;
let opnd = val | (-1 & !mask);
unsafe { T::fetch_and(meta_addr, opnd, order) };
```

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+                if self.log_num_of_bits < 3 {
+                    let lshift = meta_byte_lshift(self, data_addr);
+                    let mask = meta_byte_mask(self) << lshift;
+                    let mut old_val = unsafe { meta_addr.load::<u8>() };
+                    while let Some(next) =
+                        f(FromPrimitive::from_u8((old_val & mask) >> lshift).unwrap())
+                    {
+                        let new_val = (old_val & !mask) | ((next.to_u8().unwrap()) << lshift);
+                        match unsafe {
+                            meta_addr.compare_exchange::<AtomicU8>(
```
wks, line 584: Prefer `fetch_update` over cmpxchg loop.

wks (changes requested):
The advantage of atomic bit-and and bit-or is that they do not need a cmpxchg loop to implement, so can be more efficient than store or cmpxchg.

src/util/metadata/header_metadata.rs
```diff
@@ -29,668 +27,1014 @@ pub struct HeaderMetadataSpec {
...
+        &self,
+        object: ObjectReference,
+        val: T,
+        order: Ordering,
+    ) -> T {
+        #[cfg(debug_assertions)]
+        self.assert_spec::<T>();
+        if self.num_of_bits < 8 {
+            FromPrimitive::from_u8(
+                self.fetch_ops_on_bits(object, order, order, |x: u8| x & val.to_u8().unwrap()),
```
wks, line 418: Similar to side metadata, fetch_and doesn't need a loop.

src/util/metadata/header_metadata.rs
```diff
@@ -29,668 +27,1014 @@ pub struct HeaderMetadataSpec {
...
+        &self,
+        object: ObjectReference,
+        val: T,
+        order: Ordering,
+    ) -> T {
+        #[cfg(debug_assertions)]
+        self.assert_spec::<T>();
+        if self.num_of_bits < 8 {
+            FromPrimitive::from_u8(
+                self.fetch_ops_on_bits(object, order, order, |x: u8| x | val.to_u8().unwrap()),
```
wks, line 438: Similarly, fetch_or doesn't affect unrelated bits.

src/util/metadata/header_metadata.rs
```diff
@@ -29,668 +27,1014 @@ pub struct HeaderMetadataSpec {
...
+                let old_metadata = self.get_bits_from_u8(old_byte);
+                let old_metadata_val = FromPrimitive::from_u8(old_metadata).unwrap();
+                // new_metadata may contain overflow and need to be truncated
+                // TODO: do we really need this?
+                let new_metadata: Option<u8> =
+                    f(old_metadata_val).map(|byte| self.truncate_bits_in_u8(byte.to_u8().unwrap()));
+                if let Some(new_metadata_to_set) = new_metadata {
+                    let new_byte = self.set_bits_to_u8(old_byte, new_metadata_to_set);
+                    if unsafe {
+                        byte_addr.compare_exchange::<AtomicU8>(
```
wks, line 471: Prefer `fetch_update` over cmpxchg.

src/util/metadata/metadata_val_traits.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+impl_bits_trait!(u8);
+impl_bits_trait!(u16);
+impl_bits_trait!(u32);
+impl_bits_trait!(u64);
+impl_bits_trait!(usize);
+
+/// Describes bitwise operations.
+/// If num_traits has this, we do not need our own implementation: https://github.com/rust-num/num-traits/issues/232
+pub trait BitwiseOps {
```
wks, line 28: Is it sufficient to require the type to implement traits in `std::ops`, so that we don't have to route our own `bitand`, `bitor`, ... to their standard operators?
```rust
use std::ops::*;
pub trait BitwiseOps: BitAnd + BitOr + BitXor + Not {
}
```

src/util/alloc_bit.rs
```diff
@@ -38,17 +32,12 @@ pub fn unset_addr_alloc_bit(address: Address) {
...
 
 pub fn unset_alloc_bit(object: ObjectReference) {
     debug_assert!(is_alloced(object), "{:x}: alloc bit not set", object);
-    side_metadata::store_atomic(
-        &ALLOC_SIDE_METADATA_SPEC,
-        object.to_address(),
-        0,
-        Ordering::SeqCst,
-    );
+    ALLOC_SIDE_METADATA_SPEC.store_atomic::<u8>(object.to_address(), 0, Ordering::SeqCst);
```
wks, line 40: It is more efficient to clear bit using `fetch_and_atomic`.

src/util/alloc_bit.rs
```diff
@@ -24,12 +23,7 @@ pub fn map_meta_space_for_chunk(metadata: &SideMetadataContext, chunk_start: Add
 
 pub fn set_alloc_bit(object: ObjectReference) {
     debug_assert!(!is_alloced(object), "{:x}: alloc bit already set", object);
-    side_metadata::store_atomic(
-        &ALLOC_SIDE_METADATA_SPEC,
-        object.to_address(),
-        1,
-        Ordering::SeqCst,
-    );
+    ALLOC_SIDE_METADATA_SPEC.store_atomic::<u8>(object.to_address(), 1, Ordering::SeqCst);
```
wks, line 26: `fetch_or_atomic` can be more efficient.

src/plan/barriers.rs
```diff
@@ -81,8 +79,9 @@ impl<E: ProcessEdgesWork> ObjectRememberingBarrier<E> {
                 // We just logged the object
                 return true;
             } else {
-                let old_value =
-                    load_metadata::<E::VM>(&self.meta, object, None, Some(Ordering::SeqCst));
+                let old_value = self
+                    .meta
+                    .load_atomic::<E::VM, u8>(object, None, Ordering::SeqCst);
```
wks, line 84: Usually cmpxchg returns the old value regardless of success or failure, so it should not be necessary to load again.  Consider changing the API of cmpxchg.

Alternatively, try `fetch_and_atomic`.  It should be faster on architectures that has special instructions that never fails.

src/policy/immix/immixspace.rs
```diff
@@ -509,21 +507,19 @@ impl<VM: VMBinding> ImmixSpace<VM> {
     #[inline(always)]
     fn attempt_mark(&self, object: ObjectReference, mark_state: u8) -> bool {
         loop {
-            let old_value = load_metadata::<VM>(
-                &VM::VMObjectModel::LOCAL_MARK_BIT_SPEC,
+            let old_value = VM::VMObjectModel::LOCAL_MARK_BIT_SPEC.load_atomic::<VM, u8>(
```
wks, line 510: Depending on `mark_state`, `fetch_and` or `fetch_or` can be more efficient.

src/policy/immortalspace.rs
```diff
@@ -178,20 +174,18 @@ impl<VM: VMBinding> ImmortalSpace<VM> {
...
-                Some(Ordering::SeqCst),
+                Ordering::SeqCst,
             );
             if old_value == value {
                 return false;
             }
 
-            if compare_exchange_metadata::<VM>(
-                &VM::VMObjectModel::LOCAL_MARK_BIT_SPEC,
+            if VM::VMObjectModel::LOCAL_MARK_BIT_SPEC.compare_exchange_metadata::<VM, u8>(
```
wks, line 188: This may be refactored to use `fetch_update`.

src/policy/largeobjectspace.rs
```diff
@@ -257,25 +252,23 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
...
-                Some(Ordering::SeqCst),
+                Ordering::SeqCst,
             );
             let mark_bit = old_value & mask;
             if mark_bit == value {
                 return false;
             }
-            if compare_exchange_metadata::<VM>(
-                &VM::VMObjectModel::LOCAL_LOS_MARK_NURSERY_SPEC,
+            if VM::VMObjectModel::LOCAL_LOS_MARK_NURSERY_SPEC.compare_exchange_metadata::<VM, u8>(
```
wks, line 271: Can be refactored to `fetch_update`

src/policy/largeobjectspace.rs
```diff
@@ -289,39 +282,35 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
...
+            let old_val = VM::VMObjectModel::LOCAL_LOS_MARK_NURSERY_SPEC.load_atomic::<VM, u8>(
                 object,
                 None,
-                Some(Ordering::Relaxed),
+                Ordering::Relaxed,
             );
             let new_val = old_val & !NURSERY_BIT;
-            if compare_exchange_metadata::<VM>(
-                &VM::VMObjectModel::LOCAL_LOS_MARK_NURSERY_SPEC,
+            if VM::VMObjectModel::LOCAL_LOS_MARK_NURSERY_SPEC.compare_exchange_metadata::<VM, u8>(
```
wks, line 313: Consider refactoring to `fetch_update`.

src/policy/mallocspace/global.rs
```diff
@@ -316,9 +316,10 @@ impl<VM: VMBinding> MallocSpace<VM> {
             address,
         );
 
-        if !is_marked::<VM>(object, None) {
+        // TODO: Why do we use non-atomic load here?
+        if !unsafe { is_marked_unsafe::<VM>(object) } {
```
wks, line 320: I think it is a bug.  If it is not atomic, it is guaranteed to cause data race because tracing is parallel.

src/policy/mallocspace/metadata.rs
```diff
@@ -200,25 +197,19 @@ pub fn is_chunk_mapped(chunk_start: Address) -> bool {
...
-pub fn set_mark_bit<VM: VMBinding>(object: ObjectReference, ordering: Option<Ordering>) {
-    store_metadata::<VM>(
-        &VM::VMObjectModel::LOCAL_MARK_BIT_SPEC,
-        object,
-        1,
-        None,
-        ordering,
-    );
+pub fn set_mark_bit<VM: VMBinding>(object: ObjectReference, ordering: Ordering) {
+    VM::VMObjectModel::LOCAL_MARK_BIT_SPEC.store_atomic::<VM, u8>(object, 1, None, ordering);
```
wks, line 212: This is a one-bit metadata.  May use `fetch_or_atomic`.

wks (commented):
With the introduction of `fetch_and` and `fetch_or`, some bit operations can be done more efficiently using them, especially set bits and clear bits.

We may change all call sites of `store_atomic` into `fetch_and_atomic` or `fetch_or_atomic`.

Alternatively, we can add a special case for `store_atomic` if `bits_num_log == 1`.  In this case, we can turn `store_atomic` internally into `fetch_and` or `fetch_or`, depending on whether it is storing 1 or 0.

src/util/metadata/global.rs
```diff
@@ -31,171 +30,308 @@ impl MetadataSpec {
...
+    ///
+    /// * `metadata_spec`: is one of the const `MetadataSpec` instances from the ObjectModel trait, for the target metadata. Whether the metadata is in-header or on-side is a VM-specific choice.
+    /// * `object`: is a reference to the target object.
+    /// * `old_val`: is the expected current value of the metadata.
+    /// * `new_val`: is the new metadata value to be stored if the compare-and-exchange operation is successful.
+    /// * `mask`: is an optional mask value for the metadata. This value is used in cases like the forwarding pointer metadata, where some of the bits are reused by other metadata such as the forwarding bits.
+    /// * `success_order`: is the atomic ordering used if the operation is successful.
+    /// * `failure_order`: is the atomic ordering used if the operation fails.
+    ///
+    /// # Returns `true` if the operation is successful, and `false` otherwise.
```
wks, line 158: Cmpxchg should return the old value and whether it is successful or not.  The `AtomicU8::compare_exchange` method returns `Result<u8, u8>`. We should do it similarly here.  Some algorithm can benefit from it and eliminate a subsequent load.

wks (commented):
About the return value of cmpxchg...

src/util/metadata/global.rs
qinsoon: I moved the 'return' section to the end of the description paragraph (before `#Arguments`). I did the change for the related methods in the `metadata` and the `ObjectModel` module.

src/util/metadata/side_metadata/global.rs
qinsoon: Right. I will change this.

qinsoon:
@wks My plan is to fix whatever is about the implementation the metadata. For the usage of the metadata and any possible optimization on that, I will leave it to another PR.

src/util/metadata/side_metadata/global.rs
qinsoon: I will make the change.

src/util/metadata/side_metadata/global.rs
qinsoon: I will make the change.

src/util/metadata/side_metadata/global.rs
qinsoon: Sounds good. In case you may be interested, the following is the assembly code of `fetch_and()` for 1 bit side metadata when using `fetch_ops_on_bits` (internally uses `fetch_update`), and directly using `fetch_and`:

### with fetch_ops_on_bits
```asm
mmtk::util::alloc_bit::unset_alloc_bit_unsafe:
 mov     rcx, rdi
 shr     rdi, 6
 shr     cl, 3
 movabs  rsi, 13194139533312
 mov     al, byte, ptr, [rdi, +, rsi]
 mov     dl, -2
 rol     dl, cl
.LBB225_1:
 mov     ecx, eax
 and     cl, dl
 lock    cmpxchg, byte, ptr, [rdi, +, rsi], cl
 jne     .LBB225_1
 ret
```
### with fetch_and
```asm
mmtk::util::alloc_bit::unset_alloc_bit_unsafe:
 mov     rcx, rdi
 mov     rax, rdi
 shr     rax, 6
 shr     cl, 3
 mov     dl, -2
 rol     dl, cl
 movabs  rcx, 13194139533312
 lock    and, byte, ptr, [rax, +, rcx], dl
```

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,515 @@ impl SideMetadataSpec {
...
+                        data_addr,
+                        meta_addr,
+                        order,
+                        order,
+                        |x: u8| x | val.to_u8().unwrap(),
+                    ))
+                    .unwrap()
+                } else {
+                    unsafe { T::fetch_or(meta_addr, val, order) }
+                }
```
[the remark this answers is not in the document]
qinsoon: I will make the change.

src/util/metadata/metadata_val_traits.rs
qinsoon: I cannot get this work. The problem is that ops traits like `BitAnd` have an `Output` associate type, and we cannot set that to our integer/metadata type`MetadataValue`.

src/util/metadata/side_metadata/global.rs
qinsoon: Done

src/util/metadata/header_metadata.rs
```diff
@@ -29,668 +27,1014 @@ pub struct HeaderMetadataSpec {
...
                     loop {
-                        let old_u32_val = u32_addr.atomic_load::<AtomicU32>(order);
-                        let new_u32_val =
-                            (old_u32_val & !(mask as u32)) | (val as u32 & (mask as u32));
-                        if u32_addr
-                            .compare_exchange::<AtomicU32>(old_u32_val, new_u32_val, order, order)
+                        let old_byte_val = byte_addr.atomic_load::<AtomicU8>(order);
+                        let new_byte_val = self.set_bits_to_u8(old_byte_val, val_u8);
+                        if byte_addr
+                            .compare_exchange::<AtomicU8>(old_byte_val, new_byte_val, order, order)
```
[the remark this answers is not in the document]
qinsoon: Done

src/util/metadata/header_metadata.rs
```diff
@@ -29,668 +27,1014 @@ pub struct HeaderMetadataSpec {
...
+        } else {
+            let addr = self.meta_addr(object);
+            unsafe {
+                if let Some(order) = atomic_ordering {
+                    // if the optional mask is provided (e.g. for forwarding pointer), we need to use compare_exchange
+                    if let Some(mask) = optional_mask {
+                        loop {
+                            let old_val = T::load_atomic(addr, order);
+                            let new_val = old_val.bitand(mask.inv()).bitor(val.bitand(mask));
+                            if T::compare_exchange(addr, old_val, new_val, order, order).is_ok() {
```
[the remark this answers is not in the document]
qinsoon: Done

src/util/metadata/header_metadata.rs
```diff
@@ -29,668 +27,1014 @@ pub struct HeaderMetadataSpec {
...
-                let new_byte = (old_byte & !mask) | ((new_metadata as u8) << bit_shift);
+                let old_byte = byte_addr.atomic_load::<AtomicU8>(fetch_order);
+                let old_metadata = self.get_bits_from_u8(old_byte);
+                // new_metadata may contain overflow and need to be truncated
+                // TODO: do we really need this?
+                let new_metadata = self.truncate_bits_in_u8(update(old_metadata));
+                let new_byte = self.set_bits_to_u8(old_byte, new_metadata);
                 if byte_addr
-                    .compare_exchange::<AtomicU8>(old_byte, new_byte, order, order)
+                    .compare_exchange::<AtomicU8>(old_byte, new_byte, set_order, fetch_order)
```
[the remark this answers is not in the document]
qinsoon: Done

src/util/metadata/header_metadata.rs
qinsoon: Done.

src/util/metadata/header_metadata.rs
qinsoon: Done

src/util/metadata/header_metadata.rs
qinsoon: Done

src/util/metadata/global.rs
qinsoon: I will change this to `Result<T, T>`. For its use cases, I will leave it as `compare_exchange().is_ok()`.

src/policy/mallocspace/global.rs
qinsoon: I assume it should be a benign race, and it should not cause any actual bug. There are some discussions here: https://github.com/mmtk/mmtk-core/issues/313

qinsoon:
This pull request is ready for another review. @wks

src/policy/mallocspace/global.rs
wks: When @steveblackburn said "atomic operation" in https://github.com/mmtk/mmtk-core/issues/313#issuecomment-870346496, I think he meant atomic read-modify-write (RMW) operation, and atomic RMW operations do prevent duplicate edges.  In Java, all write operations to 32-bit variables and reference variables (even not `volatile`) are atomic, in the sense that it disallows word tearing, and if a read does not have happens-before relationship with a write, the read always sees some benign values (written by an actual store).  So we seldom use the phrase "atomic read" or "atomic write" in Java.

It is not true for C++ and Rust.  In C++ and Rust, if a non-atomic load has no happens-before relationship with a store, it has *undefined behaviour*.  The Rust counterpart of "ordinary" load/store in Java are `load_atomic` and `store_atomic` with the `Relaxed` order (actually the Java memory order is even weaker than Relaxed, but must prevent word tearing).

qinsoon: We can change this to a relaxed atomic load, which is more reasonable.

But just for the sake of argument, I don't think this causes any actual bug. The issue #313 is clear that the `if` block may be executed multiple times for the same object. So:
* if `is_marked()` returns false, but the object is actually marked: we will just execute the block again, and it is fine.
* if `is_marked()` returns true, but the object is not actually marked: I don't think this will happen, as we are monotonically set the bit from 0 to 1 at this stage.

src/plan/barriers.rs
qinsoon: Yeah, I will do it in another PR.

src/util/metadata/header_metadata.rs
```diff
@@ -29,668 +27,1004 @@ pub struct HeaderMetadataSpec {
...
+                    |raw_byte: u8| {
+                        let old_metadata = self.get_bits_from_u8(raw_byte);
+                        match f(FromPrimitive::from_u8(old_metadata).unwrap()) {
+                            Some(new_val) => {
+                                let new_metadata =
+                                    self.truncate_bits_in_u8(new_val.to_u8().unwrap());
+                                Some(self.set_bits_to_u8(raw_byte, new_metadata))
+                            }
+                            None => None,
+                        }
```
wks, line 468: This is the pattern for `Option::map`.

```suggestion
                        f(FromPrimitive::from_u8(old_metadata).unwrap()).map(|new_val| {
                                let new_metadata =
                                    self.truncate_bits_in_u8(new_val.to_u8().unwrap());
                                self.set_bits_to_u8(raw_byte, new_metadata)
                        }
```

src/util/metadata/side_metadata/global.rs
```diff
@@ -97,6 +99,534 @@ impl SideMetadataSpec {
...
+                    let mask = meta_byte_mask(self) << lshift;
+
+                    unsafe {
+                        <u8 as MetadataValue>::fetch_update(
+                            meta_addr,
+                            set_order,
+                            fetch_order,
+                            |raw_byte: u8| {
+                                let old_val = (raw_byte & mask) >> lshift;
+                                match f(FromPrimitive::from_u8(old_val).unwrap()) {
```
wks, line 604: Try `Option::map`, too.

src/util/metadata/header_metadata.rs
qinsoon: Done

src/util/metadata/side_metadata/global.rs
qinsoon: Done

src/policy/mallocspace/global.rs
qinsoon: I have changed this to a relaxed atomic load.

wks (approved):
LGTM

PR #660, opened 2022-09-06
Fix broken links in docs

wks:
Some hyperlinks in doc comments are making the Public API Check in CI fail.

-   `vm_space` is only enabled in a feature, and the module `crate::plan::global` is private.  I changed it to non-link text.
-   Use fully-qualified names for other broken links.
-   Fixed the URL of the DOI link of GenImmix.


wks:
The Public API Check still fails because my PR is in my personal repository.  Somewhere in the CI scripts should be fixed.

qinsoon (approved):
LGTM

PR #662, opened 2022-09-07
Fix API check for PRs from a forked repo

qinsoon:
This PR fixes the issue that the API check workflow will fail when a pull request is submitted from a forked repo.

### DIFF
--- a/.github/workflows/api-check.yml
+++ b/.github/workflows/api-check.yml
@@ -18,39 +18,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      # # Install nightly
-      # # The nightly version needs to match the cargo-public-api version.
-      # # Check here: https://github.com/Enselic/cargo-public-api#compatibility-matrix
-      # - uses: actions-rs/toolchain@v1
-      #   with:
-      #     toolchain: nightly-2022-09-06
-      #     profile: minimal
-
-      # # Install a specific version of cargo-public-api
-      # - name: Install cargo-public-api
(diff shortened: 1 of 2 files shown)


.github/workflows/api-check.yml
```diff
@@ -20,12 +18,35 @@ jobs:
...
+      # Install a specific version of cargo-public-api
+      - name: Install cargo-public-api
+        run: cargo install cargo-public-api --version 0.17.0
+      # We cannot compare two branches using 'cargo public-api --diff-git-checkouts ${GITHUB_BASE_REF} ${GITHUB_HEAD_REF}',
+      # as the Github checkout action does not actually fetch the branch from the forked repo, and cargo-public-api cannot
+      # check out ${GITHUB_HEAD_REF} if the PR is from a fork.
+      # So we get API dumped to text files, and diff the text files.
+      - name: Dump public API for PR
+        run: |
+          cargo +nightly-2022-09-06 public-api > pr-api.txt
```
wks, line 38: If we are sure the current HEAD is the PR commit, can we simply do the following?
```
cargo public-api --diff-git-checkouts ${GITHUB_BASE_REF} HEAD --deny=all
```

wks (approved):
LGTM

.github/workflows/api-check.yml
qinsoon: We figured out that `${{ github.event.pull_request.head.sha }}` works fine here.

PR #663, opened 2022-09-07
Support disjoint objects

wks:
At this moment, I just drafted the interface for the VM.

The VM binding shall implement `Scanning::handle_buffers`.  In that function, the VM can call back to MMTk core via the "BufferHandler" which MMTk core implements.  I documented the contract (desired behaviour of the VM binding) in the comments.

What remains to be designed/implemented includes

-   Is the current `handle_buffers` sufficient to support all plans? (non-moving plans like MS, evacuating plans like SemiSpace, and MarkCompact)
-   When is `handle_buffers` called in each plan?

I think the protocol of `Scanning::handle_buffers` should be sufficient for all plans, but the timing should be different depending on the plan.

-   MarkSweep is trivial.  Any time should be OK.
-   For SemiSpace and Immix, it should be called immediately before scanning an object (no matter if the VM chooses to do edge-enqueuing or object-enqueuing) so that the buffers are evacuated and the pointers to buffers are updated before the object is scanned.
-   For MarkCompact, it is a bit tricky.  I think `handle_buffers` may be called during the forwarding phase to work out the space needed to hold new buffers and its location.  But what should happen when the GC actually copies the buffer?

### DIFF
--- a/src/vm/scanning.rs
+++ b/src/vm/scanning.rs
@@ -1,8 +1,10 @@
 use crate::plan::Mutator;
+use crate::util::Address;
 use crate::util::ObjectReference;
 use crate::util::VMWorkerThread;
 use crate::vm::edge_shape::Edge;
 use crate::vm::VMBinding;
+use crate::AllocationSemantics;
 
 /// Callback trait of scanning functions that report edges.
 pub trait EdgeVisitor<ES: Edge> {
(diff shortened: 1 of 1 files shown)


steveblackburn:
I don't understand the motivation for `has_buffer` and `handle_buffer`.

I think that the contract should be that responsibility lies entirely with the VM to identify any buffer when it scans the parent.

wks:
> I don't understand the motivation for `has_buffer` and `handle_buffer`.
>
> I think that the contract should be that responsibility lies entirely with the VM to identify any buffer when it scans the parent.

I want to keep `Scanning::scan_object(object, edge_visitor)` serving the sole purpose of enumerating edges, but not identifying buffers or updating pointers to buffers.

It is useful in some cases.  For example, for referencing-counting GC, we sometimes need to increment the RC of all neighbouring objects of an object.  In this case, we just enumerate edges with `scan_object(object, |edge| { inc_rc(edge.load()) })`.  It is not a good time to retain buffers or merging/splitting buffers because GC may not be in progress.

PR #670, opened 2022-09-28
Add options for setting GC thread affinity

### LINKED ISSUE #299: Support pinning of GC threads

k-sareen:
This commit allows the user to set GC thread affinity by specifying a list of cores. The cores are specified similar to how a CPU list is provided to `taskset` on Linux, i.e. with discrete core ids or ranges. The user can also set how the affinity of multiple workers is resolved.

Note that this feature currently only works on Linux due to platform-specific calls.

Closes #299.


k-sareen:
I need to add some comments and documentation before it's ready for review.

k-sareen:
Actually, should we even pin the coordinator thread? Seems like a bad idea. Pinning only GC threads makes more sense.

src/scheduler/affinity.rs
```diff
@@ -0,0 +1,109 @@
...
+
+/// Represents the ID of a logical CPU on a system.
+pub type CoreId = u16;
+
+// XXX: Maybe in the future we can use a library such as https://github.com/Elzair/core_affinity_rs
+// to have an OS agnostic way of setting thread affinity.
+#[cfg(target_os = "linux")]
+/// Return the total number of (online) cores on a system.
+fn get_total_num_cpus() -> u16 {
+    unsafe { libc::sysconf(libc::_SC_NPROCESSORS_ONLN) as u16 }
```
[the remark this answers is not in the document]
k-sareen: Yes you're right. `sysconf` does not respect the number of cores that have been allocated to the program via `taskset`. I'll fix this.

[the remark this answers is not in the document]
k-sareen: Fixed in https://github.com/mmtk/mmtk-core/pull/670/commits/089a7522b5eafb09281e5e0f309782742fe065b6

Thanks for the catch.

src/util/options.rs
```diff
@@ -397,7 +469,21 @@ options! {
     work_perf_events:       PerfEventOptions     [env_var: true, command_line: true] [|_| cfg!(all(feature = "perf_counter", feature = "work_packet_stats"))] = PerfEventOptions {events: vec![]},
     // Measuring perf events for GC and mutators
     // TODO: Ideally this option should only be included when the features 'perf_counter' are enabled. The current macro does not allow us to do this.
-    phase_perf_events:      PerfEventOptions     [env_var: true, command_line: true] [|_| cfg!(feature = "perf_counter")] = PerfEventOptions {events: vec![]}
+    phase_perf_events:      PerfEventOptions     [env_var: true, command_line: true] [|_| cfg!(feature = "perf_counter")] = PerfEventOptions {events: vec![]},
+    // Set how to bind affinity to the GC Workers. Default thread affinity delegates to the OS scheduler. XXX: This option is currently only supported on Linux.
+    thread_affinity:        AffinityKind         [env_var: true, command_line: true] [always_valid] = AffinityKind::OsDefault,
```
qinsoon, line 474: This seems cleaner. 
```rust
pub enum AffinityKind {
    OsDefault,
    Fixed(CoreId),
    #[cfg(target_os = "linux)]
    RoundRobin(Vec<CoreId>),
}
```

* `cpu_list` has to be used with `thread_affinity`. It would be better not to have a separate option for `cpu_list`. 
* Supply `CoreId` with each variant of the enum is enough.
* You can let the scheduler to create the `CpuSet` from the option. And you can put `CPU_SET_IDX` to the struct (unless you actually want to this to be global, i.e. used across multiple MMTk instances).
* `CpuSet.num_cpu` is only used to verify the option so you don't need to actually save it anywhere. You can use the validation method to do it (use your validation method instead of `always_valid`).

k-sareen: Hmm fair enough. In that case would the user specify a core list as `MMTK_THREAD_AFFINITY=RoundRobin("0,3-4")`? Or I guess we could just do if it's 1 core then it's a Fixed affinity and greater than 1 core implies `RoundRobin`.

k-sareen: PS Do you have any opinions on the pinning of the coordinator thread? Currently the PR will pin it, but I think it is not required (and only complicates the pinning).

qinsoon: You can do something similar to the bounded/fixed nursery option you did before, something like `RoundRobin:0,3-4`.

k-sareen: > Or I guess we could just do if it's 1 core then it's a Fixed affinity and greater than 1 core implies RoundRobin.

I think this is simpler (for the user). Do we envision a different method of assigning cores to threads in the future? Like random maybe? If not, then I think the above will allow for a simpler user interface.

qinsoon: > PS Do you have any opinions on the pinning of the coordinator thread? Currently the PR will pin it, but I think it is not required (and only complicates the pinning).

I have no preference on this.

> Or I guess we could just do if it's 1 core then it's a Fixed affinity and greater than 1 core implies RoundRobin.

This sounds fine. Actually you don't need to deal with `Fixed` specially. `x % 1 = 0`, which is the index of the only element in the array.

k-sareen: True -- we can just remove `Fixed` in favour of just `RoundRobin` then.

k-sareen: @caizixian Thoughts on pinning GC coordinator thread? I'm inclined to remove it, but if you have a case for it, then I'll just leave it in.

k-sareen: I've addressed your comments in https://github.com/mmtk/mmtk-core/pull/670/commits/248ba47deede2ab6603060150c147b33ebca1a90, @qinsoon. Zixian and I think that pinning GC coordinator thread is not necessary, so I've removed it.

k-sareen:
Oh no. I've messed up the rebase :cry:

k-sareen:
Good to merge @caizixian @qinsoon?

src/scheduler/affinity.rs
```diff
@@ -0,0 +1,59 @@
+use super::worker::ThreadId;
+use crate::util::options::AffinityKind;
+use libc::{cpu_set_t, sched_getaffinity, sched_setaffinity, CPU_COUNT, CPU_SET, CPU_ZERO};
```
caizixian, line 3: The import should also be guarded for Linux only.

src/util/options.rs
```diff
@@ -254,6 +255,96 @@ macro_rules! options {
...
+    RoundRobin(Vec<CoreId>),
+}
+
+impl AffinityKind {
+    /// Returns an AffinityKind or String containing error. Expects the list of cores to be
+    /// formatted as numbers separated by commas, including ranges. There should be no spaces
+    /// between the cores in the list. For example: 0,5,8-11 specifies that the cores 0,5,8,9,10,11
+    /// should be used for pinning threads. Performs de-duplication of specified cores. Note that
+    /// the core list is sorted as a side-effect whenever a new core is added to the set.
+    fn parse_cpulist(cpulist: &str) -> Result<AffinityKind, String> {
```
caizixian, line 277: Should add unit tests for parsing.

caizixian (changes requested):

src/scheduler/affinity.rs
k-sareen: True. Thanks for the catch.

src/util/options.rs
k-sareen: Yes I will make some unit tests

k-sareen: I have made tests but there is a worry that the tests may spuriously fail in the CI as some of the tests depend on the existence of more than one core. If that's not true then the test may fail since the `validate()` function will fail.

k-sareen: Actually I can directly test the parsing function instead of setting the environment variable, so this is not an issue.

qinsoon: > I have made tests but there is a worry that the tests may spuriously fail in the CI as some of the tests depend on the existence of more than one core. If that's not true then the test may fail since the `validate()` function will fail.

Does that mean the `validate()` function will fail if there is only one core? That sounds like a bug in the code, rather than about the tests.

k-sareen: No let me clarify. If I do `MMTK_THREAD_AFFINITY="0-1"` but I only have 1 core available (either it's a single core machine or I'm running `taskset -c 0` on the program), then the `validate()` function will rightfully say hey I don't know what core id is `1` so this is not a valid core list.

qinsoon: Also you could get the number of CPUs, and use that to generate a valid test input.

k-sareen: Yes I was making this exact test. But I've also changed most of the tests to directly test the parsing function which is equivalent imo.

caizixian (approved):

PR #672, opened 2022-10-01
Increase lock granularity in `FreeListPageResource`

### LINKED ISSUE #300: Possible race condition in FreelistPageResource
The locking in FreelistPageResource is different from the original code in Java MMTk.

For example, in `FreelistPageResource.release_pages()`, the lock only guards a part of the function starting at https://github.com/mmtk/mmtk-core/blob/eb949afce34ccc28845963ff4ebba9c4d2654949/src/util/heap/freelistpageresource.rs#L306 for updating committed/reserved pages and updating `pages_currently_on_freelist`. In a later step, in `release_free_chunks()` -> `free_contiguous_chunk()` -> https://github.com/mmtk/mmtk-core/blob/eb949afce34ccc28845963ff4ebba9c4d2654949/src/util/heap/freelistpageresource.rs#L263, the lock is acquired again to update `pages_currently_on_freelist`. We do not hold the lock between the two lock acquisitions so it is possible someone updates `sync` between the two acquisitions. Also we **unsafely** cast an immutable reference `self` to a mutable reference and use it throughout the code. That means we have a race condition. This probably does not cause any obvious bug for us at the moment, as now `release_pages()` is only called by a single thread. However, we should carefully check this part of the code in our unsafe code audit.

In the original Java MMTk, the lock guards the rest of the function (https://github.com/JikesRVM/JikesRVM/blob/0b6002e7d746a829d56c90acfc4bb5c560faf634/MMTk/src/org/mmtk/utility/heap/FreeListPageResource.java#L216).

### LINKED ISSUE #671: `IntArrayFreeList` concurrency issue
When a *discontiguous* `FreelistPageResource` allocate pages on a `IntArrayFreeList`, it only lock the `FreelistPageResource` itself:

https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/freelistpageresource.rs#L109-L111

We expected that `IntArrayFreeList` will be locked, since `FreelistPageResource` owns it.

However, all `IntArrayFreeList` instances actually share a single free-list table:

https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/freelistpageresource.rs#L222
https://github.com/mmtk/mmtk-core/blob/master/src/util/heap/layout/map32.rs#L76
https://github.com/mmtk/mmtk-core/blob/master/src/util/int_array_freelist.rs#L41-L50
https://github.com/mmtk/mmtk-core/blob/master/src/util/int_array_freelist.rs#L60-L69

So as long as a plan has two *discontiguous* `FreelistPageResource`s, concurrent allocation or modification to a shared free-list table is incorrect.

---

Maybe I did not fully understand the code, but here are a few observations/questions:

1. This concurrency issue did have a problem if I use `Map32` and `IntArrayFreeList` on 64-bit mmtk-openjdk.
2. These code are directly ported from JikesRVM. I checked that similar to mmtk-core, JikesRVM MMTk does not lock anything (apart from the pageresource lock) before modifying the free list. How come it did not have any issue?
3. Does any GC on mmtk-jikesrvm uses two *discontiguous* `FreelistPageResource`s? Any known correctness issues?

wenyuzhao:
1. In `FreeListPageResource.release_pages`, guard free list modification operations by locks. Otherwise, the parallel page release would be incorrect. This aligns with the implementation on Java MMTk https://github.com/JikesRVM/JikesRVM/blob/master/MMTk/src/org/mmtk/utility/heap/FreeListPageResource.java#L216-L225
   * Fixes #671 
   * Fixes #300 
2. ~~Remove the `meta_data_pages_per_region` field to simplify the code. We've completely switched to side metadata and don't use this flag anymore.~~
   * Moved to #675

### DIFF
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -211,9 +211,9 @@ impl<VM: VMBinding> ImmixSpace<VM> {
         );
         ImmixSpace {
             pr: if common.vmrequest.is_discontiguous() {
-                FreeListPageResource::new_discontiguous(vm_map)
+                FreeListPageResource::new_discontiguous(0, vm_map)
             } else {
-                FreeListPageResource::new_contiguous(common.start, common.extent, vm_map)
+                FreeListPageResource::new_contiguous(common.start, common.extent, 0, vm_map)
             },
             common,
             chunk_map: ChunkMap::new(),
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -211,9 +211,9 @@ impl<VM: VMBinding> ImmixSpace<VM> {
         );
         ImmixSpace {
             pr: if common.vmrequest.is_discontiguous() {
-                FreeListPageResource::new_discontiguous(0, vm_map)
+                FreeListPageResource::new_discontiguous(vm_map)
             } else {
-                FreeListPageResource::new_contiguous(common.start, common.extent, 0, vm_map)
+                FreeListPageResource::new_contiguous(common.start, common.extent, vm_map)
             },
             common,
             chunk_map: ChunkMap::new(),
--- a/src/policy/largeobjectspace.rs
+++ b/src/policy/largeobjectspace.rs
@@ -165,9 +165,9 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
             heap,
         );
         let mut pr = if vmrequest.is_discontiguous() {
-            FreeListPageResource::new_discontiguous(vm_map)
+            FreeListPageResource::new_discontiguous(0, vm_map)
         } else {
-            FreeListPageResource::new_contiguous(common.start, common.extent, vm_map)
+            FreeListPageResource::new_contiguous(common.start, common.extent, 0, vm_map)
         };
         pr.protect_memory_on_release = protect_memory_on_release;
         LargeObjectSpace {
--- a/src/policy/largeobjectspace.rs
+++ b/src/policy/largeobjectspace.rs
@@ -165,9 +165,9 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
             heap,
         );
         let mut pr = if vmrequest.is_discontiguous() {
-            FreeListPageResource::new_discontiguous(0, vm_map)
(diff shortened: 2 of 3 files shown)


qinsoon:
This PR closes https://github.com/mmtk/mmtk-core/issues/300. 
@wenyuzhao We actually had an issue for this: https://github.com/mmtk/mmtk-core/issues/300.

wenyuzhao:
> This PR closes #300.
> @wenyuzhao We actually have an issue for this: #300.

Ah didn't realize that. Thanks.

qinsoon (commented):
The change on the lock looks good. The actual problem is the unsafe code and the bad pattern of locking in the part of code -- we will fix that eventually.

About the changes for `meta_data_pages_per_region`, I think what is done in the PR is incomplete. We also have `meta_data_pages_per_region` for `MonotonePageResource`, and we should remove that as well. I suggest splitting the changes for `meta_data_pages_per_region` to a separate PR: just revert f120698f9636d5579403606439db32b93f6d6d42, and cherry-pick that commit and create another another PR. If you don't feel motivated to do the same change for `MonotonePageResource`, I can work on that.

wenyuzhao:
> The change on the lock looks good. The actual problem is the unsafe code and the bad pattern of locking in the part of code -- we will fix that eventually.
> 
> About the changes for `meta_data_pages_per_region`, I think what is done in the PR is incomplete. We also have `meta_data_pages_per_region` for `MonotonePageResource`, and we should remove that as well. I suggest splitting the changes for `meta_data_pages_per_region` to a separate PR: just revert [f120698](https://github.com/mmtk/mmtk-core/commit/f120698f9636d5579403606439db32b93f6d6d42), and cherry-pick that commit and create another another PR. If you don't feel motivated to do the same change for `MonotonePageResource`, I can work on that.

Reverted. I'll submit a new PR to remove `meta_data_pages_per_region`.

wenyuzhao:
For the "bad pattern of locking", I plan to submit another PR in the future to support runtime address space switch based on command line args. My changes will turn `Map`, `Mmapper`, and `IntArrayFreeList` into trait objects. I'll try to implement a better lock pattern in my changes.

qinsoon (approved):
LGTM

PR #680, opened 2022-10-13
Rename alloc bit to valid-object bit (VO-bit)

wks:
We currently have a so-called "allocation bit", or "alloc bit" metadata.  This name is misleading, because the name suggests it is used to identify allocated units in spaces, which is not true.  It is actually a global one-bit-per-minimum-alignment bitmap that indicates whether there is a valid object at any given address.  It is primarily designed to help conservative GC to filter potential object references on the stack or in objects.  This PR gives it a proper name, **valid-object bit** or **VO-bit** for short.

Some existing spaces use VO-bit to identify allocated units.  Currently they are `MallocSpace` and `MarkCompactSpace`.
-   `MallocSpace` uses VO-bit to record all the memory returned by `malloc`.
-   `MarkCompactSpace` uses VO-bit to iterate through all allocation units in the space.
These two spaces should switch to local metadata.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,11 +78,11 @@ ro_space = []
 # TODO: This is not properly implemented yet. We currently use an immortal space instead, and all our spaces have execution permission at the moment.
 code_space  = []
 
-# metadata
-global_alloc_bit = []
+# Valid-object bit.
+vo_bit = []
 
 # conservative garbage collection support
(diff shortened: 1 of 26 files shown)


wks:
This PR tries to do too many things in one go.  I think it is taking way too long.  I'll do it in smaller steps.  See: https://github.com/mmtk/mmtk-core/pull/791

PR #684, opened 2022-10-17
Fix fetch_and_atomic and fetch_or_atomic

wks:
The bit operations of sub-byte fetch_and_atomic and fetch_or_atomic was incorrect, and is fixed in this commit.

### DIFF
--- a/src/util/metadata/header_metadata.rs
+++ b/src/util/metadata/header_metadata.rs
@@ -403,11 +403,11 @@ impl HeaderMetadataSpec {
         self.assert_spec::<T>();
         if self.num_of_bits < 8 {
             let (lshift, mask) = self.get_shift_and_mask_for_bits();
-            let new_val = (val.to_u8().unwrap() | !mask) << lshift;
+            let new_val = (val.to_u8().unwrap() << lshift) | !mask;
             // We do not need to use fetch_ops_on_bits(), we can just set irrelavent bits to 1, and do fetch_and
             let old_raw_byte =
                 unsafe { <u8 as MetadataValue>::fetch_and(self.meta_addr(object), new_val, order) };
-            let old_val = (old_raw_byte >> lshift) & mask;
+            let old_val = self.get_bits_from_u8(old_raw_byte);
             FromPrimitive::from_u8(old_val).unwrap()
         } else {
             unsafe { T::fetch_and(self.meta_addr(object), val, order) }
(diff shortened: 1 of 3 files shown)


src/util/metadata/header_metadata.rs
```diff
@@ -402,12 +402,12 @@ impl HeaderMetadataSpec {
         #[cfg(debug_assertions)]
         self.assert_spec::<T>();
         if self.num_of_bits < 8 {
-            let (_, mask) = self.get_shift_and_mask_for_bits();
-            let new_val = val.to_u8().unwrap() | !mask;
+            let (lshift, mask) = self.get_shift_and_mask_for_bits();
+            let new_val = (val.to_u8().unwrap() | !mask) << lshift;
```
qinsoon, line 406: `mask` is shifted to the high order bits in 
https://github.com/mmtk/mmtk-core/blob/53b6d57c2a96f6064e01e245ea0be05dd673cfad/src/util/metadata/header_metadata.rs#L85, so `mask` should be used with the high order bits. `val` is the value on the low order bits. So it is incorrect to bitor `val` with `!mask`.

Probably we should have `(val.to_u8().unwrap() << lshift) | !mask`.

src/util/metadata/header_metadata.rs
```diff
@@ -402,12 +402,12 @@ impl HeaderMetadataSpec {
...
             // We do not need to use fetch_ops_on_bits(), we can just set irrelavent bits to 1, and do fetch_and
             let old_raw_byte =
                 unsafe { <u8 as MetadataValue>::fetch_and(self.meta_addr(object), new_val, order) };
-            let old_val = self.get_bits_from_u8(old_raw_byte);
+            let old_val = (old_raw_byte >> lshift) & mask;
```
qinsoon, line 410: I think `get_bits_from_u8()` is correct.

src/util/metadata/header_metadata.rs
```diff
@@ -425,12 +425,12 @@ impl HeaderMetadataSpec {
         #[cfg(debug_assertions)]
         self.assert_spec::<T>();
         if self.num_of_bits < 8 {
-            let (_, mask) = self.get_shift_and_mask_for_bits();
-            let new_val = val.to_u8().unwrap() & mask;
+            let (lshift, mask) = self.get_shift_and_mask_for_bits();
```
qinsoon, line 428: Same problems in this method.

src/util/metadata/side_metadata/global.rs
```diff
@@ -536,12 +536,12 @@ impl SideMetadataSpec {
                 let meta_addr = address_to_meta_address(self, data_addr);
                 if self.log_num_of_bits < 3 {
                     let lshift = meta_byte_lshift(self, data_addr);
-                    let mask = meta_byte_mask(self) << lshift;
+                    let mask = meta_byte_mask(self);
```
qinsoon, line 539: I think your code on side metadata is correct. The difference is that the mask returned from `meta_byte_mask()` is not shifted to high order bits, but the mask returned from `get_shift_and_mask_for_bits()` is shifted to high order bits. This is unfortunately confusing. Unless you would like to change one of the methods to make them consistent, I would suggest at least adding some comments to those two methods to make it clear.

src/util/metadata/header_metadata.rs
qinsoon: I think it would be good if you can add at least one test for this.

qinsoon (changes requested):
The change for header metadata does not seem correct.

src/util/metadata/header_metadata.rs
wks: Yes.  I changed as you suggested.

src/util/metadata/header_metadata.rs
wks: Yes.  I changed it back.

src/util/metadata/header_metadata.rs
wks: Fixed.

src/util/metadata/side_metadata/global.rs
wks: I change it back so that "mask" is always shifted left to the bits that contains useful data.  That will be consistent with other methods.

qinsoon (approved):
LGTM

PR #695, opened 2022-11-10
Remove From<Address> and Into<Address> for Region

qinsoon:
This PR addresses the issues we discussed in https://github.com/mmtk/mmtk-core/pull/643#discussion_r1018622963. Basically, Rust expects `From<T>` to always succeed, and Rust suggests using `TryFrom<T>` if the conversion may fail. For us, turning an address into a region may fail. So we should not use `From<T>`. And we do not need the error handling with `TryFrom<T>`. Thus we just provide two methods: `Region::from_unaligned_address()` and `Region::from_aligned_address()`.

### DIFF
--- a/src/policy/immix/block.rs
+++ b/src/policy/immix/block.rs
@@ -64,30 +64,26 @@ impl BlockState {
 }
 
 /// Data structure to reference an immix block.
-#[repr(C)]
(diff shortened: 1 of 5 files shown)


wks (approved):
LGTM

PR #697, opened 2022-11-14
Fixing bug in immix allocator

udesou:
Use maximum aligned size to account for possible offset when checking if the size to allocate is larger than a line.


k-sareen:
I think you want to change this line as well:
https://github.com/mmtk/mmtk-core/blob/c453a1cabab266a27cdd9a1e4868f465fafc1df0/src/util/alloc/immix_allocator.rs#L331

qinsoon:
@udesou and I talked about this, and it looks fine to me. I think it is better to let @wenyuzhao look at it.
Can you fix the style check? @udesou

wenyuzhao (approved):
LGTM

qinsoon:
> I think you want to change this line as well:
> 
> https://github.com/mmtk/mmtk-core/blob/c453a1cabab266a27cdd9a1e4868f465fafc1df0/src/util/alloc/immix_allocator.rs#L331

I think this should be changed as well.

wenyuzhao:
> > I think you want to change this line as well:
> > https://github.com/mmtk/mmtk-core/blob/c453a1cabab266a27cdd9a1e4868f465fafc1df0/src/util/alloc/immix_allocator.rs#L331
> 
> I think this should be changed as well.

It's also possible to round up every allocation size to a multiple of MIN_ALIGMENT, so we don't need to check the size everywhere

qinsoon:
@udesou Can you fix the style check? I think we can merge this once it is fixed.

udesou:
Fixed.

PR #698, opened 2022-11-14
Checking if the referent has been cleared

udesou:
In Julia referents are set to `jl_nothing` instead of `null` to indicate that a reference has been cleared. 
I have added a default method to the ReferenceGlue trait and replaced the checks in mmtk-core to call this method instead. Therefore, bindings can overwrite the method `is_referent_cleared` to do the check properly.


qinsoon (approved):
LGTM

PR #700, opened 2022-11-14
Generalise weak reference processing for other languages

wks:
**Draft**: I still need to make it parallelisable
-   [X] Non-MarkCompact for OpenJDK
-   [X] MarkCompact for OpenJDK
-   [X] MarkSweep and Immix for Ruby
-   [X] Let VM binding instantiate tracers (wrapping `ProcessEdgesWork`) so that it can be parallelised. 

Related repos:
-   https://github.com/wks/mmtk-openjdk/tree/gen-weakref-api: I ported the `ReferenceProcessor` and `FinalizableProcessor` from mmtk-core to mmtk-openjdk and managed to get them running. However, it is a proof-of-concept, only, and shall not be merged.
-   https://github.com/wks/mmtk-ruby/tree/gen-weakref-api: With this PR, I am able to handle `obj_free`, finalisers and global weak tables in Ruby.  As a by-product, this PR also enables mmtk-ruby to fully support Ruby finalisers (`ObjectSpace::define_finalize`).  I'll merge itfor mmtk-ruby after this PR is merged.

This PR adds a mechanism to let the VM binding handle weak references and finalisers.

Changes made in this PR include:
-   Add two work bucket stages:
    -   `VMRefClosure`: expand the transitive closure by letting the VM binding trace and/or forward edges in weak references and/or tables of finalisable objects.  The decision of which edge to trace and which edge to clear is made by the binding.
    -   `VMRefForwarding`: as part of the "compaction" step in MarkCompact, it lets the VM binding forward those edges.
-   Introduce "sentinel", a work packet to be executed after a work bucket is drained, but before the work bucket is considered to be finished.
    -   The "sentinel" can add packets to the current bucket, and set another "sentinel", so that the GC stays in the same bucket stage. This allows transitive closures to be expanded multiple times, and code to be executed after each transitive closure expansion finishes.
    -   Sentinels replace the `Scheduler::closure_end` callback, and act as a more general mechanism.
    -   The `VMProcessWeakRefs` work packet is the sentinel for `VMRefClosure` and `VMRefForwarding`.  It calls `Collection::process_weak_refs`.
-   Extend the `Collection::process_weak_refs` trait method to provide extra parameters.
    -   `worker`: The current GC worker.
    -   `context`: Provide more contexts, such as whether the current GC is a nursery GC, and whether the current trace is for marking or forwarding.
    -   `tracer_factory`: It can instantiate `ObjectTracer` and call `ObjectTracer::trace_object`.  It basically wraps `ProcessEdgesWork` and its initialisation and flushing, so that the VM binding can call `trace_object` without intimate knowledge of the `ProcessEdgesWork` trait.
    -   Also use the return value to let the MMTk core call `Collection::process_weak_refs` again when transitive closure is finished again.  This allows multiple strength levels and ephemerons to be implemented.

This PR basically resurrects and extends the mechanism introduced in https://github.com/mmtk/mmtk-core/commit/2b55f89622dcc4c293bb472c634165e2ce7c2c5c This PR kept the following mechanisms:

-   `Collection::vm_release`: This can be used to do what `RefEnqueue` used to do.
    -   Also added `Collection::vm_prepare` to make the API symmetric.

Differences are:

1.  Dedicated `VMRefClosure` and `VMRefForwarding` buckets: Weak reference processing is done in those dedicated buckets instead of `Closure`.
2.  Replaced `closure_end` with "sentinels"
    -   Sentinel can be added to any bucket.  Currently we use sentinel for `VMRefClosure` and `VMRefForwarding`.
    -   `Collection::process_weak_refs` returns a `bool` and has similar semantics as the return value of `closure_end`.

# Rationale

Why do we introduce another mechanism while MMTk core has ReferenceProcessor and FinalizableProcessor?

1.  Those mechanisms are Java-centric, and does not match other languages/VMs well.  For example,
    -   V8 has ephemeron and 8 different kinds of weak references
    -   Ruby's `obj_free` is like Java `finalize()` but doesn't resurrect objects;
        Ruby's `ObjectSpace.define_finalize` is like `PhantomReference` but can be unregistered;
        Ruby's `finalizer_table`, `generic_iv_tbl_`, `id_to_obj_tbl` and `obj_to_id_tbl` are weak tables.
2.  Even for Java, our current reference/finalizable processors are brought from JikesRVM.  Although they are compatible with OpenJDK or other JVMs, they are implemented differently, and the performance and semantics have subtle differences.  For this reason, if we compare MMTk with the vanilla GC in the JVM (such as G1 in OpenJDK), the difference in performance will not only reflect the difference of the GC mechanism, but also the difference in referene/finalizable processors, which is undesirable.  We should be able to make apple-to-apple comparisons between, for example, a G1 plan in MMTk and the default G1 GC in OpenJDK, using the same reference/finalizable processors.

# Old discussions

## Known issue: can we parallelise `Collection::process_weak_refs`?

(Update: I added a "TracerFactory" so that different GC workers can instantiate `ObjectTracer` and call `trace_object`.)

In this PR, `Collection::process_weak_refs` is called in one work packet running on one GC worker.  This means `process_weak_refs` *itself* is executed sequentially.  However, after it calls `trace_object` multiple times, the underlying `ProcessEdgesWork` may split the node list into multiple work packets.  One question will be whether we should make `process_weak_refs` parallel, too, so we can process weak references in multiple work packets simultaneously.

Currently, the `ReferenceProcessor` and `FinalizableProcessor` in mmtk-core does not paralellise their `ReferenceProcessor::scan_{soft,weak,phantom}_refs` and `FinalizableProcessor::scan` methods.  So when we reimplement our current `ReferenceProcessor` and `FinalizableProcessor` in the binding on top of the new mechanism in this PR (See [1]), it has the same level of parallelism as before.

[1]: https://github.com/wks/mmtk-openjdk/blob/gen-weakref-api/mmtk/src/weak_processor/mod.rs#L45

But from my observation, the lxr branch is parallelising weak reference processing by creating multiple work packets (See [2]) that have intimate knowledge about the `ProcessEdgesWork` work packet.

[2]: https://github.com/wenyuzhao/mmtk-openjdk/blob/lxr/mmtk/src/reference_glue.rs#L151

I hesitate to expose `ProcessEdgesWork` to the VM binding, because I think it is an implementation detail of mmtk-core, and GC algorithms can actually implement node-enqueuing tracing instead of edge-enqueuing tracing.  All the VM binding needs is the ability to call `trace_object` (and I defined the `ProcessWeakRefsTracer` trait this way. See [3]).

[3]: https://github.com/mmtk/mmtk-core/blob/ba194ba0dc047a6e2900ed3b182a2ba8e6f0d9f4/src/vm/collection.rs#L25

The problem is, for our current mmtk-core, we can only implement `ProcessWeakRefsTracer` via `ProcessEdgesWork`.  In this PR, I implemented it by wrapping a **reference** to `ProcessEdgesWork` to it, calling `set_worker` before calling `Collection::process_weak_refs`, and getting its nodes list after it. See [4].

[4]: https://github.com/mmtk/mmtk-core/blob/ba194ba0dc047a6e2900ed3b182a2ba8e6f0d9f4/src/scheduler/gc_work.rs#L288

But if we wrap a `ProcessEdgesWork` **instance** inside a `ProcessWeakRefsTracer`, we will have to expose `set_worker` and the flushing to the VM binding, and I think it is not very elegant.

```rust
impl Collection for FooVMBinding {
    fn process_weak_refs(worker: &mut GCWorker, tracer: impl ProcessWeakRefsTracer) -> bool {
        trace.set_worker(worker); // This line may confuse VM binding writers.
        let work = make_another_work_packet(tracer, get_some_weak_refs());
        worker.scheduler().work_buckets[...].add(work)
    }
}

struct AnotherWorkPacket<T: ProcessWeakRefsTracer> {
    tracer: T,
}
impl GCWork for AnotherWorkPacket {
    fn do_work(&mut self, worker: &mut GCWorker<FooVMBinding>, mmtk: &'static MMTK<FooVMBinding>) {
        self.tracer.set_worker(worker); // This is confusing, too.
        self.tracer.trace_object(...);
        self.tracer.flush(); // This is confusing, too, but can be moved to `drop()`.
    }
}
```


src/vm/collection.rs
```diff
@@ -4,12 +4,25 @@ use crate::util::opaque_pointer::*;
...
 pub enum GCThreadContext<VM: VMBinding> {
     Controller(Box<GCController<VM>>),
     Worker(Box<GCWorker<VM>>),
 }
 
+/// Information related to weak reference processing.  Used by `Collection::process_weak_refs`.
+pub struct ProcessWeakRefsContext {
+    /// `true` if `process_weak_refs` is called during the forwarding phase in MarkCompact.
+    /// Always `false` if the GC is not MarkCompact.
+    pub forwarding: bool,
```
qinsoon, line 19: You could just separate `process_weak_refs` and `forward_weak_refs`. Based on your [binding code](https://github.com/wks/mmtk-openjdk/blob/c962cab7f609ecf585360e2d48305e59e8770389/mmtk/src/weak_processor/mod.rs#L59), the implementation of process and forward are different as well.

The argument from the point of view of MMTk core is that for forwarding, we should not discover new weak refs, and we do not need a boolean return value. So it is a bit confusing to merge them into one method/type/etc. This also affects `VMProcessWeakRefs`.

src/vm/collection.rs
```diff
@@ -99,8 +119,65 @@ pub trait Collection<VM: VMBinding> {
...
+    ///
+    /// Arguments:
+    /// * `worker`: The current GC worker.
+    /// * `context`: Provides more information of the current trace.
+    /// * `tracer_factory`: Use this to create an `ObjectTracer` and use it to retain and update
+    ///   weak references.
+    ///
+    /// This function shall return true if this function needs to be called again after the GC
+    /// finishes expanding the transitive closure from the objects kept alive.
+    fn process_weak_refs(
```
qinsoon, line 176: You could consider moving this to `Scanning`. `ObjectTracer` and related types/methods are all in `Scanning` -- this could be a good indication that this method should be there as well. You may also rename this as `scan_weak_refs`.

qinsoon (changes requested):
I think it is worthwhile to run some performance measurements (maybe two different runs): 1. 
 make sure that leaving the VM-side weak ref processing on (but with an empty implementation) does not cost anything (this will tell whether we need a switch for it), 2. make sure that moving the current reference processor to the binding side does not cost anything (this will tell us the efficiency of the new API). The latter is more important.

See my inline-comments for a few other issues.

src/scheduler/gc_work.rs
```diff
@@ -186,6 +188,11 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for StopMutators<E> {
             mmtk.scheduler.work_buckets[WorkBucketStage::Prepare].add(ScanStackRoot::<E>(mutator));
         });
         trace!("stop_all_mutators end");
+
+        trace!("VM prepare start");
+        <E::VM as VMBinding>::VMCollection::vm_prepare(worker.tls);
```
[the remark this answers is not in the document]
wks: It is just added for symmetry.  We had `vm_release` but not `vm_prepare`.  In `vm_release`, we do the same thing as `RefEnqueue`.

But since it is the same point as `Collection::stop_all_mutators`, I'll remove this and use `Collection::stop_all_mutators` instead.

src/vm/collection.rs
```diff
@@ -99,8 +119,65 @@ pub trait Collection<VM: VMBinding> {
...
+        _worker: &mut GCWorker<VM>,
+        _context: ProcessWeakRefsContext,
```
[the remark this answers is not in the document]
wks: Yes.  That makes sense.  I think we don't need to compute the transitive closure multiple times during forwarding because the life and death of all objects have been determined.  So it makes sense to split it into two functions.

src/scheduler/gc_work.rs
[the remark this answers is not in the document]
qinsoon: I think it is fine to add `vm_prepare` if it is called in `Prepare`. Otherwise, it is confusing.

src/vm/scanning.rs
```diff
@@ -31,6 +32,41 @@ impl<F: FnMut(ObjectReference) -> ObjectReference> ObjectTracer for F {
     }
 }
 
+/// A `QueuingTracerFactory` can create a "tracer" object local to a GC Worker to trace and
+/// update edges, and create `ScanObjects` work packets to expand the transitive closure starting
+/// from newly traced objects.  It is used during root scanning and binding-side weak reference
+/// processing.
+pub trait QueuingTracerFactory<VM: VMBinding>: Clone + Send + 'static {
```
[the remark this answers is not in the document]
wks: A "tracer" (i.e. `ObjectTracer`) provides the `trace_object` method.  "Queuing" means adding the object into a buffer if it is first visited during a transitive closure computing.  While most use cases of `trace_object` have this queuing behaviour, `LXRConcurrentTraceObjects` behaves differently because it enqueues the *children* of the visited object (See https://github.com/wenyuzhao/mmtk-core/blob/lxr/src/plan/lxr/cm.rs#L148).  But it is still a different style of enqueuing.

I think it should be safe to assume that the `trace_object` always has some kind of queuing behaviour, therefore we can omit the word "queuing" from its name.

The word "Factory" here looks odd, too, for the reason you described, although it actually creates `ObjectTracer`.  I think the name `ObjectTracerContext` is better.  `ScopedTracer` may be another option, although a `ScopedTracer` itself is not a tracer.  p.s. I originally designed it like a factory that returns `ObjectTracer` instances, but that'll need the GAT feature of newer Rust and it'll need manual flushing.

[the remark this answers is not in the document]
wks: > I am not quite comfortable about adding a new type for each API method.

I introduced `ProcessWeakRefsContext` for a purpose.  My concern is that if we extend the `process_weak_refs` method and provide more information (in addition to `forwarding` and `is_nursery`) through its parameters, existing VM bindings will need to update their source code for the added parameters even if they don't use the new information.  With a `ProcessWeakRefsContext` struct, we can safely add new fields just in case some VMs need more information.

For example, if we add a parameter:
```diff
-fn process_weak_refs(forwarding: bool) -> bool;
+fn process_weak_refs(forwarding: bool, nursery: bool) -> bool;
```

Existing VM bindings will need to add the parameter even if it is not using that information.

```diff
 impl Collection for OpenJDK {
-    fn process_weak_refs(forwarding: bool) -> bool {
+    fn process_weak_refs(forwarding: bool, _nursery: bool) -> bool {
 // ... more code here
```

But if we provide a struct in the first place, we just add fields to that struct.

```diff
 struct ProcessWeakRefsContext
     forwarding: bool,
+    nursery: bool,
 }
 trait Collection {
     fn process_weak_refs(context: ProcessWeakRefsContext) -> bool; // This doesn't change.
```

> Do you have an idea what other API methods may use this type? Can this type be used in anyway when a binding wants to process node/edges immediately in stack root scanning?

Yes.  It will be used for root scanning as well.  I'll make another PR to extend the root scanning API (the `RootsWorkFactory` trait) so that it can provide the functionality of `ObjectTracerContext`.  I'll define it like `trait RootsWorkFactory : ObjectTracerContext` so that we can reuse this `ObjectTracerContext` trait.

> My general preference is to add methods to existing types, and make them more general. I am exactly sure if that is applicable here.

There is a risk.  This is exactly why we had the `TransitiveClosure` trait 15 years ago, because Daniel wanted to introduce concurrent mark-sweep, and wanted to make a general interface (See https://wks.github.io/blog/2022/05/16/fifteen-years-transitiveclosure.html#but-why-did-we-migrate-away-from-it).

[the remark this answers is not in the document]
qinsoon: If it will be reused, then I am okay with it. 

About `ProcessWeakRefsContext`, yeah, that is the same idea as I tidied up the arguments for creating plans/spaces. I am wondering what other arguments we may pass to a binding in the context struct. One concern of mine is abstraction leakage. I don't feel that we want to the binding to know too much about intimate knowledge on the plan/GC algorithm. I know we have `plan.is_current_gc_nursery()` which is not good either.

src/vm/collection.rs
wks: Yes.  That makes sense.  I believe `process_weak_refs` should be more similar to the `scan_xxx_roots` methods, and it should be OK to move it to the `Scanning` trait.

src/vm/collection.rs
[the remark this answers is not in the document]
wks: I moved this method to the Scanning trait and split it into `process_weak_refs` and `forward_weak_refs`.

src/vm/scanning.rs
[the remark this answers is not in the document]
wks: I renamed this trait to `ObjectTracerContext`, and its method to just `with_tracer`.

I also managed to reuse the code in `ScanObjects` around the call to `Scanning::scan_object_and_trace_edges`.  They happen to have the same logic, and they both use the `ObjectTracer` trait.

src/scheduler/gc_work.rs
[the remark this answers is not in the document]
wks: I removed `vm_prepare` because I only need it to do assertion.  I moved the assertion to `stop_all_mutators`.  If we later find that we need `vm_prepare`, we can add it back.

src/scheduler/gc_work.rs
```diff
@@ -245,24 +252,122 @@ impl<VM: VMBinding> GCWork<VM> for EndOfGC {
 
 impl<VM: VMBinding> CoordinatorWork<VM> for EndOfGC {}
 
+struct ProcessEdgesWorkTracer<E: ProcessEdgesWork> {
```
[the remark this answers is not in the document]
wks: I added comments to this and `ProcessEdgesWorkTracerContext` as well.

src/vm/collection.rs
```diff
@@ -4,12 +4,25 @@ use crate::util::opaque_pointer::*;
...
 
+/// Information related to weak reference processing.  Used by `Collection::process_weak_refs`.
+pub struct ProcessWeakRefsContext {
+    /// `true` if `process_weak_refs` is called during the forwarding phase in MarkCompact.
+    /// Always `false` if the GC is not MarkCompact.
+    pub forwarding: bool,
+
+    /// `true` if the current GC is a nursery GC.
+    /// Always `false` if not using a generationl GC algorithm.
+    pub nursery: bool,
```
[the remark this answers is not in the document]
wks: I removed the nursery from the parameter.  Since I also split the closure and forwarding functions, we no longer have `ProcessWeakRefsContext` now.

src/vm/collection.rs
wks: Done.

src/vm/collection.rs
wks: Done.

src/scheduler/scheduler.rs
```diff
@@ -230,6 +205,29 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
+        // processing them together.
+
+        // VM-specific weak ref processing
+        self.work_buckets[WorkBucketStage::VMRefClosure]
+            .set_sentinel(Box::new(VMProcessWeakRefs::<C::ProcessEdgesWorkType>::new()));
+
+        if plan.constraints().needs_forward_after_liveness {
+            // VM-specific weak ref forwarding
+            self.work_buckets[WorkBucketStage::VMRefForwarding]
+                .set_sentinel(Box::new(VMForwardWeakRefs::<C::ProcessEdgesWorkType>::new()));
```
qinsoon, line 229: Does `VMForwardWeakRefs` still need to be a sentinel? I think you may just add it as a normal work packet.

src/scheduler/scheduler.rs
```diff
@@ -230,6 +205,29 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
+        // as strong references, but it is not appropriate to simply disable weak reference
+        // handling from MMTk's side.  The VM, however, may choose to do nothing in
+        // `Collection::process_weak_refs` if appropriate.
+        //
+        // It is also not sound for MMTk core to turn off weak
+        // reference processing or finalization alone, because (1) not all VMs have the notion of
+        // weak references or finalizers, so it may not make sence, and (2) the VM may
+        // processing them together.
+
+        // VM-specific weak ref processing
```
qinsoon, line 222: Maybe explain very briefly why `VMProcessWeakRefs` is a 'sentinel'.

qinsoon:
The code looks good to me. I suggest you get a review from Wenyu as well, as he wrote most of the scheduler code. Also I would like to see performance comparison between the current ref processor in MMTk core and moving the ref processor to OpenJDK using the API introduced in this PR. Optionally, you can run benchmarks to check if there is any overhead if we have an empty the VM ref process/forward work packet -- the cost of the work packet should be minimal, but I am concerned of the synchronisation cost of those two extra work buckets.

src/scheduler/gc_work.rs
```diff
@@ -245,24 +247,148 @@ impl<VM: VMBinding> GCWork<VM> for EndOfGC {
...
+            };
+            <E::VM as VMBinding>::VMScanning::process_weak_refs(worker, tracer_factory)
+        };
+
+        if need_to_repeat {
+            // Schedule Self as the new sentinel so we'll call `process_weak_refs` again after the
+            // current transitive closure.
+            let new_self = Box::new(Self::new());
+
+            worker.scheduler().work_buckets[stage].set_sentinel(new_self);
```
wenyuzhao, line 356: If I understand correctly, this is used for performing weak-ref tracing in a loop. You can use the sentinel packet to start the transitive closure again and again until reaching a steady state.

However, I think OpenJDK may also use the sentinel packet as a "state transition" to start a new phase for a different type of weak-ref processing. For example, OpenJDK does soft-ref processing first (which will start a round of closure), and then phantom processing later (and another round of closure).  If the VM is able to set a different sentinel packet itself other than `new_self `, after soft-ref processing, it can set a phantom sentinel packet to start phantom processing. Is it doable?

src/vm/scanning.rs
```diff
@@ -31,6 +32,44 @@ impl<F: FnMut(ObjectReference) -> ObjectReference> ObjectTracer for F {
...
+    ///     to know when the computing of transitive closure finishes.
+    ///
+    /// Arguments:
+    /// -   `worker`: The current GC worker.
+    /// -   `func`: A caller-supplied closure in which the created `ObjectTracer` can be used.
+    ///
+    /// Returns: The return value of `func`.
+    fn with_tracer<R, F>(&self, worker: &mut GCWorker<VM>, func: F) -> R
+    where
+        F: FnOnce(&mut Self::TracerType) -> R;
```
wenyuzhao, line 70: Just double confirming, looks like this will not introduce any dynamic dispatch?

src/vm/scanning.rs
```diff
@@ -179,4 +218,70 @@ pub trait Scanning<VM: VMBinding> {
...
+    ///
+    /// Arguments:
+    /// * `worker`: The current GC worker.
+    /// * `tracer_context`: Use this to get access an `ObjectTracer` and use it to retain and
+    ///   update weak references.
+    ///
+    /// This function shall return true if this function needs to be called again after the GC
+    /// finishes expanding the transitive closure from the objects kept alive.
+    fn process_weak_refs(
+        _worker: &mut GCWorker<VM>,
```
wenyuzhao, line 266: Not for this PR. But I think in the future, we may need to put the GC worker reference as a thread-local variable. So we can always access the current worker by calling `GCWorker::current()` instead of passing it everywhere.

src/scheduler/scheduler.rs
wks: It no longer has to.  It can be a normal work packet.  I'll change it.

src/scheduler/gc_work.rs
wks: To implement state transition, we can embed a state machine in the weak reference processor, and multiplex the operations for different states through the single `Scanning::process_weak_refs` API. See:  https://github.com/wks/mmtk-openjdk/blob/gen-weakref-api/mmtk/src/weak_processor/mod.rs#L61

It is possible to add a way to specify the next work packet to be scheduled as the "sentinel" packet.  But that may be more complicated (as it has to expose details such as "sentinel") than simply returning `true` and letting the function be called again.

src/vm/scanning.rs
wks: No.  It doesn't introduce dynamic dispatch (unless you see the "dyn" keyword).  This `with_tracer<R, F>` function is parameterised with the `<F>` type, so the call site of `func()` knows exactly which function body it is (and it is even possible to inline the call).

src/vm/scanning.rs
wks: Yes.  I am considering this very possibility.  Java 20 introduced an experimental feature "scoped value" (See https://openjdk.org/jeps/429), and I wonder if the same is possible for Rust.

wks:
>... Optionally, you can run benchmarks to check if there is any overhead if we have an empty the VM ref process/forward work packet -- the cost of the work packet should be minimal, but I am concerned of the synchronisation cost of those two extra work buckets.

I am running some benchmarks. Build1 is the master branch of mmtk-core, and build2 is this PR.  mmtk-openjdk and openjdk are the same, so the new work packets are not really used.  It is still running in progress, but the current results show that the difference in STW time is trivial for Immix, but there is a visible (but small) difference for GenImmix.  I am not sure it is really costly because the noise is also large.

URL: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2022-12-21-Wed-121614&benchmark^build^invocation^iteration^mmtk_gc&time^time.stw&|50&time.stw^1^0.1^+inf|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|40&Histogram%20(with%20CI)^build^benchmark&time.stw

src/scheduler/scheduler.rs
wks: Done. It's no longer a sentinel.

src/scheduler/scheduler.rs
wks: Done.

p.s. Strictly speaking, the first `VMProcessWeakRefs` packet added here doesn't have to be a sentinel, because it is the only packet in that bucket at that moment.  But I set it as sentinel anyway to avoid confusion, so that `VMProcessWeakRefs` is always a sentinel.

wks:
> Also I would like to see performance comparison between the current ref processor in MMTk core and moving the ref processor to OpenJDK using the API introduced in this PR. Optionally, you can run benchmarks to check if there is any overhead if we have an empty the VM ref process/forward work packet -- the cost of the work packet should be minimal, but I am concerned of the synchronisation cost of those two extra work buckets.

I ran lusearch with the following configurations:
- configs:
  - "build1|ms|s|c2|mmtk_gc-Immix|tph|probes_cp|probes"
  - "build1|ms|s|c2|mmtk_gc-Immix|tph|probes_cp|probes|rt"
  - "build2|ms|s|c2|mmtk_gc-Immix|tph|probes_cp|probes"
  - "build3|ms|s|c2|mmtk_gc-Immix|tph|probes_cp|probes"
  - "build1|ms|s|c2|mmtk_gc-GenImmix|tph|probes_cp|probes"
  - "build1|ms|s|c2|mmtk_gc-GenImmix|tph|probes_cp|probes|rt"
  - "build2|ms|s|c2|mmtk_gc-GenImmix|tph|probes_cp|probes"
  - "build3|ms|s|c2|mmtk_gc-GenImmix|tph|probes_cp|probes"
  - "build1|ms|s|c2|mmtk_gc-MarkCompact|tph|probes_cp|probes"
  - "build1|ms|s|c2|mmtk_gc-MarkCompact|tph|probes_cp|probes|rt"
  - "build2|ms|s|c2|mmtk_gc-MarkCompact|tph|probes_cp|probes"
  - "build3|ms|s|c2|mmtk_gc-MarkCompact|tph|probes_cp|probes"

where
- "build1" is mmtk-core master + mmtk-openjdk master. This is the baseline.
- "build2" is mmtk-core gen-weakref-api + mmtk-openjdk master. This enables the new buckets, but they are not used.
- "build3" is mmtk-core gen-weakref-api + mmtk-openjdk gen-weakref-api. This enables the new buckets, and the openjdk binding uses the new API instead of the existing API in mmtk-core.
- the `rt` modifier sets EnvVar `MMTK_NO_REFERENCE_TYPES=false`.

The following plot compares `build{1,2,3}` without the `rt` modifier.
(Plotty URL: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2022-12-22-Thu-134512&benchmark^build^buildstring^invocation^iteration^mmtk_gc^rt&time.stw&|10&iteration^1^4&rt^1^null|50&time.stw^2^-inf^0|20&1^invocation|30&1&mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&)

image

Comparing build1 and build2 (without the rt modifier), It shows the impact of the added empty bucket is uncertain.  It reduces the STW time for Immix, which may be due to noise because it does strictly more work than before.  It increases the STW time for GenImmix. I think GenImmix triggers GC more often, and may amplify the cost.   It increases the STW time in MarkCompact, maybe because it also has a forwarding stage that is also affected.

build3 uses the new API to handle weak references and finalisation.  Without the `rt` flag, it is not an apple-to-apple comparison, because by default mmtk-core does not enable reference processing, but my gen-weakref-api branch of mmtk-openjdk always does reference processing and finalisation.

The following plot compares build1, build1+rt, build2 and build3 under Immix.
(Plotty URL: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2022-12-22-Thu-134512&benchmark^build^buildstring^invocation^iteration^mmtk_gc^rt&time.stw&|10&iteration^1^4&mmtk_gc^1^Immix|50&time.stw^2^-inf^0|20&1^invocation|41&Histogram%20(with%20CI)^buildstring^&time.stw)

image

Enabling reference processing increases the STW time of build1 (master+master).  Comparing build1+rt vs build3 (both doing both reference processing and finalisation), we see the STW time decreased.  That may be because the new API only has one bucket (albeit using "sentinel" to multiplex multiple reference-processing phases (soft, weak, final, phantom)) while the existing API needs multiple buckets.  It could be noise, too, because there may be subtle differences between the implementation in mmtk-core and that in my mmtk-openjdk branch.

The same plot for build1, build1+rt, build2 and build3 under GenImmix
(Plotty URL: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2022-12-22-Thu-134512&benchmark^build^buildstring^invocation^iteration^mmtk_gc^rt&time.stw&|10&iteration^1^4&mmtk_gc^1^GenImmix|50&time.stw^2^-inf^0|20&1^invocation|41&Histogram%20(with%20CI)^buildstring^&time.stw)

image

(Note that the zero points are different.)

This result is similar to Immix.  

The same plot for build1, build1+rt, build2 and build3 under MarkCompact
(Plotty URL: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2022-12-22-Thu-134512&benchmark^build^buildstring^invocation^iteration^mmtk_gc^rt&time.stw&|10&iteration^1^4&mmtk_gc^1^MarkCompact|50&time.stw^2^-inf^0|20&1^invocation|41&Histogram%20(with%20CI)^buildstring^&time.stw)

image

(Note that the zero points are different.)

With MarkCompact, STW time is shorter (not by much.  Note that the scale and zero point are different) when reference processing is enabled.  My speculation is that reference processing can recycle dead weak references so that the GC only needs to compact a smaller amount of memory.

The conclusion is
-   The impact of no-op work packets is minimal.
-   The actual work executed during reference processing has more impact than the scheduling overhead.

wks:
I ran the DaCapo Chopin benchmark suite, using Immix only (it would take days if I ran all of Immix, GenImmix and MarkCompact), 20 invocations.  The following plot shows the STW time of each build, normalised to build1-null (green).

-   build1-null: mmtk-core master + mmtk-openjdk master; no reference processing
-   build1-1: mmtk-core master + mmtk-openjdk master; enabled reference processing
-   build2-null: mmtk-core gen-weak-api + mmtk-openjdk master; no reference processing
-   build3-null: mmtk-core gen-weak-api + mmtk-openjdk gen-weak-api; reference processing implemented in the binding using the new API, enabled



The difference between build1-null (green) and build2-null (yellow) shows the overhead of the unused buckets and work packets of the new API.  The histogram shows the difference is usually trivial, except h2, which is noisy anyway.

The difference between build1-null (green) and build1-1 (blue) shows the impact of only enabling reference processing in the master branches.  The impact highly depends on the concrete benchmark.

The difference between build1-1 (blue) and build3-null (purple) shows the difference of implementing reference processing in mmtk-core as in master vs implementing it in the binding using the new API.  The plot shows the STW time is usually shorter (except h2) when implementing reference processing using the new API.  But it is hard to explain why, because I copied and pasted the reference processor and finaliser processor from mmtk-core to mmtk-openjdk, with minor changes.

It is interesting to observe that enabling reference processing in mmtk-core often makes STW time longer (than build1-null), but enabling reference processing in the binding makes STW time shorter.

wks:
Well, the total time shows there is not much difference between build1-1 and build3-null (both enable reference processing, and differ only by whether it implements reference processing in mmtk-core or mmtk-openjdk).  Maybe changing the API changed the way we calculate the STW time, while the program is actually doing the same amount of work.

image

qinsoon:
> The difference between build1-1 (blue) and build3-null (purple) shows the difference of implementing reference processing in mmtk-core as in master vs implementing it in the binding using the new API. The plot shows the STW time is usually shorter (except h2) when implementing reference processing using the new API. But it is hard to explain why, because I copied and pasted the reference processor and finaliser processor from mmtk-core to mmtk-openjdk, with minor changes.

This is a bit strange. For example, for `zxing`, there seems to be a lot of work done for ref processing, as the reference processor in mmtk-core slows down the STW a lot. It is strange that the cost just magically disappeared by moving the reference processor to the binding. Can you double check that `build3-null` actually processes the weak references properly? For example, you will still need to set `no_reference_types=false` for `build3-null`, because of these lines in the binding:
https://github.com/mmtk/mmtk-openjdk/blob/b01e1cb235027b5441e46ef4e3c1969875ad2bff/mmtk/src/object_scanning.rs#L116
https://github.com/mmtk/mmtk-openjdk/blob/b01e1cb235027b5441e46ef4e3c1969875ad2bff/mmtk/src/object_scanning.rs#L139

wks:
> This is a bit strange. For example, for `zxing`, there seems to be a lot of work done for ref processing, as the reference processor in mmtk-core slows down the STW a lot. It is strange that the cost just magically disappeared by moving the reference processor to the binding. Can you double check that `build3-null` actually processes the weak references properly? For example, you will still need to set `no_reference_types=false` for `build3-null`, because of these lines in the binding: https://github.com/mmtk/mmtk-openjdk/blob/b01e1cb235027b5441e46ef4e3c1969875ad2bff/mmtk/src/object_scanning.rs#L116 https://github.com/mmtk/mmtk-openjdk/blob/b01e1cb235027b5441e46ef4e3c1969875ad2bff/mmtk/src/object_scanning.rs#L139

Indeed. That's a problem.  build3-null is still treating weak references as strong if `MMTK_NO_REFERENCE_TYPES` is not set to false.  I'll re-run zxing and lusearch with and without `rt` on all builds.

wks:
The results: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|boar-2023-01-03-Tue-054343&benchmark^buildstring^invocation^iteration&GC^time^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;build1.ms.s.c2.mmtk_gc-Immix.tph.probes_cp.probes.dacapochopin-f480064|41&Histogram%20(with%20CI)^buildstring^benchmark&

Re-run lusearch (233M heap size) and zxing (672M heap size).  This time, I apply `rt` to all three builds.

STW time (normalised to build1-null):
image

time (normalised to build1-null):
image

Number of GC (not normalised) (FYI: zxing consistently triggers GC 5 times in all invocations)
image

The result shows that the STW time is significantly greater when reference processing is enabled (`MMTK_NO_REFERENCE_TYPES=false`).  The work of processing weak references doesn't simply disappear when we move reference processing to the binding side.

The STW time without reference processing is similar among builds, and the STW time with reference processing is also similar among builds. For lusearch, enabling reference processing increases the STW time by approx. 30%, and for zxing, 60%.  I tried to decrease the heap size.  The increase in STW time is similar when the heap is smaller (160M for lusearch and 463M for zxing). zxing doesn't run with smaller heap sizes.

The STW time takes up a small fraction of total time, but is still visible when turning `rt` on and off.

There is an observable overhead in build3-rt. The following plot is an un-aggregated plot that shows the STW time of each invocation separately. We can see that the increase is consistent, and not due to outliers.

image

In build3-rt, reference processing is implemented on the binding side. I am not sure if the overhead comes from the inefficiency of this new API or any slight difference between the implementation of reference processor in mmtk-core and my mmtk-openjdk branch.  But we probably needs to reimplement the reference processor anyway, using OpenJDK's own mechanism instead of JikesRVM's, likely using the code in the lxr branch.

wks:
Added build4: mmtk-core is gen-weakref-api but without adding any existing reference/finalizer-processing work packets mmtk-core at all.  mmtk-openjdk is also gen-weakref-api, i.e. ref/finalizer-processors implemented in the binding, the same as build3.  This will remove the interference of mmtk-core's existing reference/finalizer work packets when `MMTK_NO_REFERENCE_TYPES=false`.

I ran lusearch and zxing, 40 iterations.

STW time: (normalised to build1-null)
image

Total time: (normalised to build1-null)
image

Note that the order of the bars is not 1, 2, 3 and 4.

The STW time for build4-1 (black) is slightly lower than build3-1 (blue), but still noticeably higher than build2-1 and build1-1.  This indicates that the influence of unused/no-op work packets is small, but there is another source of overhead to be discovered.

# RefEnqueue and Release

One difference I noticed is that
-   With mmtk-core's existing reference processors, the `RefEnqueue` work packet is added to the `WorkBucketStage::Release` bucket.  
-   With this PR, the `Release` work packet will
    1.  first call `VMCollection::vm_release` which the VM can implement to do what `RefEnqueue` does, and
    2.  after `VMCollection::vm_release` returns, do the rest of the work in `Release`, such as calling `plan_mut.release`.

We can see that the old method allows `RefEnqueue` to be executed in parallel with the `Release` work packet, but the new API forces `VMCollection::vm_release` to compete before running the reset of `Release`.  **If both `VMCollection::vm_release` and `plan.release` are time-consuming, this sequential constraint will likely to introduce a non-trivial overhead*.

An obvious solution is adding a `VMRelease` work packet to the `WorkBucketStage::Release` bucket so that `VMCollection::vm_release` can be executed in a different work packet than the `Release` packet.

Alternatively, the VM can spawn a new work packet in `VMCollection::vm_release` so as not to block any subsequent work.  I prefer this solution because it still allows the VM do "release" work which is guaranteed to happen before `plan_mut.release()`.  I don't know what VM needs this.

It is easy to check.  I'll run the benchmark again with a **build5** that parallelises `VMCollection::vm_release` and `Plan::release`.

There is one risk, though, because we expect the `Release` work packet to be the only work packet that accesses the `Plan` instance.  Paralellising the work may introduce a chance of racing, but the VM is not supposed to access `Plan` directly anyway.

wks:
In build6, the binding side creates another work packet in `vm_release`, making the work parallelisable.  However, experiment shows it makes the performance worse.

STW time (normalised to build1-null)
image

Total time (normalised to build1-null)
image

It is also worth noting that the total execution time has greater variance in build4 (removed existing ref-processing work packets) and build5 (parallelising ref-enqueue and plan-release).

I'll look into the details.

wks:
I constructed a micro benchmark that creates 100000 weak reference. 50000 of them point to live objects, and the rest point to dead objects, and all of them has a ReferenceQueue to enqueue to when they die.

```java
public class RefProcTest {
    public static Probe probe = new RustMMTkProbe();
    public static int unused;

    public static void doTest(int total, int live) {
        // Java doesn't like `WeakReference<Object>[total]`.
        var refs = new ArrayList<WeakReference<Object>>(total);
        var liveObjs = new ArrayList<Object>(live);
        var refQueue = new ReferenceQueue<Object>();

        for (int i = 0; i < total; i++) {
            var obj = new Object();
            var ref = new WeakReference<Object>(obj, refQueue);
            refs.add(ref);
            if (i < live) {
                liveObjs.add(obj);
            }
        }

        System.gc(); // I know this always triggers GC with MMTk.  It is not generally true, though.

        unused = refs.size() + liveObjs.size(); // just use them, in case GC treat them as garbage.
    }

        final int total;
        final int live;
        final int warmups;
        final int iterations;

        // ... parse them from args

        for (int i = 0; i < warmups; i++) {
            doTest(total, live);
        }

        probe.begin("RefProcTest", 0, false);
        for (int i = 0; i < iterations; i++) {
            doTest(total, live);
        }
        probe.end("RefProcTest", 0, false);
    }
}
```

I ran them locally with the settings of build1 (master+master) and build3 (mmtk-core has new API, and mmtk-openjdk uses the new API), with `total=100000`, `live=50000`, `warmups=10` and `iterations=100`.  `MMTK_PLAN=Immix`, `MMTK_NO_REFERENCE_TYPES=false`.  The heap is large enough so that it never triggers GC unless I call `System.gc()`.  The following table is the result. Time unit is ms.

packet|build1|build3|diff
---|---|---|---
SoftRefProcessing    |    0.998000     |   0.105000     | -0.893000
WeakRefProcessing   |   890.801000   |     0.043000     | -890.758000
Finalization   |     0.525000    |    0.381000     | -0.144000
PhantomRefProcessing  |      0.175000   |     0.047000     | -0.128000
VMProcessWeakRefs     |   0.038000   |     5.626000     | 5.588000
RefEnqueue    |   45.841000   |    0.090000     | -45.751000
Release   |     3.873000  |     37.334000     | 33.461000
PlanProcessEdges   | 36934.860000  |  39979.562000     | 3044.702000
ProcessWeakRefsWork     |       n/a    | 1008.507000     | 1008.507000

*p.s. (1) Other work packets took very little time. (2) The execution time varies a lot. Vanilla HotSpot with its default G1 GC has large variation, too, probably because of the nondeterministic nature of weak references themselves.  Despite of that, the data above is quite representative for a typical execution.*

Among the work packets, `VMProcessWeakRefs` is introduced by the new API, and it should do all the work of `{Soft,Weak,Phantom}RefProcessing` and `Finalization`.  The work of `RefEnqueue` is moved into `Release`.  

`ProcessWeakRefsWork` is special.  When doing experiment, I intentionally lets the binding create a separate work packet (`ProcessWeakRefsWork`) when processing weak references (not soft, final or phantom), just to test if my new API works if the binding splits the work into multiple work packets.  It is not necessary because I can just process weak refrences in `VMProcessWeakRefs` itself.

From that table, I can identify two sources of inefficiency.
1.  (minor) `ProcessWeakRefsWork` is not necessary. I can just let `VMProcessWeakRefs` handle weak references and eliminate one work packet creation.  But the impact of this is small.
2.  (minor) We could have a separate work packet that does the equivalent of `RefEnqueue`.  But as we can see from the table, the impact is also small.
3.  (major) The time consumed by `PlanProcessEdges` is dramatically increased.  The impact of this is greater than anything else.

I looked into the code, and I found the difference between my API and the existing `ReferenceProcessor` is how we flush `ProcessEdgesWork`.

The existing `ReferenceProcessor` simply calls `ProcessEdgesWork::flush()`:

```rust
impl<E: ProcessEdgesWork> GCWork<E::VM> for WeakRefProcessing<E> {
        let mut w = E::new(vec![], false, mmtk);
        w.set_worker(worker);
        mmtk.reference_processors.scan_weak_refs(&mut w, mmtk);
        w.flush();  // here
    }
}
```

On the other hand, my API attempted to wrap `ProcessEdgesWork` into a trait, and help dividing the enqueued nodes into chunks.

```rust

    fn with_tracer<R, F>(&self, worker: &mut GCWorker<E::VM>, func: F) -> R
    where
        F: FnOnce(&mut Self::TracerType) -> R,
    {
        let mmtk = worker.mmtk;

        // Prepare the underlying ProcessEdgesWork
        let mut process_edges_work = E::new(vec![], false, mmtk);
        // FIXME: This line allows us to omit the borrowing lifetime of worker.
        // We should refactor ProcessEdgesWork so that it uses `worker` locally, not as a member.
        process_edges_work.set_worker(worker);

        // Cretae the tracer.
        let mut tracer = ProcessEdgesWorkTracer { process_edges_work };

        // The caller can use the tracer here.
        let result = func(&mut tracer);

        // Flush the queued nodes.
        let ProcessEdgesWorkTracer {
            mut process_edges_work,
        } = tracer;

        // ============ BEGIN: PROBLEMATIC CHUNK ======================
        let next_nodes = process_edges_work.pop_nodes();
        if !next_nodes.is_empty() {
            // Divide the resulting nodes into appropriately sized packets.
            let work_packets = next_nodes
                .chunks(E::CAPACITY)
                .map(|chunk| {
                    Box::new(process_edges_work.create_scan_work(chunk.into(), false)) as _
                })
                .collect::<Vec<_>>();
            worker.scheduler().work_buckets[self.stage].bulk_add(work_packets);
        }
        // ============ END: PROBLEMATIC CHUNK ======================

        result
    }
```

And that "PROBLEMATIC CHUNK" is the culprit.  I replaced it with `process_edges_work.flush()` and it becomes faster.

packet               |          build1 |          build3     | diff
---                  |             --- |             ---     | ---
SoftRefProcessing    |        0.965000 |        0.095000     | -0.870000
WeakRefProcessing    |      841.921000 |        0.052000     | -841.869000
Finalization         |        0.462000 |        0.347000     | -0.115000
PhantomRefProcessing |        0.167000 |        0.044000     | -0.123000
VMProcessWeakRefs    |        0.037000 |        1.488000     | 1.451000
RefEnqueue           |       46.418000 |        0.094000     | -46.324000
Release              |       13.944000 |       36.644000     | 22.700000
PlanProcessEdges     |    37451.898000 |    36654.279000     | -797.619000
ProcessWeakRefsWork  |             n/a |      929.995000     | 929.995000

This managed to make `PlanProcessEdges` faster than build1, although `ProcessWeakRefsWork` still takes more time than `WeakRefProcessing` for some reason.  (I tried to merge the packet `ProcessWeakRefsWork` back to `VMProcessWeakRefs`, but the improvement is not significant.)

I'll try to run lusearch and zxing again to see if it makes a difference.

wks:
I was wrong again.  

-   Replacing the "PROBLEMATIC CHUNK" with `process_edges_work.flush()` and it becomes faster: **true**.
-   Remove the chunking logic and create one single work packet and it becomes faster: **false**.

I copied the code from `ProcessEdgesWork::flush` to here, and changed its work bucket stage to `self.stage`, and it no longer has performance increase.  How could this happen?

I then added `#[inline(always)]` to `ProcessEdgesWorkTracer::trace_object`, keeping the chunking logic, and the performance improved.

So the culprit is the hot `trace_object` method that is not inlined.  Replacing the code with `process_edges_work.flush()` somehow changed the inlining decision of the Rust compiler.  Sigh...

wks:
Experiment shows that both inlining and chunking affect the performance.

I kept build{1,2,3} of previous experiments.  The new build4 forces inlining the `trace_object` method. `build5` also removes the chunking logic.

Feature | build1 | build2 | build3 | build4 | build5
---|---|---|---|---|---
mmtk-core has the new API? | No | Yes | Yes | Yes | Yes
mmtk-openjdk uses the new API? | No | No | Yes | Yes | Yes
force inlining `trace_object`? | n/a | n/a | No | Yes | Yes
split the enqueued nodes into chunks? | No | No | Yes | Yes | No

*Note: build1 and build2 uses the built-in ref processor in mmtk-core. It directly calls `ProcessEdgesWork::trace_object`, so "force inlining" is not relevant. build{3,4,5} uses a wrapper that wraps the `E::trace_object` method, so it will hurt the performance if the wrapper is not inlined.*

*Note2: build1 and build2 uses the built-in ref processor in mmtk-core. It calls `ProcessEdgesWork::flush()` to create ScanObjects work packets, and it doesn't do any chunking. build{3,4,5} uses my wrapper which "helps" the user split the vec of enqueued nodes into smaller chunks and create multiple work packets in order to increase parallelism.*

Plotty page:

https://squirrel.anu.edu.au/plotty/wks/noproject/#0|boar-2023-01-05-Thu-143457&benchmark^build^invocation^iteration^rt&GC^time^time.stw&|60&build&rt|10&iteration^1^4|20&1^invocation|30&1&benchmark&build-rt;build1-null|40&Histogram%20(with%20CI)^build-rt^benchmark&

STW time (normalised to build1-null):
image

Total time (normalised to build1-null):
image

build5 matches the STW time and total time of build1 and build2.

The total time of build4 is shorter than other builds for zxing, but is worse than other builds for lusearch.

I still believe a good API should hide the "ideal packet size" from its user, and automatically flush the `ProcessEdgesWork` when needed.  The problem should be with the way I do the chunking, i.e. split the list when it is already populated instead of flushing as soon as it is full.

wks:
I managed to implement auto-chunking so that it will flush and create a work packet as soon as the nodes buffer in `ProcessEdgesWork` is full (the number of elements reached `ProcessEdgesWork::CAPACITY`, which is 4096 at this moment).

In the following test, the `trace_object` method is always inlined in all builds.
- **build{1,2}** are the same as before.
- In **build{3,4,5}**, mmtk-core provides the new API and mmtk-openjdk uses the new API.
-   **build3** chunks the `nodes` buffer after it is filled (if there are 100000 weak references, it will try to chunk a 100000-element vector in to many 4096-element vectors);
-   **build4** does not do the chunking (it will create one `PlanScanObjects` work packets with 100000 elemnets)
-   **build5** checks the number of elements of the `nodes` buffer in `trace_object`, and creates a `PlanScanObjects` every time 4096 objects are enqueued.

STW time (normalised to build1-null)
image

total time (normalised to build1-null)
image

In both lusearch and zxing, the STW time of build5 is very close to that of build1 and build2 when reference-processing is enabled.

When testing with my synthetic micro-benchmark, build5 consistently outperforms build1, especially in terms of time consumed in `PlanProcessEdge` work packets.

wks:
Here is the plot for build1, build2 and build5 for more tests in the DaCapo Chopin benchmark suite.

STW time (normalised to **build1-null**)

image

STW time (normalised to **build1-1**)

image

Total time (normalised to **build1-1**)

image

Number of GC (normalised to **build1-1**)

image

From the plot, we see both the STW time and the total time of build5 are almost as good as build1 on average.

As for the the STW time of zxing, build5-1 still has a noticeably overhead compared to build1-1 and build2-1, and the error is small.  It could be due to not parallelising Release and RefEnqueue, but other possibilities exist.

The behaviour of some benchmark changed significantly after turning on reference processing.

-   `lusearch`: STW time is greatly increased after enabling ref processing, and total time also increased greatly.
    -   `sunflow`, `xalan`: like `lusearch`, but milder.
-   `zxing`: STW time is greatly increased after enabling ref processing, but total time changed mildly.
-   `cassandra`, `fop`, `jython`, `luindex`, `tomcat`: STW time increased, but total time  doesn't change.
-   `h2o`, `pmd`: No obvious change before or after enabling ref processing.
-   `h2`: STW time is greatly decreased after enabling ref processing, and total time reduced noticeably.

k-sareen:
We need to be chunking work packets probably. @wks didn't you look into the size of work packets a while ago and found there were packets >>> 4096 (default size)?

wks:
Using `perf` to analyse my microbenchmark, I found two problems in my `mmtk-openjdk` repo.

1.  In my `mmtk-openjdk` branch, `add_weak_reference` attempts to borrow the `WEAK_PROCESSOR` instance immutably using `AtomicRefCell` every time it is called.  The reference processor in `mmtk-core` doesn't have this problem because ref processor is a field of `MMTK`, and an `&MMTK` is passed to `add_weak_reference` as parameter.
    This function is called during object scanning every time a reachable weak reference object is discovered.  Although `AtomicRefCell` is relatively cheap to borrow, it still involves an atomic operation.  When it is on the hot path, the cost will accumulate.  A wiser way is to let each GC worker borrow `WEAK_PROCESSOR` to its thread-local state, or change the synchronisation mechanism in other ways to eliminate this borrow.  For now, I just replaced it with an `unsafe` operation.  Also note that the main bottleneck of reference processing (of our current implementation) is the mutex lock on the `HashMap`.
2.  The closure that contains `trace_object` is not inlined.  Concretely,

```rust
                    tracer_context.with_tracer(worker, |tracer| {
                        self.reference_processors
                            .scan_weak_refs(|o| tracer.trace_object(o)); // This inner lambda is not inlined.
                    });
```

We know how hot `trace_object` is.  To force inlining it, I have to use the unusual syntax `#[inline(always)] |o| tracer.trace_object(o)` to annotate a lambda. Again the ref processor in `mmtk-core` doesn't have this problem because it passes a `&ProcessEdgesWork` directly to the reference processor.

In the following plot, build1 and build2 are like before.  Both build3 and build4 applied the auto-flushing mechanism of enqueued nodes in mmtk-core.  The difference is, build4 applied the two optimisations above in the mmtk-openjdk repo.

Feature | build1 | build2 | build3 | build4 
---|---|---|---|---
mmtk-core has the new API? | No | Yes | Yes | Yes 
mmtk-openjdk uses the new API? | No | No | Yes | Yes 
force inlining `trace_object`? | n/a | n/a | Yes | Yes 
flush enqueued nodes whenever full (4096)? | No | No | Yes | Yes 
workaround sync cost of AtomicRefCell? | No | No | No | Yes 
inline the hot lambda | No | No | No | Yes 

STW time (normalised to build1-1):
image

Total time (normalised to build1-1):
image

From the plots, we can see that the STW time and total time of **build4** (with the two optimisations above) are better than **build3** (without the optimisations), and are on par with **build1** (master+master).

Since the performance differences are a result of inlining details in `mmtk-openjdk`, I think the new weak reference API itself is efficient enough for general use.

In the future, we should deprecate the reference processor and finalisation processor in mmtk-core, and update the reference/finaliser processing mechanism in mmtk-openjdk.  The current ref processor in mmtk-core has a serious bottleneck, that is, the mutex lock on the `HashMap` that holds the discovered weak references.  This makes reference processing slower if there are more GC worker threads.

wks:
> We need to be chunking work packets probably. @wks didn't you look into the size of work packets a while ago and found there were packets >>> 4096 (default size)?

Yes.  I once found that many work packets are much smaller than 4096.  Some of them have less than 10 items (objects/edges).  The problem is that we currently have no way to merge work packets. Packets only gets smaller, unless some objects have a large fan-out, i.e. one object points to many other objects.

It should be helpful if the new API automatically chunks up the enqueued nodes into properly-sized work packets, and I have implemented that successfully.

However, for my microbenchmark, packet size is not a problem.  The mutex lock on the `HashMap` of discovered `{Soft,Weak,Phantom}Rreferences` effectively makes reference processing single-threaded when discovering references, and single-threaded when iterating over discovered references.  And an `WeakReference` instance either points to a live object or gets cleared, so there should be no objects enqueued when tracing `WeakReference`.  However, `trace_object` still needs to be called to determine whether the object the `WeakReference` points to is still alive, so whether `trace_object` is inlined still affects the problem.  `SoftReference` may be a different story, because a `SoftReference` can keep dead objects alive.

wks:
I renamed `Collection::vm_release` to `Collection::post_forwarding`, and created a `VMPostForwarding` work packet for it.

The following experiment investigates the performance impact of parallelising the reference-enqueuing work (done in `post_forwarding`, or `vm_release` in the past).

-   build1: master + master
-   build{2,3,4} uses the new API.  The differences are:
    -   build2: The `Release` work packet calls `vm_release` and then `plan_mut.release`, not parallelising at all.
    -   build3: The `Release` work packet calls `vm_release` and then `plan_mut.release`, but `vm_release` creates a work packet that does ref-enqueuing.
    -   build4: `schedule_common_work` schedules a `VMPostForwarding` work packet in the `Release` *bucket*.  The `Release` *packet* doesn't call `post_forwarding`.

STW time: (normalised to build1-1)
image

Total time: (normalised to build1-1)
image

The experiment shows build4 performs the best among build{2,3,4}, but is still within the confidence interval.  I think the reason is, the earlier the work packet is scheduled, the earlier a GC worker can start doing that work.

qinsoon:
@wks I feel you have understood and optimized enough on this. Let me know when you push all your changes, and I will take a look at this PR again.

src/scheduler/gc_work.rs
```diff
@@ -245,24 +244,190 @@ impl<VM: VMBinding> GCWork<VM> for EndOfGC {
...
+        };
+        <E::VM as VMBinding>::VMScanning::forward_weak_refs(worker, tracer_factory)
+    }
+}
+
+/// This work packet calls `Collection::post_forwarding`.
+///
+/// NOTE: This will replace `RefEnqueue` in the future.
+///
+/// NOTE: Although this work packet runs in parallel with the `Release` work packet, it does not
```
qinsoon, line 419: We cannot guarantee this. The binding may access the plan in their implementation of `Collection::post_forwarding()`.

qinsoon (approved):
LGTM. I suggest we run the binding tests before merging once the merge conflict is resolved.

src/scheduler/gc_work.rs
wks: Indeed.  While `Collection::post_forwarding()` doesn't expose the `Plan` to the binding, and the `Plan` field is declared as `pub(crate)`, many functions in the `memory_manager` module access the `plan` instance.

The timing of `RefEnqueue` (now `VMPostForwarding`) should be after all references are forwarded.  That includes the `RefForwarding` and the `FinalizerForwarding` buckets (both subsumed by the new `VMRefForwarding` bucket). And it doesn't need to access the MMTk instance.  However, it seems impossible to prevent the binding from accessing plan in our current program structure.  As long as the binding has a `&'static MMTk`, it can call functions in `memory_manager`, and indirectly access the plan.

I think what we can do is telling the VM binding not to access the `plan` indirectly. But even that sounds like a bad idea because we haven't exposed `Plan` anyway, and the binding has no way to know which API function indirectly accesses the `Plan`.  So maybe we can provide the `Collection::post_forwarding()` hook like this to make sure our ref-processing code in mmtk-core can be implemented in the binding.  But we refactor mmtk-core in the future to solve the "`plan` vs `plan_mut`" problem from the root so we can actually prevent the binding from accidentally accessing the plan.

src/scheduler/scheduler.rs
```diff
@@ -230,6 +205,31 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                     .add(ForwardFinalization::<C::ProcessEdgesWorkType>::new());
             }
         }
+
+        // We add the VM-specific weak ref processing work regardless of MMTK-side options,
```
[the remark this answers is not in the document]
wks: I have evaluated the overhead.  The cost of empty work packets (even multiple buckets with such packets, including `{Soft,Weak,Phantom}Closure`) is trivial compared to any work packet that does actual work.

PR #701, opened 2022-11-15
Change LAST_CLOSURE_BUCKET to FIRST_REF_CLOSURE_BUCKET

qinsoon:
This PR fixes a bug that was introduced in https://github.com/mmtk/mmtk-core/pull/564. See the discussion here: https://github.com/mmtk/mmtk-core/pull/564. In short, we should allow the bindings to do their own weak reference processing after we finish the normal transitive closure and before we start our weak reference processing.


qinsoon:
I am closing this. @wks is going to include some changes like this in his PR.

PR #703, opened 2022-11-21
Support for object pinning

udesou:
This is a draft PR to support object pinning. 
I have added three methods to the SFT: `pin_object`, `unpin_object`, and `is_object_pinned`. 
In policies where pinning should not be supported I've created a panic message. For policies where objects do not move, all methods simply return false and do nothing. I'm not 100% if objects can be pinned in mark compact, so I left these methods as `unimplemented()` for now. 
The PR also adds a `PinningBitSpec` to the object model, and check to the `trace_object_with_opportunistic_copy` function before attempting to forward objects or clear the forwarding status.


src/policy/immortalspace.rs
```diff
@@ -45,6 +45,15 @@ impl<VM: VMBinding> SFT for ImmortalSpace<VM> {
...
         old_value == self.mark_state
     }
+    fn pin_object(&self, _object: ObjectReference) -> bool {
+        false
+    }
+    fn unpin_object(&self, _object: ObjectReference) -> bool {
+        false
+    }
+    fn is_object_pinned(&self, _object: ObjectReference) -> bool {
+        false
```
wenyuzhao, line 55: Should this return `true`? since all objects in this space are non-movable.

src/policy/largeobjectspace.rs
```diff
@@ -43,6 +43,15 @@ impl<VM: VMBinding> SFT for LargeObjectSpace<VM> {
...
         self.test_mark_bit(object, self.mark_state)
     }
+    fn pin_object(&self, _object: ObjectReference) -> bool {
+        false
+    }
+    fn unpin_object(&self, _object: ObjectReference) -> bool {
+        false
+    }
+    fn is_object_pinned(&self, _object: ObjectReference) -> bool {
+        false
```
wenyuzhao, line 53: Same as ImmortalSpace, return `true`?

src/vm/object_model.rs
```diff
@@ -81,6 +81,8 @@ pub trait ObjectModel<VM: VMBinding> {
     const LOCAL_FORWARDING_BITS_SPEC: VMLocalForwardingBitsSpec;
     /// The metadata specification for the mark bit, used by most plans that need to mark live objects. 1 bit.
     const LOCAL_MARK_BIT_SPEC: VMLocalMarkBitSpec;
+    /// The metadata specification for the pinning bit, used by most plans that need to pin objects. 1 bit.
+    const LOCAL_PINNING_BIT_SPEC: VMLocalPinningBitSpec;
```
wenyuzhao, line 85: If this is a local metadata space and ImmixSpace may use it, this should be put inside `ImmixSpace::side_metadata_specs` as well?

wenyuzhao (approved):
LGTM. Just minor questions

src/policy/immortalspace.rs
udesou: I guess it could. It's a tricky question, the object is not pinned in the sense that the pinning bit is not set, but since the object is in a non-movable space, it is technically pinned. 
That being said It would probably not cause any harm returning true here and in the code below.

src/memory_manager.rs
```diff
@@ -745,6 +745,45 @@ pub fn add_finalizer<VM: VMBinding>(
     mmtk.finalizable_processor.lock().unwrap().add(object);
 }
 
+/// Pin an object. MMTk will make sure that the object does not move
+/// during GC. Note that action cannot happen in some plans, eg, semispace.
+/// It returns true if the pinning operation has been performed, i.e.,
```
qinsoon, line 750: I have the same question here as Wenyu. Intuitively, for this method, , we could return either:
1. if this call to `pin_object` successfully pinned the object (i.e. is the action successful?), or
2. if the object is pinned after this call (i.e. is the state as expected?).

Is there any reason that you chose 1 rather than 2?

qinsoon: For example, if a binding both use Immix (moving) and mark sweep. The binding's code that calls `pin_object` should be agnostic to the actual plan used. For Immix, it calls `pin_object`, and expects it to return `true` so the binding knows the object will not be moved. But for mark sweep, it also calls `pin_object`, and MMTk returns `false` even if the object will not be moved. That could cause some confusion.

udesou: Sometimes you want to know whether you actually did the pinning so you can safely unpin later. Say for instance an object has been pinned previously, and you have some code that does:
```
pin_object(a)
// do GC
unpin_object(a)
// do GC
``` 
Now `a` can move, even though it was pinned somewhere else and it was not supposed to move. Whereas you may have:
```
let pinned = pin_object(a)
// do GC
if pinned
  unpin_object(a)
// do GC
``` 
One option would be keeping count of pins and unpins, but that may be too expensive, at least for now (?)

udesou: Furthermore `pin_object` guarantees that the object should not move (unless it is movable and it is unpinned later), even when it returns `false`. In that sense, the 2nd piece of code from the previous comment would work for both mark sweep and immix.

qinsoon: > Furthermore `pin_object` guarantees that the object should not move (unless it is movable and it is unpinned later), even when it returns `false`. In that sense, the 2nd piece of code from the previous comment would work for both mark sweep and immix.

Yeah, this sounds reasonable.

src/policy/mallocspace/global.rs
```diff
@@ -63,6 +63,18 @@ impl<VM: VMBinding> SFT for MallocSpace<VM> {
...
+    fn pin_object(&self, _object: ObjectReference) -> bool {
+        false
+    }
+
+    fn unpin_object(&self, _object: ObjectReference) -> bool {
+        false
+    }
+
+    fn is_object_pinned(&self, _object: ObjectReference) -> bool {
+        false
```
qinsoon, line 75: `is_object_pinned` returns `true` for immortal space and large object space. Why does it return `false` for malloc space?

src/policy/sft.rs
```diff
@@ -130,6 +139,15 @@ impl SFT for EmptySpaceSFT {
         warn!("Object in empty space!");
         false
     }
+    fn pin_object(&self, _object: ObjectReference) -> bool {
```
qinsoon, line 142: I think you should panic for these 3 new methods for `EmptySFT`. Basically, MMTk tries to pin/unpin/query for an object that does not belong to any space. The work cannot be done.

src/policy/lockfreeimmortalspace.rs
```diff
@@ -50,6 +50,15 @@ impl<VM: VMBinding> SFT for LockFreeImmortalSpace<VM> {
     fn is_live(&self, _object: ObjectReference) -> bool {
         unimplemented!()
     }
+    fn pin_object(&self, _object: ObjectReference) -> bool {
+        unimplemented!()
```
qinsoon, line 54: These methods for lock free immortal space should have the same behavior as immortal space.

src/policy/markcompactspace.rs
```diff
@@ -52,6 +52,18 @@ impl<VM: VMBinding> SFT for MarkCompactSpace<VM> {
         Self::is_marked(object)
     }
 
+    fn pin_object(&self, _object: ObjectReference) -> bool {
+        unimplemented!()
```
qinsoon, line 56: I think for mark compact, its behavior on the pinning should be the same as copy space (copying is compulsory).

src/policy/immix/immixspace.rs
```diff
@@ -68,9 +68,57 @@ impl<VM: VMBinding> SFT for ImmixSpace<VM> {
             self.is_marked(object, self.mark_state) || ForwardingWord::is_forwarded::<VM>(object)
         }
     }
+    fn pin_object(&self, object: ObjectReference) -> bool {
```
qinsoon, line 71: It seems the pinning here is general, and there is nothing specific with Immix. In that case you can just implement `pin/unpin/is_pinned` on `VMLocalPinningBitSpec`, like what we did for the log bit.

https://github.com/mmtk/mmtk-core/blob/644628bcc39e649e4125c092ad325bddbd24aaaf/src/util/metadata/log_bit.rs#L6-L11

src/policy/immix/immixspace.rs
```diff
@@ -436,7 +486,11 @@ impl<VM: VMBinding> ImmixSpace<VM> {
     ) -> ObjectReference {
         let copy_context = worker.get_copy_context_mut();
         debug_assert!(!super::BLOCK_ONLY);
-        let forwarding_status = ForwardingWord::attempt_to_forward::<VM>(object);
+        let forwarding_status = if Self::is_pinned(object) {
+            0
```
qinsoon, line 490: Does `0` mean `FORWARDING_NOT_TRIGGERED_YET`?
https://github.com/mmtk/mmtk-core/blob/644628bcc39e649e4125c092ad325bddbd24aaaf/src/util/object_forwarding.rs#L9

If so, you should use the constant.

qinsoon: Another question is that do you really need any change in this method at all?I think the old code should work fine if the object is pinned.

qinsoon (changes requested):
I think we would need to discuss about the new API methods. Currently there are `pin`, `unpin` and `is_pinned`. The problem is that for a non-moving space, `is_pinned` will return `true` (which means the object is pinned) even if every call to `pin` returns `false` (which means we fail to pin the object). If necessary, we can define semantics like this. But generally, it is a bit confusing. I suggest we have some discussion first.

src/policy/immix/immixspace.rs
```diff
@@ -547,9 +605,12 @@ impl<VM: VMBinding> ImmixSpace<VM> {
...
     #[inline(always)]
-    fn is_pinned(_object: ObjectReference) -> bool {
-        // TODO(wenyuzhao): Object pinning not supported yet.
-        false
+    fn is_pinned(object: ObjectReference) -> bool {
+        use crate::mmtk::SFT_MAP;
+        use crate::policy::sft_map::SFTMap;
+        SFT_MAP
+            .get_checked(object.to_address())
+            .is_object_pinned(object)
```
wks, line 613: This is a private method, so it is not called from elsewhere.  I think we can bypass the SFT lookup (or `debug_assert!` that the input `object` is in the current ImmixSpace), and call `self.is_object_pinned`, instead.

Oh there is no `self`! Maybe we can remove this method and call `self.is_object_pinned` instead.

src/policy/immix/immixspace.rs
```diff
@@ -471,14 +525,18 @@ impl<VM: VMBinding> ImmixSpace<VM> {
                 "Forwarded object is the same as original object {} even though it should have been copied",
                 object,
             );
-            ForwardingWord::clear_forwarding_bits::<VM>(object);
+            if !Self::is_pinned(object) {
```
wks, line 528: We should be able to call `self.is_object_pinned` instead.

src/policy/immix/immixspace.rs
udesou: I'll try it out and see if that's the case. I think I had this change when the pinning bit was "shared" with the forwarding bits, but you're probably right that this method doesn't really need to change.

src/policy/mallocspace/global.rs
qinsoon: Maybe we can do this: make `pin/unpin/is_pinned` only about pinning, and additionally we change `is_movable()` to take an object reference as the argument:
https://github.com/mmtk/mmtk-core/blob/644628bcc39e649e4125c092ad325bddbd24aaaf/src/policy/sft.rs#L50

So for all the policies,
|policy|pin/unpin|is_pinned|is_movable|
|-|-|-|-|
| policy with pinning support (immix) | set/unset pin bit | check pin bit | check defrag and check pin bit |
| non moving policy (immortal/marksweep) | false | false | false |
| compulsory moving policy (copy/markcompact) | panic | false | true |

This is probably clearer.

qinsoon: You don't have to make the changes about `is_movable` in this PR. If we can agree on this solution, we just need to get `pin/unpin/is_pinned` right in this PR.

src/policy/immix/immixspace.rs
udesou: Fixed.

src/policy/mallocspace/global.rs
qinsoon: ### Updated

We define `pin/unpin` as this: if a policy does not support pin/unpin objects, it should panic. Otherwise it returns a boolean to indicate whether this call to `pin` actually pinned the object (or this call to `unpin` actually unpinned the object). 
We define `is_pinned` as this: it returns true if the object is guaranteed that it won't move (pinned, or in a non-moving policy), or returns false if the object may move.

We do not change `is_movable`.

|policy|pin/unpin|is_pinned|
|-|-|-|
| policy with pinning support (immix) | set/unset pin bit | check pin bit |
| non moving policy (immortal/marksweep) | false | true |
| compulsory moving policy (copy/markcompact) | panic | false |

wks: There is one thing I am not sure.  By "unsuccessful", it may have two different meanings.

1.  It is unsuccessful because the object was never pinned before, or another thing (thread, or scope) has already unpinned the object before this `unpin()`.  In this case, the object is guaranteed not to be pinned after `unpin()`.  Subsequent `is_pinned()` will return false.
2.  It is unsuccessful because the space is a non-moving policy.  In this case the object is still pinned.  Subsequent `is_pinned()` still returns true.

I am not sure because I cannot think of a scenario where the user (VM binding) needs to know which case the failure is.  But no matter which case it is, I think it is always a fatal failure because bug-free program should never make an unsuccessful `unpin()` call. Again I can't think of a scenario where the VM can recover from an unsuccessful `unpin()`.

wks: One possibility is to change the return value of `unpin()` to `Result<(), UnpinError>` so that the return value doesn't reflect whether the object was pinned before, or will remain pinned after.  The `UnpinError` is used to indicate the concrete error.

udesou: I can agree with this table: 

<!DOCTYPE html>

policy | pin | unpin | is_pinned
-- | -- | -- | --
policy with pinning support (immix) | set pin bit | unset pin bit | check pin bit
non moving policy (immortal/marksweep) | false | false | true
compulsory moving policy (copy/markcompact) | panic | panic | false

I think returning false when unpinning an object of a non-moving policy, makes more sense. The action of unpinning was not performed, therefore the object remains pinned in that case. For policies with pinning support, if unpinning was not successful, the object had not been pinned before (and `is_pinned` should return false afterwards). I can add that table as a comment to the documentation in the SFT methods so the semantics gets documented.
If you guys agree, I can make those changes, and we can work towards merging this.

qinsoon: That looks good to me. Probably also make it clear about our reasoning for this table as comments:
* for non moving policy, all the objects are considered as forever pinned, thus attempting to pin or unpin them will not succeed and will always return false.
* for compulsory moving policy, pin/unpin is impossible, thus we will panic.

qinsoon:
@udesou Can you fix the tests? Then we can get this merged, and https://github.com/mmtk/mmtk-core/pull/706 will not include changes about pinning.

qinsoon (approved):
LGTM

wks (commented):
I just discovered a small problem.  Variable names with an underscore in front of it, such as `_object`, indicates that they are unused.  The compiler will not give the "unused variables" warning for those variables.  We should remove those underscores since we started using them.

udesou:
> I just discovered a small problem. Variable names with an underscore in front of it, such as `_object`, indicates that they are unused. The compiler will not give the "unused variables" warning for those variables. We should remove those underscores since we started using them.

Where exactly do you mean they should be renamed? Is it in code like: 
```
    #[inline(always)]
    fn is_pinned(&self, _object: ObjectReference) -> bool {
        #[cfg(feature = "object_pinning")]
        return self.is_object_pinned(_object);

        #[cfg(not(feature = "object_pinning"))]
        false
    }
```

If so, I was just using the same idea as 

```
    fn initialize_object_metadata(&self, _object: ObjectReference, _alloc: bool) {
        #[cfg(feature = "global_alloc_bit")]
        crate::util::alloc_bit::set_alloc_bit::<VM>(_object);
    }
```

qinsoon:
> > I just discovered a small problem. Variable names with an underscore in front of it, such as `_object`, indicates that they are unused. The compiler will not give the "unused variables" warning for those variables. We should remove those underscores since we started using them.
> 
> Where exactly do you mean they should be renamed? Is it in code like:
> 
> ```
>     #[inline(always)]
>     fn is_pinned(&self, _object: ObjectReference) -> bool {
>         #[cfg(feature = "object_pinning")]
>         return self.is_object_pinned(_object);
> 
>         #[cfg(not(feature = "object_pinning"))]
>         false
>     }
> ```
> 
> If so, I was just using the same idea as
> 
> ```
>     fn initialize_object_metadata(&self, _object: ObjectReference, _alloc: bool) {
>         #[cfg(feature = "global_alloc_bit")]
>         crate::util::alloc_bit::set_alloc_bit::<VM>(_object);
>     }
> ```

The issue is how we tell Rust about arguments that are only used under some features and are unused in other cases. What we currently do for most of the cases are: leaving the argument with the underscore `_` prefix, and use it if you need to.

You could argue that the best approach is to separate the implementations, such as 
```rust
>     #[cfg(feature = "object_pinning")]
>     fn is_pinned(&self, object: ObjectReference) -> bool {
>         self.is_object_pinned(object)
>     }
>     #[cfg(not(feature = "object_pinning"))]
>     fn is_pinned(&self, object: ObjectReference) -> bool {
>         false
>     }
```

However, this may not be desirable in some cases when there are code shared.

Anyway, I think this is a minor styling issue. What has been done in this PR is consistent with our current code base, and I don't feel it is wrong -- it is just one approach to deal with it.

wks (approved):
OK. I didn't realise that the Rust compiler is not considering it as "used" when only used in some configuration.

LGTM now.

PR #705, opened 2022-11-25
Do not move root objects

udesou:
This draft PR is an attempt to make sure that root objects should not move (at least for Immix in the current implementation). I've introduced a parameter `KIND` to `PlanScanObjects` and made it explicit in implementation of `do_work_common`, such that when creating process edges work, I pass `TRACE_KIND_FAST` which should make sure that the root objects are traced with `trace_object_without_moving`. I guess other plans that allow tracing/marking objects in place could use the same strategy.
A question here is whether objects would be transitively traced without moving; as far as I understand, it wouldn't, but perhaps @wenyuzhao knows the code a bit better.


src/scheduler/gc_work.rs
```diff
@@ -869,15 +871,113 @@ impl<E: ProcessEdgesWork, P: Plan<VM = E::VM> + PlanTraceObject<E::VM>> PlanScan
...
+        // However, VMs that deliver root objects instead of root edges are incapable of updating
+        // root slots.  Like processing an edge, we call `trace_object` on those objects, and
+        // assert the GC doesn't move those objects because we cannot store back to the slots.
+        //
+        // If this is a root packet, the `scanned_root_objects` variable will hold those root
+        // objects which are traced for the first time.
+        let scanned_root_objects = self.roots().then(|| {
+            // We create an instance of E to use its `trace_object` method and its object queue.
+            let mut process_edges_work =
+                <PlanProcessEdges<E::VM, P, TRACE_KIND_FAST>>::new(vec![], false, mmtk);
```
wks, line 903: Here we hard-coded `TRACE_KIND_FAST` instead of using the `KIND` type parameter in `PlanScanObjects<E, P, KIND>`.  It is correct for Immix, but it needs to be extended to other plans, though.

qinsoon, line 903: This work packet is general for all the plans. `TRACE_KIND_FAST` is immix specific, and should not appear in general code.

src/scheduler/gc_work.rs
```diff
@@ -869,15 +871,113 @@ impl<E: ProcessEdgesWork, P: Plan<VM = E::VM> + PlanTraceObject<E::VM>> PlanScan
...
+
+            for object in buffer.iter().copied() {
+                let new_object = process_edges_work.trace_object(object);
```
wks, line 906: Here `trace_object` is a method of `process_edges_work`.  Generally speaking the `ProcessEdgesWork` (along with its `TraceKind` if relevant) is provided by the Plan (at `schedule_collection`)  But I think whether we pin the object is a different dimension.   These pinning roots needs to pinned regardless whether it is a normal collection or a defrag collection, or whether it is a marking trace or forwarding trace.  Probably it is better to add one more argument to `trace_object` itself, or add a different `trace_object_pin` method to `SFT` and `PlanTraceObject`.

wks: In other words, `TraceKind` is local to a trace (should be part of `TraceLocal` which disappeared when we ported MMTk to Rust), while whether to pin or not is specific to this `trace_object` call.

qinsoon:
This PR basically duplicates `ScanObjectsWork::do_work_common` to `PlanScanObjects::do_work_common`, and if the objects are roots, use `TRACE_KIND_FAST` for them. In this case, if we create any work with `RootsWorkFactory::create_process_node_roots_work`, they will be traced with `TRACE_KIND_FAST` (which is non moving trace object in Immix).

qinsoon:
As `TRACE_KIND_FAST` is internal to immix, this approach is not general. I think we can add a new method `trace_object_non_moving` to `ProcessEdgesWork`, and we call `trace_object_non_moving()` for root objects to ensure we do not move them.

wks:
> ... we can add a new method `trace_object_non_moving` to `ProcessEdgesWork`, and we call `trace_object_non_moving()` for root objects to ensure we do not move them.

Yes.  I agree.  Currently, `trait ProcessEdgesWork` represents the capability of tracing objects (i.e. using SFT or `PlanTraceObject` to call the underlying `trace_object` of the space).  If we are going to support read/black roots, we should augment this capability so that it can call `trace_object_without_moving` for each space, too.  Spaces should provide `trace_object_without_moving` in addition to `trace_object`, or leave it `unimplemented!()` if unsuitable.

k-sareen:
Superseded by https://github.com/mmtk/mmtk-core/pull/897

PR #706, opened 2022-12-01
Adding support for red/black roots

udesou:
This is a draft PR to properly add support for processing ***red object roots*** (the root object or any object in its transitive closure _must not_ move), and ***black object roots*** (where only the root object _must not_ move, and any object beyond it _may_ move). 

Currently, the code is very messy, and I am pretty sure it is not generic enough in multiple places, so I'd appreciate your feedback on how to do that properly.

A short list of high-level changes that I've made:
- introduced a `ClosureImmovable` work bucket that runs before `Closure`; this is necessary to make sure objects found during this closure are processed within the bucket, with the guarantee that no object processed there should move.
- introduced `TRACE_KIND_IMMOVABLE`. This means that any object traced with this kind will not move, and objects that are "found" by the trace should be put in the `ClosureImmovable` work bucket. 
- split the work in `schedule_collection` into two: in defrag we first process red roots with `TRACE_KIND_IMMOVABLE` then process the remaining roots with `TRACE_KIND_DEFRAG`. Note that I create an API function in `VMScanning` called `scan_vm_immovable_roots`, in which the binding can scan red roots, when `in_defrag` is false, these roots are processed with `TRACE_KIND_FAST` where nothing moves. 
- created a `ImmovableObjectsClosure` that creates work into `WorkBucketStage::ClosureImmovable` instead of `WorkBucketStage::Closure`. 

The rest is essentially duplicating code to enforce the separation between the two closures. 

I guess it would be interesting to discuss and agree on the implementation first, then we should discuss naming conventions (red/black, movable/immovable, etc), and definitely involve others once the code is more mature.


src/scheduler/gc_work.rs
```diff
@@ -815,6 +920,27 @@ impl<VM: VMBinding, P: PlanTraceObject<VM> + Plan<VM = VM>, const KIND: TraceKin
...
+    fn start_or_dispatch_scan_work(&mut self, work_packet: impl GCWork<Self::VM>) {
+        if Self::SCAN_OBJECTS_IMMEDIATELY {
+            // We execute this `scan_objects_work` immediately.
+            // This is expected to be a useful optimization because,
+            // say for _pmd_ with 200M heap, we're likely to have 50000~60000 `ScanObjects` work packets
+            // being dispatched (similar amount to `ProcessEdgesWork`).
+            // Executing these work packets now can remarkably reduce the global synchronization time.
+            self.worker().do_work(work_packet);
+        } else {
+            if KIND == TRACE_KIND_IMMOVABLE {
```
qinsoon, line 936: Similar to `ObjectsClosure`, you can just have a field to store which bucket new work should go to.

src/scheduler/gc_work.rs
```diff
@@ -882,10 +1014,142 @@ impl<E: ProcessEdgesWork, P: Plan<VM = E::VM> + PlanTraceObject<E::VM>> ScanObje
...
+        // root slots.  Like processing an edge, we call `trace_object` on those objects, and
+        // assert the GC doesn't move those objects because we cannot store back to the slots.
+        //
+        // If this is a root packet, the `scanned_root_objects` variable will hold those root
+        // objects which are traced for the first time.
+        let scanned_root_objects = self.roots().then(|| {
+            // We create an instance of E to use its `trace_object` method and its object queue.
+            // let mut process_edges_work = Self::E::new(vec![], false, mmtk);
+            let mut process_edges_work =
+                <PlanProcessEdges<E::VM, P, TRACE_KIND_IMMOVABLE>>::new(vec![], false, mmtk);
```
qinsoon, line 1041: Is this fix to the current `do_work_common()`?
https://github.com/mmtk/mmtk-core/blob/117000feb066edb43e8ebb94c1a465c56fbc771c/src/scheduler/gc_work.rs#L632

src/scheduler/gc_work.rs
```diff
@@ -882,10 +1014,142 @@ impl<E: ProcessEdgesWork, P: Plan<VM = E::VM> + PlanTraceObject<E::VM>> ScanObje
...
+
+        // If it is a root packet, scan the nodes that are first scanned;
+        // otherwise, scan the nodes in the buffer.
+        let objects_to_scan = scanned_root_objects.as_deref().unwrap_or(buffer);
+
+        // Then scan those objects for edges.
+        let mut scan_later = vec![];
+        {
+            if KIND == TRACE_KIND_IMMOVABLE {
+                let mut closure = ImmovableObjectsClosure::<Self::E>::new(worker);
```
qinsoon, line 1065: Similarly, if you have a field to tell you which bucket to use, then you can just pass that bucket variable between work packets, and use it when you push new work.

src/scheduler/gc_work.rs
```diff
@@ -882,10 +1014,142 @@ impl<E: ProcessEdgesWork, P: Plan<VM = E::VM> + PlanTraceObject<E::VM>> ScanObje
...
+            // Create work packets to scan adjacent objects.  We skip ProcessEdgesWork and create
+            // object-scanning packets directly, because the edges are already traced.
+            if !process_edges_work.nodes.is_empty() {
+                let next_nodes = process_edges_work.nodes.take();
+                let make_packet = |nodes| {
+                    let work_packet = self.make_another(nodes);
+                    if KIND == TRACE_KIND_IMMOVABLE {
+                        memory_manager::add_work_packet(
+                            mmtk,
+                            WorkBucketStage::ClosureImmovable,
```
qinsoon, line 1126: Same here. You just need a variable to tell you which bucket to use.

qinsoon (commented):
These are my general suggestions:
* focus on removing duplicate code
* use a field in work packets to store the bucket that new work should be pushed to
* have a switch (runtime option or feature) to enable/disable immovable closure
* let a binding to create work for immovable vs movable roots by telling `RootsWorkFactory` which kind of roots they are.

k-sareen:
Superseded by https://github.com/mmtk/mmtk-core/pull/897

PR #707, opened 2022-12-01
Mask and zero bits in metadata bytes for bulk zeroing

qinsoon:
This PR fixes a bug in `bzero_metadata()` discussed in https://github.com/mmtk/mmtk-core/pull/643#issuecomment-1331875577. When the data address range to be bulk zeroed cannot be mapped into whole bytes in the side metadata, the other bits in the bytes which should not be updated will get zeroed unexpectedly.

### DIFF
--- a/src/util/metadata/side_metadata/global.rs
+++ b/src/util/metadata/side_metadata/global.rs
@@ -146,7 +146,53 @@ impl SideMetadataSpec {
         MMAPPER.is_mapped_address(meta_addr)
     }
 
-    /// Bulk-zero a specific metadata for a chunk.
+    /// This method is used for bulk zeroing side metadata for a data address range. As we cannot guarantee
+    /// that the data address range can be mapped to whole metadata bytes, we have to deal with cases that
+    /// we need to mask and zero certain bits in a metadata byte.
+    /// The end address and the end bit are exclusive.
+    pub(super) fn zero_meta_bits(
+        meta_start_addr: Address,
+        meta_start_bit: u8,
+        meta_end_addr: Address,
+        meta_end_bit: u8,
+    ) {
+        // Start/end is the same, we don't need to do anything.
+        if meta_start_addr == meta_end_addr && meta_start_bit == meta_end_bit {
+            return;
+        }
+
+        // zeroing bytes
+        if meta_start_bit == 0 && meta_end_bit == 0 {
+            memory::zero(meta_start_addr, meta_end_addr - meta_start_addr);
+            return;
+        }
+
+        if meta_start_addr == meta_end_addr {
+            // we are zeroing selected bits in one byte
+            let mask: u8 = ((1 << meta_end_bit) - 1) - ((1 << meta_start_bit) - 1);
+            let mask = !mask;
+
+            unsafe { meta_start_addr.as_ref::<AtomicU8>() }.fetch_and(mask, Ordering::SeqCst);
+        } else if meta_start_addr + 1usize == meta_end_addr && meta_end_bit == 0 {
+            // we are zeroing the rest bits in one byte
+            let mask = u8::MAX - ((1 << meta_start_bit) - 1);
+            let mask = !mask;
+
+            unsafe { meta_start_addr.as_ref::<AtomicU8>() }.fetch_and(mask, Ordering::SeqCst);
+        } else {
+            // zero bits in the first byte
+            Self::zero_meta_bits(meta_start_addr, meta_start_bit, meta_start_addr + 1usize, 0);
+            // zero bytes in the middle
+            Self::zero_meta_bits(meta_start_addr + 1usize, 0, meta_end_addr, 0);
+            // zero bits in the last byte
+            Self::zero_meta_bits(meta_end_addr, 0, meta_end_addr, meta_end_bit);
+        }
+    }
+
+    /// Bulk-zero a specific metadata for a chunk. Note that this method is more sophisiticated than a simple memset, especially in the following
+    /// cases:
+    /// * the metadata for the range includes partial bytes (a few bits in the same byte).
+    /// * for 32 bits local side metadata, the side metadata is stored in discontiguous chunks, we will have to bulk zero for each chunk's side metadata.
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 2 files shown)


src/util/metadata/side_metadata/side_metadata_tests.rs
```diff
@@ -580,4 +580,157 @@ mod tests {
...
+
+        fill_1(start);
+        // zero the word
+        SideMetadataSpec::zero_meta_bits(start, 0, end, 0);
+        assert_eq!(unsafe { start.load::<u32>() }, 0);
+
+        fill_1(start);
+        // zero first 2 bits
+        SideMetadataSpec::zero_meta_bits(start, 0, start, 2);
+        assert_eq!(unsafe { start.load::<u32>() }, 0xFFFF_FFFC); // ....1100
```
wks, line 724: I think it is readable enough.  This is just a reminder that Rust does have binary literals, i.e. `0b11111111_11111111_11111111_11111100`.  You may use that if you think that's more readable.

src/util/metadata/side_metadata/global.rs
```diff
@@ -146,7 +146,53 @@ impl SideMetadataSpec {
...
+        if meta_start_addr == meta_end_addr {
+            // we are zeroing selected bits in one byte
+            let mask: u8 = ((1 << meta_end_bit) - 1) - ((1 << meta_start_bit) - 1);
+            let mask = !mask;
+
+            unsafe { meta_start_addr.as_ref::<AtomicU8>() }.fetch_and(mask, Ordering::SeqCst);
+        } else if meta_start_addr + 1usize == meta_end_addr && meta_end_bit == 0 {
+            // we are zeroing the rest bits in one byte
+            let mask = u8::MAX - ((1 << meta_start_bit) - 1);
+            let mask = !mask;
```
wks, line 179: Can be simplified.
```suggestion
            let mask = u8::MAX << meta_start_bit;
```

src/util/metadata/side_metadata/global.rs
```diff
@@ -146,7 +146,53 @@ impl SideMetadataSpec {
...
+        // zeroing bytes
+        if meta_start_bit == 0 && meta_end_bit == 0 {
+            memory::zero(meta_start_addr, meta_end_addr - meta_start_addr);
+            return;
+        }
+
+        if meta_start_addr == meta_end_addr {
+            // we are zeroing selected bits in one byte
+            let mask: u8 = ((1 << meta_end_bit) - 1) - ((1 << meta_start_bit) - 1);
+            let mask = !mask;
```
wks, line 173: Can be simplified, although not as intuitive as the original one:
```suggestion
            let mask: u8 = (u8::MAX << meta_end_bit) | !(u8::MAX << meta_start_bit);
```

src/util/metadata/side_metadata/global.rs
```diff
@@ -178,31 +224,21 @@ impl SideMetadataSpec {
...
+                zero(start, second_data_chunk);
+
                 let last_data_chunk = (start + size).align_down(BYTES_IN_CHUNK);
                 let last_meta_chunk = address_to_meta_address(self, last_data_chunk);
                 // bzero the last sub-chunk
-                memory::zero(
-                    last_meta_chunk,
-                    address_to_meta_address(self, start + size) - last_meta_chunk,
-                );
+                zero(last_meta_chunk, start + size);
```
wks, line 236: The `zero` function takes data address as parameter, but it passes `last_meta_chunk`.  It should be `last_data_chunk`.
```suggestion
                zero(last_data_chunk, start + size);
```

src/util/metadata/side_metadata/global.rs
```diff
@@ -157,19 +203,19 @@ impl SideMetadataSpec {
...
-        debug_assert!(start.is_aligned_to(BYTES_IN_PAGE) && meta_byte_lshift(self, start) == 0);
-
         #[cfg(feature = "extreme_assertions")]
         sanity::verify_bzero(self, start, size);
 
-        let meta_start = address_to_meta_address(self, start);
+        let zero = |data_start: Address, data_end: Address| {
+            let meta_start = address_to_meta_address(self, data_start);
+            let meta_start_shift = meta_byte_lshift(self, data_start);
+            let meta_end = address_to_meta_address(self, data_end);
```
wks, line 212: It is OK if the metadata is contiguous.  But if it is discontiguous, then `meta_start` and `meta_end` may be in two unrelated chunks.

src/util/metadata/side_metadata/global.rs
```diff
@@ -178,31 +224,21 @@ impl SideMetadataSpec {
...
-                );
+                zero(start, start + size);
             } else {
                 let second_data_chunk = start.align_up(BYTES_IN_CHUNK);
                 // bzero the first sub-chunk
-                memory::zero(
-                    meta_start,
-                    address_to_meta_address(self, second_data_chunk) - meta_start,
-                );
+                zero(start, second_data_chunk);
```
wks, line 231: Two problems:

1.  If `start` is already aligned to `BYTES_IN_CHUNK`, then `start.align_up(BYTES_IN_CHUNK)` will be equal to `start`, and it will end up calling `zero(start, start)`.
2.  If the metadata is discontiguous, passing `second_data_chunk` to the `zero` function will result in `meta_end` being in a different metadata chunk, and unrelated metadata in between will be overwritten.

src/util/metadata/side_metadata/global.rs
```diff
@@ -178,31 +224,21 @@ impl SideMetadataSpec {
...
+                zero(last_meta_chunk, start + size);
                 let mut next_data_chunk = second_data_chunk;
+
                 // bzero all chunks in the middle
                 while next_data_chunk != last_data_chunk {
-                    memory::zero(
-                        address_to_meta_address(self, next_data_chunk),
-                        metadata_bytes_per_chunk(self.log_bytes_in_region, self.log_num_of_bits),
-                    );
+                    zero(next_data_chunk, next_data_chunk + BYTES_IN_CHUNK);
```
wks, line 241: Same problem here.  `next_data_chunk + BYTES_IN_CHUNK` will be mapped to the next metadata chunk.

wks (changes requested):
One major problem is that if the metadata is discontiguous, the current way may overwrite unrelated metadata between chunks.

src/util/metadata/side_metadata/side_metadata_tests.rs
qinsoon: Yeah. I know that Rust can represent literals in binary. I didn't want to write that many 1s.

src/util/metadata/side_metadata/global.rs
qinsoon: Done. It should be `!(u8::MAX << meta_start_bit)`.

src/util/metadata/side_metadata/global.rs
qinsoon: Done

src/util/metadata/side_metadata/global.rs
qinsoon: Thanks. Fixed.

src/util/metadata/side_metadata/global.rs
qinsoon: I changed `zero` to `zero_contiguous` which uses the same implementation as what you reviewed, and added `zero_discontiguous` which computes the total bits that need to be zeroed without computing the end address.

src/util/metadata/side_metadata/global.rs
qinsoon: 1. I added a check, if the size to be zeroed is 0, then the method simply returns.
2. See my reply for the above comments.

src/util/metadata/side_metadata/global.rs
```diff
@@ -157,19 +201,66 @@ impl SideMetadataSpec {
...
+                "The data to be zeroed in discontiguous specs needs to be in the same chunk"
             );
+
+            let meta_start = address_to_meta_address(self, data_start);
+            let meta_start_shift = meta_byte_lshift(self, data_start);
+
+            // How many bits we need to zero for data_bytes
+            let meta_total_bits = (data_bytes >> self.log_bytes_in_region) << self.log_num_of_bits;
+            let meta_delta_bytes = meta_total_bits >> LOG_BITS_IN_BYTE;
+            let meta_delta_bits: u8 = (meta_total_bits % BITS_IN_BYTE) as u8;
```
wks, line 246: Nicely done.  I think these lines of code are even clearer than those in `address_to_contiguous_meta_address` and `meta_byte_lshift`.

wks (approved):
LGTM

PR #708, opened 2022-12-05
Fix BlockQueue overflow

wks:
The capacity of a BlockQueue is set to a constant 256.  However, if the block size is smaller, a chunk may have more than 256 blocks, overflowing the BlockQueue.  This commit breaks one BlockQueue into multiple instances when that happens.

This may happen when I shrink the Immix Block size from `1 << 15` to `1 << 13`.  In that case, Each `1 << 22` byte chunk may have `1 << 9` (512) blocks, overflowing the `BlockQueue::CAPACITY` of 256.


qinsoon (commented):
LGTM. An alternative is to change the capacity of a `BlockQueue` to make sure that it can always hold blocks for a chunk. Better check with @wenyuzhao on this.

wenyuzhao (approved):

PR #711, opened 2022-12-13
Turn `SFTMap` into trait object

wenyuzhao:
This PR makes MMTk use `dyn SFTMap` instead of defining multiple `type SFTMapType<'a>` based on compiler flags.

This is part of the necessary refactoring to make MMTk able to select address space at runtime. For 35-bit address space (pointer compression), MMTk should use `SFTSparseChunkMap` instead of `SFTSpaceMap`.

## Performance:

Dynamic dispatch on `dyn SFTMap` will be slower than static dispatch. But in practice, no performance drop is observed:

http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/DkTpwP

### DIFF
--- a/src/policy/sft_map.rs
+++ b/src/policy/sft_map.rs
@@ -60,17 +60,19 @@ pub trait SFTMap {
     unsafe fn clear(&self, address: Address);
 }
 
-cfg_if::cfg_if! {
-    if #[cfg(all(feature = "malloc_mark_sweep", target_pointer_width = "64"))] {
-        // 64-bit malloc mark sweep needs a chunk-based SFT map, but the sparse map is not suitable for 64bits.
-        pub type SFTMapType<'a> = dense_chunk_map::SFTDenseChunkMap<'a>;
-    } else if #[cfg(target_pointer_width = "64")] {
-        pub type SFTMapType<'a> = space_map::SFTSpaceMap<'a>;
-    } else if #[cfg(target_pointer_width = "32")] {
-        pub type SFTMapType<'a> = sparse_chunk_map::SFTSparseChunkMap<'a>;
-    } else {
-        compile_err!("Cannot figure out which SFT map to use.");
-    }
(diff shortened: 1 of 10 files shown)


src/policy/sft_map.rs
```diff
@@ -60,17 +60,19 @@ pub trait SFTMap {
...
-        pub type SFTMapType<'a> = dense_chunk_map::SFTDenseChunkMap<'a>;
-    } else if #[cfg(target_pointer_width = "64")] {
-        pub type SFTMapType<'a> = space_map::SFTSpaceMap<'a>;
-    } else if #[cfg(target_pointer_width = "32")] {
-        pub type SFTMapType<'a> = sparse_chunk_map::SFTSparseChunkMap<'a>;
-    } else {
-        compile_err!("Cannot figure out which SFT map to use.");
-    }
+pub(crate) fn create_sft_map() -> Box<dyn SFTMap> {
+    #[cfg(all(feature = "malloc_mark_sweep", target_pointer_width = "64"))]
```
qinsoon, line 64: Is it possible to use `cfg_if` within a function? I would suggest using `cfg_if` if possible. Otherwise it is very easy to get the conditions wrong. For example, in the PR code, when we enable `malloc_mark_sweep` on 64 bits machines, both `all(feature = "malloc_mark_sweep", target_pointer_width = "64")` and `target_pointer_width = "64"` is true, and the method would look like this (I hope our style check could catch this):
```rust
pub(crate) fn create_sft_map() -> Box<dyn SFTMap> {
  return Box::new(dense_chunk_map::SFTDenseChunkMap::<'static>::new());
  return Box::new(space_map::SFTSpaceMap::<'static>::new());
}
```

wenyuzhao: I see. I changed the code to use `cfg_if` instead.

qinsoon (approved):
LGTM

Issue #561, opened 2022-03-18
Support for dynamic heap resizing


steveblackburn:
We need to add support for dynamic heap resizing.   Currently MMTk only works with a fixed heap size.

This probably will involve at least the following:

1.  Extending the vm interface to support at least i) min heap, ii) max heap, and iii) no specified heap size.
2.  Adding a number of heap resizing heuristics, probably modeled on existing production systems (eg OpenJDK, PyPy, Julia).

Note: this feature is missing for historical reasons: MMTk was purely a research tool and in that setting, heap resizing is both unimportant (it is useful / essential for end users who don't want to have to specific heap sizes), and a problem (it can confound the measurement of the key time/space tradeoff).   Our focus is much broader now, and the absence of this feature is becoming a problem.

qinsoon:
The minimal we need to do is to allow bindings to set a heap size through an API during run-time. In that case, a VM can implement its own heuristics for heap resizing. To facilitate a binding's implementation, we may provide a standard heap resizing implementation. 

So for us, we may have the following:
* Fixed heap size
* Up to VM
* Dynamic heap resizing (min/max) - optional

qinsoon:
For dynamic heap resizing, we can consider using https://dl.acm.org/doi/10.1145/3563323.

Issue #641, opened 2022-08-12
Allow bindings to implement GC triggering policy


qinsoon:
MMTk currently triggers GCs based on a fixed heap size. ~~We have plans to support dynamic heap size (e.g. https://github.com/mmtk/mmtk-core/issues/561), and we wlll have different GC triggering implementations in that case. Though we will implement different GC triggering policies in MMTk, it is possible that we expose this to the binding. For example, one of the GC triggering policies is to call into the binding and let the binding to decide whether to trigger a GC with some information from MMTk (e.g. reserved pages, free pages, etc).~~ We should also allow VMs to decide whether to trigger GCs.

There are different motivations for this. 
* this could be a replacement for https://github.com/mmtk/mmtk-core/pull/639, which helps account off-heap bytes for the VM. Having an API for the binding to tell us the inc/dec of the bytes is very much not desired. Instead, the binding could implement their own GC triggering and count their allocated bytes/pages.
* some runtimes have its own GC triggering. Like Julia, they trigger GC based on the allocation volume. To have a fair performance comparison with their stock GC, we would need either implement fix heap size for Julia, or allow Julia binding to implement their own GC triggering.

wks:
# Ruby status quo

Note: This section is about the **vanilla Ruby**, *without* MMTk binding.

Ruby has its own mmtk counter.  Ruby wraps `malloc` and `free` into `ruby_xmalloc` and `ruby_xfree`, respectively.  Inside those functions, they increase or decrease the counters and possibly trigger GC.

The counter has two fields:

```c
typedef struct rb_objspace {
    // ...

    struct {
        size_t limit;
        size_t increase;
    } malloc_params;

    // ...
} rb_objspace_t;
```

where:
-   `increase` is the number of bytes allocated by `ruby_xmalloc` minus the number of bytes freed by `ruby_xfree` (saturated to 0 when underflow) since the last GC.
-   `limit` is a threshold.

If `ruby_xmalloc` caused `increase` to exceed `limit`, it will trigger GC.  After GC, it will reset `increase` to 0, and adjust the `limit`.

| variable | description | default (macro name) | adjustable? (env var name) |
| --- | --- | --- | --- |
| `limit` | threshold to trigger GC | 16MiB (`GC_MALLOC_LIMIT_MIN`) | Yes (`RUBY_GC_MALLOC_LIMIT`) |
[...]

PR #712, opened 2022-12-14
Allow different GC triggers

qinsoon:
This PR adds `GCTrigger` and forwards calls related with GC triggering to `GCTrigger`. `GCTrigger` would allow different implementations, such as fixed heap size, dynamic resizing, and runtime delegation. This PR only implements `FixedHeapSizeTrigger` and a simplified version of `MemBalancerTrigger`. This PR closes https://github.com/mmtk/mmtk-core/issues/561, and makes it easier to implement https://github.com/mmtk/mmtk-core/issues/641.

Changes:
* Adds `util::heap::gc_trigger::GCTrigger`. `Plan::poll()` is moved to `GCTrigger`. Any call to `Plan::poll()` is replaced with `GCTrigger::poll()`.
* Adds `util::options::GCTriggerSelector` and `gc_trigger` in `Options`. `gc_trigger` defaults to a fixed heap size of half of the physical memory.
* Refactoring on arguments for creating plans and spaces.
  * I needed to pass `GCTrigger` to plans and spaces so I think it would be better that I just take this chance to do the refactoring. 
  * We now have structs to organize the arguments and any plan and space will receive the same set of the arguments for their constructors. This greatly simplifies our code.

Performance:
There is very little difference for the refactoring in this PR. The following compares this PR with master, both using the same fixed heap size. It shows the refactoring and the changes have no measurable overhead.

gc_trigger
<details>
  <summary>Plotty link (not public visible)</summary>
https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|skunk-2022-12-14-Wed-054336&benchmark^build^invocation^iteration&bmtime^time^time.other^time.stw&|10&iteration^1^2|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^build^benchmark&
</details>

### DIFF
--- a/src/util/options.rs
+++ b/src/util/options.rs
@@ -626,7 +626,7 @@ options! {
     // there is no core with (perceived) id 12.
     // XXX: This option is currently only supported on Linux.
     thread_affinity:        AffinityKind         [env_var: true, command_line: true] [|v: &AffinityKind| v.validate()] = AffinityKind::OsDefault,
-    // Set the GC trigger. This defines how MMTk triggers a GC.
+    // Set the GC trigger. This defines the heap size and how MMTk triggers a GC.
     // Default to a fixed heap size of 0.3x physical memory.
     gc_trigger     :        GCTriggerSelector    [env_var: true, command_line: true] [always_valid] = GCTriggerSelector::FixedHeapSize((crate::util::memory::get_system_total_memory() as f64 * 0.3f64) as usize)
 }
--- a/src/util/options.rs
+++ b/src/util/options.rs
@@ -625,7 +625,10 @@ options! {
     // `MMTK_THREAD_AFFINITY="12" taskset -c 6-12 <program>` will not work, on the other hand, as
     // there is no core with (perceived) id 12.
     // XXX: This option is currently only supported on Linux.
-    thread_affinity:        AffinityKind         [env_var: true, command_line: true] [|v: &AffinityKind| v.validate()] = AffinityKind::OsDefault
+    thread_affinity:        AffinityKind         [env_var: true, command_line: true] [|v: &AffinityKind| v.validate()] = AffinityKind::OsDefault,
+    // Set the GC trigger. This defines how MMTk triggers a GC.
+    // Default to a fixed heap size of 0.3x physical memory.
+    gc_trigger     :        GCTriggerSelector    [env_var: true, command_line: true] [always_valid] = GCTriggerSelector::FixedHeapSize((crate::util::memory::get_system_total_memory() as f64 * 0.3f64) as usize)
 }
 
 #[cfg(test)]
--- a/src/util/options.rs
+++ b/src/util/options.rs
@@ -434,34 +434,48 @@ pub enum GCTriggerSelector {
 }
 
 impl GCTriggerSelector {
-    const K: usize = 1024;
-    const M: usize = 1024 * 1024;
-    const G: usize = 1024 * 1024 * 1024;
+    const K: u64 = 1024;
+    const M: u64 = 1024 * Self::K;
+    const G: u64 = 1024 * Self::M;
+    const T: u64 = 1024 * Self::G;
 
     /// Parse a size representation, which could be a number to represents bytes,
     /// or a number with the suffix K/k/M/m/G/g. Return the byte number if it can be
     /// parsed properly, otherwise return an error string.
     fn parse_size(s: &str) -> Result<usize, String> {
-        println!("parse: {:?}", s);
         let s = s.to_lowercase();
         if s.ends_with(char::is_alphabetic) {
             let num = s[0..s.len() - 1]
-                .parse::<usize>()
+                .parse::<u64>()
                 .map_err(|e| e.to_string())?;
-            if s.ends_with('k') {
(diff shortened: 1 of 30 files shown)


src/util/options.rs
```diff
@@ -426,6 +426,165 @@ impl Options {
...
+                    return Err(format!(
+                        "Min heap size {:?} is larger than the given max heap size {:?}",
+                        min, max
+                    ));
+                }
+                return Ok(Self::DynamicHeapSize(min, max));
+            }
+        } else if s.starts_with("Delegated") {
+            return Ok(Self::Delegated);
+        }
```
wks, line 497: Regular expression should make this easier.

src/util/options.rs
```diff
@@ -426,6 +426,165 @@ impl Options {
...
+impl GCTriggerSelector {
+    const K: usize = 1024;
+    const M: usize = 1024 * 1024;
+    const G: usize = 1024 * 1024 * 1024;
+
+    /// Parse a size representation, which could be a number to represents bytes,
+    /// or a number with the suffix K/k/M/m/G/g. Return the byte number if it can be
+    /// parsed properly, otherwise return an error string.
+    fn parse_size(s: &str) -> Result<usize, String> {
+        println!("parse: {:?}", s);
```
wks, line 445: Looks like this `println!` is a leftover for debugging.

src/util/options.rs
```diff
@@ -426,6 +426,165 @@ impl Options {
...
+pub enum GCTriggerSelector {
+    FixedHeapSize(usize),
+    DynamicHeapSize(usize, usize),
+    Delegated,
+}
+
+impl GCTriggerSelector {
+    const K: usize = 1024;
+    const M: usize = 1024 * 1024;
+    const G: usize = 1024 * 1024 * 1024;
```
wks, line 439: If we allow suffixes, I suggest we also add support for "T" (and maybe "P", and "E" as well).   1TB of RAM is quite possible for today's servers.  x86_64 doesn't really support more than 256TB (`1<<48`) of memory, so 1PB cannot be supported easily.  But it is still nice to have, just for consistency.

src/util/options.rs
```diff
@@ -426,6 +426,165 @@ impl Options {
...
+        if s.ends_with(char::is_alphabetic) {
+            let num = s[0..s.len() - 1]
+                .parse::<usize>()
+                .map_err(|e| e.to_string())?;
+            if s.ends_with('k') {
+                Ok(num * Self::K)
+            } else if s.ends_with('m') {
+                Ok(num * Self::M)
+            } else if s.ends_with('g') {
+                Ok(num * Self::G)
```
wks, line 456: Even if `num` is within the range of `usize`, this may still overflow.  For example, "1073741824G" is equal to 2 to the power of 64, and that overflows usize.  For 32-bit machines, "4G" already overflows `usize`.

Use `usize::checked_mul` instead.  Alternatively, always use `u64` or `u128` when computing, and convert back to `usize` afterwards.

src/util/options.rs
qinsoon: Yeah, this was used for debugging. Thanks.

src/util/options.rs
qinsoon: I added `T` but not `P`. It seems unlikely that we will use petabytes of RAM any time soon. Plus they can use `1024T` anyway.

src/util/options.rs
qinsoon: Done. I am parse the number and do multiplication in `u64`, and convert to usize at the end.

qinsoon:
binding-refs
OPENJDK_BINDING_REF=support-dynamic-heap-resizing
JIKESRVM_BINDING_REF=update-pr-712
V8_BINDING_REF=update-pr-712

src/util/options.rs
```diff
@@ -426,6 +426,197 @@ impl Options {
...
+                .parse::<u64>()
+                .map_err(|e| e.to_string())?;
+            let size = if s.ends_with('k') {
+                num * Self::K
+            } else if s.ends_with('m') {
+                num * Self::M
+            } else if s.ends_with('g') {
+                num * Self::G
+            } else if s.ends_with('t') {
+                num * Self::T
```
wks, line 458: I still think we should still use `checked_mul` here.  Because it is a user-facing interface, it is possible that the user will supply an unreasonably large number and cause an overflow.

The current Rust Reference says overflow will panic in debug mode, but doesn't specify whether it has undefined behaviour or wrapping behaviour.  But it is better to be safe.

wks (changes requested):

src/util/options.rs
qinsoon: Done.

wks (approved):
LGTM

MarisaKirisame:
subscribing myself to this PR. LMK if you guys have any question.

qinsoon:
> subscribing myself to this PR. LMK if you guys have any question.

Thanks a lot for offering help. This PR is mostly about the structural changes to MMTk to support different GC triggers. I will have another PR that implements mem balancer. I will tag you in that PR, and it will be great if you can help check and make sure that it is implemented properly.

PR #714, opened 2022-12-19
Generational plans set `next_gc_full_heap` in `end_of_gc`

qinsoon:
This PR adds `Plan.end_of_gc()`. Our generational plans compute whether we should do full heap GC at the end of `Plan.release()` which is incorrect. There could be other parallel release work when we execute `Plan.release()`. This PR adds `Plan.end_of_gc()`, and we compute `next_gc_full_heap` in `end_of_gc()`.


src/plan/global.rs
```diff
@@ -228,10 +228,15 @@ pub trait Plan: 'static + Sync + Downcast {
...
     fn prepare_worker(&self, _worker: &mut GCWorker<Self::VM>) {}
 
-    /// Release the plan after a GC. This is invoked at the end of a GC when most GC work is finished.
+    /// Release the plan after transitive closure. A plan can implement this method to call each policy's release,
+    /// or create any work packet that should be done in release.
     /// This is invoked once per GC by one worker thread. 'tls' is the worker thread that executes this method.
     fn release(&mut self, tls: VMWorkerThread);
 
+    /// Inform the plan about the end of a GC. It is guaranteed that there is no further work for this GC.
+    /// This is invoked once per GC by one worker thread. 'tls' is the worker thread that exectues this method.
```
k-sareen, line 237:
```suggestion
    /// This is invoked once per GC by one worker thread. `tls` is the worker thread that executes this method.
```

src/plan/global.rs
```diff
@@ -228,10 +228,15 @@ pub trait Plan: 'static + Sync + Downcast {
     /// per-worker preparation. This method is invoked once per worker by the worker thread passed as the argument.
     fn prepare_worker(&self, _worker: &mut GCWorker<Self::VM>) {}
 
-    /// Release the plan after a GC. This is invoked at the end of a GC when most GC work is finished.
+    /// Release the plan after transitive closure. A plan can implement this method to call each policy's release,
+    /// or create any work packet that should be done in release.
     /// This is invoked once per GC by one worker thread. 'tls' is the worker thread that executes this method.
     fn release(&mut self, tls: VMWorkerThread);
```
k-sareen, line 234: Personally, I find the "release" terminology a bit confusing then. What are the semantics of "release" supposed to be exactly? Most (if not all?) plans just release pages from spaces. Is the issue that the `plan.release()` is called at the start of the `Release` work bucket instead of the end? The comment for the `Release` work packet implies that there should only ever one [1] (and only one is ever made in the `scheduler_common_work` function). What is the source of the issue you have encountered? Reference processing?

\[1]: https://github.com/mmtk/mmtk-core/blob/fc144cf8dcc0809abc4d441a1bd74c6cc1efd97a/src/scheduler/gc_work.rs#L91-L97

src/scheduler/gc_work.rs
```diff
@@ -223,6 +223,11 @@ impl<VM: VMBinding> GCWork<VM> for EndOfGC {
     fn do_work(&mut self, worker: &mut GCWorker<VM>, mmtk: &'static MMTK<VM>) {
         info!("End of GC");
 
+        // We assume this is the only running work packet that accesses plan at the point of execution
+        #[allow(clippy::cast_ref_to_mut)]
+        let plan_mut: &mut dyn Plan<VM = VM> = unsafe { &mut *(&*mmtk.plan as *const _ as *mut _) };
```
k-sareen, line 228: I see this everywhere -- maybe we should just change the plan to `mut Plan` or something in the `MMTK` struct. Or am I missing something which means we can't do that?

k-sareen (commented):
I'm just a bit confused as to why this change is necessary. What are the "There could be other parallel release work when we execute `Plan.release()`"? From the scheduler code, it seems like the reference processing is done in the `Release` work bucket as well, is that the issue? If so, could you document it somewhere?

src/plan/global.rs
k-sareen: Ah I see -- it seems like MarkSweep and Immix uses the `Release` work bucket for sweeping [1, 2]. My question is then why not have a separate `Sweep` work bucket like we have with the `Compact` work bucket [3]?

[1]: https://github.com/mmtk/mmtk-core/blob/master/src/policy/marksweepspace/malloc_ms/global.rs#L472-L501
[2]: https://github.com/mmtk/mmtk-core/blob/master/src/policy/immix/immixspace.rs#L316-L340
[3]: https://github.com/mmtk/mmtk-core/blob/43a67557d148fdaff240064c490c90ccdd3e0df2/src/scheduler/work_bucket.rs#L220

k-sareen:
I think it mainly looks good to me -- now that I understand what the issue is (MarkSweep and Immix sweeping in the `Release` work bucket). But it would be good to document it, if possible. Thanks!

src/plan/global.rs
qinsoon: Yeah, right. That is the motivation. The problem is that some plans generate work in the release bucket, and we need a hook to run some stuff after every work packet is done.

What you pointed out reminds me that our current code is actually incorrect. This is a race condition: we run `Release` work (which calls `Plan.release()` with a mutable plan) and those sweeping work (which can access `Plan`).

k-sareen: The comment for the `Release` work packet above, seems to imply that it should not be executing with anything else and that "all the GC work" should have been finished before the `Release` work packet is called. If MS and Immix are performing sweeping in `Release`, then all the GC work hasn't really been finished right? Either we should fix the comment for the `Release` work packet, or make a new work bucket called `Sweep` where MS and Immix generate sweep work packets (in the latter case, the issue of "parallel release work" ceases to exist, unless reference processing is still an issue).

k-sareen: I think that "fixing" the comment is not actually a fix. Yes -- there is a race condition. The `Sweep` work bucket is the more sensible solution (unless you have a better proposal).

qinsoon: Right.

First, I think the original comment is incorrect so I changed it. 'Release' happens after transitive closure, not after a GC.
```rust
/// Release the plan after a GC. This is invoked at the end of a GC when most GC work is finished.
```

Second, we could add a `Sweep` bucket. For mark sweep and immix, they could just push work to `Sweep` instead of `Release`. In that case, our assumption about `Release` is still true. It would be good to get some input from @wks @wenyuzhao. Also, this is not an issue introduced in this PR. I do not plan to fix it in this PR.

k-sareen: > Also, this is not an issue introduced in this PR. I do not plan to fix it in this PR.

Right, but the issue is non-existent if there is just a `Sweep` work bucket right? Unless reference processing, which is the only other thing I could find executing in the `Release` work bucket, is an issue.

k-sareen: > 'Release' happens after transitive closure, not after a GC.

Yes this is the semantics I'm confused by. I thought it was after a GC, but clearly half the GC work (sweeping) is being done in the `Release` work bucket. So there is a bit of a confusion, at least for me.

k-sareen: Also assume we make a Mark-Sweep-Compact plan like many runtimes have (for example Ruby). It seems like with the current work bucket setup, compaction will happen before sweeping? Seems like it's a bit broken, no?

src/scheduler/gc_work.rs
qinsoon: This is the same pattern as our prepare/release code. This is unfortunate. 
https://github.com/mmtk/mmtk-core/blob/2300ce1d4a7c99053c5bb7055b377a71c9241e85/src/scheduler/gc_work.rs#L44-L46
https://github.com/mmtk/mmtk-core/blob/2300ce1d4a7c99053c5bb7055b377a71c9241e85/src/scheduler/gc_work.rs#L113-L115

We can fix it, but not as simple as making it `mut mmtk`. Basically `mmtk: &'static MMTK<VM>` is a non mutable reference, but we want to get a mutable reference from it -- we do so under the assumption that we are the only thread that is accessing `mmtk`. As long as the assumption holds, the code is correct.

Basically what we want is something like [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) but without overhead on borrowing. I don't think there is a safe way to do it.

k-sareen: Why is it not something like an `Arc`? Multiple work packets should be executing at the same time, all of them accessing the `mmtk` variable. Overhead of reference counting? I guess if 99% of the times it's only read like most work packets, and this is the hot loop, ugly hacks like this may be required.

k-sareen: Technically a `RWLock` could help, but I feel like adding a lock in the hottest part of our code may not be ideal.

src/plan/global.rs
qinsoon: * I am guessing the term 'release' is more neutral than 'sweep', as some GC algorithms do not 'sweep'.
* Reference processing is done before `Release`. I think it is unrelated. 
* Our mark compact only uses the marking part of mark sweep, not the sweep part. 
* Mark compact space does nothing in release. I am wondering if we really need the compact bucket. Maybe we could just do the compaction in the release bucket (?).

k-sareen: What is this work packet then: https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L217-L218

> Our mark compact only uses the marking part of mark sweep, not the sweep part.

Yes. I mean a MarkSweep plan that _occasionally_ compacts essentially. Like what Ruby does.

src/scheduler/gc_work.rs
qinsoon: `Arc` can help manage the lifetime of plan/mmtk. But it comes at the cost of a fat pointer (unknown cost) and reference counting (which should be minimal cost). But it does not solve the issue of mutability.

There are many things that we need to tidy up.

k-sareen: Ah right sorry -- you need a `RefCell` or `Mutex`/`RWLock` to have mutability. Sorry was thinking of `Arc<Mutex<T>>` which is a probably a bad idead.

src/plan/global.rs
qinsoon: > What is this work packet then: https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L217-L218

I forgot that. That is another violation of our assumption for `Plan.release()`.

k-sareen: > I am guessing the term 'release' is more neutral than 'sweep', as some GC algorithms do not 'sweep'.

Maybe we should discuss this with Steve as well if there is confusion about the term "Release". If "Sweep" is not general enough we could call it "Collect" or something. Though I will note that many GC algorithms do not compact either and we have a `Compact` work bucket.

wks: > What is this work packet then: https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L217-L218

This is for enqueuing weak references into VM-specific queues.  The same thing can be equivalently done in the `Collection::vm_release`, which is exactly what my vm-side ref processing branch does.

The following is our current `Release` work packet in `master`.

```rust
        // ...
        <C::VM as VMBinding>::VMCollection::vm_release(); // Yes.  This line alone is enough.
        // ...
    }
}
```

From the weak reference processing's point of view, the exact timing of enqueuing reference objects is not that important, because:

1.  The life and death of weak reference objects (as well as finalisation candidates) are determined at the `XxxRefClosure` (or `VMRefClosure` in my branch).  So when enqueuing weak references, it only works on live weak reference objects.  It doesn't matter if any space is not yet reclaimed.
2.  When enqueuing, it enqueues weak reference objects into queues, and the queues are live, too.  Again this step does not deal with dead objects.  So it won't hurt if sweeping is being executed in parallel.

However, it is the most convenient to do the enqueuing in the Release stage.  At that time, all references will have been forwarded, and the weak reference objects contain up-to-date pointers.  It can be done in the `XxxRefClosure` stage (or `VMRefClosure` stage in my branch), too.  (Finalisation candidates are enqueued in `FinalRefClosure`)  But we need to forward then in `RefForwarding` and `FinalizableForwarding` (or `VMRefForwarding` in my branch) when using MarkCompact.

wks: > ... That is another violation of our assumption for `Plan.release()`. But that ...

I don't know what exactly is our "assumption" for `Plan.release()`.  But I think reference enqueuing can be run concurrently with sweeping.

k-sareen: > > ... That is another violation of our assumption for `Plan.release()`. But that ...
> 
> I don't know what exactly is our "assumption" for `Plan.release()`. But I think reference enqueuing can be run concurrently with sweeping.

Seen the first comment in this thread, specifically the comment for the `Release` work packet

qinsoon: > > ... That is another violation of our assumption for `Plan.release()`. But that ...
> 
> I don't know what exactly is our "assumption" for `Plan.release()`. But I think reference enqueuing can be run concurrently with sweeping.

In the `Prepare` and `Release` work packet, we assume we are the only work that is being executed, thus it is safe for us to get a mutable reference of plan. If we have any other work packet, they can access `mmtk` in their `do_work`. And that is racy.

wks: > In the `Prepare` and `Release` work packet, we assume we are the only work that is being executed, thus it is safe for us to get a mutable reference of plan. If we have any other work packet, they can access `mmtk` in their `do_work`. And that is racy.

That makes sense.

```rust
 #[allow(clippy::cast_ref_to_mut)] 
 let plan_mut: &mut C::PlanType = unsafe { &mut *(self.plan as *const _ as *mut _) }; 
 plan_mut.prepare(worker.tls); 
```

I think this is the same issue.

It doesn't look like an issue that can be solved in one or two days.  To ensure we have only one thread having access to the plan, we must stop spreading `&'static Plan` (and &'static MMTk, too) everywhere, and have precise control of who holds reference to `Plan` (and `MMTk` as well).  A key to the solution will be the ability to "revoke" all such references during STW.   In fact, this is what STW is really meant to do, i.e. stop mutators from accessing the heap, as a form of synchronisation.

k-sareen: Yes I agree. Though with regard to this PR (or more specifically, the issue this PR is trying to address), I think the most pragmatic approach is to add a new `Sweep` or `Collect` work bucket and make it so that MS and Immix generate their sweeping work packets into this bucket. In that case we don't really need this PR as all the GC work will have been done before the `should_next_gc_full_heap()` is called.

It will also be helpful to exactly clarify what the semantics of the "Release" work bucket is, i.e. should we send the list of processed references to the binding in the "Release" work bucket or the "Sweep"/"Collect" work bucket, or even the "EndOfGC" work packet (which doesn't seem to have an associated work bucket, really).

k-sareen: > The good thing for RefEnqueue is that it never needs to access the MMTk instance. It can be a pure VM-side operation.

It technically accesses the `mmtk` instance (right now at least -- unsure of your proposed changes):

https://github.com/mmtk/mmtk-core/blob/master/src/util/reference_processor.rs#L556

wks: > It will also be helpful to exactly clarify what the semantics of the "Release" work bucket is, i.e. should we send the list of processed references to the binding in the "Release" work bucket or the "Sweep"/"Collect" work bucket, or even the "EndOfGC" work packet.

I have made an attempt to do this in my branch (see below), but not as precise as you described.
https://github.com/wks/mmtk-core/blob/165b3fab07cc788ad88876effdeefe567697aaf2/src/scheduler/work_bucket.rs#L258

wks: > > The good thing for RefEnqueue is that it never needs to access the MMTk instance. It can be a pure VM-side operation.
> 
> It technically accesses the `mmtk` instance (right now at least -- unsure of your proposed changes):
> 
> https://github.com/mmtk/mmtk-core/blob/master/src/util/reference_processor.rs#L556

It's accessing the `mmtk.reference_processor` field.  Currently it is a field of the `MMTK` struct.  In my branch, it is part of the VM binding instance.

k-sareen:
I still think we should discuss the semantics of "Release" as a group before we merge this PR

k-sareen:
Just noticed the commit message says `next_gc_full_size` instead of `next_gc_full_heap` -- maybe force push an amended commit when you update some of the documentation as mentioned above ^

src/plan/global.rs
qinsoon: Done

qinsoon:
> Just noticed the commit message says `next_gc_full_size` instead of `next_gc_full_heap` -- maybe force push an amended commit when you update some of the documentation as mentioned above ^

I will just do a squash merge, and that commit message will not be seen in the commit history.

PR #718, opened 2022-12-21
Remove unused synchronized counter

### LINKED ISSUE #677: Is `SynchronizedCounter` necessary?
Is the [`SynchronizedCounter`](https://github.com/mmtk/mmtk-core/blob/f71d1950beee79af2150f71608975aaebd1e2572/src/util/synchronized_counter.rs) necessary? Seems like nothing uses it and its functionality can be easily replicated with [`Counter`](https://github.com/mmtk/mmtk-core/blob/master/src/util/statistics/counter/mod.rs#L19).

k-sareen:
Closes #677.


qinsoon (changes requested):
The JikesRVM binding uses this: https://github.com/mmtk/mmtk-jikesrvm/blob/8cbdfd6130473dfc32ee752d500403acdd19e1b4/mmtk/src/active_plan.rs#L13

I think we can change the binding, and have an implementation on the binding side. But we cannot merge this PR as it breaks the binding.

k-sareen:
How odd. This should just be an `AtomicU32` or something -- I don't understand why a new type was required for it. I'll make the necessary changes in the JikesRVM binding.

qinsoon (approved):
LGTM

qinsoon:
I restarted a failed test. https://github.com/mmtk/mmtk-core/pull/757 should fix the problem.

Issue #206, opened 2021-01-05
Allow build on macOS and aarch64


qinsoon:
We should allow mmtk-core to build on macOS/x86_64 and linux/aarch64. At this stage, we only aim to build and run basic tests in mmtk-core on those targets - we do not provide binding support. 

This includes a few tasks:
* [x] build on macOS with CI setup (use Github hosted runner)
* [ ] build on aarch64 with CI setup (use our self-hosted ARM runner)

qinsoon:
Github provides aarch64 linux runners now. We don't need our self-hosted ARM runners to provide tier 2 support for aarch64 linux.

PR #719, opened 2022-12-21
Build MMTk on macOS

qinsoon:
This PR allows mmtk-core to build on macOS (x86_64). This PR finishes a task in https://github.com/mmtk/mmtk-core/issues/206. This PR does not guarantee that MMTk works properly on macOS. It only makes sure that we can build and run our unit tests on macOS. This should be a good first step for any following work on macOS.

Changes:
* Use `malloc_size` instead of `malloc_usable_size` on macOS
* Use `MAP_FIXED` instead of `MAP_FIXED_NOREPLACE` on macOS
* Remove warnings when building `scheduler::affinity` on macOS
* Make a few tests only run on linux (related with thread affinity and mmap)
* Test our pre-code-review test on `macos-12` and `x86_64-apple-darwin`


wks (approved):
For making it build on MacOS, it is OK as long as it builds.  In the long run, we may consider abstracting out a trait which each platform implements.

Issue #668, opened 2022-09-22
Address.store() will drop the old value


qinsoon:
https://github.com/mmtk/mmtk-core/blob/c5d62c4354fe091b7bc4d3bc39d570b7986b93de/src/util/address.rs#L235-L237

We use `*ptr = val` to store the value. However, this will trigger `Drop` on the old value. As in MMTk core, we only store `Copy` types, this does not cause any bug for us. But `Address` and the `store` method is public, and they can be used by the bindings. This could lead to unexpected behaviors.

As a workaround, a binding can do this:
```rust
let ptr = addr.to_mut_ptr();
unsafe { ptr.write(val); } // write does not drop the value
```

We should either use `ptr.write()` to implement `store`, or require the type `T` to be a `Copy` type.

PR #720, opened 2022-12-21
Use `ptr.write()` instead of directly storing to a ptr

k-sareen:
Storing values using `*ptr = value;` results in `Drop` being called on the old value. This could lead to unexpected behaviours.

Closes https://github.com/mmtk/mmtk-core/issues/668.


qinsoon (approved):
LGTM

Issue #723, opened 2022-12-22
Improper DEFAULT_MAX_NURSERY?


qinsoon:
Our current `DEFAULT_MAX_NURSERY` is set to 1G on 64bits machine.
https://github.com/mmtk/mmtk-core/blob/9edb4539826bf733059c9c33106922ad972785bf/src/util/options.rs#L83

According to `collection_required()` in `Gen`, nursery is not considered as full (to trigger a nursery GC) unless its reserved pages reach the max nursery size (1G).
https://github.com/mmtk/mmtk-core/blob/9edb4539826bf733059c9c33106922ad972785bf/src/plan/generational/global.rs#L114-L115

If the heap size is smaller than 1G and we do not set nursery size explicitly, we always trigger full heap GCs, and never trigger a nursery GC.

k-sareen:
> If the heap size is smaller than 1G and we do not set nursery size explicitly, we always trigger full heap GCs, and never trigger a nursery GC.

That's not true, right? I don't remember noticing this while I fixed the nursery size options. The function which actually controls a full-heap GC is `require_full_heap_collection` [1], not `collection_required`.

\[1]: https://github.com/mmtk/mmtk-core/blob/9edb4539826bf733059c9c33106922ad972785bf/src/plan/generational/global.rs#L147-L193

qinsoon:
If `nursery.reserved_pages` is always smaller than `nursery_max_size`, then we will not trigger a GC because of nursery being full. In that case, we will trigger a GC if `plan.reserved_pages` is larger `plan.total_pages`, and that will be a full heap GC.

k-sareen:
So in other words, only highly allocating/generational workloads work as intended... Hmm. From memory we discussed that a large nursery limit was fine since it's the "unbounded" nursery. Surely one can argue that nursery size is a tuning parameter specific to each workload and in order to get best performance, they may have to tune it?

qinsoon:
> So in other words, only highly allocating/generational workloads work as intended... 

It is not about work load. It is max nursery size and heap size. If max nursery size is larger than heap size, we will not have nursery GCs. For example, the default max nursery value (1G) is very likely to be larger than normal 2x-6x heap sizes we use for most dacapo benchmarks. If that happens, we will not see nursery GCs.

I believe we have the same issue for Java MMTk. But their default max nursery is 32M, which is more likely to be smaller than usual heap sizes we use for Dacapo.

[...]

PR #724, opened 2023-01-03
Implement mem balancer

qinsoon:
This PR implements mem balancer (https://dl.acm.org/doi/pdf/10.1145/3563323).

Changes:
* Introduce a trait `GenerationalPlan: Plan`, and move methods that are specific to generational plans to the new trait from the old `trait Plan`.
* Change the return type of `Plan::generational()` to return an option of `&dyn GenerationalPlan`. We can use this to know whether a plan is generational or not. If it is, we can further invoke methods about the generational plan.
* Add `GCTriggerPolicy::on_gc_release()` which is called in the `Release` work packet. A `GCTriggerPolicy` can use this to get stats before we reclaim memory (e.g. to get promoted size).
* Add an implementation of mem balancer to replace the current simple resize policy.
* Add some logging to help debug GC triggering.


qinsoon:
@MarisaKirisame This is the PR that implements mem balancer. I was trying to implement based on my understanding of the paper. Let me know if there is any issue. Any comment is welcome. The changes that may be interesting to you is in `src/util/heap/gc_trigger.rs`.

k-sareen:
Is it possible to implement the fix for nursery GCs we discussed in #723 in this PR? Or will we have a separate PR for that.

qinsoon:
> Is it possible to implement the fix for nursery GCs we discussed in #723 in this PR? Or will we have a separate PR for that.

Yeah, I think that is a different issue.

src/util/heap/gc_trigger.rs
```diff
@@ -139,27 +147,203 @@ pub struct MemBalancerTrigger {
     max_heap_pages: usize,
     /// The current heap size
     current_heap_pages: AtomicUsize,
+    /// Statistics
+    stats: Atomic<MemBalancerStats>,
```
wks, line 151: The Atomic type is used on a complex struct.  The underlying implementation uses a global hash map to assign a spin lock for each hash value of address.  (See:   https://docs.rs/atomic/latest/src/atomic/fallback.rs.html#101 ) This is not really lock-free, and is not efficient, either.

Given that this struct is only used at GC start/release/end when no two threads access that struct concurrently, I suggest using AtomicRefCell and it’s borrow_mut method (you can implement access_stats with them).  It panics if two threads attempt to borrow it mutably at the same time, but given the use pattern, this shouldn’t happen unless there is a bug.

GC work packet stats are implemented using AtomicRefCell, too.

Alternatively you can always use Mutex.

MarisaKirisame:
it look good. there are tricks here and there (e.g. you could try to add a constant 1MB to all heap size), but no doubt you guys will find them when doing some performance turning.

src/util/heap/gc_trigger.rs
qinsoon: Done. I changed that to `AtomicRefCell`.

src/util/options.rs
```diff
@@ -416,14 +416,24 @@ impl FromStr for NurserySize {
...
+        crate::util::conversions::bytes_to_pages_up(self.nursery.max)
+    }
+
     /// Return lower bound of the nursery size (in number of bytes)
-    pub fn get_min_nursery(&self) -> usize {
+    pub fn get_min_nursery_bytes(&self) -> usize {
         self.nursery.min
     }
+
+    /// Return lower bound of the nursery size (in number of bytes)
```
k-sareen, line 433: Should be number of pages:

```suggestion
    /// Return lower bound of the nursery size (in number of pages)
```

wks (commented):
I observed an phenomenon that some methods of `Plan` are only relevant to generational plans.  It includes:

-   The old `generational()` that returns `&Gen<Self::VM>` without `Option`
-   `is_current_gc_nursery`: Only called by the `prepare` method of generational plans and the newly added `GCTrigger`.
-   `get_mature_physical_pages_available`: Only called by `Gen`
-   Newly added `get_mature_reserved_pages`: Only called by `GCTrigger` in `generational_mem_stats_on_gc_release`.

In `Plan`, they are implemented as stubs.  `is_current_gc_nursery` returns `false`; other methods panic if the plan is not generational.

With this PR, the new `generational()` method returns an `Option<&Gen>`. I think we can refactor the code and move those generational-specific methods to `Gen`.  Existing call sites should already have access to `&Gen` instance.  If not, the `Plan::generational()` method can return an `Option` for the caller to match against.  (See my comments on the code below.)

However, I am not sure how to implement `get_mature_physical_pages_available` because it needs to know which space is the "mature space", and that is per-plan knowledge.  Intuitively, there should be another trait, such as

```rust
trait GenPlan: Plan {
    fn is_current_gc_nursery(&self) -> bool;
    fn get_mature_physical_pages_available(&self) -> usize;
    fn get_mature_reserved_pages(&self) -> usize;
}
```

Concrete plans, such as `GenCopy` and `GenImmix` shall implement `GenPlan` in addition to `Plan`.

I am trying to refactor locally, but haven't managed to get the `downcast` operation working.

What do you think?  If it is too complicated, we can defer this refactoring to another PR.

src/scheduler/gc_work.rs
```diff
@@ -108,6 +108,8 @@ impl<C: GCWorkContext> Release<C> {
 impl<C: GCWorkContext + 'static> GCWork<C::VM> for Release<C> {
     fn do_work(&mut self, worker: &mut GCWorker<C::VM>, mmtk: &'static MMTK<C::VM>) {
         trace!("Release Global");
+        self.plan.base().gc_trigger.policy.on_gc_release(mmtk);
```
wks, line 111: Not a problem for now.  This is added before the `VMCollection::vm_release()` below, but `plan_mut.release` happens after `VMCollection::vm_release()`.  It's a bit weird this way, but it doesn't matter if the VM doesn't implement `vm_release`.

 `VMCollection::vm_release()` was created for VMs to do what our current `RefEnqueue` does to handle weak references.  I am considering moving `VMCollection::vm_release()` to a dedicated work packets, because ref-enqueuing work does not depend on the `Plan` instance or the release of the `Plan`, and can therefore be parallelised.

src/util/heap/gc_trigger.rs
```diff
@@ -139,27 +147,203 @@ pub struct MemBalancerTrigger {
...
+const NO_PREV: f64 = -1f64;
+
+impl std::default::Default for MemBalancerStats {
+    fn default() -> Self {
+        let now = Instant::now();
+        Self {
+            allocation_pages_prev: NO_PREV,
+            allocation_time_prev: NO_PREV,
+            collection_pages_prev: NO_PREV,
+            collection_time_prev: NO_PREV,
```
wks, line 197: The only use case of those fields (`smooth`) checks the value against `NO_PREV`. It is obviously a good use case of `Option<T>`.

```rust
        let smooth = |prev: Option<f64>, cur, factor| {
            prev.map(|p| p * factor + cur * (1.0f64 - factor)).unwrap_or(cur)
        };
```

qinsoon: Done

qinsoon:
@wks I have done some refactoring about methods that are in `Plan` but should be specific to generational plans. See https://github.com/mmtk/mmtk-core/pull/724/commits/5aa7023fcd30f6861d457d7aa50ceef1516425f6 for the changes. In short, `Plan::generational()` now returns `Option<&dyn GenerationalPlan>` so we do not need to do downcast. Let me know if there could be any issue on this change.

src/scheduler/gc_work.rs
qinsoon: If we can run some code when we open the `Release` bucket, that would be a good timing to run this `on_gc_release()` method.

wks (changes requested):
Nice work.

Two problems remain.

One is the naming of the `Gen` struct and the `gen()` method.  It is a bit confusing to have both `trait GenerationalPlan` and `struct Gen`.  Similar is true for `generational()` and `gen()`.  I think we can rename `Gen` to `CommonGenPlan` to indicate that it plays a similar role as `CommonPlan`, i.e. `CommonGenPlan` is the common parts of most generational plans, whereas `CommonPlan` is the common parts of most plans.

The other is the `is_nursery_gc` function.  This function works for both generational plans and non-generational plans.  It is used in two places.
1.  Finalization processor. It still uses `is_nursery_gc` as a hit of which finalizable objects it needs to look at.  It is still helpful.
2.  The `ProcessModBuf` and `ProcessRegionModBuf` work packets.  Those work packets are only used by generational plans, so they can be sure that the current plan is generational.  I think they can skip the `map_or` part, and panic if the plan is not generational.

src/plan/generational/gc_work.rs
```diff
@@ -24,7 +24,7 @@ impl<VM: VMBinding> ProcessEdgesWork for GenNurseryProcessEdges<VM> {
 
     fn new(edges: Vec<EdgeOf<Self>>, roots: bool, mmtk: &'static MMTK<VM>) -> Self {
         let base = ProcessEdgesBase::new(edges, roots, mmtk);
-        let gen = base.plan().generational();
+        let gen = base.plan().generational().unwrap().gen();
```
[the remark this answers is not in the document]
qinsoon: Sounds good. I renamed it to `common_gen()`.

src/plan/generational/gc_work.rs
```diff
@@ -95,7 +95,7 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for ProcessModBuf<E> {
             );
         }
         // scan modbuf only if the current GC is a nursery GC
-        if mmtk.plan.is_current_gc_nursery() {
+        if crate::plan::is_nursery_gc(&*mmtk.plan) {
```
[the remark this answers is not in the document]
qinsoon: Sounds good. I changed it to what was suggested.

src/plan/generational/gc_work.rs
```diff
@@ -129,7 +129,7 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for ProcessRegionModBuf<E> {
     #[inline(always)]
     fn do_work(&mut self, worker: &mut GCWorker<E::VM>, mmtk: &'static MMTK<E::VM>) {
         // Scan modbuf only if the current GC is a nursery GC
-        if mmtk.plan.is_current_gc_nursery() {
+        if crate::plan::is_nursery_gc(&*mmtk.plan) {
```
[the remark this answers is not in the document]
qinsoon: Done.

wks (approved):
It looks good to me now.

PR #727, opened 2023-01-05
Fix how nursery GCs are triggered

k-sareen:
Closes https://github.com/mmtk/mmtk-core/issues/723.

### DIFF
--- a/.github/workflows/api-check.yml
+++ b/.github/workflows/api-check.yml
@@ -32,6 +32,6 @@ jobs:
 
       # Install cargo-public-api
       - name: Install cargo-public-api
-        run: cargo install cargo-public-api
+        run: cargo install cargo-public-api --version 0.26.0
       - name: API Diff
         run: cargo public-api diff origin/${GITHUB_BASE_REF}..${{ github.event.pull_request.head.sha }} --deny=all
--- a/.github/workflows/api-check.yml
+++ b/.github/workflows/api-check.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           # We need nightly for cargo-public-api to get the API output.
-          toolchain: nightly
+          toolchain: nightly-2023-01-04
           profile: minimal
           # It is not necessary to use nightly as default (which is used to install cargo-public-api and compile our code).
           # However, our current toolchain is 1.59.0, and cargo-public-api requires 1.60 at least. To make it simple,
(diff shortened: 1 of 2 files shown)


k-sareen:
I don't know how I missed it in my original PR (I guess the bug hid this behaviour), but now the `requires_full_heap_gc()` function is the exact same as it is in the original Java MMTk [1].

\[1]: https://github.com/mmtk/jikesrvm/blob/master/MMTk/src/org/mmtk/plan/generational/Gen.java#L252-L267

k-sareen:
@qinsoon I think the Public API Check CI is broken? Using an unknown option now it seems like.

qinsoon:
> @qinsoon I think the Public API Check CI is broken? Using an unknown option now it seems like.

It seems `cargo-public-api` changes their command line options. Can you change it?
https://github.com/Enselic/cargo-public-api/blob/main/docs/CI-EXAMPLES.md
Alternatively, we can lock to a specific version of `cargo-public-api` with `cargo install --version`. It seems they have breaking changes quite frequently.

k-sareen:
> > @qinsoon I think the Public API Check CI is broken? Using an unknown option now it seems like.
> 
> It seems `cargo-public-api` changes their command line options. Can you change it? https://github.com/Enselic/cargo-public-api/blob/main/docs/CI-EXAMPLES.md Alternatively, we can lock to a specific version of `cargo-public-api` with `cargo install --version`. It seems they have breaking changes quite frequently.

Yeah it does seem like it changes frequently. Apparently due to rustdoc JSON (whatever that is?) being unstable. Should I just freeze to a previously working version or should I get the latest one working here (https://github.com/Enselic/cargo-public-api/blob/main/README.md#compatibility-matrix) and then freeze it? I don't exactly recall where we landed on testing/using latest Rust versions (I think we were in agreement that bumping up the version is fine, but not sure if we've acted on it).

qinsoon:
> Should I just freeze to a previously working version or should I get the latest one working here (https://github.com/Enselic/cargo-public-api/blob/main/README.md#compatibility-matrix) and then freeze it? I don't exactly recall where we landed on testing/using latest Rust versions (I think we were in agreement that bumping up the version is fine, but not sure if we've acted on it).

Either is fine. That matrix doesn't matter. We use latest `nightly` for that particular Github action: https://github.com/mmtk/mmtk-core/blob/06237f15039d08b9faa710ce4e32a422fd98996e/.github/workflows/api-check.yml#L25
If we lock to a version, we need the version before they change `--diff-git-checkouts A B` to `diff A..B`.

k-sareen:
Done @qinsoon!

.github/workflows/api-check.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           # We need nightly for cargo-public-api to get the API output.
-          toolchain: nightly
+          toolchain: nightly-2023-01-04
```
qinsoon, line 25: If we lock to a Rust toolchain version, we'd better lock to a `cargo-public-api` version as well (in the `cargo install` below). Otherwise, when `cargo-public-api` upgrades the toolchain version, the check will fail again.

k-sareen: Oh right. Yep my bad.

qinsoon (approved):
LGTM

PR #729, opened 2023-01-05
Cancel CI runs if PRs get updated

### LINKED ISSUE #728: Cancel CI runs if PRs get updated
We should add the following to every CI in order to cancel a run for a PR if a new commit occurs. Better if we don't waste compute cycles for nothing.

```yaml
concurrency:
  # Cancels pending runs when a PR gets updated.
  group: ${{ github.head_ref || github.run_id }}
  cancel-in-progress: true
```

k-sareen:
Closes #728.

### DIFF
--- a/.github/workflows/api-check.yml
+++ b/.github/workflows/api-check.yml
@@ -5,6 +5,11 @@ on:
     branches:
       - master
 
+concurrency:
+  # Cancels pending runs when a PR gets updated.
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 # The workflow may fail if we change the public API in a pull request.
 # We allow fail on this action. But we should manually check if the changes are reasonable when we see a failed action.
 # It would be good if the workflow returns a neutral status when we find API changes. But it is currently not
--- a/.github/workflows/api-check.yml
+++ b/.github/workflows/api-check.yml
@@ -1,5 +1,6 @@
 name: Public API Check
 
+# Adding a random comment to a file to test if PR runs get canceled
 on:
   pull_request:
     branches:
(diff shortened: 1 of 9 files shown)


qinsoon:
I think you would need the group name to be a combination of head ref and job name. Otherwise only all the workflows and all the jobs share one group, and only one can run.

k-sareen:
Yeah I see. Since we have multiple workflows/CIs it's more complex. Something like this maybe:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
  cancel-in-progress: true
```

qinsoon:
> Yeah I see. Since we have multiple workflows/CIs it's more complex. Something like this maybe:
> 
> ```yaml
> concurrency:
>   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
>   cancel-in-progress: true
> ```

I think you need the job name or ID in the group name as well. One file is one workflow, but it can include multiple jobs. You still want all the jobs in one workflow to run.

k-sareen:
> > Yeah I see. Since we have multiple workflows/CIs it's more complex. Something like this maybe:
> > ```yaml
> > concurrency:
> >   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
> >   cancel-in-progress: true
> > ```
> 
> I think you need the job name or ID in the group name as well. One file is one workflow, but it can include multiple jobs. You still want all the jobs in one workflow to run.

It seems to be working right now? I'll add a dummy change to see if it's canceling as expected.

qinsoon:
If it runs, that's fine. I was just guessing based on their document.

k-sareen:
Cool! Seems like runs actually get canceled if you update the PR.

qinsoon (approved):
LGTM

PR #731, opened 2023-01-08
Add tcmalloc library for malloc MarkSweep

k-sareen:


k-sareen:
As an aside, why was the Hoard dependency removed? And if we no longer have Hoard, then we should remove the `malloc_hoard` feature from our code.

PR #732, opened 2023-01-09
`Map` and `Mmapper` as trait objects

wenyuzhao:
This PR convert `Map` and `Mmapper` instances into trait objects, which builds an abstraction layer over the different VM allocation and mapping policy. This is useful for compressed pointers or 39-bit virtual space support, as they will need to use a different  `Map` or `Mmapper` other than the default 64-bit implementation.


src/util/heap/layout/map64.rs
```diff
@@ -28,15 +28,20 @@ pub struct Map64 {
...
-impl Map for Map64 {
-    type FreeList = RawMemoryFreeList;
-
-    fn new() -> Self {
+impl Map64 {
+    pub fn new() -> Self {
         let mut high_water = vec![Address::ZERO; MAX_SPACES];
         let mut base_address = vec![Address::ZERO; MAX_SPACES];
 
+        assert!(
```
qinsoon, line 36: If `Map64` does not work on 32 bits, we can guard the entire module with `#[cfg(target_pointer_width = "64")]`.

src/util/heap/layout/mod.rs
```diff
@@ -3,17 +3,11 @@ pub mod heap_parameters;
 pub mod vm_layout_constants;
 pub mod mmapper;
 pub use self::mmapper::Mmapper;
-#[cfg(target_pointer_width = "32")]
 mod byte_map_mmapper;
-#[cfg(target_pointer_width = "32")]
 pub use self::byte_map_mmapper::ByteMapMmapper;
-#[cfg(target_pointer_width = "64")]
 mod fragmented_mapper;
-#[cfg(target_pointer_width = "64")]
```
qinsoon, line 12: Does `FragmentedMmapper` work for 32 bits? If not, we do not need to compile it for 32 bits.

src/mmtk.rs
```diff
@@ -31,10 +31,19 @@ lazy_static! {
     // TODO: We should refactor this when we know more about how multiple MMTK instances work.
 
     /// A global VMMap that manages the mapping of spaces to virtual memory ranges.
-    pub static ref VM_MAP: VMMap = VMMap::new();
+    pub static ref VM_MAP: Box<dyn Map> =  if cfg!(target_pointer_width = "32") {
```
qinsoon, line 34: If you cannot use `#[cfg(...)]` inside `lazy_static!`, you can do this
```rust
lazy_static! { ... }
lazy_static! { ... }
```

qinsoon (commented):
Generally good. But I don't think we should use conditional compilation, and avoid including 64bits mmapper and vmmap in 32 bits if they don't work on 32 bits.

src/util/heap/layout/mod.rs
wenyuzhao: Both FragmentedMmapper and Map64 are removed from the 32-bit build.

qinsoon (approved):
LGTM

Issue #733, opened 2023-01-09
Investigate use and performance of `inline` directives


k-sareen:
We're fairly liberal with our usage of `inline` directives (be it `inline(always)` or `inline(never)`). Ideally we let the compiler make inlining decisions except for the hot loop (like `trace_object`) which we may have to tune for maximum performance.

Inlining too much can lead to too many register spills and an increase in code size which can lead to very hard to debug performance issues (for example, we may see a random performance drop if code moves around).

qinsoon:
As discussed, we probably will remove the current compiler directives, and use [profile-guided optimisation](https://rustc-dev-guide.rust-lang.org/profile-guided-optimization.html) to generate Rust binary (or use it to force inlining etc). We can just use one benchmark to profile, we may either have one profile per plan, or use the default 'production' plan (the fastest plan).

k-sareen:
I'll close this issue after having investigated the affects of different profiling steps. I will update the best practices in both mmtk-core and mmtk-openjdk then.

k-sareen:
[Internal link](http://squirrel.anu.edu.au/plotty/kunals/bench-test/p/3Fnxqd). We ran an experiment which tried different ways to generate the PGO profile. The different profiles were:

1. `master` (106fc714), no inlining, no PGO
2. PGO profile generated using `GenImmix` with a stress factor of 4 MB and 3 invocations
3. PGO profile generated using `GenImmix` with a stress factor of 4 MB and 1 invocation
4. PGO profile generated using `GenImmix` with a stress factor of 16 MB and 1 invocation
5. PGO profile generated using `GenImmix`, `Immix, and `SemiSpace` with a stress factor of 4 MB and 1 invocation for each profiling GC
6. PGO profile generated using `GenImmix`, `Immix, and `SemiSpace` with a stress factor of 16 MB and 1 invocation for each profiling GC

From the results we can see that for the stop-the-world time, `Immix` and `SemiSpace` perform worse than the baseline (build 1) for builds 2-4, while exceed the baseline for builds 5-6 by around 5-6%. For `GenImmix` all PGO builds perform better than the baseline, however, build 1 and build 6 are the only ones with 5%+ improvements to the stop-the-world time. Overall **build 6 is the clear winner** in terms of performance across GCs.

[...]

PR #735, opened 2023-01-16
Remove all #[inline] attributes

k-sareen:


qinsoon:
According to the results we have (https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/inline.20directives/near/321556606), simply removing all the inline directives will slow down our GC by a few percentages. And its performance impact on the mutator time is unknown (we measured with a JIT-generated allocation fastpath). 

Our goal on https://github.com/mmtk/mmtk-core/issues/733 is to improve performance (or at least not degrade performance), and remove as many inline directives as possible (so that we can have a cleaner code base). Simply removing all the inline directives won't achieve the goal.

k-sareen:
Technically if the PGO build works, then we never have to touch any `inline` directives at all. So I guess we should merge this PR after we know PGO works.

k-sareen:
This commit is stale so don't merge it yet. I'll add a script that directly builds the PGO build as well. Though maybe that should sit in the binding repo?

k-sareen:
@qinsoon How should I lay out the two PRs? Should I make a PGO section in the READMEs of both mmtk-core and mmtk-openjdk?

qinsoon:
> @qinsoon How should I lay out the two PRs? Should I make a PGO section in the READMEs of both mmtk-core and mmtk-openjdk?

This is just what I am thinking about.

For mmtk-core, you can 
1. add a subsection in the Build section in mmtk-core, mentioning that we recommend using PGO when you try build and measure performance with MMTk. You can just include a link to the Rust PGO guide. 
2. add some comments in the `lib.rs` to mention that we deliberately remove all the inline directives, and rely on the compiler and PGO for inlining.

For the OpenJDK binding, you can include a more detailed guide about how to build the OpenJDK binding with PGO.

k-sareen:
Alright makes sense to me. I'll do that. Note that I've also bumped the Rust version to 1.66.1 (due to LLVM tool compatibility) in this PR. That's why we're seeing some new style checks I presume.

qinsoon:
> Alright makes sense to me. I'll do that. Note that I've also bumped the Rust version to 1.66.1 (due to LLVM tool compatibility) in this PR. That's why we're seeing some new style checks I presume.

I would suggest not upgrading the Rust toolchain in this PR. I assume you do not have to mention the exact version in the build description.

k-sareen:
Right -- the results above were based on Rust v1.66.1. So we should just keep that in mind. I think we have `llvm-tools` installed for the older Rust version as well.

qinsoon:
> Right -- the results above were based on Rust v1.66.1. So we should just keep that in mind. I think we have `llvm-tools` installed for the older Rust version as well.
> 
> EDIT: @qinsoon Actually no. We don't have the `llvm-tools` installed for the older Rust version. Only for v1.66.1.

OK. Then we probably have to upgrade the Rust version.

k-sareen:
So the previous result mistakenly compiled `master` with Rust v1.59.0 as opposed to Rust v1.66.1. This result compiles `master` with Rust v1.66.1: http://squirrel.anu.edu.au/plotty/kunals/bench-test/p/QTMP3M. GenImmix and Immix see performance benefits (or at least no performance regressions), but there is a 10% slowdown for SemiSpace.

As an aside, changing Rust version to v1.66.1 has a big benefit as we can see in this result: http://squirrel.anu.edu.au/plotty/kunals/bench-test/p/WDptqC

k-sareen:
@qinsoon I don't know if I should fix all the clippy errors in this PR. It's unrelated to remove `inline` attributes. I think the best way to approach this is to have a PR that changes Rust version to v1.66.1 that will fix all the clippy errors as well and then we merge this PR after that (same for the mmtk-openjdk PR). Does that make sense to you?

qinsoon:
> @qinsoon I don't know if I should fix all the clippy errors in this PR. It's unrelated to remove `inline` attributes. I think the best way to approach this is to have another PR that changes Rust version to v1.66.1 that will fix all the clippy errors as well and then we merge this PR after that (same for the mmtk-openjdk PR). Does that make sense to you?

Yes. I agree.

README.md
```diff
@@ -41,6 +41,24 @@ The artefact produced produced by the build process can be found under
...
 
+### Profile-Guided Optimized Build
+
+For the best performance, we recommend using a profile-guided optimized (PGO)
+build. Rust supports [PGO builds](https://doc.rust-lang.org/rustc/profile-guided-optimization.html)
+by hooking into the LLVM profiling infrastructure. The idea is that we gather
+profiling data by running a representative benchmark and then later use the
+profiling data as a feedback on making optimization decisions.
+
+It is recommended to choose the best-performing GC for the profiling step. For
```
qinsoon, line 52: This is not true according to your results. You can update this section once we have a conclusion.

qinsoon (approved):
LGTM

README.md
k-sareen: Yeah -- for the time-being it's fine, I think. I can edit it to say something like "We use our best-performing GC for the profiling step in order to get the best performance results", if you want.

Issue #669, opened 2022-09-28
Slow initialization for VM map


qinsoon:
Following our discussion in https://github.com/mmtk/mmtk-core/pull/632#issuecomment-1257457406, after fixing the initialisation cost for SFT map in https://github.com/mmtk/mmtk-core/pull/632, our boot time is still significantly slower than the stock OpenJDK.

See the table below (from https://github.com/mmtk/mmtk-core/pull/632#issuecomment-1257457406):
|                              | Elapsed wall clock time (s) | Maximum resident set size (kbytes) | thread attach log time |
|------------------------------|-----------------------------|------------------------------------|--|
| Stock OpenJDK                | 0.17                        | 60696                              | 0.021s |
| MMTk - sparse chunk map (master)                | 0.52                        | 839596                             | 0.349s |
| ... | ...                        | ...                            | ... |
| MMTk - space map (PR)        | 0.30                        | 314604                             | 0.151s |

`perf` suggests the time is spent in initializing the MMTk instance (`MMTK::new()`). I instrumented the method, and found that 93% of the time in `MMTK::new()` is spent in `VMMap::new()`. 

Our `Map64` is a straightforward port of https://github.com/JikesRVM/JikesRVM/blob/master/MMTk/src/org/mmtk/utility/heap/layout/Map64.java from the Java MMTk. I am not familiar with the code, and I haven't looked further into the problem.

PR #737, opened 2023-01-17
Fix slow vec allocation during start-up.

wks:
Use alloc::alloc_zeroed to manually allocate the buffer for the vector of `SpaceDescriptor(usize)` to elide the storing of zeroes to its elements during start-up.

This is one of the two parts for solving the problem of slow start-up time.

Related issues:
-   https://github.com/mmtk/mmtk-core/issues/669
-   https://github.com/mmtk/mmtk-ruby/issues/13

The problem is, Rust has specialised Vec creation mechanisms for built-in types, such as `usize`.  This means `vec![0usize; 33554432]` will use pre-zeroed buffers to create the Vec, and will not store `0` values to its elements.  This is very fast.  However, Rust cannot do the same optimisation for `vec![SpaceDescriptor(0); 33554432]`. It will have to store a `SpaceDescriptor(0)` to each of its 33554432 elements, and that'll be very slow.

This PR implements our own Vec creation function using `std::alloc::alloc_zeroed`.  This change alone can reduce the execution time of `MMTK::new()` from 140ms to 60ms on my machine.


src/util/mod.rs
```diff
@@ -65,6 +65,8 @@ pub(crate) mod statistics;
 pub(crate) mod test_util;
 /// A treadmill implementation.
 pub(crate) mod treadmill;
+/// Fast allocation of large zeroed arrays
+pub(crate) mod zeroed_alloc;
```
qinsoon, line 69: There is an existing module `util::rust_util` where we put a few stuff that we wish Rust provides could but it does not provide yet. This could be moved to `rust_util`, but I don't have a strong opinion on this. It is okay to keep this as a separate module.

qinsoon (approved):
LGTM

src/util/mod.rs
wks: I think `rust_util` is a good place for this because it contains things that work around limitations of the Rust language.  I moved `zeroed_alloc` to `util::rust_util::zeroed_alloc`.

PR #738, opened 2023-01-17
mmap adjacent chunks

wks:
This avoids calling mmap too many times.

This is part 2/2 for solving the slow start-up problem. 

Related issues:
-   The first part: https://github.com/mmtk/mmtk-core/pull/737
-   https://github.com/mmtk/mmtk-core/issues/669
-   https://github.com/mmtk/mmtk-ruby/issues/13

Currently `FragmentedMaper::quarantine_address_range` calls `mmap` once for each 4MiB chunk.  While each `mmap` call only takes about 1us, the large address space will require it to quarantine 50000+ such chunks, and the cost will accumulate.  It introduces a 60ms delay on start-up, which is noticeable for light-weight scripting languages like Ruby.

This PR will find adjacent chunks with state = `MapState::Unmapped` and call `mmap` once for those chunks.

As a result, MMTk only calls `mmap` 8 times (instead of 50000+ times) for quarantining memory.  When both this PR and the previous "part 1" are applied, the start-up time for Ruby is reduced from 140ms to 10-15ms.

Issues to consider:

-   Should we coalesce chunks for adjacent slabs too?
    -   Yes. Without any form of coalescing, it calls mmap 50000+ times during startup; when coalescing only chunks within a slab, it reduces mmap to 100+ times; when coalescing between slabs, it calls mmap only 8 times for quarantining.  That is equal to the number of calls to `FragmentedMapper::quarantine_address_range`.  I have already implemented it in this PR.
-   Should we do this kind of coalescing for `FragmentedMapper::ensure_mapped`, too?
    -   No, at least not necessary now.  From the log, `ensure_mapping` is usually called at page granularity, or several pages at a time.  As a result, it only calls `mmap` to map one chunk at a time, if the chunk is not already mapped.  Observing this use pattern, I think it is unnecessary to optimise for `ensure_mapped` at this time.  But I have abstracted out the algorithm for grouping elements in a slice of slices (i.e. a list of slabs) so that is is easy to adapt this algorithm to `ensure_mapping`, too.
-   Should we change the algorithm so that we don't need to map the memory (even with `PROT_NONE`) on start-up?
    -   We can discuss that in the future.  I think it is legit to mmap large chunk of memory with `PROT_NONE` because it doesn't actually request physical memory from the operating system.


src/util/rust_util/sos_util.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        Some(rv)
+    }
+}
+
+/// An iterator that interates through groups of elements in a slice of slice.
+///
+/// Each group contains a contiguous range of elements that may span multiple inner slices.
+/// All elements in a group share the same key.
+pub(crate) struct SliceOfSlicesGrouper<'s, T, K, F>
```
qinsoon, line 71: It took me a while to understand what this does. I think essentially it does [`flatten()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flatten) on the 'slice of slice', and [`group_by()`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.group_by) (from `itertools`) to group the resulting slice by keys. 

I would suggest try using those library methods and see if they work. In that case, we do not need to maintain this type. If they do not work for some reason, I suggest just adding comments for this type to make it clear that it does both `flatten` and `group_by`.

src/util/rust_util/sos_util.rs
```diff
@@ -0,0 +1,274 @@
...
+            self.maybe_prev_key = None;
+
+            Some(group)
+        } else {
+            None
+        }
+    }
+}
+
+#[test]
```
qinsoon, line 198: Probably put the tests to a `#[cfg(test)]` module to avoid including them in the normal generated binary.

wks: Done. I moved tests into a `tests` module.

src/util/rust_util/sos_util.rs
wks: Yes.  But one problem with the `group_by` provided by `Itertools` is that it is a one-way iterator.  But our mmapper needs to traverse each group twice to coalesce `mmap` calls:

1. First it finds all `MapState` elements with the same state, in order to compute the size of the memory range with the same state.  This size will be passed to `mmap`.
2. Then after calling `mmap`, it will visit every `MapState` in this group again to transition the state to `Quarantined`.

`Itertools::group_by` does not allow me to do this.

Fortunately, `flatten()` still does what we need.  I removed this `sos_util` module, and replaced with a simpler `RevisitableGroupBy`.  It is similar to `Itertools::group_by`, but it pre-iterate through the group to find its length before letting the user iterate through the group again.  It also play nicely with `.flatten()`, and this greatly simplified the implementation.

qinsoon (approved):
LGTM

PR #741, opened 2023-01-24
Sticky immix

### LINKED ISSUE #683: Clarify DEFRAG in Immix (vs non-moving)
We have a compile-time constant `DEFRAG` in Immix.
https://github.com/mmtk/mmtk-core/blob/53b6d57c2a96f6064e01e245ea0be05dd673cfad/src/policy/immix/mod.rs#L19

Generally Immix allows *opportunistic copying*, and defrag GC is one of the reasons Immix will do opportunistic copying. In our current code base, defrag GC is the only case where we do opportunistic copying in Immix. Thus we interchangeably use non-defrag and non-moving, which is wrong.

We should clarify this, especially when we have sticky immix (https://github.com/mmtk/mmtk-core/issues/682):
* no opportunistic copying === non moving
* no defrag !== no opportunistic copying

### LINKED ISSUE #742: `harness_begin` does not force a full heap GC
https://github.com/mmtk/mmtk-core/blob/a272c2399df39177361788449b1f96d060a86080/src/mmtk.rs#L144-L146

It only does a full heap GC if the option `full_heap_system_gc` is `true` (but its default is `false`):
https://github.com/mmtk/mmtk-core/blob/a272c2399df39177361788449b1f96d060a86080/src/util/options.rs#L660

This may affect our performance measurement for generational plans.

qinsoon:
This PR adds `StickyImmix` as a plan (https://github.com/mmtk/mmtk-core/issues/682). 

Changes:
* Refactoring to current code for sticky plans
  * Introduce more methods to `trait GenerationalPlan` and add `trait GenerationalPlanExt`
  * Refactor `GenObjectBarrierSemantics` and `GenNurseryProcessEdges` so they use the generational plan traits, and work for sticky plans. 
  * Macro-generated post scan will delegate the call to 'parent' if the object is not in any space in the current plan.
  * Extract work scheduling code to `Immix::schedule_immix_full_heap_collection` so all immix-related plans can use it.
  * Add `ImmixSpaceArgs` that changes some Immix behaviors and make it possible to use as a nursery/mature space for sticky immix.
* Add sticky immix
* Refactor `trait Plan` and move some methods to `trait GenerationalPlan`.
* Refactor sanity GC to allow check object sanity with plans and bindings.
* Bug fix: all the plans that use object barrier may have duplicated edges (from memory slice copy barrier)
* Bug fix: clarify non-defrag vs non-moving in Immix (this closes #683)
* Bug fix: `harness_begin` will force a full heap GC (this closes #742)
* Add some logging
  * info log used/total pages before and after GC
  * info log elapsed time for each GC
  * trace log some info about edge tracing, object scan, and a few other places.

### DIFF
--- a/src/scheduler/controller.rs
+++ b/src/scheduler/controller.rs
@@ -14,7 +14,7 @@ use crate::vm::VMBinding;
 use crate::MMTK;
 use atomic::Ordering;
 
-use super::{GCWork, GCWorkScheduler, GCWorker};
+use super::{CoordinatorWork, GCWorkScheduler, GCWorker};
 
 /// The thread local struct for the GC controller, the counterpart of `GCWorker`.
 pub struct GCController<VM: VMBinding> {
@@ -64,22 +64,9 @@ impl<VM: VMBinding> GCController<VM> {
 
     /// Process a message. Return true if the GC is finished.
     fn process_message(&mut self, message: CoordinatorMessage<VM>) -> bool {
-        let worker = &mut self.coordinator_worker;
-        let mmtk = self.mmtk;
         match message {
             CoordinatorMessage::Work(mut work) => {
-                work.do_work_with_stat(worker, mmtk);
-                let old_count = self
-                    .scheduler
-                    .pending_coordinator_packets
-                    .fetch_sub(1, Ordering::SeqCst);
-                if old_count == 1 {
-                    // When the coordinator finishes executing all coordinator work packets,
-                    // it is a chance to open more work buckets.
-                    // Notify one worker so it can open buckets.
-                    let _guard = self.scheduler.worker_monitor.0.lock().unwrap();
-                    self.scheduler.worker_monitor.1.notify_one();
-                }
+                self.execute_coordinator_work(work.as_mut(), true);
                 false
             }
             CoordinatorMessage::Finish => {
@@ -97,7 +84,7 @@ impl<VM: VMBinding> GCController<VM> {
     pub fn do_gc_until_completion(&mut self) {
         let gc_start = std::time::Instant::now();
         // Schedule collection.
-        ScheduleCollection.do_work_with_stat(&mut self.coordinator_worker, self.mmtk);
+        self.initiate_coordinator_work(&mut ScheduleCollection, true);
 
         // Tell GC trigger that GC started - this happens after ScheduleCollection so we
         // will know what kind of GC this is (e.g. nursery vs mature in gen copy, defrag vs fast in Immix)
@@ -124,7 +111,14 @@ impl<VM: VMBinding> GCController<VM> {
                 CoordinatorMessage::Finish => {}
             }
         }
-        self.scheduler.deactivate_all();
+
+        {
+            // Note: GC workers may spuriously wake up, examining the states of work buckets and
+            // trying to open them.  Use lock to ensure workers do not wake up when we deactivate
+            // buckets.
+            let _guard = self.scheduler.worker_monitor.0.lock().unwrap();
+            self.scheduler.deactivate_all();
+        }
 
         // Tell GC trigger that GC ended - this happens before EndOfGC where we resume mutators.
         self.mmtk.plan.base().gc_trigger.policy.on_gc_end(self.mmtk);
@@ -134,11 +128,57 @@ impl<VM: VMBinding> GCController<VM> {
         //       Otherwise, for generational GCs, workers will receive and process
         //       newly generated remembered-sets from those open buckets.
         //       But these remsets should be preserved until next GC.
-        EndOfGC {
+        let mut end_of_gc = EndOfGC {
             elapsed: gc_start.elapsed(),
-        }
-        .do_work_with_stat(&mut self.coordinator_worker, self.mmtk);
+        };
+        self.initiate_coordinator_work(&mut end_of_gc, false);
 
         self.scheduler.debug_assert_all_buckets_deactivated();
+    }
+
+    /// The controller uses this method to start executing a coordinator work immediately.
+    ///
+    /// Note: GC workers will start executing work packets as soon as individual work packets
+    /// are added.  If the coordinator work (such as `ScheduleCollection`) adds multiple work
+    /// packets into different buckets, workers may open subsequent buckets while the coordinator
+    /// work still has packets to be added to prior buckets.  For this reason, we use the
+    /// `pending_coordinator_packets` to prevent the workers from opening any work buckets while
+    /// this coordinator work is being executed.
+    ///
+    /// # Arguments
+    ///
+    /// -   `work`: The work to execute.
+    /// -   `notify_workers`: Notify one worker after the work is finished. Useful for proceeding
+    ///     to the next work bucket stage.
+    fn initiate_coordinator_work(
+        &mut self,
+        work: &mut dyn CoordinatorWork<VM>,
+        notify_workers: bool,
+    ) {
+        self.scheduler
+            .pending_coordinator_packets
+            .fetch_add(1, Ordering::SeqCst);
+
+        self.execute_coordinator_work(work, notify_workers)
+    }
+
+    fn execute_coordinator_work(
+        &mut self,
+        work: &mut dyn CoordinatorWork<VM>,
+        notify_workers: bool,
+    ) {
+        work.do_work_with_stat(&mut self.coordinator_worker, self.mmtk);
+
+        self.scheduler
+            .pending_coordinator_packets
+            .fetch_sub(1, Ordering::SeqCst);
+
+        if notify_workers {
+            // When a coordinator work finishes, there is a chance that all GC workers parked, and
+            // no work packets are added to any open buckets.  We need to wake up one GC worker so
+            // that it can open more work buckets.
+            let _guard = self.scheduler.worker_monitor.0.lock().unwrap();
+            self.scheduler.worker_monitor.1.notify_one();
+        };
     }
 }
--- a/.github/workflows/post-review-ci.yml
+++ b/.github/workflows/post-review-ci.yml
@@ -34,6 +34,9 @@ jobs:
           repository: ${{ needs.binding-refs.outputs.v8_binding_repo }}
           path: mmtk-v8
           ref: ${{ needs.binding-refs.outputs.v8_binding_ref }}
+      - name: Use mmtk-core Rust toolchain for bindings
+        run: |
+          cp mmtk-core/rust-toolchain mmtk-v8/mmtk
       - name: Overwrite MMTk core in V8 binding
         run: |
           rm -rf mmtk-v8/repos/*
@@ -68,6 +71,9 @@ jobs:
           repository: ${{ needs.binding-refs.outputs.openjdk_binding_repo }}
           path: mmtk-openjdk
           ref: ${{ needs.binding-refs.outputs.openjdk_binding_ref }}
+      - name: Use mmtk-core Rust toolchain for bindings
+        run: |
+          cp mmtk-core/rust-toolchain mmtk-openjdk/mmtk
       - name: Setup
         run: |
           cd mmtk-openjdk
@@ -76,7 +82,8 @@ jobs:
           sed -i 's/^mmtk[[:space:]]=/#ci:mmtk=/g' mmtk/Cargo.toml
           sed -i 's/^#[[:space:]]mmtk/mmtk/g' mmtk/Cargo.toml
       - name: Overwrite MMTk core in openjdk binding
-        run: cp -r mmtk-core mmtk-openjdk/repos/
+        run: |
+          cp -r mmtk-core mmtk-openjdk/repos/
       - name: Test
         run: |
           cd mmtk-openjdk
@@ -99,6 +106,9 @@ jobs:
           repository: ${{ needs.binding-refs.outputs.jikesrvm_binding_repo }}
           path: mmtk-jikesrvm
           ref: ${{ needs.binding-refs.outputs.jikesrvm_binding_ref }}
+      - name: Use mmtk-core Rust toolchain for bindings
+        run: |
+          cp mmtk-core/rust-toolchain mmtk-jikesrvm/mmtk
       - name: Setup
         run: |
           cd mmtk-jikesrvm
@@ -131,6 +141,9 @@ jobs:
           repository: ${{ needs.binding-refs.outputs.julia_binding_repo }}
           path: mmtk-julia
           ref: ${{ needs.binding-refs.outputs.julia_binding_ref }}
+      - name: Use mmtk-core Rust toolchain for bindings
+        run: |
+          cp mmtk-core/rust-toolchain mmtk-julia/mmtk
       - name: Setup
         run: |
           cd mmtk-julia
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,58 @@
-0.16.0 (2022-12-006)
+0.17.0 (2023-02-17)
+===
+
+Plan
+---
+* Fix a bug where `SemiSpace::get_available_pages` returned unused pages as 'available pages'. It should return half of the unused pages as 'available'.
+* Fix a bug where generational plans may make full-heap GC decision before all the releasing work is done. Now plans have a `Plan::end_of_gc()` method
+  that is executed after all the GC work is done.
+* Fix a bug where generational plans tended to trigger full heap GCs when heap is full. Now we properly use Appel-style nursery and allow
+  the nursery to extend to the heap size.
+* Add a feature `immix_zero_on_release` as a debug feature for Immix to eagerly zero any reclaimed memory.
+* Fix a bug where the alloc bits for dead objects were still set after the objects were reclaimed. Now Immix eagerly clears the alloc bits
+  once the memory is reclaimed.
+* Fix a bug where Immix uses forwarding bits and forwarding pointers but did not declare it.
+
+
+Allocator
+---
+* Fix a bug about the maximum object size in `FreeListAllocator`. The allowed maximum object size depends on both the block size and the largest bin size.
+
+
+Scheduler
+---
+* Add a 'sentinel' mechanism to execute a specified 'sentinel' work packet when a bucket is drained, and prevent the GC from entering the next stage depending
+  on the sentinel packet. This makes it possible to expand the transitive closure multiple times. It replaces `GCWorkScheduler::closure_end()`.
+
+
+API
+---
+* Add a new set of APIs to support binding-specific weak reference processing:
+  * Add `Scanning::process_weak_refs()` and `Scanning::forward_weak_refs()`. They both supply an `ObjectTracerContext` argument for retaining and
+    updating weak references. `Scanning::process_weak_refs()` allows a boolean return value to indicate if the method should be called again
+    by MMTk after finishing transitive closures for the weak references, which can be used to implement ephemeron.
+  * Add `Collection::post_forwarding()` which is called by MMTk after all weak reference related work is done. A binding can use this call
+    for any post processing for weak references, such as enqueue cleared weak references to a queue (Java).
+  * These replace old methods for the same purpose, like `Collection::process_weak_refs()`, `Collection::vm_release()`, and `memory_manager::on_closure_end`.
+
+
+Misc
+---
+* Upgrade the Rust toolchain we use to 1.66.1 and upgrade MSRV to 1.61.0.
+* Improve the boot time significantly for MMTk:
+  * Space descriptor map is created as an zeroed vector.
+  * Mmapper coalesces the mmap requests for adjancent chunks, and reduces our `mmap` system calls at boot time.
+* Add `GCTrigger` to allow different heuristics to trigger a GC:
+  * Implement `FixedHeapSizeTrigger` that triggers the GC when the maximum heap size is reached (the current approach).
+  * Implement `MemBalancerTrigger` (https://dl.acm.org/doi/pdf/10.1145/3563323) as the dynamic heap resize heuristic.
+* Refactor `SFTMap` so it is dynamically created now.
+* Fix a bug where `Address::store` dropped the value after store.
+* Fix a bug about an incorrect pointer cast in unsafe code in `CommonFreeListPageResource`.
+* Remove all inline directives from our code base. We rely on Rust compiler and PGO for inlining decisions.
+* Remove `SynchronizedCounter`.
+
+
+0.16.0 (2022-12-06)
 ===
 
 Plan
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mmtk"
-version = "0.16.0"
+version = "0.17.0"
 authors = ["The MMTk Developers <>"]
 edition = "2021"
 license = "MIT OR Apache-2.0"
@@ -10,7 +10,7 @@ repository = "https://github.com/mmtk/mmtk-core"
 readme = "README.md"
 categories = ["memory-management"]
 keywords = ["gc", "garbage", "collection", "garbage-collection", "allocation"]
-rust-version = "1.57.0"
+rust-version = "1.61.0"
 build = "build.rs"
 
 [lib]
@@ -20,22 +20,22 @@ doctest = false
 
 [dependencies]
 # MMTk macros
-mmtk-macros = { version = "0.16.0", path = "macros/" }
+mmtk-macros = { version = "0.17.0", path = "macros/" }
 
 libc = "0.2"
-jemalloc-sys = { version = "0.3.2", features = ["disable_initial_exec_tls"], optional = true }
+jemalloc-sys = { version = "0.5.3", features = ["disable_initial_exec_tls"], optional = true }
 mimalloc-sys = { version = "0.1.6", optional = true }
 lazy_static = "1.1"
 log = { version = "0.4", features = ["max_level_trace", "release_max_level_off"] }
 crossbeam = "0.8.1"
 num_cpus = "1.8"
-enum-map = "=2.1"
+enum-map = "2.4.2"
 downcast-rs = "1.1.1"
-atomic-traits = "0.2.0"
+atomic-traits = "0.3.0"
 atomic = "0.5.1"
 num-traits = "0.2"
-spin = "0.5.2"
-env_logger = "0.8.2"
+spin = "0.9.5"
+env_logger = "0.10.0"
 pfm = { version = "0.1.0-beta.1", optional = true }
 atomic_refcell = "0.1.7"
 strum = "0.24"
@@ -46,11 +46,11 @@ sys-info = "0.9"
 regex = "1.7.0"
 
 [dev-dependencies]
-rand = "0.7.3"
+rand = "0.8.5"
 paste = "1.0.8"
 
 [build-dependencies]
-built = { version = "0.5.1", features = ["git2"] }
+built = { version = "0.6.0", features = ["git2"] }
 
 [features]
 default = []
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ We maintain an up to date list of the prerequisite for building MMTk and its bin
 
 ## Build
 
-MMTk can build with a stable Rust toolchain. The minimal supported Rust version is 1.57.0, and MMTk is tested with 1.66.1.
+MMTk can build with a stable Rust toolchain. The minimal supported Rust version is 1.61.0, and MMTk is tested with 1.66.1.
 
 ```console
 $ cargo build
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mmtk-macros"
 # the macro crate uses the same version as mmtk-core
-version = "0.16.0"
+version = "0.17.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "MMTk macros provides procedural macros used by mmtk-core."


src/plan/immix/global.rs
```diff
@@ -81,21 +82,25 @@ impl<VM: VMBinding> Plan for Immix<VM> {
...
-        );
-
-        // The blocks are not identical, clippy is wrong. Probably it does not recognize the constant type parameter.
-        #[allow(clippy::if_same_then_else)]
-        if in_defrag {
-            scheduler.schedule_common_work::<ImmixGCWorkContext<VM, TRACE_KIND_DEFRAG>>(self);
-        } else {
-            scheduler.schedule_common_work::<ImmixGCWorkContext<VM, TRACE_KIND_FAST>>(self);
-        }
+        // let in_defrag = self.immix_space.decide_whether_to_defrag(
```
caizixian, line 85: We probably want to remove these commented out sections?

qinsoon: Yes, I will. This is just a draft. If you are curious, you are welcome to look at the changes and leave any comments. But do not spend much time on a detailed review yet.

qinsoon:
In terms of performance, I have one run [here (internal url)](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|ferret-2023-02-06-Mon-235911&benchmark^buildstring^invocation^iteration&GC^time^time.other^time.stw&|10&iteration^1^2|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ix.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.no_biased_locking.no_class_unloading.dacapochopin-0d047f55|41&Histogram%20(with%20CI)^buildstring^benchmark&).
* Basically the sticky immix plan has the same generational behavior as generational immix (they are both faster or slower in a benchmark). 
* Sticky immix is faster than generational immix generally in the measurement, which is strange. It suggests there are some issues with our generational immix implementation -- I do not have time to look into that right now.
* `eclipse` is the only benchmark that sticky immix is slower than generational immix. The issue is that sticky immix cannot properly implement the address check for nursery objects needed the memory slice copy barrier (see the discussion here: https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/sticky-immix-array-copy-barrier/near/327476810)
* `luindex` and `pmd` fail for both generational and sticky immix in both `master` and this PR. I haven't figured out the issue.

I will do a new run with PGO and with new Rust version soon, and post the results here. However, I would like to start the review process right now while waiting for the new performance results.

src/plan/immix/global.rs
```diff
@@ -160,4 +162,37 @@ impl<VM: VMBinding> Immix<VM> {
...
+        );
+
+        if in_defrag {
+            scheduler.schedule_common_work::<DefragContext>(plan);
+        } else {
+            // The type of plan is `DefragContext::PlanType`, and we need it as `FastContext::PlanType`.
+            // They should be the same plan. But I don't find a way to tell Rust compiler that
+            // those PlanTypes are the same. So just do a unsafe transmute here.
+            let fast_plan = unsafe { std::mem::transmute(plan) };
+            scheduler.schedule_common_work::<FastContext>(fast_plan);
```
wks, line 191: We can tell Rust that both `FastContext` and `DefragContext` has the same `PlanType` by adding an additional type parameter.  Here is a patch:

[same-type.patch.gz](https://github.com/mmtk/mmtk-core/files/10729156/same-type.patch.gz)

qinsoon: Thanks. This is very helpful. I will make this change.

src/policy/immix/immixspace.rs
```diff
@@ -511,24 +566,36 @@ impl<VM: VMBinding> ImmixSpace<VM> {
...
+            {
                 self.attempt_mark(object, self.mark_state);
                 ForwardingWord::clear_forwarding_bits::<VM>(object);
                 Block::containing::<VM>(object).set_state(BlockState::Marked);
                 object
             } else {
-                ForwardingWord::forward_object::<VM>(object, semantics, copy_context)
+                let new_object =
+                    ForwardingWord::forward_object::<VM>(object, semantics, copy_context);
+                Block::containing::<VM>(new_object).set_state(BlockState::Marked);
```
wks, line 579: Both the `if` and the `else` branches set the marked state.  We may merge them and move them after the `if` statement.

src/policy/immix/immixspace.rs
```diff
@@ -646,14 +721,25 @@ pub struct PrepareBlockState<VM: VMBinding> {
...
+            side.bzero_metadata(self.chunk.start(), Chunk::BYTES);
+        }
+        if self.space.space_args.reset_log_bit_in_major_gc {
+            if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC {
+                // We zero all the log bits in major GC, and for every object we trace, we will mark the log bit again.
+                side.bzero_metadata(self.chunk.start(), Chunk::BYTES);
+            } else {
+                // If the log bit is not in side metadata, we cannot bulk zero. We can either
+                // clear the bit for dead objects in major GC, or clear the log bit for new
+                // objects. In both cases, we do not need to set log bit at tracing.
```
wks, line 737: In English, it is more natural to use "in either case" if the main part is negative, i.e. "we do *not*".

```suggestion
                // objects. In either case, we do not need to set log bit at tracing.
```

See: https://forum.wordreference.com/threads/in-either-case-vs-in-both-case.2243816/

src/policy/largeobjectspace.rs
```diff
@@ -196,9 +196,15 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
             object
         );
         let nursery_object = self.is_in_nursery(object);
+        trace!(
+            "LOS object {} {} a nursery object",
+            object,
+            if nursery_object { "is" } else { "is not " }
```
wks, line 202:
```suggestion
            if nursery_object { "is" } else { "is not" }
```

src/scheduler/controller.rs
```diff
@@ -130,6 +126,8 @@ impl<VM: VMBinding> GCController<VM> {
         }
         self.scheduler.deactivate_all();
 
+        let gc_elapsed = std::time::Instant::now() - gc_start;
```
wks, line 129:
```suggestion
        let gc_elapsed = gc_start.elapsed();
```

src/scheduler/gc_work.rs
```diff
@@ -219,11 +219,16 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for StopMutators<E> {
 impl<E: ProcessEdgesWork> CoordinatorWork<E::VM> for StopMutators<E> {}
 
 #[derive(Default)]
-pub struct EndOfGC;
+pub struct EndOfGC(pub std::time::Duration);
```
wks, line 222: I recommend giving the field a name.  It is not very obvious what that duration means when looking at this work packet alone.

src/policy/immix/immixspace.rs
qinsoon: I agree. There is an assertion below to check if the block is marked. But the original code does not mark the block if we forward the object. I don't know how it works. We would better check with @wenyuzhao.

src/scheduler/gc_work.rs
wks: > You can't give field names inside of a tuple struct in Rust. Have to make it a proper struct.

Yes. That's what I mean.  Make it a struct with named fields.

qinsoon:
I made changes as you suggested @wks, except the module-level comments and `set_state`.

wks (approved):
LGTM.

qinsoon:
> In terms of performance, I have one run [here (internal url)](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0%7Cferret-2023-02-06-Mon-235911&benchmark%5Ebuildstring%5Einvocation%5Eiteration&GC%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E2%7C20&1%5Einvocation%7C30&1&benchmark&buildstring;jdk-mmtk-master.ix.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.no_biased_locking.no_class_unloading.dacapochopin-0d047f55%7C41&Histogram%20(with%20CI)%5Ebuildstring%5Ebenchmark&).
> 
> * Basically the sticky immix plan has the same generational behavior as generational immix (they are both faster or slower in a benchmark).
> * Sticky immix is faster than generational immix generally in the measurement, which is strange. It suggests there are some issues with our generational immix implementation -- I do not have time to look into that right now.
> * `eclipse` is the only benchmark that sticky immix is slower than generational immix. The issue is that sticky immix cannot properly implement the address check for nursery objects needed the memory slice copy barrier (see the discussion here: https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/sticky-immix-array-copy-barrier/near/327476810)
> * `luindex` and `pmd` fail for both generational and sticky immix in both `master` and this PR. I haven't figured out the issue.
> 
> I will do a new run with PGO and with new Rust version soon, and post the results here. However, I would like to start the review process right now while waiting for the new performance results.

This is the performance run (with Rust 1.66.1 and PGO): [internal link](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|ferret-2023-02-14-Tue-050817&benchmark^build^invocation^iteration&GC^time^time.other^time.stw&|10&iteration^1^2|20&1^invocation|50&GC^1^1^+inf|30&1&benchmark&build;jdk-mmtk-master-ix|41&Histogram%20(with%20CI)^build^benchmark&). It is similar to my previous measurement.

Total time

stickyix-time

GC time

stickyix-stw-time

src/mmtk.rs
```diff
@@ -142,7 +142,10 @@ impl<VM: VMBinding> MMTK<VM> {
     }
 
     pub fn harness_begin(&self, tls: VMMutatorThread) {
-        // FIXME Do a full heap GC if we have generational GC
+        // For generational plans, force a full heap collection.
+        if let Some(gen) = self.plan.generational() {
+            gen.force_full_heap_collection();
```
wenyuzhao, line 147: What about passing another argument to the `handle_user_collection_request` call below to force a full GC?

src/plan/global.rs
```diff
@@ -678,7 +693,9 @@ impl<VM: VMBinding> BasePlan<VM> {
             .store(emergency_collection, Ordering::Relaxed);
 
         if emergency_collection {
-            plan.force_full_heap_collection();
+            if let Some(gen) = plan.generational() {
+                gen.force_full_heap_collection();
```
wenyuzhao, line 697: it's possible to call `plan.force_full_heap_collection();` all the time. For non-generational full-heap collectors this is just a no-op.

src/plan/generational/immix/global.rs
```diff
@@ -225,11 +236,13 @@ impl<VM: VMBinding> GenImmix<VM> {
...
-        let immix_space = ImmixSpace::new(plan_args.get_space_args(
-            "immix_mature",
-            true,
-            VMRequest::discontiguous(),
-        ));
+        let immix_space = ImmixSpace::new(
+            plan_args.get_space_args("immix_mature", true, VMRequest::discontiguous()),
+            ImmixSpaceArgs {
+                reset_log_bit_in_major_gc: false,
+                unlog_object_when_traced: false,
```
wenyuzhao, line 243: shouldn't this be true?

src/plan/generational/global.rs
```diff
@@ -298,21 +300,53 @@ impl<VM: VMBinding> CommonGenPlan<VM> {
...
+/// to be object safe.
 pub trait GenerationalPlan: Plan {
-    /// Return the common generational implementation [`crate::plan::generational::global::CommonGenPlan`].
-    fn common_gen(&self) -> &CommonGenPlan<Self::VM>;
+    /// Is the current GC a nursery GC? If a GC is not a nursery GC, it will be a full heap GC.
+    /// This should only be called during GC.
+    fn is_current_gc_nursery(&self) -> bool;
+
+    /// Is the object in the nursery?
+    fn is_object_in_nursery(&self, object: ObjectReference) -> bool;
```
wenyuzhao, line 311: This function only works based on the assumption that we only have two generations in the heap (i.e. young-gen and old-gen). If the number of generations is larger than 2, shall we do something different to handle this?

As an example, G1 has three generations: eden, survived, and old.

src/policy/immix/immixspace.rs
wenyuzhao: Blocks are marked when they're allocated from the page-resource or poped out from the reusable blocks queue.  https://github.com/mmtk/mmtk-core/blob/master/src/policy/immix/block.rs#L158-L162 . The store of the mark byte only happens once per block. The `set_state` here is unnecessary and also less efficient. We should remove this.

src/policy/immix/immixspace.rs
```diff
@@ -511,24 +566,36 @@ impl<VM: VMBinding> ImmixSpace<VM> {
...
+                Block::containing::<VM>(new_object).set_state(BlockState::Marked);
+                new_object
             };
             debug_assert_eq!(
                 Block::containing::<VM>(new_object).get_state(),
                 BlockState::Marked
             );
             queue.enqueue(new_object);
             debug_assert!(new_object.is_live());
+            self.unlog_object_if_needed(new_object);
```
wenyuzhao, line 588: A better approach is to bulk zero the unlog bits when a clean block is allocated, or when reusing a reusable hole, instead of unlog each object here. Unlog every individual object may not be faster than bulk zeroing.

src/policy/immix/immixspace.rs
```diff
@@ -646,14 +721,25 @@ pub struct PrepareBlockState<VM: VMBinding> {
...
         // NOTE: We reset the mark bits because cyclic mark bit is currently not supported, yet.
         // See `ImmixSpace::prepare`.
         if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::LOCAL_MARK_BIT_SPEC {
-            side.bzero_metadata(chunk.start(), Chunk::BYTES);
+            side.bzero_metadata(self.chunk.start(), Chunk::BYTES);
+        }
+        if self.space.space_args.reset_log_bit_in_major_gc {
+            if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC {
+                // We zero all the log bits in major GC, and for every object we trace, we will mark the log bit again.
+                side.bzero_metadata(self.chunk.start(), Chunk::BYTES);
```
wenyuzhao, line 733: When do you need to bulk zero all the unlog bits in the heap? This sounds expensive.

src/mmtk.rs
qinsoon: That sounds good. But we can only add the argument in our internal call, rather than expose the idea of full heap GC to the user.

src/plan/global.rs
qinsoon: That is what the current code does, and `force_full_heap_collection()` is a method of `trait Plan`. I moved the method to `trait GenerationalPlan`. I think it is cleaner, as not all plans are generational, and the idea of full heap does not apply to them.

src/plan/generational/immix/global.rs
qinsoon: There is a slight difference.

In generational immix, we only need to unlog objects in GC when we copy an object to the mature/immix space (no matter from nursery or from the mature space). When we copy from nursery to mature, we do not invoke `ImmixSpace::trace_object`. So setting `unlog_object_when_traced` to `true` will not affect the log bit for those promoted object (as `ImmixSpace::trace_object` is not called on them). We use `CopySemantics::PromoteToMature` or `CopySemantics::Mature` to set the log bit in the `post_copy`.

In sticky immix, any object we trace in the immix space (no matter copied or not) is considered promoted, and we need to unlog it. For example, when we trace a nursery object, even if we do not copy it, it is 'promoted' to a mature object, and we need to unlog it. In this case, using `post_copy` won't capture all the cases here. So I added this option, and for sticky immix, we can just unlog every object when we trace it.

qinsoon: I can add some comments to make it clear why we do not need it to be `true`.

src/plan/generational/global.rs
qinsoon: If we need to support multiple generations, we will have to do some refactoring to allow it. For example, `gen_plan.is_object_in_nursery(object, Generationa::Eden)`. Alternatively, we may use a new trait, like `MultiGenerationalPlan`. I don't know what would be a better way.

The changes in this PR is to allow us to reuse the generational code for sticky plans. I think if we want to support multiple generations, we would also want to reuse the current generational code as much as we can.

src/policy/immix/immixspace.rs
qinsoon: Thanks @wenyuzhao  for pointing to the code. The assertion failure actually failed for sticky immix. I think the following is the problem.

We have two immix allocators in the worker, one used for defrag (`copy` is `true`), and the other one for normal copy (`copy` is `false`)
https://github.com/mmtk/mmtk-core/blob/106fc71472e0d960e78da977ef2d43d54bbc5113/src/policy/immix/immixspace.rs#L803-L806

For a defrag immix GC, we use the defrag allocator, and the defrag allocator will initiate the block as marked.
https://github.com/mmtk/mmtk-core/blob/106fc71472e0d960e78da977ef2d43d54bbc5113/src/policy/immix/immixspace.rs#L776-L780
https://github.com/mmtk/mmtk-core/blob/106fc71472e0d960e78da977ef2d43d54bbc5113/src/policy/immix/block.rs#L158-L162

For a fast immix GC, we use the normal allocator which does not initiate the blocks as marked, but we use `trace_object_without_moving()` which marks blocks when we trace objects.
https://github.com/mmtk/mmtk-core/blob/106fc71472e0d960e78da977ef2d43d54bbc5113/src/policy/immix/immixspace.rs#L455

So in short, for fast GC, we mark each block at tracing, and for defrag GC, we initiate each block as marked. It is an optimization so we do not need to mark objects if they are copied in the GC.

For sticky immix, this no longer works. A nursery GC in sticky immix is not a defrag GC (so blocks are not initiated as marked), but it may move objects (so it uses `trace_object_with_opportunistic_copying()` which does not mark blocks if an object is copied). So there are some blocks that are not marked if we do not add the line, and we will see the assertion failure.

qinsoon: I will think about what to do for this.

src/policy/immix/immixspace.rs
qinsoon: Can you elaborate on this a bit more? I am confused. Bulk zeroing is setting bits to 0, but here we set the bit to 1 (making the object a mature object). At the end of a nursery GC, all the mature objects should have bits as 1.

Also bear in mind that in sticky immix, an object becomes mature as long as we trace it in a nursery GC (we do not have to copy it to make it a mature object). So usually we unlog an object when we copy it, but in sticky immix, we need to unlog an object when we trace it.

src/policy/immix/immixspace.rs
qinsoon: `PrepareBlockState` is only scheduled for full heap GCs. And this bulk zeroing is only for sticky immix, it does not affect other immix variants.

src/mmtk.rs
qinsoon: Done

src/plan/generational/immix/global.rs
qinsoon: I added some comments to explain this.

src/policy/immix/immixspace.rs
qinsoon: I checked the Java MMTk. And it seems we do not need two different immix copying allocators for most immix plans (only generational immix needs two copying allocators). So I renamed the current `ImmixCopyContext` (with two copying allocators) to `ImmixHybridCopyContext`, and this is used by generational immix. I also added a `ImmixCopyContext` (with one copying allocator), and this is used by immix and sticky immix. In this case, we no longer need the extra block marking in `trace_object`. I will measure performance again and post here.

k-sareen: That's not quite right. StickyImmix uses `nurseryCopy` as well as `copy`. Note that it extends `ImmixCollector` so it already has the `ImmixAllocator copy` field.

qinsoon: I didn't find where `nurseryCopy` is used. Is it used anywhere? It inherits `ImmixCollector`, but does not override `allocCopy`.

k-sareen: Huh -- very odd... You're right. It's initialized but not used. `nurseryCopy.reset()` is called but nothing is ever alloced in it. I'll check git logs to see if it was ever used

k-sareen: It was never used from the initial commit which added Immix (https://github.com/mmtk/jikesrvm/commit/f10ba059ebdb801cc66e2cb6552a6f4a3a0f193d). How odd.

qinsoon:
@wenyuzhao I have addressed your comments. I am running performance measurement for the changes in https://github.com/mmtk/mmtk-core/pull/741#discussion_r1112381496. I am not doing other work for this PR right now. If there is any other requested change, just let me know. Otherwise this PR will be merged if the new performance results are similar to the previous one (https://github.com/mmtk/mmtk-core/pull/741#issuecomment-1432284832) after I get them.

src/policy/immix/immixspace.rs
qinsoon: > I checked the Java MMTk. And it seems we do not need two different immix copying allocators for most immix plans (only generational immix needs two copying allocators). So I renamed the current `ImmixCopyContext` (with two copying allocators) to `ImmixHybridCopyContext`, and this is used by generational immix. I also added a `ImmixCopyContext` (with one copying allocator), and this is used by immix and sticky immix. In this case, we no longer need the extra block marking in `trace_object`. I will measure performance again and post here.

The results are [here](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|ferret-2023-02-21-Tue-004454&benchmark^build^invocation^iteration&GC^time^time.other^time.stw&|10&iteration^1^2|20&1^invocation|50&GC^1^1^+inf|30&1&benchmark&build;jdk-mmtk-master-ix|41&Histogram%20(with%20CI)^build^benchmark&). I am still running it. For what I have now, the performance looks similar to my previous run (https://github.com/mmtk/mmtk-core/pull/741#issuecomment-1432284832).

src/policy/immix/immixspace.rs
wenyuzhao: >  Bulk zeroing is setting bits to 0, but here we set the bit to 1 (making the object a mature object)

yes I mean bulk-memset the unlog table to 1

> So usually we unlog an object when we copy it, but in sticky immix, we need to unlog an object when we trace it.

For object left in place without moving, yes you have to unlog them without bulk-memset.

But bulk-memset can be applied to to-space. For this case, the object is already copied. The to-space memory should be bulk initialized to 1 in allocation slow-path.

wenyuzhao: Say if you're copying 1k promoted objects to a 32 KB clean block, bulk initializing the unlog table once is definitely faster then setting the unlog table 1k times for each copied object.

src/policy/immix/immixspace.rs
wenyuzhao: Sorry I mean for sticky immix, why all the unlog table need to be reset before a full-heap GC?

wenyuzhao:
> @wenyuzhao I have addressed your comments. I am running performance measurement for the changes in [#741 (comment)](https://github.com/mmtk/mmtk-core/pull/741#discussion_r1112381496). I am not doing other work for this PR right now. If there is any other requested change, just let me know. Otherwise this PR will be merged if the new performance results are similar to the previous one ([#741 (comment)](https://github.com/mmtk/mmtk-core/pull/741#issuecomment-1432284832)) after I get them.

All other parts looks good to me. I'm only concerning the performance of the unlog table related code. Mainly two parts: (1) setting unlog bits when trace objects (2) bulking zeroing all the unlog table before a full GC. These do not affect correctness, but are performance critical.

src/policy/immix/immixspace.rs
qinsoon: The thing is that we may not copy an object at all in sticky immix. The object is allocated with `log bit === 0`, but when we trace it, we need to set its log bit to 1. However, I can try this: for objects that are copied, bulk set their log bit, for objects that are not copied, set the log bit per object.

src/policy/immix/immixspace.rs
qinsoon: We need a way to set the log bit back to `0` for dead objects after a full heap GC. We can either clear the log bit when we allocate new objects in mutator phase, or just bulk zero the log bit and set the bit for the live objects. As I already need to set the bit for live objects (see our discussion in the other comment), I think it is natural to just bulk zero the log bit.

qinsoon:
> > @wenyuzhao I have addressed your comments. I am running performance measurement for the changes in [#741 (comment)](https://github.com/mmtk/mmtk-core/pull/741#discussion_r1112381496). I am not doing other work for this PR right now. If there is any other requested change, just let me know. Otherwise this PR will be merged if the new performance results are similar to the previous one ([#741 (comment)](https://github.com/mmtk/mmtk-core/pull/741#issuecomment-1432284832)) after I get them.
> 
> All other parts looks good to me. I'm only concerning the performance of the unlog table related code. Mainly two parts: (1) setting unlog bits when trace objects (2) bulking zeroing all the unlog table before a full GC. These do not affect correctness, but are performance critical.

I can try this implementation and see how it performs:
* bulk set log bit to 1 in copying
* set log bit to 1 per object if an object is not copied
* bulk set log bit to 0 in allocation

Is this what you suggest?

qinsoon:
> > > @wenyuzhao I have addressed your comments. I am running performance measurement for the changes in [#741 (comment)](https://github.com/mmtk/mmtk-core/pull/741#discussion_r1112381496). I am not doing other work for this PR right now. If there is any other requested change, just let me know. Otherwise this PR will be merged if the new performance results are similar to the previous one ([#741 (comment)](https://github.com/mmtk/mmtk-core/pull/741#issuecomment-1432284832)) after I get them.
> > 
> > 
> > All other parts looks good to me. I'm only concerning the performance of the unlog table related code. Mainly two parts: (1) setting unlog bits when trace objects (2) bulking zeroing all the unlog table before a full GC. These do not affect correctness, but are performance critical.
> 
> I can try this implementation and see how it performs:
> 
> * bulk set log bit to 1 in copying
> * set log bit to 1 per object if an object is not copied
> * bulk set log bit to 0 in allocation
> 
> Is this what you suggest?

We discussed this in today's meeting:
* Bulk setting log bits to 1 may not be necessary for object remembering barriers. But it could be an important optimisation for field remembering barriers.
* For object remembering barriers, we could simply set the byte to `0xff` rather than masking and setting bits. It will set adjacent bits to 1, but it is still correctness and it improves performance.
* With the discussion above, we will need to unlog every object we trace in a nursery GC, and we need to unlog every 'nursery' object in a mature GC as well. To make it simple, we can just zero all the log bits in mature GC, and rely on the same code path in object tracing to unlog every object in a mature GC. An alternative approach is to bulk zero log bit in allocation (lazily) or in releasing (eagerly) (but we still need to unlog nursery objects in tracing). This could be more efficient.

What I will do:
* Set byte to unlog objects
* Try bulk zeroing log bit for dead objects after major GC (instead of bulk zeroing log bits for all blocks in major GC).

qinsoon:
This is the PR for setting byte for log bit: https://github.com/mmtk/mmtk-core/pull/764. I measured with GenImmix, and it seems no measurable difference in GC time.

I don't think it works with sticky immix, as in sticky immix, nursery objects can be adjacent to mature objects. If we simply set byte for log bit in a GC, the new allocated nursery objects may have its log bit as `1`.

wenyuzhao:
> Also I don't think it works with sticky immix, as in sticky immix, nursery objects can be adjacent to mature objects. If we simply set byte for log bit in a GC, the new allocated nursery objects may have its log bit as 1.

I don't think this is true. Nursery and mature objects cannot sit in the same line. A line is 256 B, which maps to 32 bits in the unlog table (if the min object size is 8 bytes). Relax-storing a byte (or even an `i32`) should be safe.

qinsoon:
> > Also I don't think it works with sticky immix, as in sticky immix, nursery objects can be adjacent to mature objects. If we simply set byte for log bit in a GC, the new allocated nursery objects may have its log bit as 1.
> 
> I don't think this is true. Nursery and mature objects cannot sit in the same line. A line is 256 B, which maps to 32 bits in the unlog table (if the min object size is 8 bytes). Relax-storing a byte (or even an `i32`) should be safe.

I can try and see if it works/improves performance for sticky immix. From my previous runs, this change does not improve performance for gen immix.

qinsoon:
> What I will do:
> 
> * Set byte to unlog objects

There are some links for the results. It is still running. 
* Sticky immix: [internal link](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|fisher-2023-02-24-Fri-022521&benchmark^buildstring^invocation^iteration&time^time.other^time.stw&|10&iteration^1^2|10&buildstring^2^jdk-mmtk-log-bit-set-byte.genix.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.no_biased_locking.no_class_unloading.dacapochopin-0d047f55|10&buildstring^2^jdk-mmtk-log-bit-set-bit.genix.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.no_biased_locking.no_class_unloading.dacapochopin-0d047f55|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-log-bit-set-bit.stickyix.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.no_biased_locking.no_class_unloading.dacapochopin-0d047f55|41&Histogram%20(with%20CI)^buildstring^benchmark&)
* Gen immix: [internal link](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|fisher-2023-02-24-Fri-022521&benchmark^buildstring^invocation^iteration&time^time.other^time.stw&|10&iteration^1^2|10&buildstring^2^jdk-mmtk-log-bit-set-byte.stickyix.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.no_biased_locking.no_class_unloading.dacapochopin-0d047f55|10&buildstring^2^jdk-mmtk-log-bit-set-bit.stickyix.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.no_biased_locking.no_class_unloading.dacapochopin-0d047f55|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-log-bit-set-bit.genix.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.no_biased_locking.no_class_unloading.dacapochopin-0d047f55|41&Histogram%20(with%20CI)^buildstring^benchmark&)

> * Try bulk zeroing log bit for dead objects after major GC (instead of bulk zeroing log bits for all blocks in major GC).

This is the link for the results (still running): [internal link](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|ferret-2023-02-24-Fri-022105&benchmark^build^invocation^iteration&time^time.other^time.stw^work.PlanProcessEdges.time.total^work.PrepareBlockState.time.total^work.SweepChunk.time.total&|10&iteration^1^2|20&1^invocation)

I will update this post once I get results. If the results are positive, I will adopt those changes. Otherwise, I do not plan to spend more time on this PR. I don't think there is any major performance issue for this PR. If we do want to optimize further, we can do it in separate PRs.

qinsoon:
binding-refs
JULIA_BINDING_REF=add-immix-no-nursery-copy

PR #747, opened 2023-02-06
Minor fixes for mimalloc

micahsinclair:
This pull request includes two minor changes regarding the implementation of `mimalloc`:

- In `policy::marksweepspace::native_ms::block_list.rs` the maximum large object size now is adjusted to match the block/page size as defined in `Block::LOG_BYTES`. This ensures that when the block size is changed, `native_ms` will still be able to run. It is worth noting that the bin sizes are still hard-coded.
- In `util::malloc::library.rs` the page accounting for library mimalloc is now set to 4KiB, the size of one OS page. In the library mimalloc code, only 4KiB is touched at first, even though mimalloc blocks/pages are usually 64KiB (https://github.com/microsoft/mimalloc/blob/dd7348066fe40e8bf372fa4e9538910a5e24a75f/src/page.c#L586).


src/util/malloc/library.rs
```diff
@@ -36,8 +36,8 @@ mod jemalloc {
 
 #[cfg(feature = "malloc_mimalloc")]
 mod mimalloc {
-    // MiMalloc 64K Page
-    pub const LOG_BYTES_IN_MALLOC_PAGE: u8 = 16;
+    // Normal 4K page accounting
+    pub const LOG_BYTES_IN_MALLOC_PAGE: u8 = 12;
```
k-sareen, line 40: I think it would be better to match what other malloc libraries do:
```suggestion
    pub const LOG_BYTES_IN_MALLOC_PAGE: u8 = crate::util::constants::LOG_BYTES_IN_PAGE;
```

micahsinclair: Yes, that makes sense. Thanks!

src/policy/marksweepspace/native_ms/block_list.rs
```diff
@@ -167,7 +167,7 @@ pub(crate) const MI_BIN_FULL: usize = MAX_BIN + 1;
 pub(crate) const MAX_BIN: usize = 48;
 
 /// Largest object size allowed with our mimalloc implementation, in bytes
-pub(crate) const MI_LARGE_OBJ_SIZE_MAX: usize = MAX_BIN_SIZE;
+pub(crate) const MI_LARGE_OBJ_SIZE_MAX: usize = Block::BYTES;
```
qinsoon, line 170: `MAX_BIN_SIZE` is `8194 * MI_INTPTR_SIZE` (64K on 64bits, 32K on 32bits). So the max LO size is 64K on 64bits, and 32K on 32bits. After this change, the max LO size is always 64K. In that case, any object with its size between 32K and 64K is allocated with the free list allocator. But the allocator does not have bins for those sizes. That's why we see the out-of-bound error in the CI tests.

qinsoon: This could be a simple fix.
```suggestion
pub(crate) const MI_LARGE_OBJ_SIZE_MAX: usize = crate::util::rust_util::min_of_usize(Block::BYTES, MAX_BIN_SIZE);
```

qinsoon (approved):
LGTM

PR #748, opened 2023-02-06
Immix: Clear alloc bits during sweeping

wks:
No longer clear alloc bit when an object is being moved.  This fixes an assertion error when trace_object is called on the same object twice during defrag GC.

Clear alloc bits eagerly during sweeping instead of during allocation. This ensures an address has the alloc bit (i.e. valid-object bit) if and only if that address corresponds to a valid object reference during mutator execution.

### DIFF
--- a/src/policy/immix/block.rs
+++ b/src/policy/immix/block.rs
@@ -241,6 +241,9 @@ impl Block {
                         holes += 1;
                     }
 
+                    #[cfg(feature = "global_alloc_bit")]
+                    crate::util::alloc_bit::bzero_alloc_bit(line.start(), Line::BYTES);
+
                     #[cfg(feature = "immix_zero_on_release")]
                     crate::util::memory::zero(line.start(), Line::BYTES);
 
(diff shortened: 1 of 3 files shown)


qinsoon (approved):
LGTM

qinsoon:
As long as the OpenJDK tests pass, we can merge this. The V8 build is expected to fail. The JikesRVM fail is new to me -- I need to look into it, but it should not be related with this PR.

wks:
> As long as the OpenJDK tests pass, we can merge this. The V8 build is expected to fail. The JikesRVM fail is new to me -- I need to look into it, but it should not be related with this PR.

I reproduced the same error when compiling JikesRVM on my ArchLinux (a rolling distribution).  The compilation will succeed on my LXC container with Ubuntu 22.04.  The error happens while configuring GNU Classpath.

wks:
The error is likely because the `javac` is not from Java 8.  The line is testing "if $JAVAC works".  It is testing by compiling a small Java program with javac.  But the testing script (`m4/ac_prog_javac_works.m4`) adds `-source 1.5 -target 1.5` to the command line, which later versions of `javac` don't support.

PR #749, opened 2023-02-07
Test julia binding

qinsoon:
This PR adds binding tests in our CI for Julia.


qinsoon:
binding-refs
JULIA_BINDING_REF=test-with-mmtk-core

.github/workflows/post-review-ci.yml
```diff
@@ -115,3 +115,42 @@ jobs:
...
+        run: |
+          cd mmtk-julia
+          ./.github/scripts/ci-checkout.sh
+          ./.github/scripts/ci-setup.sh
+          sed -i 's/^mmtk[[:space:]]=/#ci:mmtk=/g' mmtk/Cargo.toml
+          sed -i 's/^#[[:space:]]mmtk/mmtk/g' mmtk/Cargo.toml
+      - name: Overwrite MMTk core in Julia binding
+        run: |
+          mkdir -p mmtk-julia/repos/mmtk-core
+          cp -r mmtk-core/* mmtk-julia/repos/mmtk-core
```
wks, line 144: One alternative is to use the commands
```
cd mmtk-core
git worktree add ../mmtk-julia/repos/mmtk-core origin/master
```
to checkout another work tree.  Inside the work tree, the `git` command still considers it to be in a git repository, so our `build.rs` should still be able to encode a git SHA1 hash into the version string.

But isn't it easier to simply change `Cargo.toml` to point to `../../mmtk-core`?

wks (approved):
It looks good to me, as it is mostly identical to other bindings.

I have a comment about the use of `cp` to copy `mmtk-core` over.

.github/workflows/post-review-ci.yml
qinsoon: Yes. Those are all possible, and simpler. Initially, `mmtk-core` was a submodule of the binding, and we put it under `repos/mmtk-core`. But the submodule was removed, and we kept the path. In that case, we just manually copy `mmtk-core` to the path.

PR #750, opened 2023-02-07
Fix invalid pointer casting

wenyuzhao:
`fl` has type of `&&CommonFreeListPageResource`.

`fl as *const _ as *mut _` will first implicitly convert `fl` to  `*const &CommonFreeListPageResource`, then to `*mut CommonFreeListPageResource`. The second conversion from `*const` to `*mut` is incorrect.


caizixian (approved):
I’ve been debugging this with Wenyu. This looks good to me. Before this fix, the write is incorrect, and could sometimes be out of bound of the vector, corrupting the header of adjacent malloc’ed cells.

PR #752, opened 2023-02-07
Bump Rust version to v1.66.1

k-sareen:


k-sareen:
I've pretty much just directly applied `cargo clippy --fix`. The most interesting fix was in `src/policy/marksweepspace/native_ms/block.rs` where we convert a `usize` into a `*mut BlockList`.

src/scheduler/work_bucket.rs
```diff
@@ -43,6 +43,8 @@ impl<VM: VMBinding> BucketQueue<VM> {
     }
 }
 
+// Allow type_complexity as Rust support for trait aliases is still experimental
+#[allow(clippy::type_complexity)]
```
qinsoon, line 47: I think it is complaining about `can_open: Option<Box<dyn (Fn(&GCWorkScheduler<VM>) -> bool) + Send>>`, which is indeed too complex. We can create a type alias:
```rust
pub type BucketOpenCondiiton<VM> = Box<dyn (Fn(&GCWorkScheduler<VM>) -> bool) + Send>;
```
This should make the warning disappear.

k-sareen: I tried doing that but then Rust says "type aliases can't be used as traits"

k-sareen: Ah I see. I made a new trait called `trait BucketOpenCondiiton<VM: VMBinding> Fn(&GCWorkScheduler<VM>) -> bool) + Send` and Rust was complaining about that. I'll use your fix :+1:

k-sareen: In the future we may want to make this a trait alias instead of a type alias so that the `set_open_condition` function can also use the trait directly.

k-sareen:
@qinsoon Seems like the Rust v1.59.0 run is required but we've moved to Rust v1.66.1 in this PR. Don't see how to change the requirement anywhere.

qinsoon:
> @qinsoon Seems like the Rust v1.59.0 run is required but we've moved to Rust v1.66.1 in this PR. Don't see how to change the requirement anywhere.

It is set in branch protection rules in `Settings`. I just removed that requirement. Maybe we should omit the version from the test names so we have a fixed test name regardless of the Rust version we use, then add back the rule.

k-sareen:
I'm not an owner so I can't see the Settings. Yeah that makes sense to me.

wks:
This PR is bumping the version to 1.66.1, but the latest Rust version is 1.67.0 (released in 24 Jan 2023).  Is this intended?

k-sareen:
We have v1.66.1installed on our machines.

k-sareen:
I think the failing v8 and jikesrvm tests are expected, right? Can we merge this so we can merge the remove all inline PR after?

qinsoon:
> I think the failing v8 and jikesrvm tests are expected, right? Can we merge this so we can merge the remove all inline PR after?

Yeah.

qinsoon (approved):
LGTM

PR #753, opened 2023-02-07
Allow forwarding bits to be on the side for Immix

wks:
Now ImmixSpace can use on-the-side forwarding bits metadata.  When on the side, the forwarding bits are eagerly cleared when sweeping lines.

It is useful for adding Immix support to an existing VM where we can't squeeze any bits out of the existing object header for GC.

How to test: We can use mmtk-openjdk, and change `FORWARDING_BITS_METADATA_SPEC` to a side metadata.

```diff
 pub(crate) const FORWARDING_BITS_METADATA_SPEC: VMLocalForwardingBitsSpec =
-    VMLocalForwardingBitsSpec::in_header(FORWARDING_BITS_OFFSET);
+    VMLocalForwardingBitsSpec::side_after(MARKING_METADATA_SPEC.as_spec());
```

### DIFF
--- a/src/util/copy/mod.rs
+++ b/src/util/copy/mod.rs
@@ -100,8 +100,18 @@ impl<VM: VMBinding> GCWorkerCopyContext<VM> {
     /// * `bytes`: The size of the object in bytes.
     /// * `semantics`: The copy semantic used for the copying.
     pub fn post_copy(&mut self, object: ObjectReference, bytes: usize, semantics: CopySemantics) {
-        // Clear forwarding bits.
-        object_forwarding::clear_forwarding_bits::<VM>(object);
+        // Clear forwarding bits if the forwarding bits are in the header.
+        if !VM::VMObjectModel::LOCAL_FORWARDING_BITS_SPEC.is_on_side() {
+            object_forwarding::clear_forwarding_bits::<VM>(object);
+        } else {
+            // NOTE: If the forwarding bits were on the side, they would have been cleared when
+            // preparing the space.  See:
+            // - `CopySpace::prepare` and
+            // - `PrepareBlockState::reset_object_mark`.
+            debug_assert!(!object_forwarding::is_forwarded_or_being_forwarded::<VM>(
+                object
+            ));
+        }
         // If we are copying objects in mature space, we would need to mark the object as mature.
         if semantics.is_mature() && self.config.constraints.needs_log_bit {
             // If the plan uses unlogged bit, we set the unlogged bit (the object is unlogged/mature)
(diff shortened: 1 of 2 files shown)


src/policy/immix/block.rs
```diff
@@ -241,6 +242,11 @@ impl Block {
                         holes += 1;
                     }
 
+                    object_forwarding::bzero_forwarding_bits_if_on_side::<VM>(
```
qinsoon, line 245: Is this a bug fix, or just an optimization? See the comments in `post_copy`.

src/util/copy/mod.rs
```diff
@@ -100,8 +100,16 @@ impl<VM: VMBinding> GCWorkerCopyContext<VM> {
     /// * `bytes`: The size of the object in bytes.
     /// * `semantics`: The copy semantic used for the copying.
     pub fn post_copy(&mut self, object: ObjectReference, bytes: usize, semantics: CopySemantics) {
-        // Clear forwarding bits.
-        object_forwarding::clear_forwarding_bits::<VM>(object);
```
qinsoon, line 104: This is called for copying in both `CopySpace` and `ImmixSpace`. But the bulk zeroing is only implemented for `ImmixSpace` in this PR. With this change, we are not clearing side forwarding bits for `CopySpace`.

Also this happens when we copy an object (lazily clear forwarding bits). The change about bulk zeroing eagerly clears the forwarding bits in bulk. If it is meant to be an optimization, It is hard to say it is a clear win.

wks: I reverted the change of eagerly zeroing while sweeping.  I initially did it this way, similar to how I handled VO bits in the previous PR.  Then I realised there is a difference.  It is necessary to eagerly clear the VO bits for correctness, otherwise mutators may see the VO bit set for dead objects.  The forwarding bits are not observed by mutators, so it is not necessary to clear them during sweeping.

Now I clear the forwarding bits at the same time when clearing the mark bit.

qinsoon (approved):
LGTM

src/util/copy/mod.rs
wks: I also updated the comment to mention CopySpace, too, because both CopySpace and ImmixSpace clear the forwarding bits during preparation.

wks:
This PR is ready.  However, there is a known issue. If the VM actually declare the forwarding pointer to be on the side, the fragmented mapper will attempt to quarantine a memory region that is large enough to exceed the slab table size.  It will happen even if the heap size is small, such as 15MB, in which case the metadata should simply need another 15MB to hold the forwarding pointers.

I think the root cause is the way we handle memory mapping, not what we put in side metadata.  It should be fixed separately.

PR #754, opened 2023-02-07
Other changes related Immix for Ruby

wks:
These are small changes related to getting Ruby working with Immix.

**Implemented `get_forwarded_object` for ImmixSpace.**  `mmtk-core` didn't use the `get_forwarded_object` method.  In theory, we can let Ruby support copying GC without that method.  The `trace_object` method is sufficient to forward all fields.  However, it feels odd to have that method exposed as a method of `ObjectReference` but not implemented for Immix.  It is a nice debugging tool.  It is also useful during weak reference processing where the VM wants to forward weak fields, but not expand the transitive closure.

**Removed an assertion that the plan must be non-moving when scanning non-moving roots.**  Now that we already have object pinning for Immix, it is no longer relevant.

### DIFF
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -52,7 +52,6 @@ impl<VM: VMBinding> SFT for ImmixSpace<VM> {
         self.get_name()
     }
 
-    #[inline(always)]
     fn get_forwarded_object(&self, object: ObjectReference) -> Option<ObjectReference> {
         if !Block::containing::<VM>(object).is_defrag_source() {
             return None;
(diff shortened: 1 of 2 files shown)


src/policy/immix/immixspace.rs
```diff
@@ -51,6 +51,20 @@ impl<VM: VMBinding> SFT for ImmixSpace<VM> {
     fn name(&self) -> &str {
         self.get_name()
     }
+
+    #[inline(always)]
```
k-sareen, line 55: Probably remove the `inline(always)` since we've been investigating this :laughing:

qinsoon (approved):
LGTM

src/policy/immix/immixspace.rs
wks: You're right.  I'll remove that :)

Issue #734, opened 2023-01-10
PageAccounting::clear_reserved() called after reset() in GenImmix


wks:
GenImmix is very likely to crash. I can reproduce the crash using the following command:

```
RUST_BACKTRACE=1 MMTK_PLAN=GenImmix /path/to/openjdk/build/linux-x86_64-normal-server-slowdebug/jdk/bin/java -XX:+UseThirdPartyHeap -Xm{s,x}50M -jar dacapo-9.12-MR1-bach.jar lusearch -t 1
```

It may crash with both debug and release build alike.

What happened is, the counter `PageAccounting::reserved` underflowed (attempted to subtract from 0).

At the end of a GC, `PageAccounting::reset()` is called because `CopySpace::release()` calls `self.pr.reset()`.  `reserved` is cleared to 0.

Then GC finishes.  A mutator wakes up, and calls `pr.clear_request(pages_reserved)` with a non-zero `page_reserved`.  This will call `PageAccounting::clear_reserved(pages)` with a non-zero argument, and it will call `self.reserved.fetch_sub` when the current value of `reserved` is zero.

The `pr.clear_request` statement is in `src/policy/space.rs` line 72:

```rust
// src/policy/space.rs:66
        trace!("Polling ..");

        if should_poll && self.get_gc_trigger().poll(false, Some(self.as_space())) {
            debug!("Collection required");
            assert!(allow_gc, "GC is not allowed here: collection is not initialized (did you call initialize_collection()?).");
            VM::VMCollection::block_for_gc(VMMutatorThread(tls)); // We have checked that this is mutator
            pr.clear_request(pages_reserved); // clear the pages after GC. We need those reserved pages so we can compute new heap size properly.
            unsafe { Address::zero() }
        } else {
            debug!("Collection not required");
```

A stack trace when this happened is shown below.  Note that the "attempt to subtract with overflow" happens because I added a logging statement:

```diff
 // src/util/heap/accounting.rs
     pub fn clear_reserved(&self, pages: usize) {
         let result = self.reserved.fetch_sub(pages, Ordering::Relaxed); // `fetch_sub` never crashes when overflowing.
+        eprintln!("[clear_reserved] {} -> {}", result, result - pages);  // `result-pages` will overflow in debug mode.
     }
```

The stack trace:

```
thread '<unnamed>' panicked at 'attempt to subtract with overflow', /home/wks/projects/mmtk-github/mmtk-core/src/util/heap/accounting.rs:42:56
[...]

PR #755, opened 2023-02-09
Notify GC trigger about pending allocation

qinsoon:
This PR reverts a change about where to call `pr.clear_request()` in `Space.acquire()` in https://github.com/mmtk/mmtk-core/pull/712. The change caused issues like https://github.com/mmtk/mmtk-core/issues/734. This PR reverts the change. We now clear the reserved pages before GC (the same as before, and the same as Java MMTk), and additionally we inform the GC trigger about pending allocation pages. This closes https://github.com/mmtk/mmtk-core/issues/734.

### DIFF
--- a/src/util/heap/gc_trigger.rs
+++ b/src/util/heap/gc_trigger.rs
@@ -79,6 +79,11 @@ impl<VM: VMBinding> GCTrigger<VM> {
 /// GC start/end so they can collect some statistics about GC and allocation. The policy needs to
 /// decide the (current) heap limit and decide whether a GC should be performed.
 pub trait GCTriggerPolicy<VM: VMBinding>: Sync + Send {
+    /// Inform the triggering policy that we have pending allocation.
+    /// Any policy with dynamic heap size should take this into account when calculating a new heap size.
+    /// Failing to do so may result in unnecessay GCs, or result in an infinite loop if the new heap size
+    /// can never accomodate the pending allocation.
+    fn on_pending_allocation(&self, _pages: usize) {}
     /// Inform the triggering policy that a GC starts.
     fn on_gc_start(&self, _mmtk: &'static MMTK<VM>) {}
     /// Inform the triggering policy that a GC is about to start the release work. This is called
@@ -147,6 +152,9 @@ pub struct MemBalancerTrigger {
     max_heap_pages: usize,
     /// The current heap size
     current_heap_pages: AtomicUsize,
+    /// The number of pending allocaiton pages. The allocation requests for them have failed, and a GC is triggered.
+    /// We will need to take them into consideration so that the new heap size can accomodate those allocations.
+    pending_pages: AtomicUsize,
     /// Statistics
     stats: AtomicRefCell<MemBalancerStats>,
 }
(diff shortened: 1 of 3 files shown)


src/util/heap/gc_trigger.rs
```diff
@@ -147,6 +152,9 @@ pub struct MemBalancerTrigger {
     max_heap_pages: usize,
     /// The current heap size
     current_heap_pages: AtomicUsize,
+    /// The number of pending allocaiton pages. The allocation requests for them have failed, and a GC is triggered.
```
wks, line 155:
```suggestion
    /// The number of pending allocation pages. The allocation requests for them have failed, and a GC is triggered.
```

src/util/heap/gc_trigger.rs
```diff
@@ -467,6 +480,10 @@ impl MemBalancerTrigger {
             (live as f64 * 4096f64).sqrt()
         };
 
+        // Take pending allocations into consideration
+        extra_reserve += self.pending_pages.load(Ordering::SeqCst);
```
wks, line 484: I think it is clearer to simply create a new variable for pending pages,

```suggestion
        let pending_pages = self.pending_pages.load(Ordering::SeqCst);
```
... and then add it to `optimal_heap`.

```diff
-let optimal_heap = live + e as usize + extra_reserve;
+let optimal_heap = live + e as usize + extra_reserve + pending_pages;
```

And we don't need to make the `extra_reserve` parameter mutable.

src/util/heap/gc_trigger.rs
```diff
@@ -79,6 +79,11 @@ impl<VM: VMBinding> GCTrigger<VM> {
 /// GC start/end so they can collect some statistics about GC and allocation. The policy needs to
 /// decide the (current) heap limit and decide whether a GC should be performed.
 pub trait GCTriggerPolicy<VM: VMBinding>: Sync + Send {
+    /// Inform the triggering policy that we have pending allocation.
+    /// Any policy with dynamic heap size should take this into account when calculating a new heap size.
```
wks, line 83: We use the word "policy" for both spaces (CopySpace, ImmixSpace, etc.) and GC trigger (GCTriggerPolicy).  But because this method is called by Space, I thought the "policy" here refers to spaces when I first read this sentence.

```suggestion
    /// Any GC trigger policy with dynamic heap size should take this into account when calculating a new heap size.
```

wks (changes requested):
Mostly good except some minor problems.

qinsoon:
I have made the changes as suggested. @wks

wks (approved):
LGTM

qinsoon:
I made another change: the pending pages is always cleared at the end of GC (in every GC, nursery and mature), rather than after computing the new heap size (only in mature GCs). It should be correct. I will merge this PR tomorrow. If there is any issue, just let me know. @wks

wks:
I think the new approach (clearing pending pages at the end of every GC) is better.  Triggering nursery GC doesn't mean the heap size cannot accommodate the requested object.  Moreover, if nursery GC is triggered 10 times before a full-heap is triggered, the old approach may think mutators are requesting 10 times as much memory as they actually requested the last time.

Issue #693, opened 2022-11-08
Update rust-toolchain


wks:
Currently, as of 8 Nov 2022, the file `mmtk-core/rust-toolchain` locks the Rust toolchain version to 1.59.0, but the latest stable Rust toolchain is 1.65.0.  This means the toolchain we use is now 9 months behind the latest stable.

Locking the toolchain version has certain advantages, such as

-   If it is tested on the CI, it probably works on other machines because they use the same compiler.
-   The performance comparison between mmtk-core revisions is consistent because they use the same compiler.

However, the problems include

-   Some programs may break on newer toolchain versions, because
    -   The language semantics or standard library may change, deprecating old features, or
    -   Newer compilers may have different behaviour for programs with undefined behaviours.
        -   (See https://github.com/mmtk/mmtk-core/issues/691)
-   Some third-party library mmtk-core depends on require newer language or stdlib features.
    -   (See [this line](https://github.com/mmtk/mmtk-core/blob/83fa8ec95a220be22fb38d5b6b3b3030a9f47f0c/Cargo.toml#L33) and [this PR comment](https://github.com/mmtk/mmtk-core/pull/507#discussion_r879081998))
-   Newer `cargo clippy` can provide warnings of more smelly code than before.

Before we bump the version, we need to:

-   [x] Fix bugs that break with the new toolchain
    -   [x] https://github.com/mmtk/mmtk-core/issues/691
-   [ ] Update dependency versions.
    -   [ ] `enum-map = "=2.1"` (newest version is 2.4.1)
-   [x] Fix (or suppress when necessary) new `clippy` warnings.
-   [x] (ANU-only) Ensure development and testing machines are equipped with the new toolchain.

PR #756, opened 2023-02-10
Update dependencies versions

wks:
Some dependencies have new versions.  It is a good chance to bump the dependency versions, too, after we bump the Rust toolchain versions.

One notable dependency is `enum-map`.  Its latest version is 2.4.2, but we locked its version to 2.1.0 because it required a newer Rust toolchain.  Now we can depend on its latest version, instead.  We also changed the `rust-version` property in `Cargo.toml` to `1.61.0` because `enum-map-derive` depends on that, and `1.61.0` is not new compared to the `1.66.1` we use.  See: https://github.com/mmtk/mmtk-core/pull/507#discussion_r879081998

This closes https://github.com/mmtk/mmtk-core/issues/693


qinsoon:
LGTM. This PR upgrades our minimal supported Rust version. I think we should update our bindings first, then merge this. Otherwise, our bindings cannot run with the latest mmtk-core. We will have a release at the end of this week, and we could update the binding's MSRV when we bump the version for the bindings.

qinsoon:
@wks I pushed a change so when we test bindings with a PR, we use the Rust toolchain version in the PR for the bindings. I think this probably makes more sense, as the tests are initiated from mmtk-core. The MSVR and the toolchain specified in the binding should be updated in the binding when the binding uses a new mmtk-core commit as their dependency.

qinsoon:
binding-refs
JULIA_BINDING_REF=enum-map-deps

qinsoon:
The binding tests should pass (except v8). I will keep an eye on this, and fix if there is any problem. Once everything works, I will merge it.

qinsoon (approved):

Issue #758, opened 2023-02-15
Attempted to schedule sentinel work while bucket is not open


qinsoon:
We observed this assertion failure in both Julia and Ruby.

```console
panicked at 'Attempted to schedule sentinel work while bucket is not open', /home/wks/.cargo/git/checkouts/mmtk-core-3306bdeb8eb4322b/b682c8b/src/scheduler/work_bucket.rs:216:9
Backtrace is disabled.
run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

We don't know what caused this, and we do not have a way to reproduce it yet.

wks:
If I add the following sleep statements and spurious wakeups, it will almost guarantee to trigger the bug.

```patch
diff --git a/src/scheduler/controller.rs b/src/scheduler/controller.rs
index c39aea05e..e8a14059b 100644
--- a/src/scheduler/controller.rs
+++ b/src/scheduler/controller.rs
@@ -128,6 +128,8 @@ impl<VM: VMBinding> GCController<VM> {
                 CoordinatorMessage::Finish => {}
             }
         }
+        std::thread::sleep(std::time::Duration::from_millis(10));
+
         self.scheduler.deactivate_all();
 
         // Tell GC trigger that GC ended - this happens before EndOfGC where we resume mutators.
diff --git a/src/scheduler/scheduler.rs b/src/scheduler/scheduler.rs
index b6bbe4588..657efe2db 100644
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -443,7 +443,10 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                 // coordinator work packets.
             }
             // Wait
-            guard = self.worker_monitor.1.wait(guard).unwrap();
+            guard = {
+                let (new_guard, _timeout_result) = self.worker_monitor.1.wait_timeout(guard, std::time::Duration::from_millis(10)).unwrap();
+                new_guard
+            };
             // The worker is unparked here where `parking_guard` goes out of scope.
         }
 
diff --git a/src/scheduler/work_bucket.rs b/src/scheduler/work_bucket.rs
index 9b3c18dfe..2ff1e7c28 100644
--- a/src/scheduler/work_bucket.rs
+++ b/src/scheduler/work_bucket.rs
@@ -213,6 +213,7 @@ impl<VM: VMBinding> WorkBucket<VM> {
     }
 
[...]

PR #759, opened 2023-02-16
Fix races related to OS scheduling and spurious wakeup

wks:
This PR fixes two race conditions.

1.  If a GC worker spuriously wakes up after the `Finish` coordinator message is sent, the coordinator may deactivate work packets while the GC worker is trying to schedule sentinels or open new buckets.

2.  If `Plan::schedule_collection` runs too slowly due to OS scheduling, workers may open subsequent work buckets while `Plan::schedule_collection` still has pending packets to be added in prior buckets.

Fixes: https://github.com/mmtk/mmtk-core/issues/758

### DIFF
--- a/src/scheduler/controller.rs
+++ b/src/scheduler/controller.rs
@@ -14,7 +14,7 @@ use crate::vm::VMBinding;
 use crate::MMTK;
 use atomic::Ordering;
 
-use super::{GCWork, GCWorkScheduler, GCWorker};
+use super::{CoordinatorWork, GCWorkScheduler, GCWorker};
 
 /// The thread local struct for the GC controller, the counterpart of `GCWorker`.
 pub struct GCController<VM: VMBinding> {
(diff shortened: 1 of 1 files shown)


src/scheduler/controller.rs
```diff
@@ -128,7 +143,11 @@ impl<VM: VMBinding> GCController<VM> {
                 CoordinatorMessage::Finish => {}
             }
         }
-        self.scheduler.deactivate_all();
+
+        {
+            let _guard = self.scheduler.worker_monitor.0.lock().unwrap();
```
qinsoon, line 148: Maybe add some comments about why we need to acquire the lock here (like 'make sure no worker is trying to update buckets or schedule sentinel while we are deactivating buckets').

qinsoon (approved):
LGTM

src/scheduler/controller.rs
```diff
@@ -101,7 +101,22 @@ impl<VM: VMBinding> GCController<VM> {
     /// Coordinate workers to perform GC in response to a GC request.
     pub fn do_gc_until_completion(&mut self) {
         // Schedule collection.
+        // Note: GC workers will start executing work packets as soon as individual work packets
+        // are added.  This means workers may open subsequent buckets while `schedule_collection`
+        // still has packets to be added to prior buckets.
+        // We use the`pending_coordinator_packets` to prevent the workers from opening any work
+        // buckets.
+        self.scheduler
```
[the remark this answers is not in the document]
wks: I have changed the code accordingly.  I added an `initiate_coordinator_work` function for the controller itself to start a coordinator work packet, and an `execute_coordinator_work` function to execute coordinator work packets sent from workers.

src/scheduler/controller.rs
wks: I added a comment.

PR #764, opened 2023-02-22
Set entire byte for side log bit when we copy objects to mature space

qinsoon:
This PR adds `SideMetadataSpec:: set_raw_byte_atomic` and use it to implement `VMGlobalLogBitSpec ::mark_byte_as_unlogged`. We use the new method to unlog objects for two scenarios: 1. copy objects to a mature space, and 2. unlog immix objects during tracing in sticky immix.

### DIFF
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -591,6 +591,19 @@ impl<VM: VMBinding> ImmixSpace<VM> {
 
     fn unlog_object_if_needed(&self, object: ObjectReference) {
         if self.space_args.unlog_object_when_traced {
+            // Make sure the side metadata for the line can fit into one byte. For smaller line size, we should
+            // use `mark_as_unlogged` instead to mark the bit.
+            const_assert!(
+                Line::BYTES
+                    >= (1
+                        << (crate::util::constants::LOG_BITS_IN_BYTE
+                            + crate::util::constants::LOG_MIN_OBJECT_SIZE))
+            );
+            const_assert_eq!(
+                crate::vm::object_model::specs::VMGlobalLogBitSpec::LOG_NUM_BITS,
+                0
+            ); // We should put this to the addition, but type casting is not allowed in constant assertions.
+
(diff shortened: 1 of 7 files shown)


qinsoon:
This is the results for the PR: [internal link](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|ferret-2023-02-22-Wed-224722&benchmark^build^invocation^iteration&GC^time^time.other^time.stw&|10&iteration^1^2|20&1^invocation|50&GC^1^1^+inf|30&1&benchmark&build;jdk-mmtk-log-bit-set-bit|41&Histogram%20(with%20CI)^build^benchmark&).

I measured with GenImmix.It seems it has no measurable difference for this PR in GC time.
log-bit

qinsoon:
I used `SeqCst` store for the run above. I changed that to `Relaxed` store, and the results showed there is ~2% speedup of STW time for generational immix and sticky immix (https://github.com/mmtk/mmtk-core/pull/741#issuecomment-1442771354).

src/policy/immix/immixspace.rs
```diff
@@ -591,7 +591,11 @@ impl<VM: VMBinding> ImmixSpace<VM> {
 
     fn unlog_object_if_needed(&self, object: ObjectReference) {
         if self.space_args.unlog_object_when_traced {
-            VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.mark_as_unlogged::<VM>(object, Ordering::SeqCst);
+            // Every immix line is 256 bytes, which is mapped to 4 bytes in the side metadata.
```
wks, line 594: A line is 256 bytes now, but I anticipate someone may want to do experiments by changing the line size to be extremely large or extremely small.  If it is too small, this optimisation will no longer be valid.  We can

-   Do an assertion that the line is sufficiently large, or
-   Use `if-else` statement to conditionally use this optimisation if the line is sufficiently large.

wks (approved):
LGTM

PR #768, opened 2023-03-07
Collect live bytes during GC

qinsoon:
This PR adds a feature `count_live_bytes_in_gc`. When enabled, each worker will maintain a counter for live bytes when scanning an object, and in `Release`, we get live bytes from all the workers, and sum it up. We also provide `memory_manager::live_bytes_in_last_gc()` for users to query this value. This is usually used to debug fragmentation.


src/scheduler/worker.rs
```diff
@@ -31,6 +31,11 @@ pub fn current_worker_ordinal() -> Option<ThreadId> {
 pub struct GCWorkerShared<VM: VMBinding> {
     /// Worker-local statistics data.
     stat: AtomicRefCell<WorkerLocalStat<VM>>,
+    /// Accumulated bytes for live objects in this GC. When each worker scans
+    /// objects, we increase the live bytes. We get this value from each worker
+    /// at the end of a GC, and reset this counter.
+    #[cfg(feature = "count_live_bytes_in_gc")]
+    live_bytes: AtomicUsize,
```
caizixian, line 38: Do we need atomic for this? It's worker-local right?

caizixian: Ah OK. I was just thinking if we can make it cheap enough by not using atomic for worker local stats, we could enable this feature by default.

k-sareen: Why not make it worker-local and each worker reports the live bytes after their transitive closure? Then we don't have to explicitly iterate through all the workers, we just collate what the numbers we get from the transitive closure.

wks: It could be any stage after the transitive closure stages (including weak reference processing stages).  `ReleaseMutator` should be an ideal place to collect worker-local data.

wks: We have a `ReleaseCollector` work packet.  It currently releases only the copy context.  We can also use this stage to add worker-local bytes to the global `AtomicUsize`, and clear worker-local data.

qinsoon: We need to schedule a 'designated' work packet for each worker (which makes sure the work packet is executed for every worker, and only once for each worker). @wks I think you meant `ReleaseCollector`, not `ReleaseMutator` (which is not 'designated work').

But what should we do for mark compact? For mark compact, we need to either reset the live bytes after the first transitive closure, or not count live bytes in the fist transitive closure. Currently, I can easily iterate through all the workers, and clear their counters. With this change, we would need another designated work packet to reset the counter for mark compact between the two transitive closures.

But anyway, I don't feel we want to enable this by default. 
1. This is only intended as a debug feature to know about fragmentation. Unlike GCs in some languages, we do not count allocated bytes during allocation. This can hardly be a useful stat even if we turn it on by default.
2. we need to call to the binding to get object size, which may not be cheap for some bindings. Enabling this by default does not sound like a good idea.

Let me know what you think.

qinsoon: > Maybe it's easier to defer the "live byte count" to each policy as opposed to each worker? But then it stops being worker-local so it would need synchronization.

Yeah. It need at least the same level synchronization as the current code. And the counting code is scattered to each policy.

qinsoon: We could use `AtomicUsize::as_mut_ptr` and do non-atomic add with unsafe code when we do the counting. But I still question if we want to make this enabled by default.

src/scheduler/gc_work.rs
```diff
@@ -323,6 +357,16 @@ impl<E: ProcessEdgesWork> ObjectTracerContext<E::VM> for ProcessEdgesWorkTracerC
...
+    {
+        self.with_tracer_and_worker(worker, |tracer, _| func(tracer))
+    }
+}
+
+impl<E: ProcessEdgesWork> ProcessEdgesWorkTracerContext<E> {
+    // This Also exposes worker to the callback function. This is not a public method.
+    fn with_tracer_and_worker<R, F>(&self, worker: &mut GCWorker<E::VM>, func: F) -> R
+    where
+        F: FnOnce(&mut ProcessEdgesWorkTracer<E>, &mut GCWorker<E::VM>) -> R,
```
wks, line 369: Strictly speaking, it is unsafe to give the user access to both the tracer and the worker.  The tracer is a closure that calls `trace_object` underneath, and the `worker` is also one of the arguments of `trace_object`.  I planned to forbid this by adding a lifetime annotation, but I couldn't do it without the General Associated Type (GAT) feature before Rust 1.65.

See: https://github.com/mmtk/mmtk-core/blob/813fa99fbf5b5af603d58bd3b54576c141188c37/src/vm/scanning.rs#L53

src/scheduler/worker.rs
k-sareen: I personally don't have a strong opinion. Keeping track of fragmentation is definitely useful though. If it's possible to do it cheaply for every GC, may as well do it. If it's not possible then I don't think should spend more time on it.

src/scheduler/gc_work.rs
wks: Another problem with this is that VM bindings can use `ObjectTracerContext` to get temporary access to `trace_object` during `Scanning::process_weak_refs` and `Scanning::forward_weak_refs`.  Adding a `with_tracer_and_worker` will allow `ScanObjects` to record object sizes, but not for VM-specific weak reference processing.

qinsoon: `with_tracer` takes a mutable reference of `worker`, so I cannot use `worker` any more in `ScanObjects::do_work_common`. We need to a way to get the `worker` back from the tracer. Another option is to expose some methods from `ObjectTracer` so we can get the worker or directly increase the counter.

src/scheduler/gc_work.rs
```diff
@@ -826,6 +870,12 @@ pub trait ScanObjectsWork<VM: VMBinding>: GCWork<VM> + Sized {
                     // If an object supports edge-enqueuing, we enqueue its edges.
                     <VM as VMBinding>::VMScanning::scan_object(tls, object, &mut closure);
                     self.post_scan_object(object);
+
+                    #[cfg(feature = "count_live_bytes_in_gc")]
+                    closure
+                        .worker
+                        .shared
+                        .increase_live_bytes(VM::VMObjectModel::get_current_size(object));
```
wks, line 878: This should be hoisted before the `if` statement, because we will count the object's size regardless of whether it supports edge enqueuing or not.

src/scheduler/gc_work.rs
```diff
@@ -855,6 +905,11 @@ pub trait ScanObjectsWork<VM: VMBinding>: GCWork<VM> + Sized {
                         object_tracer,
                     );
                     self.post_scan_object(object);
+
+                    #[cfg(feature = "count_live_bytes_in_gc")]
+                    _worker
+                        .shared
+                        .increase_live_bytes(VM::VMObjectModel::get_current_size(object));
```
wks, line 912: If we already counted the object sizes in the previous loop, this is unnecessary.

wks (changes requested):
We could count the object sizes in one loop instead of splitting the objects into those that supports edge enqueuing and those that don't.  By doing this, the `with_tracer_and_worker` is also unnecessary.

src/scheduler/gc_work.rs
qinsoon: Why? I thought the previous loop is for edges, and this loop is for nodes.

src/scheduler/gc_work.rs
qinsoon: > Another problem with this is that VM bindings can use `ObjectTracerContext` to get temporary access to `trace_object` during `Scanning::process_weak_refs` and `Scanning::forward_weak_refs`. Adding a `with_tracer_and_worker` will allow `ScanObjects` to record object sizes, but not for VM-specific weak reference processing.

The remaining question is how to count live bytes for weak references. As the weak reference processing is totally at the binding side, we may have to expose a method and let the binding to call it when they scan weak refs.

src/scheduler/worker.rs
wks: @qinsoon Sorry about the typo. I meant `ReleaseCollector`.  Yes.  It will be a problem for `MarkCompact`.  If we make the counter a private field of a `Worker`, it will require another rendezvous (designated works) between the marking phase and the forwarding phase.  Given this situation, I don't mind if we use `AtomicUsize` for now.

src/scheduler/gc_work.rs
wks: > The remaining question is how to count live bytes for weak references. As the weak reference processing is totally at the binding side, we may have to expose a method and let the binding to call it when they scan weak refs.

If we count the object size at `ScanObjectsWork`, it will not be a problem.  The binding keeps objects alive during the weak reference processing stage using `trace_object` (via `ObjectTracer::trace_object`, which is implemented with `ProcessEdgesWork::trace_object` underneath and eventually gets to the space-specific `trace_object`).  Any object visited by `trace_object` will eventually be visited by `ScanObjectsWork`.

src/scheduler/gc_work.rs
qinsoon: I have made this change.

src/plan/tracing.rs
```diff
@@ -74,7 +74,7 @@ impl ObjectQueue for VectorQueue<ObjectReference> {
 /// A transitive closure visitor to collect all the edges of an object.
 pub struct ObjectsClosure<'a, E: ProcessEdgesWork> {
     buffer: VectorQueue<EdgeOf<E>>,
-    worker: &'a mut GCWorker<E::VM>,
+    pub(crate) worker: &'a mut GCWorker<E::VM>,
```
wks, line 77: This no longer needs to be `pub(crate)`.

wks (changes requested):
Only a small problem remains.

src/plan/tracing.rs
qinsoon: I think we still need this. When we do `worker.shared.increase_live_bytes()` in `ScanObjectsWork::do_work_common()`, we already created `ObjectsClosure` which takes `&mut GCWorker`. We can't use `worker` directly, and we have to use it through `ObjectsClosure`.

wks: Oh, sorry. I didn't notice that.

wks (approved):
LGTM

caizixian:
I think we should merge this. We can also add a tracepoint for this.

PR #769, opened 2023-03-07
Add missing set_worker

wks:
The worker was not set after creating the `ProcessEdgesWork`.  It didn't crash because that `ProcessEdgesWork` was used to trace node roots.  Node roots are pinned, so it never used the copy context in the worker.  But the newly added `debug_assert!` in `ImmixSpace::trace_object` accesses the `worker`, and it will crash fail if `worker` is null.


qinsoon (approved):
LGTM

PR #771, opened 2023-03-07
Use saturating_sub in membalancer

qinsoon:
This PR changes subtraction in mem balancer to `saturating_sub`. Using plain sub will wrap around the result, thus may result in a huge number. This will lead to unexpected statistics to be used in mem balancer heuristics. This PR fixes this. In cases where we may have an overflow, `saturating_sub` will make sure the value is 0.


wks (approved):
LGTM

Issue #770, opened 2023-03-07
Trying to open buckets when there are pending coordinator packets


wks:
I observed an error when running Ruby bootstrap tests in the debug mode.  The following is a stack trace.

```
   panicked at 'GCWorker attempted to open buckets when there are pending coordinator work packets', /home/wks/projects/mmtk-github/mmtk-core/src/scheduler/scheduler.rs:242:9
   [2023-03-07T08:09:40Z INFO  mmtk::scheduler::gc_work] End of GC (251/1276 pages, took 155 ms)
      0: mmtk_ruby::handle_gc_thread_panic
                at /home/wks/projects/mmtk-github/mmtk-ruby/mmtk/src/lib.rs:95:14
      1: mmtk_ruby::set_panic_hook::{{closure}}
                at /home/wks/projects/mmtk-github/mmtk-ruby/mmtk/src/lib.rs:124:13
      2: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
                at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/alloc/src/boxed.rs:2032:9
      3: std::panicking::rust_panic_with_hook
                at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:692:13
      4: std::panicking::begin_panic_handler::{{closure}}
                at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:577:13
      5: std::sys_common::backtrace::__rust_end_short_backtrace
                at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/sys_common/backtrace.rs:137:18
      6: rust_begin_unwind
                at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:575:5
      7: core::panicking::panic_fmt
                at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:64:14
      8: mmtk::scheduler::scheduler::GCWorkScheduler<VM>::are_buckets_drained
                at /home/wks/projects/mmtk-github/mmtk-core/src/scheduler/scheduler.rs:242:9
      9: mmtk::scheduler::scheduler::GCWorkScheduler<VM>::new::{{closure}}
[...]

PR #772, opened 2023-03-08
Workaround a synchronization bug.

wks:
If a worker thread spuriously wakes up, it may observe the pending coordinator work counter going from 0 to 1 when the EndOfGC work packet starts to be executed.  That'll trigger an assertion failure.

It is a workaround because a full solution should explicitly address the need to prevent opening more work packets when the coordinator (or any worker) is adding work packets to related buckets.

Closes: https://github.com/mmtk/mmtk-core/issues/770

### DIFF
--- a/src/scheduler/controller.rs
+++ b/src/scheduler/controller.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 
 use crate::plan::gc_requester::GCRequester;
 use crate::scheduler::gc_work::{EndOfGC, ScheduleCollection};
-use crate::scheduler::CoordinatorMessage;
+use crate::scheduler::{CoordinatorMessage, GCWork};
 use crate::util::VMWorkerThread;
 use crate::vm::VMBinding;
 use crate::MMTK;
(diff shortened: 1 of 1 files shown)


k-sareen:
I would say it's better that we don't close the issue if the fix is just a workaround.

wks:
> I would say it's better that we don't close the issue if the fix is just a workaround.

I created an issue to discuss the full fix in detail. https://github.com/mmtk/mmtk-core/issues/774  I think it is OK to close #770 while we can continue discussion in #774

src/scheduler/controller.rs
```diff
@@ -131,7 +131,16 @@ impl<VM: VMBinding> GCController<VM> {
...
-        self.initiate_coordinator_work(&mut end_of_gc, false);
+
+        // Note: We cannot use `initiate_coordinator_work` here.  If we increment the
+        // `pending_coordinator_packets` counter when a worker spuriously wakes up, it may try to
+        // open new buckets and result in an assertion error.  That counter was there to prevent
+        // GC workers from opening more work packets while the coordinator is scheduling more work
+        // packets for the workers.  The only two other coordinator, i.e. `ScheduleCollection` and
+        // `StopMutators`, belong to that category.  `EndOfGC` doesn't add new work packets, so it
+        // doesn't need to prevent the workers from opening buckets (as there are none to be
+        // opened.)
```
qinsoon, line 142: Eventually we still want to use `initiate_coordinator_work` here, right? Can you put the link to both issues in the comments, and add a `FIXME`?

qinsoon (approved):
It is a pretty hacky workaround. But if this is necessary to make sure we do not break bindings, it is okay with me.

src/scheduler/controller.rs
wks: Probably not.  The `initiate_coordinator_work` function is intended for preventing workers from opening more packets.  Maybe `run_coordinator_work_and_prevent_opening_buckets` is a more meaningful (but verbose) function name.  From the diagram I drew in https://github.com/mmtk/mmtk-core/issues/774, `ScheduleCollection` and `StopMutators` are preconditions of some buckets, but `EndOfGC` is not a precondition of anything.  So it may not make sense to wrap all CoordinatorWork with the function currently named `initiate_coordinator_work`.

Anyway, I'll add links to both issues so that we can fix it more properly later.

PR #773, opened 2023-03-08
Add feature to stress copying for Immix

wks:
Added two features:

-   immix_stress_defrag: make every GC a defrag GC
-   immix_stress_copy: additionally make every block a defrag source

These features will make Immix copy as many objects as possible.  They are useful for debugging object copying for VMs that cannot use Semispace for this purpose.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,7 +90,7 @@ is_mmtk_object = ["global_alloc_bit"]
 # Enable object pinning, in particular, enable pinning/unpinning, and its metadata
 object_pinning = []
 
-# The following two features are useful for using Immix for VMs that do not support moving GC.
+# The following features are useful for using Immix for VMs that do not support moving GC.
 
 # Disable defragmentation for ImmixSpace.  This makes Immix a non-moving plan.
 immix_no_defrag = []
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -90,17 +90,31 @@ is_mmtk_object = ["global_alloc_bit"]
...
 immix_no_nursery_copy = []
 # Reduce block size for ImmixSpace.  This mitigates fragmentation when defrag is disabled.
 immix_smaller_block = []
+
+# The following features are useful for debugging moving GC when using Immix.
+# If the VM cannot use Semispace for this purpose (for example, if the VM needs object pinning),
+# these features will make Immix copy as many objects as possible.
+
+# Make every GC a defragment GC.
+immix_stress_defrag = []
```
qinsoon, line 107: Why do you think we should have features rather than just having two `const boolean`s? I have no objection for the PR. But we should have a rule about what becomes a 'feature', and check through our feature list.

wks: Good point.  After a second thought, I think we don't need to make them features.

"non-moving Immix" is a feature because some VMs cannot support copying GC (in production or at some stage of development).  They need to be built against `mmtk-core` as is.  So we have `immix_no_defrag` so that they can disable defrag in the `Cargo.toml` of the VM binding.

Stress testing defragmentation should not be a feature because it is only used in testing environments, and never used in production.  It is always possible to clone `mmtk-core` to a local directory and modify the source code of `mmtk-core` when testing.  It is convenient to add a parameter that can be tuned in the VM binding repository without changing `mmtk-core`, but it is unnecessary.

src/policy/immix/immixspace.rs
```diff
@@ -791,11 +791,10 @@ impl<VM: VMBinding> GCWork<VM> for PrepareBlockState<VM> {
...
             }
             // Check if this block needs to be defragmented.
-            if super::DEFRAG && defrag_threshold != 0 && block.get_holes() > defrag_threshold {
-                block.set_as_defrag_source(true);
-            } else {
-                block.set_as_defrag_source(false);
-            }
+            block.set_as_defrag_source(
+                super::DEFRAG_EVERY_BLOCK ||
+                super::DEFRAG && defrag_threshold != 0 && block.get_holes() > defrag_threshold,
```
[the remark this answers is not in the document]
wks: Yes.  It is a good idea to put `super::DEFRAG` in the front.  But `cargo fmt` indents this piece of code in a very unreadable way.  I replaced it with a sequence of if-else statements with comments, and I hope that can make the code more readable.

qinsoon (approved):
LGTM

Issue #776, opened 2023-03-10
Assertion failure in post_copy


wks:
I observed the following assertion failed once, but I cannot reproduce it.

```
    pub fn post_copy(&mut self, object: ObjectReference, bytes: usize, semantics: CopySemantics) {
        if !VM::VMObjectModel::LOCAL_FORWARDING_BITS_SPEC.is_on_side() {
            object_forwarding::clear_forwarding_bits::<VM>(object);
        } else {
            debug_assert!(!object_forwarding::is_forwarded_or_being_forwarded::<VM>(    // FAILED!
                object
            ));
        }
```

Unfortunately, I don't have a stack trace (I forgot to set `RUST_BACKTRACE=1`).  I was running a Ruby version of GCBench, using the Immix plan, and I used the debugging feature introduced in https://github.com/mmtk/mmtk-core/pull/773 to make Immix defrag every allocated block at every GC.  Unlike mmtk-openjdk, the `ObjectModel::LOCAL_FORWARDING_BITS_SPEC` metadata is a side metadata.

The only reason I can think of is that we forgot to clear the forwarding bits.  This is not necessary if the forwarding bits are in the header, but when on the side, the forwarding bits need to be explicitly cleared.  Currently, on-the-side forwarding bits are cleared in the `PrepareBlockState` work packet at chunk level, cleared together with marking bits.

I'll try to find a way to reproduce it.

wks:
OK. It's simpler than I thought.  `PrepareBlockState` only clears forwarding bits of *Allocated* chunks, but not *Free* chunks.  If all objects in a chunk are moved, it will become *Free*, but forwarded bits in this chunk are not cleared.  When objects are copied, it may be copied to new chunk, and the new chunk will contain stale forwarded bits.  That will trigger the assertion failure.

It is very hard to reproduce because the condition to reproduce this is having a whole chunk becoming free.  As a chunk is 4MiB in size, it is hard to have such a large amount of objects in such a large area being moved in one single GC.  If any object remains in that chunk, it will not become `Free`, and the existing `PrepareBlockState` work packet will clear the forwarded bits metadata before GC.

[...]

PR #777, opened 2023-03-10
Fix stale forwarding bits bug

wks:
When copying, we erroneously assumed that the forwarding bits of the new copy must not have been set if the forwarding bits metadata is on the side.  If the new copy is allocated in a previously free chunk, it may contain stale forwarding bits from the previous GC.

Now we clear forwarding bits in post_copy regardless whether the forwarding bits metadata is in the header or on the side.

Given that post_copy always clears forwarding bits, PrepareBlockState now only clears forwarding bits for defrag source blocks of ImmixSpace.

Closes: https://github.com/mmtk/mmtk-core/issues/776

### DIFF
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -769,12 +769,6 @@ impl<VM: VMBinding> PrepareBlockState<VM> {
                 unimplemented!("We cannot bulk zero unlogged bit.")
             }
         }
-        if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::LOCAL_FORWARDING_BITS_SPEC {
-            side.bzero_metadata(self.chunk.start(), Chunk::BYTES);
-        }
-        // NOTE: We don't need to reset the forwarding pointer metadata because it is meaningless
-        // until the forwarding bits are also set, at which time we also write the forwarding
-        // pointer.
     }
 }
 
@@ -808,6 +802,18 @@ impl<VM: VMBinding> GCWork<VM> for PrepareBlockState<VM> {
             block.set_state(BlockState::Unmarked);
             debug_assert!(!block.get_state().is_reusable());
             debug_assert_ne!(block.get_state(), BlockState::Marked);
+            // Clear forwarding bits if necessary.
+            if is_defrag_source {
+                if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::LOCAL_FORWARDING_BITS_SPEC {
+                    // Clear on-the-side forwarding bits.
+                    // NOTE: In theory, we only need to clear the forwarding bits of occupied lines of
+                    // blocks that are defrag sources.
+                    side.bzero_metadata(block.start(), Block::BYTES);
+                }
+            }
+            // NOTE: We don't need to reset the forwarding pointer metadata because it is meaningless
+            // until the forwarding bits are also set, at which time we also write the forwarding
+            // pointer.
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/util/copy/mod.rs
```diff
@@ -107,18 +107,8 @@ impl<VM: VMBinding> GCWorkerCopyContext<VM> {
...
-            // NOTE: If the forwarding bits were on the side, they would have been cleared when
-            // preparing the space.  See:
-            // - `CopySpace::prepare` and
-            // - `PrepareBlockState::reset_object_mark`.
-            debug_assert!(!object_forwarding::is_forwarded_or_being_forwarded::<VM>(
-                object
-            ));
-        }
+        // Clear forwarding bits.
+        object_forwarding::clear_forwarding_bits::<VM>(object);
```
qinsoon, line 111: This reverts the changes in https://github.com/mmtk/mmtk-core/commit/d341506b70a4465dbbc05e87d27d0b7bd23844bb. It is okay.

src/policy/immix/immixspace.rs
```diff
@@ -808,6 +802,18 @@ impl<VM: VMBinding> GCWork<VM> for PrepareBlockState<VM> {
             block.set_state(BlockState::Unmarked);
             debug_assert!(!block.get_state().is_reusable());
             debug_assert_ne!(block.get_state(), BlockState::Marked);
+            // Clear forwarding bits if necessary.
+            if is_defrag_source {
+                if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::LOCAL_FORWARDING_BITS_SPEC {
```
qinsoon, line 807: This moves the bzero code from `reset_object_mark()` which does bzero unconditionally for the entire chunk to here, which is conditional (`is_defrag_source` and not `Unallocated`), and zeroed at block granularity. Is this an optimization?

src/util/copy/mod.rs
wks: Yes.  The change in https://github.com/mmtk/mmtk-core/commit/d341506b70a4465dbbc05e87d27d0b7bd23844bb was an optimisation, but was wrong.  I didn't take free chunks into consideration.

src/policy/immix/immixspace.rs
wks: Yes.  This is an optimisation.  It will still be correct if we zero entire chunks unconditionally.

Previously, it was necessary to clear all forwarding bits of all chunks because `post_copy` assumed all forwarding bits for all chunks were cleared.  Since `post_copy` now unconditionally clears the forwarding bits of new copies, we no longer need to clear the forwarding bits of all chunks.  We only need to ensure all objects that can be moved have their forwarding bits (of the old copies) cleared before copying them.

An alternative solution is letting `PrepareBlockState` clear all forwarding bits of all Allocated and Free chunks so that `post_copy` can still elide the clearing of forwarding bits.  That would be too costly because very few blocks are defrag sources.  If the number of copied objects is small, it should be faster to clear the forwarding bits for less blocks in `PrepareBlockState` and clear the forwarding bits for more objects in `post_copy`.

qinsoon: I see. The bulk zeroing was also introduced in d341506b70a4465dbbc05e87d27d0b7bd23844bb. So this PR basically reverts both changes in that commit, and does bulk zeroing in another place to fix the original problem. It is probably worth mentioning the commit d341506b70a4465dbbc05e87d27d0b7bd23844bb in the description (which will be the merge commit message).

qinsoon (approved):
LGTM

qinsoon:
The Julia tests failed. But it seems to be a network issue. I don't think it is related with this PR.

src/policy/immix/immixspace.rs
wks: Sure. I'll mention it when merging.

Issue #775, opened 2023-03-09
Clean up cargo features: feature vs option vs const boolean


qinsoon:
I am talking about [cargo features](https://github.com/mmtk/mmtk-core/blob/7750fc0de76a5527cf8d9c03355e73c4ddc0fc88/Cargo.toml#L56), [our runtime options](https://github.com/mmtk/mmtk-core/blob/7750fc0de76a5527cf8d9c03355e73c4ddc0fc88/src/util/options.rs#L637), and `const boolean`s in the code. They all change MMTk's behaviours. But they have obvious differences, and I think we should be careful about this. 

The following is what I propose:
1. Adding `const boolean`s is a trivial change, as they are not public to the users. We should feel free to add as long as they are useful to developers in the future. 
3. We have to use cargo features if
    1. we need conditional compilation (e.g. optional base spaces, alloc bit).
    2. our users (bindings) want something out of box for every build (e.g. non-moving immix).
4. We have to use runtime options if we want to configure something at runtime.
5. If we can freely choose between cargo features and runtime options, we should prefer using runtime options as they do not require a rebuild.

With this proposal, we may consider making the following changes:
* `sanity`: feature -> runtime option. It just involves scheduling sanity work. There is a check in every `ProcessEdgesWork`, we need to make sure this does not introduce overhead.
* `perf_counter`: feature -> (removed). `perf_counter` only works on linux, we can use `cfg(target_os = "linux")` instead. We also have runtime options to set perf events, which should be sufficient.
* `object_pining`: feature -> (removed). We can just include all the methods about object pinning. 
* `nogc_no_zeroing`: feature -> `const boolean`.
* `single_worker`: feature -> (removed). We can just use the options `threads` to set worker threads. 
* `immix_zero_on_release`: feature -> `const boolean`. This is just an option for developing and debugging.

wks:
`object_pinning` requires an additional pinning bit metadata for some spaces.  For Immix, it also requires a metadata load during `trace_object_with_opportunistic_copy`.  So there is a run-time cost.

We may add a `const USE_OBJECT_PINNING: bool` in `ObjectModel` so that `ImmixSpace::is_pinned` can elide the load from `VM::VMObjectModel::LOCAL_PINNING_BIT_SPEC` when `USE_OBJECT_PINNING` is false.

[...]

PR #781, opened 2023-03-16
Introduce immix_non_moving to replace the current immix_no_defrag and immix_no_nursery_copy

qinsoon:
This PR introduces a feature `immix_non_moving`. Based on the proposal https://github.com/mmtk/mmtk-core/issues/775, defrag or nursery copy could simply be a constant controlled by the feature `immix_non_moving`. This also solves the confusion that a binding needs to enable `immix_no_nursery_copy` to make the Immix policy non-moving even if they use Immix which does not do nursery copy.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,10 +92,9 @@ object_pinning = []
 
 # The following two features are useful for using Immix for VMs that do not support moving GC.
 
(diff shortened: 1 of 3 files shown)


wks (approved):
LGTM.  It's better to provide one single easy-to-use user-visible feature for non-moving Immix.  When debugging, we can still modify individual constants manually.

Issue #778, opened 2023-03-14
Still trying to open buckets when there are pending coordinator packets


wks:
The assertion error observed in https://github.com/mmtk/mmtk-core/issues/770 can still be reproduced in rarer cases when GC is triggered very often.  One possibility is that GC workers may spuriously wake up when a mutator triggers GC and the coordinator starts executing `ScheduleCollection` and increments the `pending_coordinator_work` counter.

Perhaps we should not let GC workers do too much work other than getting executable work packets and executing them.

wks:
This error can be easily triggered by manually triggering GC (`handle_user_collection_request`) repeatedly.

The problem goes away if i add a sleep statement (20ms is enough) between consecutive GCs.

Issue #774, opened 2023-03-08
Condition of opening work packets


wks:
I recently found a synchronisation issue: https://github.com/mmtk/mmtk-core/issues/770  Although a workaround exists (https://github.com/mmtk/mmtk-core/pull/772), it reveals some problems in the design of the work packet system.

-   When can more buckets be opened?
-   How to prevent opening new work packets if more work packets are being added?

Currently, buckets form a linear (totally ordered) relation.  A bucket can be opened if (1) all workers are idle, and (2) the coordinator doesn't have pending work, and (3) all previous buckets are empty.

Condition (1) exists because a bucket is only considered finished if all of its work packets are finished, but a worker may be executing one of its work packet.

Condition (1) and (2) exist because one work packet (coordinator packet or ordinary packet) may add more work packets to buckets.  If a subsequent bucket is opened before work packets are added to a prior bucket, it will be an error.  That's why we use the counter `pending_coordinator_packets` to prevent buckets from being opened if there are coordinator packets.

The conditions (1) (2) and (3) are reasonably sound, but are too complicated to test.  As can be seen in the `GCWorkScheduler::poll_slow` function, the condition variable `worker_monitor.1` is not guarded by any concrete variables.  It is very different from typical Condvar use cases like the following:

```rust
let mut guard = mutex.lock();
while !can_open_more_buckets {
    guard = condvar.wait(guard);
}
```

Actually, all other lines of code in `poll_slow` other than the `worker_monitor.1.wait()` itself are computing the condition of "can_open_more_buckets".  Those lines of code include:

1.  Getting schedulable work (`self.poll_schedulable_work(worker)`)
2.  Getting worker-specific work (`self.worker_group.has_designated_work()`)
3.  Checking if any open bucket has sentinels (`self.schedule_sentinels()`)
4.  Checking if the opening condition of any bucket is met (`self.update_buckets()`).  Currently, the opening condition is `scheduler.are_buckets_drained(&cur_stages)` where `cur_stages` includes all previous stages.

All of them need to be checked every time the last worker has parked.  Even with such care, we still encounter synchronisation bugs like https://github.com/mmtk/mmtk-core/issues/770

I think the root problem is that the opening condition is way too complicated than it should be.

A previous issue https://github.com/mmtk/mmtk-core/issues/546 identified that the synchronisation is too complicated, and proposed a major refactoring where there is only one BinaryHeap to hold all poll-able work.

Here I am making an orthogonal proposal of when to open new buckets.

[...]

PR #782, opened 2023-03-16
Let the coordintator thread open buckets

wks:
Note: This PR is **not** intended to do any full-scale refactoring on the work packet framework as described in https://github.com/mmtk/mmtk-core/issues/774  Instead, this PR tries to fix https://github.com/mmtk/mmtk-core/issues/778 cleanly while not changing the overall structure.

The main problem is that `Condvar::wait` is allowed to spuriously wake up at any time even without another thread notifying it.  Currently, work buckets are opened and closed under the assumption that all GC workers have parked.  Due to spurious wake-ups, any GC worker can wake up at any time, and break the assumption.

This PR makes the following changes:

1.  Changed the worker-to-coordinator channel so that
    -   Workers now report the "all workers parked" condition instead of "GC finished".  Now the coordinator determines whether GC has finished.
    -   The channel is also changed from a strict FIFO queue to a shared data structure.  The coordinator always executes all pending coordinator work packets before handling the "all workers parked" condition.
2.  Only the coordinator can open more buckets (with the exception of `Prepare`).
3.  After all workers parked, a boolean flag `group_sleep` is set.  No GC workers can unpark until this flag is explicitly cleared by the coordinator.  It is cleared if the coordinator detected that some workers still have pending work, or the coordinator made more work packets available by adding any work packets to open buckets, or opening new buckets.
4.  Removed the `pending_coordinator_packets` counter.

Part (3) above is the key to handle spurious wake-ups gracefully.  If the `group_sleep` flag is set, all workers will call `Condvar::wait` in a `while` loop until the flag is unset, making the workers immune to spurious wake-ups.

Known issue:

1.  There are two objects for synchronisation, `WorkerMonitor` and `controller::channel::Channel`.  Both of them already existed, except that the former was implemented as a tuple of `(Mutex<()>, Condvar)` while the latter was a sender-receiver pair of the `mpsc` module.  But after this PR, it is obvious that the condition of "all workers have parked" is important for both waking up the coordinator and waking up workers.  We may merge the two objects in the future.
2.   Is there a better name for "group sleeping"?  It should describe a state that once the group of threads entered that state, they cannot get out of that state by themselves, and must be release by an external object (coordinator).


wks:
I ran `lusearch` from Dacapo Chopin (without PGO).  With this PR, the STW time is 1.013 (-2.76%, +2.83%) times the `master` branch, and the total time is 1.002 (-0.89%, +0.90%) of `master`.  This data shows the difference is small, but doesn't show whether there is any structural difference in the scheduling of work packets (for example, load balance).

wenyuzhao:
> Currently, work buckets are opened and closed under the assumption that all GC workers have parked. Due to  wake-ups, any GC worker can wake up at any time, and break the assumption.

work buckets are opened based on only one assumption that all the previous buckets are empty, including all the worker local queues. The only reason it want all the workers to reach the parked state is to check if all the local queues are empty. So it is corrected to open more buckets as long as all the workers reached the "all parked" state once. It probably doesn't matter that some workers spuriously wake up before a new bucket is open.

Same as work bucket close condition. As long as all the buckets are empty, and all the workers reached the "all parked" state once, it is correct to finish the current GC and close all the buckets.

This makes me wondering that probably [this assertion](https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L242-L245) is not strictly correct. I think it's okay for workers to poll for more packets, or even try to open more buckets when the coordinator thread is doing other jobs -- for example when doing `StopMutators`, the worker is totally safe to do other things, like scanning some stacks before all are yielded. Does this spurious wakeups actually cause any correctness issues or crashes without this assertion?

qinsoon:
> work buckets are opened based on only one assumption that all the previous buckets are empty, including all the worker local queues.

This does not sound 100% right. Let's clarify one thing: if a worker is executing a packet, the packet is not in any bucket, and it is not in any local queue, right? Under this definition, If all the active buckets are empty but a worker is executing a work packet, we still cannot open new buckets, as the worker that is executing a work packet may generate more work packets.

> So it is corrected to open more buckets as long as all the workers reached the "all parked" state once.

This makes sense if all the packets are executed by the workers. But what about the coordinator? A coordinator work also belongs to a bucket. If an coordinator is executing a packet, we should not open new buckets, right? The coordinator may generate new work as well.

wenyuzhao:
I'm curious whether there are any real correctness issues or crashes after removing these strong assertions. If so, that means the current worker synchronization mechanism is broken. Otherwise, it is probably less efficient to enforce a strong sequential order between coordinator packets and normal work packets.

qinsoon:
> I'm curious whether there are any real correctness issues or crashes after removing these strong assertions. If so, that means the current worker synchronization mechanism is broken. Otherwise, it is probably less efficient to enforce a strong sequential order between coordinator packets and normal work packets.

What exactly do you mean by 'strong assertions'?

If you meant 'we cannot open new buckets if we have `pending_coordinator_packets`', I think this was used to prevent an issue in `ScheduleCollection`. When the coordinator is executing `ScheduleCollection`, most of the GC packets are not yet added to the buckets, and a worker may see the active buckets as empty, and start opening subsequent buckets.

wenyuzhao:
> What exactly do you mean by 'strong assertions'?

I was talking about this one [1] But probably there are others

(I just noticed one more below)

> When the coordinator is executing `ScheduleCollection`, most of the GC packets are not yet added to the buckets, and a worker may see the active buckets as empty, and start opening subsequent buckets.

When doing `ScheduleCollection` no buckets should be active. I remember the first bucket is open at the end of `StopMutators` [2] .

Even if the workers see all the active buckets are empty, it still cannot open buckets because we already have a check here [3] and prevent buckets from opening before pending coordinator work goes to zero.
* I don't think this check [3] is necessary as well. It should still be correct if we remove this, because workers cannot open the first stw bucket.

[1] https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L242-L245
[2] https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L482-L489
[3] https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L425-L432
[4] https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L81

qinsoon:
> > What exactly do you mean by 'strong assertions'?
> 
> I was talking about this one [1] But probably there are others
> 
> (I just noticed one more below)
> 
> > When the coordinator is executing `ScheduleCollection`, most of the GC packets are not yet added to the buckets, and a worker may see the active buckets as empty, and start opening subsequent buckets.
> 
> When doing `ScheduleCollection` no buckets should be active. I remember the first bucket is open at the end of `StopMutators` [2] .
> 
> Even if the workers see all the active buckets are empty, it still cannot open buckets because we already have a check here [3] and prevent buckets from opening before pending coordinator work goes to zero.
> 
> * I don't think this check [3] is necessary as well. It should still be correct if we remove this, because workers cannot open the first stw bucket [4].
> 
> [1] https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L242-L245 [2] https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L482-L489 [3] https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L425-L432 [4] https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L81

If we don't have checks for `pending_coordinator_packets` for opening buckets, when the coordinator executes `ScheduleCollection`, `StopMutators` could be added to the `Unconstrained` bucket, and it could start execution in parallel with `ScheduleCollection` . In that case, if `StopMutators` finishes before `ScheduleCollection`, a worker may open any bucket if there is no work in them (`ScheduleCollection` hasn't yet pushed work to them). That is the reason we would need such a check.

The next problem is that even with the check for `pending_coordinator_packets` before opening buckets, the assertion you mentioned (https://github.com/mmtk/mmtk-core/blob/master/src/scheduler/scheduler.rs#L242-L245) may fail. That could still cause the problem I discussed above. 

Generally I think it is troublesome to allow every worker to open new buckets and also require them to coordinate with the coordinator. It is way simpler to just let the coordinator to do this kind of global work like opening new buckets.

wenyuzhao:
Well the current work packet system is problematic. But in the original implementation, the coordinator thread can only receive four signals:
1. GC is triggered (and it will call `ScheduleCollection` to initiate the GC)
2. Stop the mutators (because openjdk expects the same thread to stop and wake mutators up. So this cannot be done by an arbitrary worker.
3. `CoordinatorMessage::AllWorkerParked` or `CoordinatorMessage::BucketDrained` to open more buckets
4. `CoordinatorMessage::Finish` to finish a GC and wake the mutaturs up

Note the signal (3) is a part of the things this PR want to do -- don't make workers do too much things other than executing packets. This was removed lately in the [work stealing PR](https://github.com/mmtk/mmtk-core/pull/507/files#diff-561a0acf6a922ed5201f43a704c5a68bfc2431156ceb7b9bd8fa5041917f67c6L367) because on a machine with more cores (say 32 cores), it's already expensive to do a synchronization between buckets. It will be more expensive to communicate with one more thread to do one more synchronization to open buckets.

Also, I believe signal (2) can now be moved back to the normal workers as well. We have a VM companion thread in the openjdk binding to trigger safepoints now.

If we remove (2) and (3) out of the coordinator, the coordinator can be greatly simplified: it can only initiate a GC, and finish a GC. We don't need to care about most of the synchronization issues between coordinator and workers anymore.

Given that (2) and (3) can be removed, the coordinator message queue and pending work counter can be removed as well. We only need the coordinator to have three states, and it triggers a GC when the state transits from `Not In GC` to `GC In Progress`, and finishes the GC when the state transits back to `Not In GC`.

qinsoon:
> > If COORDINATOR_ONLY_STW is false, a worker will execute the work packet. And we may see the problem.
> 
> This is still not a problem. No worker can open the first stw bucket. Thus no worker can open any buckets, until stop mutators work is finished.

During the execution of `StopMutators`, the first stw bucket is activated. 
https://github.com/mmtk/mmtk-core/blob/f8469b0a019bea58cba26e9fb5c70fb7398ccd31/src/scheduler/gc_work.rs#L191

After this, a worker can open new buckets if our defined requirement is met. At this point, the coordinator may be still executing `ScheduleCollection`. This is the problem I talked about -- we need some sort of synchronisation between workers and coordinator. 

Checking `pending_coordinator_packets` before opening buckets attempts to solve the problem. I don't think we can just remove the assertions or the check. However, our implementation with `pending_coordinator_packets` still has issues. That is why we have this PR to simply the synchronization.

wenyuzhao:
> After this, a worker can open new buckets if our defined requirement is met. At this point, the coordinator may be still executing ScheduleCollection. This is the problem I talked about -- we need some sort of synchronisation between workers and coordinator.

With the current `pending_coordinator_packets` checks we have, this will never happen. The counter will not go to zero before ScheduleCollection finishes.

If you are really worrying about this issue, a simpler solution would be taking `ScheduleCollection` out of the coordinator thread, and treating it as a normal work packet executed by a worker thread.

> Checking pending_coordinator_packets before opening buckets attempts to solve the problem. I don't think we can just remove the assertions or the check. However, our implementation with pending_coordinator_packets still has issues. That is why we have this PR to simplify the synchronization.

See my other proposal above. I believe all of the jobs can be moved out of the coordinator, apart from initiating a GC and ending a GC. If you think `pending_coordinator_packets` is not perfect for solving the problem, you can completely remove it now:)

qinsoon (commented):
LGTM. Once Wenyu approves this PR, we can merge it.

wenyuzhao (approved):
lgtm

PR #785, opened 2023-03-21
Add C2 slowpath allocation barrier

wenyuzhao:
Add the missing C2 slow-path allocation barrier.

The C2 slow-path allocation code can do deoptimization after the allocation and before returning to C2 compiled code. The deoptimization itself contains a safepoint.

For generational plans, if a GC happens at this safepoint, the allocated object will be promoted, and all the subsequent field initialization should be recorded. The solution is to unconditionally treat this object as being written if the object is not in nursery, and record this object in the modbuf. This is also the solution for G1: https://github.com/mmtk/openjdk/blob/e1e420b3c4fd627218e45ec977972561c559acec/src/hotspot/share/gc/shared/cardTableBarrierSet.cpp#L136-L160

[mmtk-openjdk PR](https://github.com/mmtk/mmtk-openjdk/pull/205)


src/plan/barriers.rs
```diff
@@ -94,6 +94,14 @@ pub trait Barrier<VM: VMBinding>: 'static + Send + Downcast {
...
     /// Full post-barrier for array copy
     fn memory_region_copy_post(&mut self, _src: VM::VMMemorySlice, _dst: VM::VMMemorySlice) {}
+
+    /// OpenJDK C2 slowpath allocated object
+    ///
+    /// The C2 slowpath allocation code can do deoptimization after the allocation and before
+    /// returning to C2 compiled code. The deoptimization itself contains a safepoint. For generational
+    /// plans, if a GC happens at this safepoint, the allocated object will be promoted, and all the
+    /// subsequent field initialization should be recorded.
+    fn on_slowpath_allocation_exit(&mut self, _obj: ObjectReference) {}
```
qinsoon, line 104: I have a few questions on this method:
* If I understand correct, what happens for OpenJDK is: 1. slowpath allocate so we have an object but its fields are uninitialized, 2. deoptimize (which has a safe point and may trigger a GC here), 3. this method is called after the safe point to record the object (there might have been a GC so the object is promoted), 4. initialize the fields of the object (which updates a mature object and should be recorded). Is this correct?
* How is this method different from a normal write barrier, except that you only know the src object, not the target?
* The name of the method is very specific to OpenJDK and where it is called while the actual semantic of the method is fairly general. Julia also uses write barrier only with src object. I would suggest renaming it to be more general.

qinsoon:
binding-refs
OPENJDK_BINDING_REPO=wenyuzhao/mmtk-openjdk
OPENJDK_BINDING_REF=slowpath-allocation-barrier

src/plan/barriers.rs
wenyuzhao: > If I understand correct, what happens for OpenJDK is ...

Yes this is correct.

> How is this method different from a normal write barrier, except that you only know the src object, not the target?

To object barrier, there is no difference. But this has to be a special call for field barrier. We don't know which field is going to be modified. So for field barrier, we have to assume all the fields are modified.

> I would suggest renaming it to be more general.

Do you have suggestions on a better name? G1 simply call it `invalidate` as it wants to mark the whole memory region occupied by the object as dirty, so this can be processed lately for card scanning. For our case, perhaps `log_object` or `invalidate_object`?

wenyuzhao: Yeah I think `object_anonymous_write` is a good name. This can also merge the object copy semantics. The only reason we may want object clone barrier is that we don't know which field(s) in the clone target do we need to are about.

For my use case, we may want to split this into two barriers, a pre-barrier and a post-barrier. LXR needs this to happen before the actual write or object clone. But generational barriers need this to happen after the write.

wenyuzhao: Include @steveblackburn in the discussion

wenyuzhao: After discussing with Steve, we agreed to use `object_probable_write` as the barrier name. This indicates that some fields of the object might be modified soon without a write barrier. This barrier should strictly be a pre-barrier. I'll update the code soon.

wenyuzhao: done

src/plan/barriers.rs
```diff
@@ -94,6 +94,20 @@ pub trait Barrier<VM: VMBinding>: 'static + Send + Downcast {
...
     /// Full post-barrier for array copy
     fn memory_region_copy_post(&mut self, _src: VM::VMMemorySlice, _dst: VM::VMMemorySlice) {}
+
+    /// A pre-barrier indicating that some fields of the object will probably be modified soon.
+    /// Specifically, the caller should ensure that:
+    ///     * The barrier must called before any field modification.
+    ///     * Some fields (unknown at the time of calling this barrier) might be modified soon, without a write barrier.
+    ///     * There are no safepoints between the barrier call and the field writes.
+    /// 
+    /// Currently, this is only used by the OpenJDK binding. The OpenJDK C2 slowpath allocation code
```
qinsoon, line 104: Preferably use the OpenJDK case as an example, rather than a special case. You can decide whether this goes to `rustdoc` (with `///`), or just normal comments (with `//`). Either sounds okay to me.

src/plan/barriers.rs
```diff
@@ -94,6 +94,20 @@ pub trait Barrier<VM: VMBinding>: 'static + Send + Downcast {
...
+    /// can do deoptimization after the allocation and before returning to C2 compiled code.
+    /// The deoptimization itself contains a safepoint. For generational plans, if a GC
+    /// happens at this safepoint, the allocated object will be promoted, and all the
+    /// subsequent field initialization should be recorded.
+    /// TODO: Review any potential use cases for other VM bindings.
```
qinsoon, line 109: `TODO` should be in the normal comments (`//`).

qinsoon (commented):
LGTM. Just a minor thing about the comments. I will review the OpenJDK PR.

src/plan/barriers.rs
wenyuzhao: Done

src/plan/barriers.rs
wenyuzhao: Done

qinsoon (approved):
LGTM

PR #789, opened 2023-04-04
Fix bugs in StickyImmix

k-sareen:
This commit fixes two bugs in StickyImmix:

  1. The first bug is that StickyImmix double counts the Immix defragmentation headroom in the collection reserve.
  2. The second bug is that StickyImmix had inverted the condition for when a full heap GC is to be performed.

This commit also renames the ImmixSpace in GenImmix to `immix_space` to be more descriptive.

### DIFF
--- a/src/plan/generational/global.rs
+++ b/src/plan/generational/global.rs
@@ -273,7 +273,7 @@ impl<VM: VMBinding> CommonGenPlan<VM> {
         let min_nursery = plan.base().options.get_min_nursery_pages();
         let next_gc_full_heap = available < min_nursery;
         trace!(
-            "next gc will be full heap? {}, availabe pages = {}, min nursery = {}",
+            "next gc will be full heap? {}, available pages = {}, min nursery = {}",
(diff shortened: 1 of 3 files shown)


k-sareen:
Forgot to make a PR for the StickyImmix fixes. Wanted to get it done before we release the next version.

qinsoon:
The collection reserve in sticky immix is computed in the same way as Java MMTk: https://github.com/JikesRVM/JikesRVM/blob/5072f19761115d987b6ee162f49a03522d36c697/MMTk/src/org/mmtk/plan/stickyimmix/StickyImmix.java#L183. I am guessing this could be related with nursery copying? Have you confirmed with Steve that this is a bug?

k-sareen:
Yes I did confirm with Steve that this was a bug. It's double counting the Immix defrag headroom. The original MMTk probably just copy-pasted it without fixing it.

qinsoon (approved):
LGTM

PR #802, opened 2023-05-03
Introduce VMSpace, and allow VMSpace to be set lazily

qinsoon:
This PR allows users to set the start address and size for VM space, and also allows VM space to be initialised lazily.

* Add `VMSpace` which wraps the normal `ImmortalSpace`. The policy is mostly the same as `ImmortalSpace` except that the space may be uninitialized and can be initialized later.
* Add option `vm_space_start` and `vm_space_size`.
* Modify the heap start constant for 32 bits, so JikesRVM's VM space does not overlap with our heap range.
* Fix a subtraction overflow issue in `align_allocation`.


qinsoon:
binding-refs
JIKESRVM_BINDING_REF=set-vm-space-with-options
JULIA_BINDING_REF=use-mmtk-vm-space

src/util/heap/layout/mod.rs
```diff
@@ -34,3 +34,37 @@ pub fn create_mmapper() -> Box<dyn Mmapper> {
...
+
+use crate::util::Address;
+
+/// Return true if the given address in our heap range
+pub fn address_in_heap(addr: Address) -> bool {
+    addr >= vm_layout_constants::HEAP_START && addr < vm_layout_constants::HEAP_END
+}
+
+/// Return true if the given address in our available heap range (where we manage and allocate)
+pub fn address_in_avialable_range(addr: Address) -> bool {
```
wks, line 46:
```suggestion
pub fn address_in_available_range(addr: Address) -> bool {
```

src/policy/space.rs
```diff
@@ -524,7 +522,32 @@ impl<VM: VMBinding> CommonSpace<VM> {
...
+                let start_in_range = if start > layout::vm_layout_constants::AVAILABLE_START {
+                    start
+                } else {
+                    layout::vm_layout_constants::AVAILABLE_START
+                };
+                let end_in_range = if start + extent < layout::vm_layout_constants::AVAILABLE_END {
+                    start + extent
+                } else {
+                    layout::vm_layout_constants::AVAILABLE_END
+                };
```
wks, line 542: This piece of code is somewhat hard to read because they span a few lines and involves multiple variables, although the logic is simple (using the max of two lower bounds as the lower bound, and using the min of two upper bounds as the upper bound). We could use `max` and `min` methods instead of the `if` statements.  But I suggest using `Range` (i.e. the `a..b` expression) for this purpose.  See my comments on the `range_overlaps_available_range` method.

src/util/heap/layout/mod.rs
```diff
@@ -34,3 +34,37 @@ pub fn create_mmapper() -> Box<dyn Mmapper> {
...
+/// Return true if the given range overlaps with our heap range
+pub fn range_overlaps_heap(addr: Address, size: usize) -> bool {
+    !(addr >= vm_layout_constants::HEAP_END || addr + size <= vm_layout_constants::HEAP_START)
+}
+
+/// Return true if the given range overlaps with our available heap range
+pub fn range_overlaps_available_range(addr: Address, size: usize) -> bool {
+    !(addr >= vm_layout_constants::AVAILABLE_END
+        || addr + size <= vm_layout_constants::AVAILABLE_START)
+}
```
wks, line 59: Instead of implementing methods for testing whether an address is within range, I suggest we use the `Range<Address>` type (i.e. `start..end`) and provide a few functions (or constants) that return ranges, for example,

```rust
pub fn heap_range() -> Range<Address> {
    vm_layout_constants::HEAP_START..vm_layout_constants::HEAP_END
}
```

Then we can use standard methods for testing range inclusion

```rust
if heap_range().contains(addr) { ... }
```

We can implement a method for computing the intersection of two ranges to make the code more readable:

```rust
impl Range<Address> {
    fn intersection(&self, other: &Range<Address>) -> Range<Address> {
        self.start.max(other.start)..self.end.min(other.min)
    }
}
```

(Note that there is a `range` crate that provides the same method (https://docs.rs/range/latest/range/struct.Range.html#method.intersect), but it seems to be unmaintained for quite a few years.)

Then we can test whether two ranges overlap by testing if their intersection is empty

```rust
let overlap = (start..end).intersect(available_range());
if !overlap.is_empty() {
    args.plan_args.vm_map.insert(overlap.start, overlap.len(), rtn.descriptor);
}

Comment (qinsoon) [reply src/util/heap/layout/mod.rs]: Good point. I have made changes as suggested.

File: src/plan/global.rs [hunk 4]
```diff
@@ -176,6 +178,7 @@ pub trait Plan: 'static + Sync + Downcast {
     }
 
     fn base(&self) -> &BasePlan<Self::VM>;
+    fn base_mut(&mut self) -> &mut BasePlan<Self::VM>;
```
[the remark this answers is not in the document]
qinsoon: I agree that we should move some states from `Plan` (or `BasePlan`) to `MMTK`, but I am not sure if those base/common spaces are one of the things that should be moved to `MMTK`. As plans will have their specific spaces, it makes sense to let plans to own all the spaces (plan-specific, and base/common ones).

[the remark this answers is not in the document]
wks: > ... let plans to own all the spaces (plan-specific, and base/common ones).

Yes.  That makes sense for the status quo.  Our current model is that plans own spaces.

But "plan" merely means *how* we do GC.  It is a description of a GC algorithm.  A GC algorithm defines what spaces it *needs*, but it does not have to include *all* the spaces present in MMTk, especially the spaces that are not specific to any GC algorithms.  I would argue that the `MMTk` instance owns spaces, but some spaces are customised by the concrete `Plan` instance.  `MMTk` *delegates* some of its operations (such as finding which space contains a given object) to the `Plan` which is determined at run time, and this is why operations like `trace_object` needs a dispatch (may be implemented as SFT, or `match` statement).

src/util/alloc/allocator.rs
```diff
@@ -66,7 +66,7 @@ pub fn align_allocation_inner<VM: VMBinding>(
 
     let mask = (alignment - 1) as isize; // fromIntSignExtend
     let neg_off = -offset; // fromIntSignExtend
-    let delta = (neg_off - region_isize) & mask;
+    let delta = neg_off.wrapping_sub(region_isize) & mask; // Use wrapping_sub to avoid overflow
```
wks, line 69: Since Rust 1.66.0 (our rust-toolchain is 1.66.1), we can use `isize::wrapping_sub_usize` and eliminate the `region_isize` variable.

```rust
    let delta = neg_off.wrapping_sub_usize(region) & mask; // Use wrapping subtraction to avoid overflow
```

Alternatively, we can use `usize::wrapped_add_signed` to eliminate all variables of the `isize` type.

```rust
    let mask = alignment - 1;
    let delta = region.wrapping_neg().wrapping_add_signed(neg_off) & mask; // Use wrapping operations to avoid overflow
```

or

```rust
    let mask = alignment - 1;
    let delta = region.wrapping_add_signed(offset).wrapping_neg() & mask; // Use wrapping operations to avoid overflow
```

I am not sure if any of them is a good idea because none of them is significantly more readable than others.  Mathematically, they are just various ways to express that "`delta ≡ - region - offset` (mod `alignment`)"

src/policy/vmspace.rs
```diff
@@ -0,0 +1,249 @@
...
+        self.space().pin_object(object)
+    }
+    #[cfg(feature = "object_pinning")]
+    fn unpin_object(&self, object: ObjectReference) -> bool {
+        self.space().unpin_object(object)
+    }
+    #[cfg(feature = "object_pinning")]
+    fn is_object_pinned(&self, object: ObjectReference) -> bool {
+        self.space().is_object_pinned(object)
+    }
```
wks, line 43: If this pattern (`self.space().xxxx()`) repeats, we may introduce the `delegate` crate: https://crates.io/crates/delegate

src/policy/space.rs
```diff
@@ -524,7 +522,22 @@ impl<VM: VMBinding> CommonSpace<VM> {
...
+        // they are definitely in our heap range. But for VM space, a runtime could give us an arbitrary range. We only
+        // insert into our vm map if the range overlaps with our heap.
+        {
+            use crate::util::heap::layout;
+            let overlap =
+                Address::range_intersection(&(start..start + extent), &layout::available_range());
+            if !overlap.is_empty() {
+                args.plan_args.vm_map.insert(
+                    overlap.start,
+                    overlap.end - overlap.start,
```
wks, line 536: We can use the `len()` method.
```suggestion
                    overlap.len(),
```

wks (commented):
Only some minor suggestions.

src/policy/space.rs
qinsoon: I dont think we can use `len()` for `Range<Address>`. `len()` is in the trait `ExactSizeIterator `, and Rust only implements it for some common range types, like `Range<u8>` etc.

src/util/alloc/allocator.rs
qinsoon: This is a good idea. I gave it a try. But this requires us to bump our minimal support Rust version to 1.66. I feel it is not worth bumping MSRV merely for this change. I added a TODO for the line, and still kept the original code.

src/policy/vmspace.rs
qinsoon: I made the changes as suggested, using the `delegate` crate. It works well.

qinsoon:
@wks Is the PR good to merge? Let me know if other changes are needed.

src/policy/vmspace.rs
```diff
@@ -0,0 +1,233 @@
...
+            Self {
+                inner: Some(Self::create_space(args, None)),
+                args: args_clone,
+            }
+        } else {
+            Self {
+                inner: None,
+                args: args_clone,
+            }
+        }
```
wks, line 138: If we want to further reduce repetition, we can use the `bool::then` method.  But since the `Self` struct is not very big, the existing code looks pretty readable.  

```suggestion
        let inner = (!args.global_args.options.vm_space_start.is_zero()).then(|| {
            Self::create_space(args, None)
        });
        Self {
            inner,
            args: args_clone,
        }
```

wks (approved):
LGTM, but I still added one minor comment.

src/policy/vmspace.rs
qinsoon: Done

PR #806, opened 2023-05-08
Some fixes for sanity GC

qinsoon:
This PR fixes a few issues for sanity GC.

1. We should not cache roots (for sanity GC) if we are doing sanity GC.
2. We do not need to schedule the work `VMSpecificRoots` in sanity GC. The roots are cached.
3. Reset edge logger before sanity GC. In sanity GC, we will visit the object and edges again. We have to clear what we have logged before.


wks (approved):
LGTM

PR #807, opened 2023-05-08
Reset nursery_index in finalizable processor if we remove objects from candidates

qinsoon:


src/util/finalizable_processor.rs
```diff
@@ -72,6 +72,7 @@ impl<F: Finalizable> FinalizableProcessor<F> {
         // Keep the finalizable objects alive.
         self.forward_finalizable(e, nursery);
 
+        // Set nursery_inex to the end of the candidates (the candidates before the index are scanned)
```
wks, line 75:
```suggestion
        // Set nursery_index to the end of the candidates (the candidates before the index are scanned)
```

wks (changes requested):
Just a minor spelling mistake.

src/util/finalizable_processor.rs
qinsoon: Done

wks (approved):
LGTM

PR #808, opened 2023-05-08
Allow bulk set side metadata

qinsoon:
This PR adds bulk set for side metadata. Similar to bulk zeroing, bulk set will set the side metadata for a given data address range to all 1s (maximum value).

### DIFF
--- a/src/util/memory.rs
+++ b/src/util/memory.rs
@@ -13,8 +13,12 @@ pub fn result_is_mapped(result: Result<()>) -> bool {
 }
 
 pub fn zero(start: Address, len: usize) {
+    set(start, 0, len);
+}
+
+pub fn set(start: Address, val: u8, len: usize) {
     let ptr = start.to_mut_ptr();
-    wrap_libc_call(&|| unsafe { libc::memset(ptr, 0, len) }, ptr).unwrap()
+    wrap_libc_call(&|| unsafe { libc::memset(ptr, val as i32, len) }, ptr).unwrap()
 }
 
 /// Demand-zero mmap:
--- a/src/util/memory.rs
+++ b/src/util/memory.rs
@@ -17,8 +17,9 @@ pub fn zero(start: Address, len: usize) {
 }
 
 pub fn set(start: Address, val: u8, len: usize) {
-    let ptr = start.to_mut_ptr();
-    wrap_libc_call(&|| unsafe { libc::memset(ptr, val as i32, len) }, ptr).unwrap()
+    unsafe {
+        std::ptr::write_bytes::<u8>(start.to_mut_ptr(), val, len);
+    }
 }
 
 /// Demand-zero mmap:
--- a/src/util/metadata/side_metadata/global.rs
+++ b/src/util/metadata/side_metadata/global.rs
@@ -182,26 +182,58 @@ impl SideMetadataSpec {
         }
     }
 
-    /// Bulk-zero a specific metadata for a chunk. Note that this method is more sophisiticated than a simple memset, especially in the following
-    /// cases:
-    /// * the metadata for the range includes partial bytes (a few bits in the same byte).
-    /// * for 32 bits local side metadata, the side metadata is stored in discontiguous chunks, we will have to bulk zero for each chunk's side metadata.
-    ///
-    /// # Arguments
-    ///
-    /// * `start`: The starting address of a memory region. The side metadata starting from this data address will be zeroed.
-    /// * `size`: The size of the memory region.
-    ///
-    pub fn bzero_metadata(&self, start: Address, size: usize) {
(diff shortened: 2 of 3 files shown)


src/util/memory.rs
```diff
@@ -13,8 +13,12 @@ pub fn result_is_mapped(result: Result<()>) -> bool {
 }
 
 pub fn zero(start: Address, len: usize) {
+    set(start, 0, len);
+}
+
+pub fn set(start: Address, val: u8, len: usize) {
     let ptr = start.to_mut_ptr();
-    wrap_libc_call(&|| unsafe { libc::memset(ptr, 0, len) }, ptr).unwrap()
+    wrap_libc_call(&|| unsafe { libc::memset(ptr, val as i32, len) }, ptr).unwrap()
```
wks, line 21: Can we use `std::ptr::write_bytes<u8>` instead?

It seems the underlying LLVM compiler can treat `libc::memset` as an intrinsic function, but I am not sure whether it always behaves this way.

qinsoon: What are the concerns of using `memset`?

wks: It just feels strange to me that we have a function in the Rust standard library for this, but we used something in libc instead.

wks: Performance-wise, I wrote a microbenchmark that sets the contents of a byte array.  There is no performance difference between `write_bytes` and `memset` no matter whether I write the whole array with one `write_bytes`/`memset` or write each element individually in a loop using those functions.

qinsoon: > It just feels strange to me that we have a function in the Rust standard library for this, but we used something in libc instead.

Sounds reasonable. We can change it to the equivalent in Rust standard library. I guess we knew we wanted `memset`, and we could access it from `libc`. So we did not try to find if there is an equivalent in the Rust library

src/util/metadata/side_metadata/sanity.rs
```diff
@@ -433,6 +433,34 @@ pub fn verify_bzero(metadata_spec: &SideMetadataSpec, start: Address, size: usiz
...
+    let end = align_to_region_start(metadata_spec, start + size);
+    let max_value = (1 << (1 << metadata_spec.log_num_of_bits)) - 1;
+    match sanity_map.get_mut(metadata_spec) {
+        Some(spec_sanity_map) => {
+            let mut cursor = start;
+            let step: usize = 1 << metadata_spec.log_bytes_in_region;
+            while cursor < end {
+                spec_sanity_map.insert(cursor, max_value);
+                cursor += step;
+            }
```
wks, line 456:
```suggestion
            let step: usize = 1 << metadata_spec.log_bytes_in_region;
            for cursor in (start..end).step_by(step) {
                spec_sanity_map.insert(cursor, max_value);
            }
```

src/util/metadata/side_metadata/global.rs
```diff
@@ -182,64 +182,92 @@ impl SideMetadataSpec {
...
-    /// * `size`: The size of the memory region.
-    ///
-    pub fn bzero_metadata(&self, start: Address, size: usize) {
-        #[cfg(feature = "extreme_assertions")]
-        let _lock = sanity::SANITY_LOCK.lock().unwrap();
+    /// This method is used for bulk setting side metadata for a data address range. As we cannot guarantee
+    /// that the data address range can be mapped to whole metadata bytes, we have to deal with cases that
+    /// we need to mask and set certain bits in a metadata byte.
+    /// The end address and the end bit are exclusive.
+    pub(super) fn set_meta_bits(
```
wks, line 189: Just like you parameterised `bulk_update_metadata` with the `update_meta_bits` parameter, you can merge `set_meta_bits` and `zero_meta_bits` and parameterise the setting/clearing operation with a parameter, too.

src/util/metadata/side_metadata/sanity.rs
qinsoon: This requires `std::iter::Step` implemented for `Address`. But `std::iter::Step` itself is a nightly-only trait. I don't think we can implement it for `Address`.

src/util/memory.rs
qinsoon: Done. `memory::set` is now implemented with `ptr::write_bytes<u8>`.

src/util/metadata/side_metadata/global.rs
qinsoon: Done. I added `update_meta_bits` which takes two function arguments, `update_bytes` and `update_bits`. `set_meta_bits` and `zero_meta_bits` both use `update_meta_bits` with different update bytes/bits closures.

wks (approved):
LGTM

PR #809, opened 2023-05-08
Fix an issue that the aligned VM space may not match the original location

qinsoon:
This PR fixes an issue that if the given VM space start and size is not aligned to chunks, the aligned start and size is calculated wrong.


wks (approved):
LGTM

PR #812, opened 2023-05-10
Use extreme assertion for metadata mapped assert

wks:
The `assert_metadata_mapped` method uses the `mmap` system call and expects a failure.  This is too costly even for debug builds.  We disable it unless the "extreme_assertions" feature is enabled.

This issue has a greater impact on VMs that use the valid-object bit (VO bit) metadata because the metadata is accessed on every object allocation, and there is such an assertion in metadata access.

The `rr` debugging tool seems to record every system call.  This PR also makes `rr` *much* faster (e.g. 14.20s vs 0.202s when running hello world in Ruby).

# How to test

Compile mmtk-core with debug build, and use `strace` to run the VM.  Before this PR, a hello world program in Ruby (using mmtk-ruby) shall have 400000+ `mmap` invocations returning -1 with `errno = EEXIST`; with this PR, they will all disappear.


qinsoon (approved):
LGTM

PR #814, opened 2023-05-15
Remove Collection::COORDINATOR_ONLY_STW

### LINKED ISSUE #813: Remove Scanning::COORDINATOR_ONLY_STW
We introduced the constant `Scanning::COORDINATOR_ONLY_STW` so that VMs can request the MMTk core to use a dedicated thread (the coordinator thread) to stop mutators and resume mutators.  The OpenJDK binding used to require that feature, but it found a different solution, therefore `Scanning::COORDINATOR_ONLY_STW` is no longer necessary.

After this PR: https://github.com/mmtk/mmtk-core/pull/794, we removed the concept of "coordinator-only work" completely, and that constant `Scanning::COORDINATOR_ONLY_STW` became useless.  We left it there as "deprecated" to give bindings a grace period to remove their uses.  When no bindings are using that constant, we can remove that from MMTk-core.

Checklist:
-   [x] `mmtk-openjdk` (https://github.com/mmtk/mmtk-openjdk/pull/217)
-   [x] `mmtk-jikesrvm` (The JikesRVM binding does not set the constant).
-   [ ] ~~`mmtk-v8`~~ (Note: we cannot build with V8 right now.)
-   [x] `mmtk-ruby` (Done in [30f95ff4d5022b15625e6f4efeeda5374ff672b2](https://github.com/mmtk/mmtk-ruby/commit/30f95ff4d5022b15625e6f4efeeda5374ff672b2))
-   [x] `mmtk-julia` (The Julia binding does not set the constant).

ArberSephirotheca:
Close #813
```
mmtk-core/src/vm/ChangeLog:
    * collection.rs: Remove deprecated constant 'COORDINATOR_ONLY_STW'.
```

### DIFF
--- a/src/vm/collection.rs
+++ b/src/vm/collection.rs
@@ -12,10 +12,6 @@ pub enum GCThreadContext<VM: VMBinding> {
 
 /// VM-specific methods for garbage collection.
 pub trait Collection<VM: VMBinding> {
-    /// This const existed for historical reasons.  It has no effect now.
-    #[deprecated]
-    const COORDINATOR_ONLY_STW: bool = true;
-
     /// Stop all the mutator threads. MMTk calls this method when it requires all the mutator to yield for a GC.
     /// This method is called by a single thread in MMTk (the GC controller).
     /// This method should not return until all the threads are yielded.


wks (approved):
LGTM

k-sareen:
JikesRVM CI fails due to the missing dependency as well (see https://github.com/mmtk/mmtk-core/pull/816). Need to fix it here: https://github.com/mmtk/mmtk-jikesrvm/blob/master/.github/scripts/ci-setup.sh

qinsoon:
> JikesRVM CI fails due to the missing dependency as well (see #816). Need to fix it here: https://github.com/mmtk/mmtk-jikesrvm/blob/master/.github/scripts/ci-setup.sh

I created a PR here: https://github.com/mmtk/mmtk-jikesrvm/pull/140

k-sareen:
JikesRVM CI ran for 6 hrs apparently? Just re-running it to see if it times out again.

PR #815, opened 2023-05-15
Remove a warning in sticky immix trace_object_nursery

qinsoon:
This PR removes an unnecessary `warn` in `trace_object_nursery` in sticky immix. The `warn` logging was used for debugging in OpenJDK where we only allocate objects in the default space and LOS. The logging actually logs every encountered object that is not in the immix space and LOS, which is unnecessary.


wks (approved):
LGTM.

PR #820, opened 2023-05-18
Use min nursery as mem balancer's extra memory

qinsoon:
This PR changes the calculation of extra memory in mem balancer from max nursery to min nursery. This closes https://github.com/mmtk/mmtk-core/issues/790.


k-sareen (approved):
LGTM. Thanks!

Issue #825, opened 2023-05-21
Potential UB in safe function `store_block_list`


shinmao:
https://github.com/mmtk/mmtk-core/blob/cf91f0a55da8edaf92c08a1fef538725f1b4752a/src/policy/marksweepspace/native_ms/block.rs#L167-L172
In line 169, pointers to struct is transmuted to integers which is [undefined behavior](https://doc.rust-lang.org/std/mem/fn.transmute.html). If we want to work on the address, the better alternative could be `as`, isn't it?

Please correct me if I have any misunderstanding;)

qinsoon:
Yes. I submitted a PR to fix this. Thanks for reporting the issue.

shinmao:
Thanks for the quick response. Do you think it worth creating an advisory about this soundness issue?

qinsoon:
Considering that our project is still in its early stages and our releases are currently pre-release versions, I think an advisory may not be necessary at this point. Our current focus is on active development and resolving any issues as they arise.

Again, thank you for your contribution to help improve the safety and reliability of the project.

shinmao:
Sure! The PR seems great. I will close the issue now.

wks:
Let's keep it open until the PR is merged.

The documentation says "Transmuting pointers to integers *in a `const` context* is undefined behavior."  But this is not a `const` context.  But, anyway, the documentation also recommended against using `transmute` outside the `const` context.  I think the PR https://github.com/mmtk/mmtk-core/pull/826 is good.

PR #826, opened 2023-05-22
Remove some uses of mem::transmute in marksweep block

qinsoon:
This PR removes a few unnecessary uses of `mem::transmute` in mark sweep `Block`. This closes https://github.com/mmtk/mmtk-core/issues/825.


wks (approved):
LGTM

PR #827, opened 2023-05-22
Use sysinfo instead of sys-info-rs

k-sareen:


qinsoon:
The tests failed on i686-linux. @k-sareen

k-sareen:
Huh. That's very odd. How do I run 32-bit tests locally?

qinsoon:
> Huh. That's very odd. How do I run 32-bit tests locally?

Roughly `cargo test --target i686-unknown-linux-gnu`.

k-sareen:
It fails because of `u64` -> `usize` conversion on 32-bit. We should return u64 always I feel like

k-sareen:
Fixed @qinsoon. 32-bit tests pass now

src/util/memory.rs
```diff
@@ -192,17 +193,13 @@ pub fn get_process_memory_maps() -> String {
...
-        Ok(mem_info) => mem_info.total as usize,
-        Err(e) => {
-            warn!(
-                "Failed to get sys_info::mem_info: {:?}. Return 1G in get_system_total_memory()",
-                e
-            );
-            1024 * 1024 * 1024
-        }
-    }
+pub(crate) fn get_system_total_memory() -> u64 {
```
qinsoon, line 196: Changing to `u64` is fine. However, this suggests we should not use `get_system_total_memory()` for setting a default heap size. Maybe we should have something like [`get_contraimed_memory()`](https://docs.rs/libuv/latest/libuv/misc/fn.get_constrained_memory.html) instead (comparing to [`get_total_memory()`](https://docs.rs/libuv/latest/libuv/misc/fn.get_total_memory.html)).

qinsoon (approved):
LGTM

src/util/memory.rs
k-sareen: Ah right. The main issue is that you can run 32-bit binaries on 64-bit computers. So you generally have more than 4 GB memory visible to a 32-bit build. Maybe we can just have a compare against 4 GB and if more then cap it to 4 GB for 32-bit builds?

k-sareen: Actually you're right. `get_constrained_memory()` is a better solution as one can reduce the memory limit for a specific process and MMTk should respect that.

qinsoon: I think the current PR is fine. It changes the use of `sys_info_rs` to `sysinfo` which works with more platforms. However, if you are interested to do more changes in this PR, that sounds good.

k-sareen: I might make a separate PR for that later if I work on it. I just wanted to quickly remove the sys-info-rs dependency as its build step is a bit non-standard.

k-sareen:
I'll merge after the tests finish running

Issue #824, opened 2023-05-21
Native Mark-Sweep: block sweeping anomaly


chiarameiohas:
Not all blocks seem to be swept in Mark-Sweep (at the end of the GC of eager-sweeping, or in between collections in lazy-sweeping).

### allocated object points to an unallocated object

I am running my own GC code where we mark all referents of an allocated object (ex. object A is allocated and A->B and A->C, then B&C are the referents and will be marked). 
 I added an assertion that checks if we try to mark an **unallocated** object, and this assertion fails. This means that an allocated object points to an unallocated object (say A->B) ... this can happen only if the unallocated object (B) was freed in the previous collection and (A) was not freed (even though A was not marked). Although I didn't provide a detailed explanation of our GC algorithm,  an allocated object should never point to an unallocated object. This assertion does not fail in Mark-Sweep, because any live object is also allocated, so we mark only allocated objects.

I also copied the mark bitmap to the alloc bitmap at the end of the collection, only then the assertion does not fail...which again makes me think that some blocks are not freed. 
NOTE: the alloc bit of dead objects is zeroed while sweeping the block see https://github.com/mmtk/mmtk-core/blob/cf91f0a55da8edaf92c08a1fef538725f1b4752a/src/policy/marksweepspace/native_ms/block.rs#L278).

###  lazy sweeping sweeps more blocks than eager sweeping

Lazy-Sweeping sweeps more blocks in between collections than Eager-Sweeing at the end of a collection.
Additionally MS with eager-sweeping performs more collections than MS with lazy sweeping (I ran this multiple times to make sure it wasn't a one time wonder).

In a clean MMTK-core image (without my code) I printed how many blocks were swept: how many blocks called ```sweep``` in block.rs 
https://github.com/mmtk/mmtk-core/blob/cf91f0a55da8edaf92c08a1fef538725f1b4752a/src/policy/marksweepspace/native_ms/block.rs#L245
The logs below are from the dacapo biojava benchmark, running Mark-Sweep with a 1GB heap. I activated eager-sweeping by compiling with the ```eager_sweeping``` feature. 

[biojava_eager_1G.log](https://github.com/mmtk/mmtk-core/files/11524538/biojava_eager_1G.log)
[biojava_lazy_1G.log](https://github.com/mmtk/mmtk-core/files/11524539/biojava_lazy_1G.log)

Mark-Sweep with eager sweeping:

```
===== DaCapo evaluation-git-29a657f biojava starting =====
[...]

PR #830, opened 2023-05-25
Sweep abandoned blocks in eager sweeping

qinsoon:
This PR fixes part of the issues in https://github.com/mmtk/mmtk-core/issues/824:
* Remove the incorrect `abandon_blocks()` call in `FreeListAllocator::reset()` for eager sweeping. The incorrect call caused blocks to be pushed to the global abandoned list before they are swept locally by the allocator.
* Add global sweeping that sweeps the abandoned list.


src/util/alloc/free_list_allocator.rs
```diff
@@ -487,11 +472,10 @@ impl<VM: VMBinding> FreeListAllocator<VM> {
...
             }
 
-            bin += 1;
+            // For eager sweeping, we should not have unswept blocks
+            assert!(self.unswept_blocks[bin].is_empty());
 
-            if Self::ABANDON_BLOCKS_IN_RESET {
-                self.abandon_blocks();
-            }
+            bin += 1;
```
wks, line 478: It looks like this pattern occurs in many places.

wks (approved):
It looks good to me.

It looks like the current implementation prefers the C-style while loop over the idiomatic for loop over range.  It's a minor style problem.  We can refactor it later.

qinsoon:
> It looks good to me.
> 
> It looks like the current implementation prefers the C-style while loop over the idiomatic for loop over range. It's a minor style problem. We can refactor it later.

I changed a few occurrences of C-style loop to Rust for loop.

PR #833, opened 2023-05-25
Add extreme assertion for barrier slow path calls

### LINKED ISSUE #685: Adding extreme assertions for barrier slow path calls
One assertion could be that the obj argument for an object barrier should point to a valid object. I had an OpenJDK bug today that passes interior pointers of objects instead of object references. This causes some objects to not be kept alive, and MMTk to scan from the middle of objects.

ArberSephirotheca:
Close #685 
```
mmtk-core/ChangeLog:
        * src/plan/barrier.rs: add extreme assertion to check valid object bit in slow path.
	* vmbindings/dummyvm/src/tests/assertion_barrier.rs: add unit test for new changes.
	* vmbindings/dummyvm/src/tests/mod.rs: add unit test for new changes.
```
I am not quite sure if more extreme assertions are needed, so please suggest any places where we can add assertions for an easier debugging experience.


vmbindings/dummyvm/src/tests/barrier_slow_path_assertion.rs
```diff
@@ -0,0 +1,30 @@
...
+use crate::{api::*, edges};
+use atomic::Atomic;
+use mmtk::util::{Address, ObjectReference, VMMutatorThread, VMThread};
+use mmtk::vm::edge_shape::SimpleEdge;
+use mmtk::AllocationSemantics;
+
+
+#[test]
+#[should_panic]
+fn test_assertion_barrier() {
```
k-sareen, line 13: Is this test supposed to panic? Seems like the barrier is getting called on `objref` which is a valid object. Maybe try something like `objref + 0x4` or something?

ArberSephirotheca: The test should panic(and it panic on my local machine)because the `post_alloc` is not called for the object, which makes the object's metadata remain unitialized. So the assertion `debug_assert!(is_vo_bit_set::<S::VM>(object), "object bit is unset");` fails.

qinsoon: I think the test will panic. It did not panic in the CI tests, as the CI script did not extract the features correctly, and ran the test without any feature.

However, I still find the test a bit hacky. We expect the users to call `post_alloc` for every object they allocate. This test deliberately does not call `post_alloc` to create a scenario that the alloc bit is not set. We could just create an object normally (with `alloc` and `post_alloc`), then create an internal reference and use that as the object reference, which of course is not a valid object reference and its alloc bit is not set. In addition to this, we may also want a test case that passes.

I will push some changes to the PR.

k-sareen (approved):
LGTM. Thanks!

PR #834, opened 2023-05-25
Comment FreeList and its implementations

wks:
I copied/pasted the comments from JikesRVM and reformatted for Rust.

# About style

Rust uses MarkDown for doc comments, and it doesn't have the structured `@param` and `@return` mark-ups.  I expressed the same idea using the `# Parameters` and `# Returns` headers, which may make the comment look a bit verbose.  Rust's standard library doesn't seem to have dedicated "parameters" and "returns" sections for function doc-comments.

# JikesRVM's EPLv1 License

One problem, however, is that JikesRVM is licensed under EPL.  If I include an EPL-licensed component, and I modify it, I need to release the source code of it.  In this case, I modified the comments, and I released those comments to the public.  If one of our down-stream consumers uses mmtk-core and releases its software in the binary form, they will need to open-source the comments if they modified them.  This is a bit silly, but could be problematic.  The safest way is reading comments from my branch, but does not merge the comment into the master.

Alternatively, we can rewrite the comments from scratch.

I think we have not had this problem because the Rust version of mmtk-core is re-written from scratch in Rust.  Maybe we have been intentionally avoid copying code from JikesRVM directly.

### DIFF
--- a/src/util/freelist.rs
+++ b/src/util/freelist.rs
@@ -14,16 +14,116 @@ const MULTI_MASK: i32 = 1 << (TOTAL_BITS - 1);
 const COALESC_MASK: i32 = 1 << (TOTAL_BITS - 2);
 const SIZE_MASK: i32 = (1 << UNIT_BITS) - 1;
 
+/// This is a very simple, generic malloc-free allocator.  It works abstractly, in "units", which
+/// the user may associate with some other allocatable resource (e.g. heap blocks).  The user
+/// issues requests for N units and the allocator returns the index of the first of a contiguous
+/// set of N units or fails, returning -1.  The user frees the block of N units by calling `free()`
+/// with the index of the first unit as the argument.
+///
+/// Properties/Constraints:
+///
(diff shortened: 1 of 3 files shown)


qinsoon:
Currently most of our code base uses `# Arguments` (as suggested in https://doc.rust-lang.org/rust-by-example/meta/doc.html), and uses no specific section for return values (just mention it in the description paragraph if necessary). For example,
https://github.com/mmtk/mmtk-core/blob/440b57c5bcadedf9477ed469dc7ef527832d6a70/src/memory_manager.rs#L336-L342

Given that Rust does not nominate a more structured way for arguments and returns, I think it is fine that we just stick to our current style, rather than introducing a different style.

wks:
I am closing this PR because of two problems.  One is the license.  The other is, the comments may not reflect what the code is actually doing, given that the code may have been changed after being ported, and given that the comments in the original JikesRVM is not 100% complete.  We should rewrite the comments for the Rust MMTk, using the comments from JikesRVM as a guideline rather than copying and pasting directly from JikesRVM.

Issue #835, opened 2023-05-26
System::new_all() takes 400ms to initialise


wks:
We switched to the `sysinfo` crate in https://github.com/mmtk/mmtk-core/pull/827  Now the start-up time of MMTk is greatly increased, even in release builds.  The single most time-consuming function call is `System::new_all()`.  Using the following code to measure:

```rs
pub(crate) fn get_system_total_memory() -> u64 {
    let time1 = Instant::now();
    let sys = System::new_all();
    let time2 = Instant::now();
    let result = sys.total_memory();
    let time3 = Instant::now();

    eprintln!("Time spent in System::new_all(): {}ns", (time2 - time1).as_nanos());
    eprintln!("Time spent in sys.total_memory(): {}ns", (time3 - time2).as_nanos());

    result
}
```

The result is:

```
$ time ./miniruby --mmtk --mmtk-plan=Immix /home/wks/code/ruby/gctest/hello.rb
Time spent in System::new_all(): 402819994ns
Time spent in sys.total_memory(): 65ns
Hello world!

________________________________________________________
Executed in  432.45 millis    fish           external
   usr time  240.00 millis  911.00 micros  239.09 millis
   sys time  366.58 millis  317.00 micros  366.27 millis
```

Note that this is using the release build.

This problem has greater impact on short-running scripts, such as Ruby which runs lots of small test scripts during testing.

wks:
Replacing `System::new_all()` with `System::new_with_specifics(RefreshKind::new().with_memory())` will reduce the initialisation time to 108247ns (0.1ms) in debug build and 65510ns (0.06ms) in release build.  This overhead is trivial compared to the 20ms time needed to execute "Hello world".

I'll make a PR for that.

k-sareen:
Apologies. Thanks for catching that.

k-sareen:
We should make a CI test which makes sure no regressions happen to startup time.

PR #836, opened 2023-05-26
Let sysinfo only load memory-related info

wks:
This avoids slowing down the start-up time of MMTk by only loading necessary components of sysinfo.

Closes: https://github.com/mmtk/mmtk-core/issues/835


k-sareen (approved):
LGTM. Thanks!

PR #842, opened 2023-06-01
Fix `is_live` for ImmixSpace

wks:
To save time during `Prepare`, we only clear the on-the-side forwarding bits for defrag sources.  As a result, blocks which were defrag sources in the previous GC may still have stale forwarding bits from the previous GC.  It makes it unreliable to use the forwarding bits alone to test if a given object is forwarded or not.

This commit lets `is_alive` also test whether the block is a defrag source before looking at the forwarding bits.  If it is not a defrag source, it must not have been forwarded.


qinsoon (approved):
LGTM

PR #855, opened 2023-06-18
Allow plan-specific bucket config

qinsoon:
This PR adds `WorkBucketStageConfig` which can be defined by each plan. Thus each plan can create their own custom buckets. Currently only mark compact needs a custom bucket for compaction.
* Add `WorkBucketStageConfig` which defines all the bucket stages, and the first STW stage. Each plan may provide its own config.
* Add `WorkBucketStage::Custom(usize)`, and remove `WorkBucketStage::Compact`. The mark compact plan create their own custom bucket stage.
* Use `HashMap`, instead of `EnumMap` for buckets. All `buckets[stage]` uses need to be replaced with `buckets[&stage]` due to the difference in indexing the map.

### DIFF
--- a/src/plan/markcompact/global.rs
+++ b/src/plan/markcompact/global.rs
@@ -44,6 +44,9 @@ pub const MARKCOMPACT_CONSTRAINTS: PlanConstraints = PlanConstraints {
     ..PlanConstraints::default()
 };
 
+/// A custom stage for doing compaction for mark compact. It happens after all the tracing stages, and before Release
+pub const COMPACT_STAGE: WorkBucketStage = WorkBucketStage::Custom(0);
+
 impl<VM: VMBinding> Plan for MarkCompact<VM> {
     type VM = VM;
 
@@ -96,21 +99,21 @@ impl<VM: VMBinding> Plan for MarkCompact<VM> {
         //     );
 
         // Stop & scan mutators (mutator scanning can happen before STW)
-        scheduler.work_buckets[WorkBucketStage::Unconstrained]
+        scheduler.work_buckets[&WorkBucketStage::Unconstrained]
             .add(StopMutators::<MarkingProcessEdges<VM>>::new());
 
         // Prepare global/collectors/mutators
-        scheduler.work_buckets[WorkBucketStage::Prepare]
+        scheduler.work_buckets[&WorkBucketStage::Prepare]
             .add(Prepare::<MarkCompactGCWorkContext<VM>>::new(self));
 
-        scheduler.work_buckets[WorkBucketStage::CalculateForwarding]
+        scheduler.work_buckets[&WorkBucketStage::CalculateForwarding]
             .add(CalculateForwardingAddress::<VM>::new(&self.mc_space));
         // do another trace to update references
-        scheduler.work_buckets[WorkBucketStage::SecondRoots].add(UpdateReferences::<VM>::new());
-        scheduler.work_buckets[WorkBucketStage::Compact].add(Compact::<VM>::new(&self.mc_space));
+        scheduler.work_buckets[&WorkBucketStage::SecondRoots].add(UpdateReferences::<VM>::new());
+        scheduler.work_buckets[&COMPACT_STAGE].add(Compact::<VM>::new(&self.mc_space));
 
         // Release global/collectors/mutators
-        scheduler.work_buckets[WorkBucketStage::Release]
+        scheduler.work_buckets[&WorkBucketStage::Release]
             .add(Release::<MarkCompactGCWorkContext<VM>>::new(self));
 
         // Reference processing
@@ -118,19 +121,19 @@ impl<VM: VMBinding> Plan for MarkCompact<VM> {
             use crate::util::reference_processor::{
                 PhantomRefProcessing, SoftRefProcessing, WeakRefProcessing,
             };
-            scheduler.work_buckets[WorkBucketStage::SoftRefClosure]
+            scheduler.work_buckets[&WorkBucketStage::SoftRefClosure]
                 .add(SoftRefProcessing::<MarkingProcessEdges<VM>>::new());
-            scheduler.work_buckets[WorkBucketStage::WeakRefClosure]
+            scheduler.work_buckets[&WorkBucketStage::WeakRefClosure]
                 .add(WeakRefProcessing::<MarkingProcessEdges<VM>>::new());
-            scheduler.work_buckets[WorkBucketStage::PhantomRefClosure]
+            scheduler.work_buckets[&WorkBucketStage::PhantomRefClosure]
                 .add(PhantomRefProcessing::<MarkingProcessEdges<VM>>::new());
 
             use crate::util::reference_processor::RefForwarding;
-            scheduler.work_buckets[WorkBucketStage::RefForwarding]
+            scheduler.work_buckets[&WorkBucketStage::RefForwarding]
                 .add(RefForwarding::<ForwardingProcessEdges<VM>>::new());
 
             use crate::util::reference_processor::RefEnqueue;
-            scheduler.work_buckets[WorkBucketStage::Release].add(RefEnqueue::<VM>::new());
+            scheduler.work_buckets[&WorkBucketStage::Release].add(RefEnqueue::<VM>::new());
         }
 
         // Finalization
@@ -139,33 +142,33 @@ impl<VM: VMBinding> Plan for MarkCompact<VM> {
             // finalization
             // treat finalizable objects as roots and perform a closure (marking)
             // must be done before calculating forwarding pointers
-            scheduler.work_buckets[WorkBucketStage::FinalRefClosure]
+            scheduler.work_buckets[&WorkBucketStage::FinalRefClosure]
                 .add(Finalization::<MarkingProcessEdges<VM>>::new());
             // update finalizable object references
             // must be done before compacting
-            scheduler.work_buckets[WorkBucketStage::FinalizableForwarding]
+            scheduler.work_buckets[&WorkBucketStage::FinalizableForwarding]
                 .add(ForwardFinalization::<ForwardingProcessEdges<VM>>::new());
         }
 
         // VM-specific weak ref processing
-        scheduler.work_buckets[WorkBucketStage::VMRefClosure]
+        scheduler.work_buckets[&WorkBucketStage::VMRefClosure]
             .set_sentinel(Box::new(VMProcessWeakRefs::<MarkingProcessEdges<VM>>::new()));
 
         // VM-specific weak ref forwarding
-        scheduler.work_buckets[WorkBucketStage::VMRefForwarding]
+        scheduler.work_buckets[&WorkBucketStage::VMRefForwarding]
             .add(VMForwardWeakRefs::<ForwardingProcessEdges<VM>>::new());
 
         // VM-specific work after forwarding, possible to implement ref enququing.
-        scheduler.work_buckets[WorkBucketStage::Release].add(VMPostForwarding::<VM>::default());
+        scheduler.work_buckets[&WorkBucketStage::Release].add(VMPostForwarding::<VM>::default());
 
         // Analysis GC work
         #[cfg(feature = "analysis")]
         {
             use crate::util::analysis::GcHookWork;
-            scheduler.work_buckets[WorkBucketStage::Unconstrained].add(GcHookWork);
+            scheduler.work_buckets[&WorkBucketStage::Unconstrained].add(GcHookWork);
         }
         #[cfg(feature = "sanity")]
-        scheduler.work_buckets[WorkBucketStage::Final]
+        scheduler.work_buckets[&WorkBucketStage::Final]
             .add(crate::util::sanity::sanity_checker::ScheduleSanityGC::<Self>::new(self));
     }
 
@@ -217,6 +220,17 @@ impl<VM: VMBinding> MarkCompact<VM> {
             .verify_side_metadata_sanity(&mut side_metadata_sanity_checker);
 
         res
+    }
+
+    pub fn work_bucket_stage_config() -> WorkBucketStageConfig {
+        let mut default = WorkBucketStageConfig::default();
+        // Add compact stage before release
+        crate::util::rust_util::vec_insert_before(
+            &mut default.stages,
+            COMPACT_STAGE,
+            &WorkBucketStage::Release,
+        );
+        default
     }
 }
 
--- a/src/plan/markcompact/global.rs
+++ b/src/plan/markcompact/global.rs
@@ -44,8 +44,24 @@ pub const MARKCOMPACT_CONSTRAINTS: PlanConstraints = PlanConstraints {
     ..PlanConstraints::default()
 };
 
-/// A custom stage for doing compaction for mark compact. It happens after all the tracing stages, and before Release
-pub const COMPACT_STAGE: WorkBucketStage = WorkBucketStage::Custom(0);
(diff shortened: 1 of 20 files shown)


wks:
I think `CalculateForwarding`, `SecondRoot`, `RefForwarding`, `FininalzableForwarding`, `VMRefForwarding`, `Compact` are all specific to MarkCompact.

qinsoon:
binding-refs
JULIA_BINDING_REF=update-mmtk-core-855

qinsoon:
These are some performance results for the PR. I compared two builds, `master` with the PR, and measured immix and mark compact. There is no improvement or degradation for both plans. This is expected. Though we have some mark compact specific buckets for plans like immix, there is no actual overhead. When the schedule opens an empty bucket, it will not wake up workers. Instead, it just keeps opening the next bucket until it finds a bucket with some work. 

The change from `EnumMap` to `HashMap` for buckets may slow things down, as indexing in `EnumMap` is essentially indexing in an array which is cheaper. But it seems no measurable overhead on this. If requested, I can make changes so we can index into the bucket map as an array.

Immix [link](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|skunk-2023-06-20-Tue-012101&benchmark^build^buildstring^invocation^iteration^ix^mc&time^time.other^time.stw&|10&ix^1^1|10&iteration^1^2|10&benchmark^2^batik&benchmark^2^jme|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^buildstring^benchmark&)
plan-specific-bucket-ix

MarkCompact [link](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|skunk-2023-06-20-Tue-012101&benchmark^build^buildstring^invocation^iteration^ix^mc&time^time.other^time.stw&|10&ix^1^null|10&iteration^1^2|10&benchmark^2^batik&benchmark^2^jme|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^buildstring^benchmark&)
plan-specific-bucket-mc

wks:
This PR attempts to reduce the number of (globally available) buckets by introducing plan-specific buckets.  Currently plan-specific buckets are used for MarkCompact because many buckets, from CalculatingForwarding to Compact, are all specific to MarkCompact.

However, from the high level, I think this PR is only "scratching itch" because it solves the immediate problem (i.e. some buckets are specific to some plans), but does not address the fundamental problem.

## The proper way to describe a GC algorithm

The fundamental difference between MarkCompact and other plans we currently have is the *high-level steps of GC*.  For MarkCompact, the actual high-level idea those many buckets are trying to express is that *MarkCompact does two transitive closures*.  Concretely, the steps can be described as:

1.  Do the first transitive closure from roots and process weak references during the course. (from Prepare to VMRefClosure)
2.  Compute the new addresses of all objects. (CalculateForwarding)
3.  Do another transitive closure from roots and forward all edges by tracing, including objects that are kept alive during weak reference processing. (from SecondRoot to VMRefForwarding)
4.  Move objects to their new locations (Compact)
5.  Call `release` on spaces, mutators, collectors, etc.
6.  (if sanity GC is enabled) Do yet another transitive closure to validate objects.

While other GC algorithms are simpler:
-   A.  Do the first transitive closure from roots and process weak references during the course. (from Prepare to VMRefClosure)
-   B.  Call `release` on spaces, mutators, collectors, etc.
-   C.  (if sanity GC is enabled) Do another transitive closure to validate objects.

Step (1) and (A) are the same, and step (5) and (B) are the same, while step (2), (3) and (4) are specific to MarkCompact. (If sanity GC is enabled, the sanity GC is also similar between GCs). This PR captures this apparent difference, and allows plans (currently MarkCompact) to add more buckets according to their own need.  The interface for extending the default set of buckets is `bulk_insert_before`.  For example: 

```rust
        config.bulk_insert_before(
            vec![
                // Calculate forwarding addresses for live objects
                CALCULATE_FORWARDING_STAGE,
                // Second round transitive closure
                SECOND_ROOTS_STAGE,
                // Forwarding weak/final references
                REF_FORWARDING_STAGE,
                FINALIZABLE_FORWARDING_STAGE,
                VM_REF_FORARDING_STAGE,
                // Compaction
                COMPACT_STAGE,
            ],
            WorkBucketStage::Release,
        );
```

This piece of code conveys the message that "MarkCompact does this, this and this before Release".  I don't think this is a good way to express a GC algorithm.  MarkCompact should be described as the five steps shown above, i.e. two transitive closures with operations in between.

## Not general enough

I agree that buckets should be plan-specific.  Work buckets are categories of tasks (packets) that should be executed at different stages of GC.

I even believe the set of buckets (and their topology) should even be specific to each GC.  For example, minor GC and full GC should be different.

But currently, it just happened that all of our generational plans can be expressed in a similar way.  Minor GCs differ from major GCs only in the root set and whether to trace through mature objects.  Therefore, both nursery and full GCs can be expressed with the step (A) and (B) shown above, namely transitive closure followed by releasing, although the transitive closures are different in details.

But it's not generally true that minor GC and major GCs touch the same set of buckets.  We don't have RC-based plans, yet.  But if we introduce RC Immix, then minor GCs will only include reference counting, while major GCs include transitive closure computing.  If we introduce an RC-based algorithms without back-up tracing, or with trial deletion, then we may argue that `Closure`, `{Soft,Weak,Final,Pantom,VMRef}Closure` are all plan-specific.  The only buckets that are fully general may be Prepare and Release, but even that could be questionable if an on-the-fly concurrent GC never stops all threads at the same time.

## Conclusion

This PR addressed the problem of plan-specific buckets, but I think this approach is not general enough.

I think the proper way to describe the bucket sequence (better if it is abucket graph) is introducing a domain-specific language (or a highly-readable API) that can express high-level ideas that "do a transitive closure" which constructs multiple buckets in a row (or a subgraph).  We may allow each plan to define several bucket sequences (or graph) (for example, one for major GC and another for minor GCs), or define one sequence (or graph) for each GC.

I think we need some input from LXR.  That could demand a very different set of buckets.

Another minor problem of this PR that we no longer have one place where we can see all the buckets we have (previously the `enum WorkBucketStage` which contains all stages for all plans).  The current way to define plan-specific buckets in MarkCompact is a bit awkward because it relies on constants instead of enum cases to extend `Custom(usize)`.  I don't think the profits this PR brings (moving plan-specific buckets to plans, but no obvious performance advantage) is worth sacrificing the simplicity (having all buckets in one place).

## Other things to consider

Currently sanity GC happens after the regular part of the GC finishes.  But previous discussions mentioned the possibility to do sanity GC before and after the regular GC.  I think in the future we should make that possible by allowing the bucket sequence (or graph) to be constructed in a more flexible way.

qinsoon:
> However, from the high level, I think this PR is only "scratching itch" because it solves the immediate problem (i.e. some buckets are specific to some plans), but does not address the fundamental problem.

Yes, that is the intention of this PR. It tries to prevent our global bucket list keep growing when we add more plans and features that need new buckets. It offers more flexibility to allow us introduce buckets in certain circumstances. It does not try to solve any fundamental issue. As https://github.com/mmtk/mmtk-core/issues/774#issuecomment-1462964165, Steve made it clear that Wenyu is going to reimplement the work bucket.

> Another minor problem of this PR that we no longer have one place where we can see all the buckets we have (previously the `enum WorkBucketStage` which contains all stages for all plans). 

I don't agree with this. If a bucket is specific to a plan, it should not be even visible to other plans (only the scheduler and the plan itself need to know about the bucket). Current code is awkward as we have all the buckets (general or plan-specific) defined in a global `enum`. This will get worse and worse when we need some buckets in certain conditions (like `ClearVOBits`).

> The current way to define plan-specific buckets in MarkCompact is a bit awkward because it relies on constants instead of enum cases to extend `Custom(usize)`. 

Do you have any concrete suggestion? Alternatively we can just remove the `enum` and always use an integer-size struct to encode a stage (like what Java MMTk does: https://github.com/JikesRVM/JikesRVM/blob/5072f19761115d987b6ee162f49a03522d36c697/MMTk/src/org/mmtk/plan/Simple.java#L48). The point is that I think we should allow new buckets and stages conditionally (determined by plan, feature, or option), thus the representation of the stage needs to be extensible. C-style enum is not extensible.

> I don't think the profits this PR brings (moving plan-specific buckets to plans, but no obvious performance advantage) is worth sacrificing the simplicity (having all buckets in one place).

I wouldn't call the current code simple. I think it is just wrong. Why would the scheduler check if the `Compact` bucket should be opened or not when we run `Immix` GC? That is just a workaround as we currently only allow a fixed set of buckets. This PR tries to correct that.

src/plan/markcompact/global.rs
```diff
@@ -44,6 +44,25 @@ pub const MARKCOMPACT_CONSTRAINTS: PlanConstraints = PlanConstraints {
...
+pub const REF_FORWARDING_STAGE: WorkBucketStage = WorkBucketStage::Custom(2);
+
+/// A stage to forward finalizable objects. This happens after the second transitive closure.
+pub const FINALIZABLE_FORWARDING_STAGE: WorkBucketStage = WorkBucketStage::Custom(3);
+
+/// A stage to allow VM bindings to do ref forwarding. This happens after other ref forwarding work.
+pub const VM_REF_FORARDING_STAGE: WorkBucketStage = WorkBucketStage::Custom(4);
+
+/// A custom stage to do compaction. It happens after all the tracing stages, and before Release
+pub const COMPACT_STAGE: WorkBucketStage = WorkBucketStage::Custom(5);
```
wks, line 64: May organise them into an enum, such as
```
#[derive(enum_map::Enum)]
enum MarkCompactStage {
    CalculateForwarding,
    SecondRoots,
    RefForwarding,
    FinalizableForwarding,
    VMRefForwarding,
    Compact,
}
```
and automatically map them to and from `WorkBucketStage::Custom(...)`.  That should be simpler.

src/plan/markcompact/global.rs
```diff
@@ -218,6 +238,27 @@ impl<VM: VMBinding> MarkCompact<VM> {
...
+                CALCULATE_FORWARDING_STAGE,
+                // Second round transitive closure
+                SECOND_ROOTS_STAGE,
+                // Forwarding weak/final references
+                REF_FORWARDING_STAGE,
+                FINALIZABLE_FORWARDING_STAGE,
+                VM_REF_FORARDING_STAGE,
+                // Compaction
+                COMPACT_STAGE,
+            ],
```
wks, line 257: If we have a `MarkCompactStage` that implements `enum_map::Enum`, we can convert all of its variants into `WorkBucketStage` automatically, such as: `            (0..MarkCompactStage::LENGTH).map(|i|MarkCompactStage::from_usize(i).as_stage()).collect_vec()
`

src/scheduler/gc_work.rs
```diff
@@ -599,7 +599,7 @@ pub trait ProcessEdgesWork:
             // Executing these work packets now can remarkably reduce the global synchronization time.
             self.worker().do_work(work_packet);
         } else {
-            self.mmtk.scheduler.work_buckets[WorkBucketStage::Closure].add(work_packet);
+            self.mmtk.scheduler.work_buckets[&WorkBucketStage::Closure].add(work_packet);
```
wks, line 602: If this `&` occurs frequently, we may consider this as a chance to turn the `work_buckets` vec into a private field, and force all accesses through a method such as `scheduler.add_work_packet(stage, packet);`  We can let the `stage` argument be `impl Into<WorkBucketStage>` so that we can pass `PlanSpecificStage` instances directly.

wenyuzhao:
> Current code is awkward as we have all the buckets (general or plan-specific) defined in a global enum. This will get worse and worse when we need some buckets in certain conditions (like ClearVOBits).

> Why would the scheduler check if the Compact bucket should be opened or not when we run Immix GC?

Another simpler solution: You can also choose to define all buckets in a global enum, but only enabling / disabling a subset of them in each specific plan. 

https://github.com/wenyuzhao/mmtk-core/blob/lxr/src/plan/lxr/global.rs#L852-L873

> If a bucket is specific to a plan, it should not be even visible to other plans (only the scheduler and the plan itself need to know about the bucket).

I believe buckets should be general and shouldn't be private to a specific plan. This is just a workaround for gc thread synchronization. This shouldn't be private to any plan.

wenyuzhao:
> Why would the scheduler check if the Compact bucket should be opened or not when we run Immix GC?

Actually there is nothing wrong with that. The check itself is cheap and efficient. The real cost is the opening / closing of an empty bucket and the unnecessary gc thread wake ups and yields. If we are able to totally skip the gc thread wakeups for empty threads, the check on an empty bucket is acceptable. It can also help to make the interface simpler.

qinsoon:
> > Current code is awkward as we have all the buckets (general or plan-specific) defined in a global enum. This will get worse and worse when we need some buckets in certain conditions (like ClearVOBits).
> 
> > Why would the scheduler check if the Compact bucket should be opened or not when we run Immix GC?
> 
> Another simpler solution: You can also choose to define all buckets in a global enum, but only enabling / disabling a subset of them in each specific plan.
> 
> This is what I did to LXR. I enable different subsets of the buckets for different types of GC pauses: https://github.com/wenyuzhao/mmtk-core/blob/lxr/src/plan/lxr/global.rs#L852-L873

That is also a reasonable approach. We can define all the buckets in an enum, and let plan to choose which ones they would like to sue.

> > If a bucket is specific to a plan, it should not be even visible to other plans (only the scheduler and the plan itself need to know about the bucket).
> 
> I believe buckets should be general and shouldn't be private to a specific plan. This is just a workaround for gc thread synchronization. This shouldn't be private to any plan.

I am not sure if I understand you. For example, what is your proposal for mark compact? Mark compact currently uses 5 buckets that are only specific for mark compact. How do you propose to generalize the buckets?

qinsoon:
@wenyuzhao I don't quite understand your point of buckets being general. In my understanding, if we need to specify the order of work packets, for example, execute `A`, then `B`, then `C`, we just create 3 buckets, and put `A/B/C` to different buckets. In this case, buckets can hardly be general, and will be specific to plans, features, and options that are in use.

If you mean we can use other mechanisms to guarantee the order of work packets within the same bucket, like the [epilogue](https://github.com/mmtk/mmtk-core/blob/02b57a907e76fdfc65ed1c3b1bbf127d8e40638c/src/policy/immix/immixspace.rs#L844), or [sentinel](https://github.com/mmtk/mmtk-core/blob/02b57a907e76fdfc65ed1c3b1bbf127d8e40638c/src/scheduler/work_bucket.rs#L61), or other mechanisms that we should implement, yes, I would agree with that. With that, we can just have a few limited number of general buckets, and let the work packets to sort out their execution order. For example, for mark compact, we can just use one `Closure` bucket, we schedule work for the marking transitive closure. When that is done, we schedule work to `Closure` for calculating forwarding pointers. When that is done, we schedule work to `Closure` for a forwarding transitive closure. This sounds reasonable to me.

qinsoon:
@wenyuzhao I am waiting for your response on this PR

wenyuzhao:
> That is also a reasonable approach. We can define all the buckets in an enum, and let plan to choose which ones they would like to use.

I guess this is also the easier approach.

> In this case, buckets can hardly be general, and will be specific to plans, features, and options that are in use.

Right. Yes, private buckets are probably better if some of them can't be general. But actually, all work buckets are identical. It's just the names that are hard to generalize, not the buckets. Making each plan define its own `bucket name => index` mapping can also solve the problem.

In terms of work bucket generality, I think any changes or no change at the moment is fine. Our ultimate goal is to remove work buckets in the future.

qinsoon:
I closed this PR as both @wenyuzhao and @wks felt negative about this change. I would rather just wait for the rework.

PR #863, opened 2023-07-03
Use atomic and/or for marking and forwarding

wks:
Use atomic and/or operation instead of atomic store for marking and forwarding.

Atomic store is currently implemented with compare-exchange or fetch_update.  Atomic and/or operations may be much faster because they are unconditional and do not need a loop.

### DIFF
--- a/src/policy/largeobjectspace.rs
+++ b/src/policy/largeobjectspace.rs
@@ -251,37 +251,19 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
     /// succeeds, the method will return true, meaning the object is marked by this invocation.
     /// Otherwise, it returns false.
     fn test_and_mark(&self, object: ObjectReference, value: u8) -> bool {
-        loop {
-            let mask = if self.in_nursery_gc {
-                LOS_BIT_MASK
-            } else {
-                MARK_BIT
-            };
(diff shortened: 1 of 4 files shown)


wks:
I ran `lusearch` on mole.moma with 2.4x min heap size using several plans.  PGO is enabled, and the profile is collected from running all of those plans.  VO bit is disabled.

https://squirrel.anu.edu.au/plotty/wks/noproject/#0|mole-2023-07-03-Mon-051852&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

image

There is slight improvement in STW time for GenImmix, almost no effect for Immix and StickyImmix, but there is slowdown for MarkCompact and serious slow down for SemiSpace.  The strange thing is there is slowdown in mutator time for SemiSpace, too.

Not sure what caused this.  Maybe unconditionally storing to the metadata using atomic AND/OR caused more cache misses than the conditional cmpxchg which doesn't need to write if the old value is already expected.  Maybe PGO is adding uncertainty to the measurement.

Anyway, I'll not prioritise this PR for now.

wks:
The slowdown is likely a result of replacing test-and-test-and-set with a single atomic-or.  We see improvements when we do the opposite (https://github.com/mmtk/mmtk-core/pull/1438).

PR #864, opened 2023-07-04
Add ExternalPageResource and allow discontiguous VM space

qinsoon:
This PR changes `VMSpace` to multiple discontiguous regions as VM space. This PR is used in https://github.com/mmtk/mmtk-julia/pull/71 to support system and package images.
* add an `ExternalPageResource` to manage the discontiguous regions.
* implement `VMSpace` with `ExternalPageResource`. `VMSpace` no longer uses `ImmortalSpace`.
* rename `lazy_init_vm_space` to `set_vm_space`, as we allow setting vm space multiple times with non overlapping regions.


qinsoon:
Currently we cannot run with MSRV 1.61 at the moment. We have a dependency chain `mmtk->env_logger->is_terminal`, `is_terminal` updated to MSRV 1.63 in `0.4.8` and `env_logger` uses `is_terminal 0.4.0` (including any point versions). We can wait until they resolve the issue. Otherwise, we can bump our MSRV. https://github.com/rust-cli/env_logger/issues/273

qinsoon:
From https://github.com/rust-cli/env_logger/issues/273, the options for us are:
1. Bump our MSRV to 1.63 -- this will resolve the issue for now, but we may see the same problem in the future until cargo/Rust improves on version resolution for MSRV.
2. Check in our `Cargo.lock` and specify a version that is compatible with `MSRV 1.61`. This is not recommended by the current Rust documentation, but is in use by some Rust official projects (like `env_logger`) and there is an issue that suggest to make it recommended to check in lock file for libraries https://github.com/rust-lang/cargo/issues/8728.

wks:
We can also play our old trick of locking a dependency to a specific version until the problem is solved.  We once did this for `enum_map`.

```diff
diff --git a/Cargo.toml b/Cargo.toml
index db4da419d..6f07eb851 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ delegate = "0.9.0"
 downcast-rs = "1.1.1"
 enum-map = "2.4.2"
 env_logger = "0.10.0"
+is-terminal = "=0.4.7"  # latest version requires MSRV 1.63
 itertools = "0.10.5"
 jemalloc-sys = { version = "0.5.3", features = ["disable_initial_exec_tls"], optional = true }
 lazy_static = "1.1"
```

src/policy/vmspace.rs
```diff
@@ -112,118 +144,104 @@ impl<VM: VMBinding> crate::policy::gc_work::PolicyTraceObject<VM> for VMSpace<VM
...
 
-    fn create_space(
-        args: &mut CreateSpecificPlanArgs<VM>,
-        location: Option<(Address, usize)>,
-    ) -> ImmortalSpace<VM> {
-        use crate::util::heap::layout::vm_layout_constants::BYTES_IN_CHUNK;
+    pub fn add_external_pages(&self, start: Address, size: usize) {
+        let start = start.align_down(BYTES_IN_PAGE);
+        let end = (start + size).align_up(BYTES_IN_PAGE);
+        let size = end - start;
```
wks, line 176: When we calculate `end`, `start` has already been overwritten.  If size is a multiple of pages, `end` will not include the last page the original `start` and `size` covered.

Because there are variables named `chunk_start` and `chunk_end` below, I suggest prefixing the computed values with `page_`.

```suggestion
    pub fn add_external_pages(&self, start: Address, start: usize) {
        let page_start = start.align_down(BYTES_IN_PAGE);
        let page_end = (start + start).align_up(BYTES_IN_PAGE);
        let page_size = page_end - page_start;
```

src/policy/vmspace.rs
```diff
@@ -112,118 +144,104 @@ impl<VM: VMBinding> crate::policy::gc_work::PolicyTraceObject<VM> for VMSpace<VM
...
-            vm_space_start + vm_space_bytes,
-            vm_space_start_aligned,
-            vm_space_end_aligned
-        );
-
-        let space_args = args.get_space_args(
-            "vm_space",
-            false,
-            VMRequest::fixed(vm_space_start_aligned, vm_space_bytes_aligned),
+            start, end, chunk_start, chunk_end
```
wks, line 195: Maybe we should use `page_start` and `page_end` here, but `start` and `end` are still useful information to show.

src/policy/vmspace.rs
```diff
@@ -112,118 +144,104 @@ impl<VM: VMBinding> crate::policy::gc_work::PolicyTraceObject<VM> for VMSpace<VM
...
+        // self.common.vm_map.insert(chunk_start, chunk_size, self.common.descriptor);
+        // Update SFT
+        assert!(SFT_MAP.has_sft_entry(chunk_start), "The VM space start (aligned to {}) does not have a valid SFT entry. Possibly the address range is not in the address range we use.", chunk_start);
+        unsafe {
+            SFT_MAP.eager_initialize(self.as_sft(), chunk_start, chunk_size);
+        }
 
-    fn space(&self) -> &ImmortalSpace<VM> {
-        self.inner.as_ref().unwrap()
+        self.pr.add_new_external_pages(ExternalPages { start, end });
```
wks, line 213:
```suggestion
        self.pr.add_new_external_pages(ExternalPages { start: page_start, end: page_end });
```

wks (changes requested):

src/memory_manager.rs
```diff
@@ -86,9 +86,13 @@ pub fn mmtk_init<VM: VMBinding>(builder: &MMTKBuilder) -> Box<MMTK<VM>> {
     Box::new(mmtk)
 }
 
+/// Add an externally mmapped region to the VM space. A VM space can be set through MMTk options (`vm_space_start` and `vm_space_size`),
+/// and can also be set through this function call. A VM space can be discontiguous. This function can be called multiple times,
+/// and all the address ranges passed as arguments in the function will be considered as part of the VM space.
+/// Currently we do not allow remove regions from VM space.
```
k-sareen, line 92:
```suggestion
/// Currently we do not allow removing regions from VM space.
```

qinsoon:
# binding-refs
JULIA_BINDING_REF=update-julia-upstream-43bf2c8

qinsoon:
This PR is ready for review again. @wks

wks (approved):
LGTM

PR #866, opened 2023-07-14
Counting VM-allocated pages into heap size.

wks:
Some VMs allocate memory outside the MMTk heap, using `malloc` or other allocation methods.  Those memory can usually be reclaiming by the finalizers of dead object in the MMTk heap.  This commit allows the VM to report the amount of such off-heap memory so that MMTk can trigger GC more promptly to reclaim such memory.

to-do list:
-   [x] Ensure it works with dynamic heap size.

# Related work

The "**malloc_counted_size**" feature: Introduced in https://github.com/mmtk/mmtk-core/pull/608, that feature was intended for implementing malloc/free in MMTk's spaces, although the current implementation simply wraps the `malloc` function from libc.  This PR, on the other hand, is agnostic of how the VM allocates memory outside the MMTk heap.  This PR is useful if the VM already does similar accounting.  For example, Ruby accounts for memory allocated by `xmalloc`, a wrapper of `malloc`.

### DIFF
--- a/src/plan/global.rs
+++ b/src/plan/global.rs
@@ -259,6 +259,9 @@ pub trait Plan: 'static + Sync + Downcast {
         let used_pages = self.get_used_pages();
         let collection_reserve = self.get_collection_reserved_pages();
         let vm_live_bytes = <Self::VM as VMBinding>::VMCollection::vm_live_bytes();
+        // Note that `vm_live_bytes` may not be the exact number of bytes in whole pages.  The VM
+        // binding is allowed to return an approximate value if it is expensive or impossible to
+        // compute the exact number of pages occupied.
         let vm_live_pages = conversions::bytes_to_pages_up(vm_live_bytes);
         let total = used_pages + collection_reserve + vm_live_pages;
 
--- a/src/plan/global.rs
+++ b/src/plan/global.rs
@@ -252,10 +252,23 @@ pub trait Plan: 'static + Sync + Downcast {
     // work fine for non-copying plans. For copying plans, the plan should override any of these methods
     // if necessary.
 
-    /// Get the number of pages that are reserved, including used pages and pages that will
-    /// be used (e.g. for copying).
+    /// Get the number of pages that are reserved, including pages used by MMTk spaces, pages that
+    /// will be used (e.g. for copying), and VM-allocated pages as reported by the VM binding.
     fn get_reserved_pages(&self) -> usize {
-        self.get_used_pages() + self.get_collection_reserved_pages()
+        let used_pages = self.get_used_pages();
+        let collection_reserve = self.get_collection_reserved_pages();
+        let vm_allocated_pages = <Self::VM as VMBinding>::VMCollection::vm_allocated_pages();
+        let total = used_pages + collection_reserve + vm_allocated_pages;
+
+        trace!(
+            "Reserved pages = {}, used pages: {}, collection reserve: {}, VM-allocated pages: {}",
+            total,
+            used_pages,
+            collection_reserve,
+            vm_allocated_pages,
+        );
+
+        total
     }
 
     /// Get the total number of pages for the heap.
@@ -266,6 +279,9 @@ pub trait Plan: 'static + Sync + Downcast {
     /// Get the number of pages that are still available for use. The available pages
     /// should always be positive or 0.
     fn get_available_pages(&self) -> usize {
+        let reserved_pages = self.get_reserved_pages();
+        let total_pages = self.get_total_pages();
+
         // It is possible that the reserved pages is larger than the total pages so we are doing
         // a saturating subtraction to make sure we return a non-negative number.
         // For example,
@@ -274,15 +290,14 @@ pub trait Plan: 'static + Sync + Downcast {
         //    the reserved pages is larger than total pages after the copying GC (the reserved pages after a GC
         //    may be larger than the reserved pages before a GC, as we may end up using more memory for thread local
         //    buffers for copy allocators).
+        let available_pages = total_pages.saturating_sub(reserved_pages);
         trace!(
             "Total pages = {}, reserved pages = {}, available pages = {}",
-            self.get_total_pages(),
-            self.get_reserved_pages(),
-            self.get_reserved_pages()
-                .saturating_sub(self.get_reserved_pages())
+            total_pages,
+            reserved_pages,
+            available_pages,
         );
-        self.get_total_pages()
-            .saturating_sub(self.get_reserved_pages())
+        available_pages
     }
 
     /// Get the number of pages that are reserved for collection. By default, we return 0.
--- a/src/plan/global.rs
+++ b/src/plan/global.rs
@@ -25,7 +25,7 @@ use crate::util::metadata::side_metadata::SideMetadataSpec;
 use crate::util::options::Options;
 use crate::util::options::PlanSelector;
 use crate::util::statistics::stats::Stats;
-use crate::util::ObjectReference;
+use crate::util::{conversions, ObjectReference};
 use crate::util::{VMMutatorThread, VMWorkerThread};
 use crate::vm::*;
 use downcast_rs::Downcast;
@@ -253,19 +253,21 @@ pub trait Plan: 'static + Sync + Downcast {
     // if necessary.
 
     /// Get the number of pages that are reserved, including pages used by MMTk spaces, pages that
-    /// will be used (e.g. for copying), and VM-allocated pages as reported by the VM binding.
+    /// will be used (e.g. for copying), and live pages allocated by the VM as reported by the VM
+    /// binding.
     fn get_reserved_pages(&self) -> usize {
         let used_pages = self.get_used_pages();
         let collection_reserve = self.get_collection_reserved_pages();
-        let vm_allocated_pages = <Self::VM as VMBinding>::VMCollection::vm_allocated_pages();
-        let total = used_pages + collection_reserve + vm_allocated_pages;
+        let vm_live_bytes = <Self::VM as VMBinding>::VMCollection::vm_live_bytes();
+        let vm_live_pages = conversions::bytes_to_pages_up(vm_live_bytes);
+        let total = used_pages + collection_reserve + vm_live_pages;
 
         trace!(
-            "Reserved pages = {}, used pages: {}, collection reserve: {}, VM-allocated pages: {}",
+            "Reserved pages = {}, used pages: {}, collection reserve: {}, VM live pages: {}",
             total,
             used_pages,
             collection_reserve,
-            vm_allocated_pages,
+            vm_live_pages,
         );
 
         total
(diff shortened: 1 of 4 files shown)


wks:
MSRV-related tests are failing, but we already worked around that issue previously.

src/vm/collection.rs
```diff
@@ -112,28 +112,43 @@ pub trait Collection<VM: VMBinding> {
...
     fn post_forwarding(_tls: VMWorkerThread) {}
 
-    /// Return the amount of memory (in pages) the VM allocated off-heap which could be released
-    /// by doing GC.  MMTk core will trigger a GC if the amount of allocated memory in MMTk spaces
-    /// plus this amount exceeds a certain limit.
+    /// Return the amount of memory (in bytes) the VM allocated (but not released yet) outside the
+    /// MMTk heap which can be released by doing GC.
+    ///
+    /// MMTk core will add this amount to the amount of memory allocated or reserved in MMTk
+    /// spaces, and will trigger a GC if the sum exceeds a certain limit.
```
qinsoon, line 119: How to trigger a GC depends on the internal GC trigger code. You could simply say that MMTk core will consider the reported memory as part of MMTk heap for the purpose of heap size accounting.

src/vm/collection.rs
```diff
@@ -112,28 +112,43 @@ pub trait Collection<VM: VMBinding> {
     /// * `tls_worker`: The thread pointer for the worker thread performing this call.
     fn post_forwarding(_tls: VMWorkerThread) {}
 
-    /// Return the amount of memory (in pages) the VM allocated off-heap which could be released
-    /// by doing GC.  MMTk core will trigger a GC if the amount of allocated memory in MMTk spaces
-    /// plus this amount exceeds a certain limit.
+    /// Return the amount of memory (in bytes) the VM allocated (but not released yet) outside the
+    /// MMTk heap which can be released by doing GC.
```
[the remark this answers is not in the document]
wks: Yes.  "Kept alive by heap objects" sounds better.

src/vm/collection.rs
```diff
@@ -112,28 +112,43 @@ pub trait Collection<VM: VMBinding> {
...
+    ///
+    /// MMTk does not specify what memory this amount should includes.  However, because this
+    /// amount is used to trigger GC, it is advisable to include memory that can be released by
+    /// executing finalizers or other language-specific cleaning-up routines that are executed when
+    /// an in-heap object is dead.  For example, if a language implementation allocates array
+    /// headers in the MMTk heap, but allocates their underlying buffers that hold the actual
+    /// elements using `malloc`, then those buffers should be included in this amount.  When the GC
+    /// finds such an array dead, its finalizer shall `free` the buffer and reduce this amount.
+    ///
+    /// This function is a hint for the MMTk core to trigger GC, therefore does not have to be
```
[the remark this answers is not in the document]
wks: But the "min heap size" is an empirical value, too. It depends on the plan, the number of CPUs (which determines the number of GC threads) and other specifics of the machine the program is running on.  I think it is OK to say it is a hint, given that the exact set of "objects kept alive by heap objects" is not precisely defined.  Another important point is that the VM should not spend too much time computing the exact value because this function is called very often (by `gc_poll`).  An approximate value should be good enough for MMTk to trigger GC promptly and keep the memory footprint under control.

src/vm/collection.rs
```diff
@@ -112,28 +112,43 @@ pub trait Collection<VM: VMBinding> {
...
+    /// pages occupied by off-heap objects, and report the number of bytes of those whole pages
+    /// instead of individual objects.  Because the underlying operating system manages memory at
+    /// page granularity, the occupied pages (instead of individual objects) determine the memory
+    /// footprint of a process, and how much memory MMTk spaces can obtain from the OS.
+    ///
+    /// However, if the VM is incapable of accouting off-heap memory by pages (for example, if the
+    /// VM uses `malloc` and the implementation of `malloc` is opaque to the VM), the VM binding
+    /// can simply return the total number of bytes of those off-heap objects as an approximation.
+    ///
+    /// The default implementation which returns 0 should also work if the VM only allocates a
```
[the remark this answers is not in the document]
wks: Yes. That's reasonable.

src/vm/collection.rs
wks: Yes. It is the GC trigger that determines when to trigger GC, but I want the VM binding developer to know that the return value of this function is additive (rather than multiplicative or having other non-linear relation) to MMTk core's "used pages" when considered to trigger GC, so that the developer can have a basic idea of what effect the return value of this function will have on triggering GC.  At least the current implementation is additive for both fixed heap size and the MemBalancer.  The MemBalancer changes the heap limit dynamically, but the way to compute occupied pages is still adding the used pages of MMTk spaces and the result of this function.  I can emphasise that being additive is only a detail of the current implementation.

src/vm/collection.rs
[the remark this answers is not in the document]
qinsoon: I see what you mean. The value does not need to be precise for the sake of performance. But I woudn't say it is a 'hint'. A 'hint' means something like an suggestion. Like Java's `System.gc()` is a hint, we may do or may not do a GC. Julia's `--heap-size-hint` is a hint, the actual heap size can be larger than the given value. But in our case, the number returned by the method is not a hint, and we do count it into our heap size and we will always count it -- that is the contract of this method, the memory is counted into our heap size.

You have another paragraph talking about performance. If you would like to emphasis further on the tradeoff between performance and being precise, you can expand that section.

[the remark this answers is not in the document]
wks: Sounds good.  I'll remove the "hint" part and merge it with the paragraph about performance.

src/vm/collection.rs
qinsoon: > I want the VM binding developer to know that the return value of this function is additive (rather than multiplicative or having other non-linear relation) to MMTk core's "used pages" when considered to trigger GC

I think this is incorrect. GC triggering does not necessarily relate to heap sizes. We could implement a GC trigger that uses allocation volume -- in which case, we do not directly use the vm allocated pages returned by the method, we may use a diff of value or simply ignore it. Sticky immix does not trigger nursery GC based on heap sizes, LXR has similar behaviors as well.
https://github.com/mmtk/mmtk-core/blob/40c977308cb24d8bdf46b82916a388c10fa956e7/src/plan/sticky/immix/global.rs#L153

This is what I suggest about the contract of this method.
> MMTk core will consider the reported memory as part of MMTk heap for the purpose of heap size accounting.

But we do not guarantee that GC triggering is related with heap size.

wks: I removed the second paragraph and actually give recommendations for VMs about what to include instead of saying "MMTk doesn't specify what to include except that number is considered by the GC trigger".  In this way, newer GC triggers can decide whether to take that reported number into consideration when triggering GC.

qinsoon (approved):
LGTM

src/vm/collection.rs
```diff
@@ -111,4 +111,38 @@ pub trait Collection<VM: VMBinding> {
...
+    /// memory as part of MMTk heap for the purpose of heap size accounting.
+    ///
+    /// This amount should include memory that is kept alive by heap objects and can be released by
+    /// executing finalizers (or other language-specific cleaning-up routines) that are executed
+    /// when the heap objects are dead.  For example, if a language implementation allocates array
+    /// headers in the MMTk heap, but allocates their underlying buffers that hold the actual
+    /// elements using `malloc`, then those buffers should be included in this amount.  When the GC
+    /// finds such an array dead, its finalizer shall `free` the buffer and reduce this amount.
+    ///
+    /// If possible, the VM should account off-heap memory by pages.  That is, count the number of
```
k-sareen, line 126:
```suggestion
    /// If possible, the VM should account off-heap memory in pages.  That is, count the number of
```

src/vm/collection.rs
```diff
@@ -111,4 +111,38 @@ pub trait Collection<VM: VMBinding> {
...
+    /// pages occupied by off-heap objects, and report the number of bytes of those whole pages
+    /// instead of individual objects.  Because the underlying operating system manages memory at
+    /// page granularity, the occupied pages (instead of individual objects) determine the memory
+    /// footprint of a process, and how much memory MMTk spaces can obtain from the OS.
+    ///
+    /// However, if the VM is incapable of accouting off-heap memory by pages (for example, if the
```
k-sareen, line 132:
```suggestion
    /// However, if the VM is incapable of accounting off-heap memory in pages (for example, if the
```

src/vm/collection.rs
```diff
@@ -111,4 +111,38 @@ pub trait Collection<VM: VMBinding> {
...
+    /// VM uses `malloc` and the implementation of `malloc` is opaque to the VM), the VM binding
+    /// can simply return the total number of bytes of those off-heap objects as an approximation.
+    ///
+    /// # Performance note
+    ///
+    /// This function will be called when MMTk polls for GC.  It happens every time the MMTk
+    /// allocators allocated a certain amount of memory, usually one or a few blocks.  Because this
```
k-sareen, line 139:
```suggestion
    /// allocators have allocated a certain amount of memory, usually one or a few blocks.  Because this
```

src/plan/global.rs
```diff
@@ -252,10 +252,25 @@ pub trait Plan: 'static + Sync + Downcast {
...
-    /// be used (e.g. for copying).
+    /// Get the number of pages that are reserved, including pages used by MMTk spaces, pages that
+    /// will be used (e.g. for copying), and live pages allocated by the VM as reported by the VM
+    /// binding.
     fn get_reserved_pages(&self) -> usize {
-        self.get_used_pages() + self.get_collection_reserved_pages()
+        let used_pages = self.get_used_pages();
+        let collection_reserve = self.get_collection_reserved_pages();
+        let vm_live_bytes = <Self::VM as VMBinding>::VMCollection::vm_live_bytes();
+        let vm_live_pages = conversions::bytes_to_pages_up(vm_live_bytes);
```
k-sareen, line 262: Could you add a one line comment here just saying that the live bytes may not be actual pages?

src/vm/collection.rs
k-sareen: Actually, maybe "in terms of pages used" is more appropriate. It is more verbose though. Maybe "as pages used" instead? Same for comment below

src/plan/global.rs
wks: Yes. That's reasonable because the live bytes may be occupied bytes instead of actual pages, and may sometimes be an approximate value.

wks:
@k-sareen I have made changes according to your suggestions.

src/plan/global.rs
```diff
@@ -252,10 +252,28 @@ pub trait Plan: 'static + Sync + Downcast {
     // work fine for non-copying plans. For copying plans, the plan should override any of these methods
     // if necessary.
 
-    /// Get the number of pages that are reserved, including used pages and pages that will
-    /// be used (e.g. for copying).
+    /// Get the number of pages that are reserved, including pages used by MMTk spaces, pages that
+    /// will be used (e.g. for copying), and live pages allocated by the VM as reported by the VM
```
k-sareen, line 256: It's not live pages allocated by the VM necessarily. Maybe slightly rephrase it to say something like "We also take into account the live bytes as reported by the VM. For more details, refer to the [`vm_live_bytes`] function."

wks: I rephrased it a bit.

PR #868, opened 2023-07-20
Expose posix_memalign. Rename counted free/realloc

qinsoon:


qinsoon:
I am wondering if we still need the 'counted' interface. With https://github.com/mmtk/mmtk-core/pull/866, we may be able to completely remove the current 'counted malloc' methods.

wks:
> I am wondering if we still need the 'counted' interface. With #866, we may be able to completely remove the current 'counted malloc' methods.

I think it is still part of our plan to provide an implementation of `malloc` using MMTk's allocators and collectors.  So we probably still need those interfaces.  But whether the interface should remain in its current form is a different question.

And its implementation needs to be improved, too.  Currently it wraps the `malloc` from naive libraries (such as libc) and we account the memory by bytes.  But currently I don't see any urgent need (at least for the Ruby binding) for an implementation of `malloc` provided by the MMTk core.

We may need to address issues such as providing an implicit global instance of MMTK (or just the necessary part) so that we can replace the global `malloc` function (and probably allow using `LD_PRELOAD` to replace the `malloc` implementation).  We may also need to allow per-MMTk `malloc` implementation so that the user can `malloc` into a given MMTk instance (which is like the current API).

qinsoon:
We need the malloc interface. But we do not necessarily need the 'counted malloc' interface. I will see if I can change the Julia binding to use the `vm_live_bytes` to replace the current use of 'counted malloc'.

qinsoon:
Closing this PR. The julia binding is not using the counted malloc APIs (https://github.com/mmtk/mmtk-julia/pull/80), and we do not need to add more. We may consider removing those APIs.

Issue #871, opened 2023-07-28
MMTk should signal an error in release mode for incorrect metadata placement


playXE:
Right now if you build binary in release mode it won't signal an error but just continue to run like it's not a problem, on debug builds error does happen: 
```rust
error[E0080]: evaluation of constant value failed
   --> /home/adel/.cargo/git/checkouts/mmtk-core-3306bdeb8eb4322b/47ee126/src/vm/object_model.rs:559:5
    |
559 |     define_vm_metadata_spec!(VMLocalForwardingBitsSpec, false, 1, LOG_MIN_OBJECT_SIZE);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'assertion failed: side_spec.is_global == Self::IS_GLOBAL', /home/adel/.cargo/git/checkouts/mmtk-core-3306bdeb8eb4322b/47ee126/src/vm/object_model.rs:559:5
    |

```

PR #873, opened 2023-07-31
Replace debug_assert in side_after with assert

qinsoon:
Closes https://github.com/mmtk/mmtk-core/issues/871


k-sareen (approved):
LGTM. Thanks!

Issue #496, opened 2021-11-08
Clarify stack root scanning in API


qinsoon:
If `VMScanning::SCAN_MUTATORS_IN_SAFEPOINT` is false, mmtk-core does not schedule any stack root scanning work (`ScanStackRoots` or `ScanStackRoot`). MMTk expects each thread to scan its own stack once the thread is stopped, thus MMTk does not schedule any stack root scanning packet. However, this is not documented anywhere.

k-sareen:
See some further discussion here: https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/SCAN_MUTATORS_IN_SAFEPOINT/near/364428734

@wks Suggested making the boolean into an `enum` to clarify intent and add more documentation.

wks:
For example, we can change it to

```rs
/// When do we scan stacks
enum StackScanStrategy {
    /// Create stack-scanning work packets after all mutators stopped.
    AfterStopMutators,
    /// Create a stack-scanning work packet for a mutator as soon as the
    /// `mutator_visitor` callback of `Collection::stop_all_mutators` is called.
    /// The VM binding may choose not to call back, in which case mmtk-core
    /// will not create any stack-scanning work packets.  It is useful if the VM
    /// prefers scanning stacks in its own way during `stop_all_mutators`
    /// (i.e. not using the `ScanStackRoot{,s}` work packets), or
    /// scanning stacks during `Scanning::scan_vm_specific_roots`.
    OnCallback,
}
```

k-sareen:
Maybe rename the `OnCallback` to `HandledByVm` or some other better name given as you point out in the comments that the callback may never actually be used if the VM scans stacks in the `scan_vm_specific_roots` function.

wks:
> ... each thread to scan its own stack once the thread is stopped...

Currently each time the callback `mutator_visitor: impl FnMut(&'static mut Mutator<VM>)` is called, mmtk-core crates a `ScanStackRoot` work packet for the mutator given.  But the `ScanStackRoot` work packet is added to the `Prepare` bucket which is not open until `stop_all_mutators` returns and `mmtk.scheduler.notify_mutators_paused` is called.  This makes it impossible to achieve the desired effect.

> Maybe rename the `OnCallback` to `HandledByVm` or some other better name given as you point out in the comments that the callback may never actually be used if the VM scans stacks in the `scan_vm_specific_roots` function.

[...]

PR #875, opened 2023-07-31
Tidy up mutator scan API

qinsoon:
This PR closes https://github.com/mmtk/mmtk-core/issues/496.
* Remove Scanning::SCAN_MUTATORS_IN_SAFEPOINT, and Scanning::SINGLE_THREAD_MUTATOR_SCANNING.
* Remove Scanning::scan_roots_in_all_mutator_threads.
* Remove Collection::prepare_mutator
* Update comments for a few methods

### DIFF
--- a/src/scheduler/gc_work.rs
+++ b/src/scheduler/gc_work.rs
@@ -165,8 +165,6 @@ impl<VM: VMBinding> GCWork<VM> for ReleaseCollector {
 
 /// Stop all mutators
 ///
-/// Schedule a `ScanStackRoots` immediately after a mutator is paused
-///
 /// TODO: Smaller work granularity
 #[derive(Default)]
 pub struct StopMutators<ScanEdges: ProcessEdgesWork>(PhantomData<ScanEdges>);
@@ -184,7 +182,8 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for StopMutators<E> {
         <E::VM as VMBinding>::VMCollection::stop_all_mutators(worker.tls, |mutator| {
             // TODO: The stack scanning work won't start immediately, as the `Prepare` bucket is not opened yet (the bucket is opened in notify_mutators_paused).
             // Should we push to Unconstrained instead?
-            mmtk.scheduler.work_buckets[WorkBucketStage::Prepare].add(ScanStackRoot::<E>(mutator));
+            mmtk.scheduler.work_buckets[WorkBucketStage::Prepare]
+                .add(ScanMutatorRoots::<E>(mutator));
         });
         trace!("stop_all_mutators end");
         mmtk.scheduler.notify_mutators_paused(mmtk);
@@ -410,11 +409,11 @@ impl<VM: VMBinding> GCWork<VM> for VMPostForwarding<VM> {
     }
 }
 
-pub struct ScanStackRoot<Edges: ProcessEdgesWork>(pub &'static mut Mutator<Edges::VM>);
+pub struct ScanMutatorRoots<Edges: ProcessEdgesWork>(pub &'static mut Mutator<Edges::VM>);
 
-impl<E: ProcessEdgesWork> GCWork<E::VM> for ScanStackRoot<E> {
+impl<E: ProcessEdgesWork> GCWork<E::VM> for ScanMutatorRoots<E> {
     fn do_work(&mut self, worker: &mut GCWorker<E::VM>, mmtk: &'static MMTK<E::VM>) {
-        trace!("ScanStackRoot for mutator {:?}", self.0.get_tls());
+        trace!("ScanMutatorRoots for mutator {:?}", self.0.get_tls());
         let base = &mmtk.plan.base();
         let mutators = <E::VM as VMBinding>::VMActivePlan::number_of_mutators();
         let factory = ProcessEdgesWorkRootsWorkFactory::<E>::new(mmtk);
--- a/src/scheduler/gc_work.rs
+++ b/src/scheduler/gc_work.rs
@@ -182,33 +182,12 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for StopMutators<E> {
         trace!("stop_all_mutators start");
         mmtk.plan.base().prepare_for_stack_scanning();
         <E::VM as VMBinding>::VMCollection::stop_all_mutators(worker.tls, |mutator| {
+            // TODO: The stack scanning work won't start immediately, as the `Prepare` bucket is not opened yet (the bucket is opened in notify_mutators_paused).
+            // Should we push to Unconstrained instead?
             mmtk.scheduler.work_buckets[WorkBucketStage::Prepare].add(ScanStackRoot::<E>(mutator));
         });
         trace!("stop_all_mutators end");
         mmtk.scheduler.notify_mutators_paused(mmtk);
-        if <E::VM as VMBinding>::VMScanning::SCAN_MUTATORS_IN_SAFEPOINT {
-            // Prepare mutators if necessary
-            // FIXME: This test is probably redundant. JikesRVM requires to call `prepare_mutator` once after mutators are paused
-            if !mmtk.plan.base().stacks_prepared() {
-                for mutator in <E::VM as VMBinding>::VMActivePlan::mutators() {
-                    <E::VM as VMBinding>::VMCollection::prepare_mutator(
-                        worker.tls,
-                        mutator.get_tls(),
-                        mutator,
-                    );
-                }
-            }
-            // Scan mutators
-            if <E::VM as VMBinding>::VMScanning::SINGLE_THREAD_MUTATOR_SCANNING {
-                mmtk.scheduler.work_buckets[WorkBucketStage::Prepare]
-                    .add(ScanStackRoots::<E>::new());
-            } else {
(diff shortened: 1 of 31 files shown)


k-sareen:
I also just realized that the GC state is never updated to `GcProper` in case the binding does not use the callback and does not implement `scan_roots_in_mutator_thread` , i.e. the case where the binding prefers not to or cannot differentiate stack roots from other kinds of roots. It's not a big deal, but thought I'd mention it

qinsoon:
> I also just realized that the GC state is never updated to `GcProper` in case the binding does not use the callback and does not implement `scan_roots_in_mutator_thread` , i.e. the case where the binding prefers not to or cannot differentiate stack roots from other kinds of roots. It's not a big deal, but thought I'd mention it

Right. We should require the binding to use the callback. Bindings should use the callback, but do not have to implement `scan_roots_in_mutator_thread`.

k-sareen:
That feels like a bit of a waste though? We'll end up scheduling empty `ScanStackRoot` work packets

k-sareen:
But I also realized that if the binding doesn't execute the `ScanStackRoot` work packets, then the remembered sets for mutators won't be flushed

k-sareen:
Perhaps we should just flush the mutators in `prepare_mutator` instead of doing it in the `ScanStackRoot` work packet

src/vm/collection.rs
```diff
@@ -16,9 +15,12 @@ pub trait Collection<VM: VMBinding> {
     /// This method is called by a single thread in MMTk (the GC controller).
     /// This method should not return until all the threads are yielded.
     /// The actual thread synchronization mechanism is up to the VM, and MMTk does not make assumptions on that.
+    /// MMTk provide a callback function and expects the binding to use the callback for each mutator when it
```
k-sareen, line 18:
```suggestion
    /// MMTk provides a callback function and expects the binding to use the callback for each mutator when it
```

k-sareen (commented):
My understanding of the discussion was that if the binding/VM can't tell us when its threads are ready for scanning, then we should still provide a way to flush each mutator. Or is that the binding's job? If it is, then it should be documented.

qinsoon:
> My understanding of the discussion was that if the binding/VM can't tell us when its threads are ready for scanning, then we should still provide a way to flush each mutator. Or is that the binding's job? If it is, then it should be documented.

I think we conclude that a binding will tell us when the thread is ready to be scanned. It can tell us as soon as a thread is stopped, or after all the threads are stopped. But the binding has to tell us that. 

Flushing mutator is not a binding's job. MMTk will do that. We require a binding to tell us if a thread can be scanned, and we schedule `ScanStackRoot` for the thread. In `ScanStackRoot`, we ask the binding for stack roots, and also flush mutators.

k-sareen:
Right. I'm getting hung up on the case where the binding/VM doesn't differentiate between thread roots and other kinds of roots. In that case the binding will just dump all roots into the `scan_vm_specific_roots` function. But that won't flush mutators unless the binding also creates empty `ScanStackRoot` work packets.

Also can we rename the `ScanStackRoot` work packet while we're at it? It's ambiguous. Renaming to `ScanStackRoots` (given that we're removing the `ScanStackRoots` work packet) or `ScanStackRootsForMutator` would be better imo.

src/scheduler/gc_work.rs
```diff
@@ -182,33 +182,12 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for StopMutators<E> {
         trace!("stop_all_mutators start");
         mmtk.plan.base().prepare_for_stack_scanning();
         <E::VM as VMBinding>::VMCollection::stop_all_mutators(worker.tls, |mutator| {
+            // TODO: The stack scanning work won't start immediately, as the `Prepare` bucket is not opened yet (the bucket is opened in notify_mutators_paused).
+            // Should we push to Unconstrained instead?
```
wks, line 186: Someone may want some work to be done before scanning any stacks, but currently there is no such things.  For Ruby, "potential pinning parents" (PPP) needs to be handled before transitive closure, but can be done at the same time as stack scanning.

We also had some issue letting OpenJDK report the yielding of individual threads.  See https://github.com/mmtk/openjdk/pull/9  There used to be something called `report_java_thread_yield`, but we removed that.  So the current status is that OpenJDK can only report that all threads have come to a stop.

I think we can leave the code as is for now.  After we figure out how to report the yielding of each thread (for OpenJDK 11, 17 and 22), we can make another attempt and move the `ScanStackRoot` work packet to `Unconstrained` or another bucket.

k-sareen: There is the `prepare_mutator` function for VM-specific behaviour. But this PR would remove that.

qinsoon: > Someone may want some work to be done before scanning any stacks, but currently there is no such things. For Ruby, "potential pinning parents" (PPP) needs to be handled before transitive closure, but can be done at the same time as stack scanning.

The binding can do whatever work is needed before it tells MMTk that the thread is ready.

qinsoon: > There is the `prepare_mutator` function for VM-specific behaviour. But this PR would remove that.

Yeah. I think it is confusing about when we should invoke it. For the old timing where we call `prepare_mutator`, the binding can do it before returning `stop_all_mutators`. JikesRVM is the only binding we know that uses the method. I changed it to https://github.com/mmtk/mmtk-jikesrvm/pull/147/files#diff-54f4f04a18788baf9bf66cff8c4256151b680087e3490fb8e8f62e2d60ad9fabR30.

wks:
> Also can we rename the `ScanStackRoot` work packet while we're at it? It's ambiguous. Renaming to `ScanStackRoots` (given that we're removing the `ScanStackRoots` work packet) or `ScanStackRootsForMutator` would be better imo.

I would suggest simply calling it `ScanMutatorRoots`.  It's not only stack roots, but may include other thread-local states as well.  For example, for OpenJDK, it calls `JavaThread::oops_do` which scans not only Java frames, but also thread-local JNI handles, pending exceptions, monitors, etc.

src/vm/collection.rs
```diff
@@ -16,9 +15,12 @@ pub trait Collection<VM: VMBinding> {
...
+    /// is ready for stack scanning. Usually a stack can be scanned as soon as the thread stops in the yieldpoint.
     ///
     /// Arguments:
     /// * `tls`: The thread pointer for the GC controller/coordinator.
+    /// * `mutator_visitor`: A callback for each mutator to notify MMTk when a mutator is ready to be stack scanned.
```
wks, line 23: In English, "... for x to y" means "we do ... so that x can do y", and x becomes the subject.  This comment means "a callback so that each mutator can notify MMTk that it is ready".  But it may not always be the mutator that calls this callback (That would require `F` to implement `Send` and/or `Sync`).  It can be the current GC thread, too.  The GC thread can wait for some condition variable and notify MMTk on behalf of the mutator.

```suggestion
    /// * `mutator_visitor`: A callback.  Call it with a mutator as argument to notify MMTk that the mutator is ready to be scanned.
```

qinsoon: Done.

qinsoon:
> > Also can we rename the `ScanStackRoot` work packet while we're at it? It's ambiguous. Renaming to `ScanStackRoots` (given that we're removing the `ScanStackRoots` work packet) or `ScanStackRootsForMutator` would be better imo.
> 
> I would suggest simply calling it `ScanMutatorRoots`. It's not only stack roots, but may include other thread-local states as well. For example, for OpenJDK, it calls `JavaThread::oops_do` which scans not only Java frames, but also thread-local JNI handles, pending exceptions, monitors, etc.

I have renamed the packet to `ScanMutatorRoots`.

src/vm/scanning.rs
```diff
@@ -203,20 +195,12 @@ pub trait Scanning<VM: VMBinding> {
...
-    /// * `factory`: The VM uses it to create work packets for scanning roots.
-    fn scan_roots_in_all_mutator_threads(
-        tls: VMWorkerThread,
-        factory: impl RootsWorkFactory<VM::VMEdge>,
-    );
-
-    /// Scan one mutator for roots.
+    /// Scan one mutator for stack roots. MMTk assumes a binding is able to scan
+    /// any given thread for its stack roots. If a binding cannot scan the stack for
+    /// a given thread, it can leave this method empty, and deal with stack
```
wks, line 200: The word "given" (as an adjective) is a bit ambiguous.  In mathematics, "given" refers to conditions that are known to be true.  So "a binding cannot scan the stack for a given thread" can be interpreted as "there exists one thread in the VM that the binding cannot scan (but it can scan other threads)".  The binding must be able to scan all threads (stacks). I think this comment is contrasting two cases: (1) Whenever MMTk gives the binding a thread reference, the binding can scan it; and (2) The binding can only enumerate threads by itself, and scan them while enumerating.

```suggestion
    /// Scan one mutator for stack roots.
    ///
    /// Some VM bindings may not be able to implement this method.
    /// For example, the VM binding may only be able to enumerate all threads and
    /// scan them while enumerating, but cannot scan stacks individually when given
    /// the references of threads.
    /// In that case, it can leave this method empty, and deal with stack
```

qinsoon: Sounds good.

wks (approved):
LGTM.  But bindings need to adapt to this change, too.

Issue #874, opened 2023-07-31
PR #849 broke JikesRVM RBaseBaseSemiSpace


qinsoon:
Running https://github.com/mmtk/mmtk-core/pull/849 with the latest JikesRVM binding (4c47fdf904cbb99362f5d1ffa2ecea28150ed6ad). I got a sigsegv when running `RBaseBaseSemiSpace`:

Build
```
./bin/buildit localhost RBaseBaseSemiSpace --answer-yes --use-third-party-heap=../mmtk-jikesrvm --use-third-party-build-configs=../mmtk-jikesrvm/jikesrvm/build/configs/ --use-external-source=../mmtk-jikesrvm/jikesrvm/rvm/src --m32
```
Run
```
$ dist/RBaseBaseSemiSpace_x86_64_m32-linux/rvm -jar /usr/share/benchmarks/dacapo/dacapo-2006-10
-MR2.jar fop
[2023-07-31T02:45:35Z INFO  mmtk::memory_manager] Initialized MMTk with SemiSpace (FixedHeapSize(104857600))
Segmentation fault (core dumped)
```

The CI tests did not catch this, because of a mistake in JikesRVM `ci-test.sh` (no `set -e` in the script).

qinsoon:
BaseBase Marksweep and NoGC also segfaults.

wks:
BaseBase Semispace panicked because of "VO bit not set" on copyspace.rs:222:9.

I guess the reason is that the allocation fast path did not set the VO bit.  OpenJDK's JIT compiler generates the allocation code specially if VO_BIT=1.

qinsoon:
But JikesRVM should not enable `vo_bit` for any of its build.

wks:
> But JikesRVM should not enable `vo_bit` for any of its build.

Oops.  I thought you mean you enabled VO bits for JikesRVM and found it crashing.  I'll test again without the VO bits.

PR #876, opened 2023-07-31
Work around stack overflow in array_from_fn

wks:
It should fix a crash in JikesRVM caused by overflown stack overwriting objects in the VM space.

Fixes: https://github.com/mmtk/mmtk-core/issues/874


qinsoon:
binding-refs
JIKESRVM_BINDING_REF=set-e-in-ci-test

qinsoon (approved):
LGTM. Thanks for the fix.

PR #879, opened 2023-08-01
Fix UB in SFTMap implementations

playXE:
1) Replaces casts of references to mutable references with `UnsafeCell<T>`. 
2) Spaces are stored inside each of implementation as pointers in order to make Rust compiler happy because of lifetimes.

### DIFF
--- a/src/policy/sft_map.rs
+++ b/src/policy/sft_map.rs
@@ -136,7 +136,7 @@ mod space_map {
             }
 
             //*mut_self.sft.get_unchecked_mut(index) = space;
-            *(*self.sft.get()).get_unchecked_mut(index) = unsafe { std::mem::transmute(space) };
+            *(*self.sft.get()).get_unchecked_mut(index) = std::mem::transmute(space);
         }
 
         unsafe fn clear(&self, addr: Address) {
--- a/src/policy/sft_map.rs
+++ b/src/policy/sft_map.rs
@@ -64,11 +64,11 @@ pub(crate) fn create_sft_map() -> Box<dyn SFTMap> {
     cfg_if::cfg_if! {
         if #[cfg(all(feature = "malloc_mark_sweep", target_pointer_width = "64"))] {
             // 64-bit malloc mark sweep needs a chunk-based SFT map, but the sparse map is not suitable for 64bits.
-            Box::new(dense_chunk_map::SFTDenseChunkMap::<'static>::new())
+            Box::new(dense_chunk_map::SFTDenseChunkMap::new())
         } else if #[cfg(target_pointer_width = "64")] {
             Box::new(space_map::SFTSpaceMap::new())
         } else if #[cfg(target_pointer_width = "32")] {
-            Box::new(sparse_chunk_map::SFTSparseChunkMap::<'static>::new())
+            Box::new(sparse_chunk_map::SFTSparseChunkMap::new())
         } else {
             compile_err!("Cannot figure out which SFT map to use.");
         }
--- a/src/policy/sft_map.rs
+++ b/src/policy/sft_map.rs
@@ -34,7 +34,7 @@ pub trait SFTMap {
     /// # Safety
     /// The address must have a valid SFT entry in the map. Usually we know this if the address is from an object reference, or from our space address range.
     /// Otherwise, the caller should check with `has_sft_entry()` before calling this method.
-    unsafe fn update(&self, space: &(dyn SFT + Sync + 'static), start: Address, bytes: usize);
+    unsafe fn update(&self, space: *const (dyn SFT + Sync + 'static), start: Address, bytes: usize);
 
     /// Eagerly initialize the SFT table. For most implementations, it could be the same as update().
     /// However, we need this as a seprate method for SFTDenseChunkMap, as it needs to map side metadata first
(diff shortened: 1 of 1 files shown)


qinsoon:
> 1. Replaces casts of references to mutable references with `UnsafeCell<T>`.
> 2. Spaces are stored inside each of implementation as pointers in order to make Rust compiler happy because of lifetimes.

Thanks for the PR. The use of `UnsafeCell` looks fine as a workaround for the UB (eventually we would like to have a safe implementation). Anyway, it looks good. But I don't understand the second point -- is the new version of Rust compiler complaining about the lifetime? The current `SFTMap` implementation does not assume the lifetime of `&SFT`, and the creation of `SFTMap` states `&SFT` is `'static`. This allows us to easily change the scope of `SFTMap` in the future (e.g. we can confine the lifetime of `SFTMap` to the same lifetime as `MMTK`).

src/policy/sft_map.rs
```diff
@@ -128,17 +134,19 @@ mod space_map {
                     assert!(start + bytes <= space_start + MAX_SPACE_EXTENT);
                 }
             }
-            *mut_self.sft.get_unchecked_mut(index) = space;
+
+            //*mut_self.sft.get_unchecked_mut(index) = space;
```
k-sareen, line 138: Minor: Could you remove the commented out code here? I think there are a couple of other places too.

playXE:
> > 1. Replaces casts of references to mutable references with `UnsafeCell<T>`.
> > 2. Spaces are stored inside each of implementation as pointers in order to make Rust compiler happy because of lifetimes.
> 
> Thanks for the PR. The use of `UnsafeCell` looks fine as a workaround for the UB (eventually we would like to have a safe implementation). Anyway, it looks good. But I don't understand the second point -- is the new version of Rust compiler complaining about the lifetime? The current `SFTMap` implementation does not assume the lifetime of `&SFT`, and the creation of `SFTMap` states `&SFT` is `'static`. This allows us to easily change the scope of `SFTMap` in the future (e.g. we can confine the lifetime of `SFTMap` to the same lifetime as `MMTK`).

The second point arises from the fact that `mut_self` is removed. Before it would create a reference to SFTMap implementation that had different lifetime and `space` that is passed to `ugprade` could get assigned to it, now when replaced with UnsaafeCell lifetimes do not match

src/policy/sft_map.rs
playXE: Should be fixed now

k-sareen: Thanks. I think there are some other locations where the previous code has been commented out too. Could you remove those lines as well?

playXE: Should be done now for sure

src/policy/sft_map.rs
```diff
@@ -348,49 +348,51 @@ mod dense_chunk_map {
...
-        #[allow(clippy::cast_ref_to_mut)]
-        #[allow(clippy::mut_from_ref)]
-        unsafe fn mut_self(&self) -> &mut Self {
-            &mut *(self as *const _ as *mut _)
-        }
-
         pub fn addr_to_index(addr: Address) -> u8 {
             SFT_DENSE_CHUNK_MAP_INDEX.load_atomic::<u8>(addr, Ordering::Relaxed)
         }
+        #[inline(always)]
```
k-sareen, line 366: Could you remove the `#[inline(always)]` here and for `index_map`? We've discovered that adding inline pragmas isn't always ideal and we prefer using profile-guided optimizations nowadays.

src/policy/sft_map.rs
k-sareen: Thanks!

src/policy/sft_map.rs
playXE: Done

k-sareen: Thanks :)

playXE:
Hmm I would not merge this PR for now, I'm going to make storage of pointers in SFT vectors Atomic as well, using Atomic<T> from [atomic](https://docs.rs/atomic)

playXE:
> Hmm I would not merge this PR for now, I'm going to make storage of pointers in SFT vectors Atomic as well, using Atomic from [atomic](https://docs.rs/atomic)

So far @qinsoon told me that update method is invoked behind the lock so probably `push()` and update of spaces inside it are fine without atomic pointers. In that case merging this PR should be fine.

qinsoon:
> > Hmm I would not merge this PR for now, I'm going to make storage of pointers in SFT vectors Atomic as well, using Atomic from [atomic](https://docs.rs/atomic)
> 
> So far @qinsoon told me that update method is invoked behind the lock so probably `push()` and update of spaces inside it are fine without atomic pointers. In that case merging this PR should be fine.

Can you fix the style check? Then I will start a test run with our bindings to test the PR, then we can merge it.

playXE:
> > > Hmm I would not merge this PR for now, I'm going to make storage of pointers in SFT vectors Atomic as well, using Atomic from [atomic](https://docs.rs/atomic)
> > 
> > 
> > So far @qinsoon told me that update method is invoked behind the lock so probably `push()` and update of spaces inside it are fine without atomic pointers. In that case merging this PR should be fine.
> 
> Can you fix the style check? Then I will start a test run with our bindings to test the PR, then we can merge it.

Hmm, weirdly enough I don't get the same error when I git clone it and run cargo fmt check and clippy, could you maybe restart pre code review check?

qinsoon (approved):
LGTM

PR #880, opened 2023-08-01
Expose Allocators type to public API

playXE:
Without it getting offsets of `Allocators` fields is impossible in Rust, other bindings just declare that type in header file but in Rust it should be as simple as using memoffset crate + Allocators type from mmtk-core itself.


qinsoon (approved):
LGTM

PR #883, opened 2023-08-03
Add USDT tracepoints for key GC activities

caizixian:
This work is described in Improving Garbage Collection Observability with Performance Tracing by Claire Huang, Stephen M. Blackburn, and Zixian Cai, to appear in MPLR'23.

These tracepoints are compiled into `nop`s with minimal overhead when not attached as measured in the paper. The tracing tools that make use of these tracepoints will be released separately in another PR.

### DIFF
--- a/src/scheduler/gc_work.rs
+++ b/src/scheduler/gc_work.rs
@@ -545,6 +545,10 @@ impl<VM: VMBinding> ProcessEdgesBase<VM> {
     pub fn pop_nodes(&mut self) -> Vec<ObjectReference> {
         self.nodes.take()
     }
+
+    pub fn is_roots(&self) -> bool {
+        self.roots
+    }
 }
 
 /// A short-hand for `<E::VM as VMBinding>::VMEdge`.
@@ -631,6 +635,7 @@ pub trait ProcessEdgesWork:
     }
 
     fn process_edges(&mut self) {
+        probe!(mmtk, process_edges, self.edges.len(), self.is_roots());
         for i in 0..self.edges.len() {
             self.process_edge(self.edges[i])
         }
--- a/src/scheduler/worker.rs
+++ b/src/scheduler/worker.rs
@@ -345,12 +345,26 @@ impl<VM: VMBinding> GCWorker<VM> {
     /// Entry of the worker thread. Resolve thread affinity, if it has been specified by the user.
     /// Each worker will keep polling and executing work packets in a loop.
     pub fn run(&mut self, tls: VMWorkerThread, mmtk: &'static MMTK<VM>) {
+        probe!(mmtk, gcworker_run);
         WORKER_ORDINAL.with(|x| x.store(Some(self.ordinal), Ordering::SeqCst));
         self.scheduler.resolve_affinity(self.ordinal);
         self.tls = tls;
         self.copy = crate::plan::create_gc_worker_context(tls, mmtk);
         loop {
+            // Instead of having work_start and work_end tracepoints, we have
+            // one tracepoint before polling for more work and one tracepoint
+            // before executing the work.
+            // This allows measuring the distribution of both the time needed
+            // poll work (between work_poll and work), and the time needed to
+            // execute work (between work and next work_poll).
+            // If we have work_start and work_end, we cannot measure the first
+            // poll.
+            probe!(mmtk, work_poll);
             let mut work = self.poll();
+            // probe! expands to an empty block on unsupported platforms
+            #[allow(unused_variables)]
+            let typename = work.get_type_name();
+            probe!(mmtk, work, typename.as_ptr(), typename.len());
             work.do_work_with_stat(self, mmtk);
         }
     }
--- a/src/util/alloc/allocator.rs
+++ b/src/util/alloc/allocator.rs
@@ -219,7 +219,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
             } else {
                 // If we are not doing precise stress GC, just call the normal alloc_slow_once().
                 // Normal stress test only checks for stress GC in the slowpath.
-                self.alloc_slow_once(size, align, offset)
+                self.alloc_slow_once_traced(size, align, offset)
             };
 
             if !is_mutator {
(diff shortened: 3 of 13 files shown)


caizixian:
The clippy warnings on macOS seems to be because the `probe!` macro compiles to an empty block on macOS. I think this is fine.

qinsoon:
> The clippy warnings on macOS seems to be because the `probe!` macro compiles to an empty block on macOS. I think this is fine.

Yeah. Just suppress the warning.

src/plan/global.rs
```diff
@@ -710,12 +710,14 @@ impl<VM: VMBinding> BasePlan<VM> {
     pub fn set_gc_status(&self, s: GcStatus) {
         let mut gc_status = self.gc_status.lock().unwrap();
         if *gc_status == GcStatus::NotInGC {
+            probe!(mmtk, gc_start);
```
qinsoon, line 713: Why not put this to the controller loop which is more reliable than `GcStatus`? 
https://github.com/mmtk/mmtk-core/blob/5d22c2b3ca6c0f3393f9c8cbd8b47a28868a48cc/src/scheduler/controller.rs#L54

We set `GcStatus::GcProper` in these places, and we have executed some packets before that.
https://github.com/mmtk/mmtk-core/blob/5d22c2b3ca6c0f3393f9c8cbd8b47a28868a48cc/src/scheduler/gc_work.rs#L452
https://github.com/mmtk/mmtk-core/blob/5d22c2b3ca6c0f3393f9c8cbd8b47a28868a48cc/src/scheduler/gc_work.rs#L475

src/scheduler/worker.rs
```diff
@@ -350,7 +350,10 @@ impl<VM: VMBinding> GCWorker<VM> {
         self.tls = tls;
         self.copy = crate::plan::create_gc_worker_context(tls, mmtk);
         loop {
+            probe!(mmtk, work_poll);
```
qinsoon, line 353: I was curious when I used the code: why use `work_poll` as a signal for work end. You can just have a work_end tracepoint after `do_work_with_stat`.

src/util/alloc/allocator.rs
```diff
@@ -318,6 +318,19 @@ pub trait Allocator<VM: VMBinding>: Downcast {
...
     fn alloc_slow_once(&mut self, size: usize, align: usize, offset: usize) -> Address;
 
+    /// A wrapper method for [`alloc_slow_one`](Allocator::alloc_slow_once) to insert USDT tracepoints.
+    ///
+    /// Arguments:
+    /// * `size`: the allocation size in bytes.
+    /// * `align`: the required alignment in bytes.
+    /// * `offset` the required offset in bytes.
+    fn alloc_slow_once_traced(&mut self, size: usize, align: usize, offset: usize) -> Address {
+        probe!(mmtk, alloc_slow_once_start);
```
qinsoon, line 328: These trace points are pretty clear. You can use the same pattern for gc start/end, and work start/end.

src/plan/global.rs
caizixian: Fixed

src/scheduler/worker.rs
caizixian: Explained in the comment

src/util/alloc/allocator.rs
caizixian: Done

src/util/alloc/allocator.rs
```diff
@@ -318,6 +318,19 @@ pub trait Allocator<VM: VMBinding>: Downcast {
     /// * `offset` the required offset in bytes.
     fn alloc_slow_once(&mut self, size: usize, align: usize, offset: usize) -> Address;
 
+    /// A wrapper method for [`alloc_slow_one`](Allocator::alloc_slow_once) to insert USDT tracepoints.
```
k-sareen, line 321:
```suggestion
    /// A wrapper method for [`alloc_slow_once`](Allocator::alloc_slow_once) to insert USDT tracepoints.
```

k-sareen:
Did you have the performance overheads of the extra indirection via `_trace` functions? I presume it's none since the compiler would inline it

caizixian:
> Did you have the performance overheads of the extra indirection via `_trace` functions? I presume it's none since the compiler would inline it

I believe so.

wks:
Can you also include an example bpftrace script somewhere in the `mmtk-core` repo?  For example, in `mmtk-core/examples/bpftrace/gcvis.bt`?  It would be more helpful if we also include a script that can convert its result into a trace format that can be opened by perfetto or other well-known tools.

caizixian:
> Can you also include an example bpftrace script somewhere in the `mmtk-core` repo? For example, in `mmtk-core/examples/bpftrace/gcvis.bt`? It would be more helpful if we also include a script that can convert its result into a trace format that can be opened by perfetto or other well-known tools.

I plan to place the tracing scripts under `tool/` following the convention of bpftrace and bcc. 
I want to this in a separate PR, because I want to clean up the scripts to not have hardcoded paths etc.

wks:
Can we also include USDT trace points for the following functions in mmtk-core?

-   `harness_begin`
-   `harness_end`
-   `GCController::run`
-   `GCWorker::run`

We usually need to initialise some states in the bpftrace trace.  Since mmtk-core doesn't provide a official C API (the bindings may create `extern "C"` functions if they need to), we can't directly use `uprobe` or `uretprobe` for those points.  We can simply add USDT in those functions so that we can trace those points in a VM-agnostic way.

wks (changes requested):
I think this PR looks good in general.  The only missing peaces that I think we must include in this PR are the USDT trace points in the beginning of GC threads and `harness_{begin,end}`, without which it would be incomplete.

I am OK with postponing the splitting of `process_edges` trace point to another PR, because the current method is known to work.

I have another general suggestion, but doesn't have to be included in this PR, either.  We may let `GCWork` implement a very general `fn size(&self) -> Option<usize>` method so that when we do eBPF tracing, we can add a size argument to every work packet if it makes sense.  I think it makes sense for quite many packet types, including `ProcessEdgesWork`, `ScanObjects`, `Process{,Region}ModBuf` as well as weak reference and finalizer processing.  VM bindings can also use that to describe the size of their work packets, too, if that makes sense.  Examples include the number of "potential pinning parents" a Ruby VM has, the number of elements in weak global tables, such as "string tables" in OpenJDK and `fstring_table` in Ruby, etc.

src/scheduler/gc_work.rs
```diff
@@ -631,6 +635,7 @@ pub trait ProcessEdgesWork:
     }
 
     fn process_edges(&mut self) {
+        probe!(mmtk, process_edges, self.edges.len(), self.is_roots());
```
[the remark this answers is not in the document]
caizixian: Yeah, let's split into other PRs. We can always refine the tracepoints we have when we have a compelling use case.

wks (approved):
LGTM

PR #885, opened 2023-08-04
Run CI build/unit test with latest stable Rust toolchain

qinsoon:
This PR adds checks to build and unit test MMTk core with the latest Rust stable version. We do not require the tests with latest stable Rust to pass in order to merge a PR. See discussion in https://github.com/mmtk/mmtk-core/pull/881#issuecomment-1664925112.


wks (approved):
This PR looks good.  From the CI result, it worked, but there are some test failures for the stable release.  We can fix it for the stable release separately.

Issue #886, opened 2023-08-04
Misaligned access in align_offset_alloc


wks:
The function `util::malloc::malloc_ms_util::align_offset_alloc(size: usize, align: usize, offset: usize) -> Address` allocates `size` bytes of memory, and ensures `(result + offset) % align == 0`.

It also stores the original malloc result at `result - BYTES_IN_ADDRESS`.

```rs
    let malloc_res_ptr: *mut usize = (result - BYTES_IN_ADDRESS).to_mut_ptr();
    unsafe { *malloc_res_ptr = address.as_usize() };
```

But the result is not aligned if `offset % align != 0`.  It is an error in Rust to dereference unaligned pointers (See https://doc.rust-lang.org/std/primitive.pointer.html):

> However, when a raw pointer is dereferenced (using the * operator), it must be non-null and aligned.

Currently this error is causing the tests to fail on the stable Rust release.

qinsoon:
https://github.com/mmtk/mmtk-core/pull/887 provided a workaround for the issue using unaligned memory access functions on Rust `ptr`. We probably still want to properly remove the unaligned access, and store the address into an aligned slot.

wks:
It will be better if we can do so.  But I won't worry too much about it if we only need to load it when we free it.  The `free()` function will probably be the bottleneck anyway.

PR #887, opened 2023-08-04
Fix unaligned access

wks:
`align_offset_alloc` and related functions now use `ptr::{read,write}_unaligned` to access the prepended address.

When calculating the address or offset of metadata, we ensure its address or offset is at least word-aligned.

Fixes: https://github.com/mmtk/mmtk-core/issues/886
Fixes: https://github.com/mmtk/mmtk-core/issues/888


qinsoon (approved):
LGTM.

I think we probably would like to keep https://github.com/mmtk/mmtk-core/issues/886 open, as this PR is just a workaround for the issue.

Issue #658, opened 2022-09-01
Problems with the public API


wks:
The public API of the mmtk-core to the VM is defined as all public items in the `mmtk-core` crate.  Now they are fully listed here: https://www.mmtk.io/mmtk-core/public-doc/plan/index.html

We can now observe problems by looking at that document.

# memory_manager

It is a legacy from JikesRVM MMTk.  It contains many static (Java) or top-level (Rust) functions.  They used to be the set of functions MMTk provide to JikesRVM.  The problems of the `memory_manager` module in Rust MMTk are:

1.  They are not idiomatic in Rust.  A Rust crate exposes its interface through modules, structs, enums, traits, functions and constants.  Although Rust is not as object-oriented as C++ and Java, it is still more natural for Rust to use abstract data types and expose methods around types.
2.  They are incomplete.  Some VMs still need to interact with types and functions in other modules to use needed functionalities of MMTk.

I suggest make transition gradually

1.  Deprecate the functions in `memory_manager`
2.  Implement those functions in relevant parts of MMTk.
    -   If a function is related to a MMTk instance, make it a method of `MMTK`.
        -   Example: `memory_manager::bind_mutator` to `MMTK::bind_mutator`
    -   Similarly, if it is related to another object, make it a method of that.
        -   Example: `memory_manager::alloc` to `Mutator::alloc`
    -   If it is a top-level function, make it a top-level function of a related module.
        -   Example: `memory_manager::malloc` to `util::malloc::malloc`.  Expose `util::malloc` as `pub`.
3.  Migrate bindings to remove deprecation warnings.
4.  Remove the `memory_manager` module.

# Things that shouldn't have been exported

The following are about plans:

-   `plan::Plan`: Contains `dyn` methods of a plan.  A VM shouldn't directly interact with plans.
-   `plan::PlanConstraints`: Constraints of a plan.  VMs probably never access them directly, but the VM does need to **filter plans based on some criteria**, such as whether a plan can pin objects.

The following are for work packets and computing the transitive closure.  The VM [used to create work packets](https://github.com/mmtk/mmtk-openjdk/commit/ebf6e740ab7e264f0dd4e4a452d0bf8bf6068637#diff-54f4f04a18788baf9bf66cff8c4256151b680087e3490fb8e8f62e2d60ad9fabL16) which involves some of tracing, but we [refactored the API before](https://github.com/mmtk/mmtk-core/pull/611), and the VMs never see work packets since then. Now they shouldn't be exposed.

[...]

PR #889, opened 2023-08-05
Implement AllocatorInfo

playXE:
This type adds `AllocatorInfo` enum that can be used to query offset of allocator fields for fast-path generation. Currently all bindings have to re-declare `*Allocator` types with the same layout as in mmtk-core in order to get offset of these fields. This enum simplifies that, all you need to do is invoke `get_allocator_info(selector)`.


src/util/alloc/mod.rs
```diff
@@ -10,6 +10,7 @@ pub use allocator::Allocator;
 pub(crate) mod allocators;
 pub use allocators::AllocatorSelector;
 pub use allocators::Allocators;
```
qinsoon, line 12: With this PR, the changes in https://github.com/mmtk/mmtk-core/pull/880 can be reverted.

src/memory_manager.rs
```diff
@@ -350,6 +358,60 @@ pub fn get_allocator_mapping<VM: VMBinding>(
     mmtk.plan.get_allocator_mapping()[semantics]
 }
 
+/// Return an AllocatorInfo for the given allocator selector. This method is provided
+/// so that VM compilers may generate allocator fast-path and load fields for the fast-path.
+///
+/// Arguments:
+/// * `selector`: The allocator selector to query.
+pub fn get_allocator_info<VM: VMBinding>(selector: AllocatorSelector) -> AllocatorInfo {
```
qinsoon, line 366: Better move the actual logic to `AllocatorInfo`. We try to keep methods in `memory_manager` simple -- mostly just forwarding calls. We have an intention to remove the module completely (https://github.com/mmtk/mmtk-core/issues/658).

src/util/alloc/allocators.rs
```diff
@@ -175,3 +175,33 @@ impl Default for AllocatorSelector {
...
+pub enum AllocatorInfo {
+    BumpPointer {
+        limit_offset: usize,
+        top_offset: usize,
+    },
+    MarkCompact {
+        limit_offset: usize,
+        top_offset: usize,
+    },
+    Immix {
```
qinsoon, line 192: I think you can just remove the `Immix` variant as well. Immix's `overflow_alloc` (which uses the large cursor/limit) and `alloc_slow_hot` (which tries to find a recyclable line) is considered as the 'intermediate' path (i.e. between fastpath and slowpath). Normally the compiler don't need to inline it. In that case, generating allocation fastpath for Immix would be the same as `BumpPointer` fastpath (just checking cursor/limit, and adjust cursor).

src/memory_manager.rs
```diff
@@ -350,6 +358,60 @@ pub fn get_allocator_mapping<VM: VMBinding>(
...
+            let bump_offset = base_offset + offset_of!(MarkCompactAllocator<VM>, bump_allocator);
+            let limit_offset = bump_offset + offset_of!(BumpAllocator<VM>, limit);
+            let top_offset = bump_offset + offset_of!(BumpAllocator<VM>, cursor);
+            AllocatorInfo::MarkCompact {
+                limit_offset,
+                top_offset,
+            }
+        }
+
+        _ => AllocatorInfo::None,
```
qinsoon, line 411: Probably better to differentiate `Unimplemented` and `None`. For example, free list allocator has a fastpath, but it is not implemented in this PR. But large object allocator does not have a fastpath.

qinsoon (commented):
I think the changes look good. This is something we need to properly support compiler generated allocation fastpath.

src/util/alloc/bumpallocator.rs
```diff
@@ -18,9 +18,9 @@ pub struct BumpAllocator<VM: VMBinding> {
     /// [`VMThread`] associated with this allocator instance
     pub tls: VMThread,
     /// Current cursor for bump pointer
-    cursor: Address,
+    pub(crate) cursor: Address,
     /// Limit for bump pointer
-    limit: Address,
+    pub(crate) limit: Address,
```
wks, line 23: We can make a `#[repr(C)] struct` for these two fields since it appears so often.  Then we only need to return one offset for those two fields.

src/memory_manager.rs
```diff
@@ -350,6 +358,60 @@ pub fn get_allocator_mapping<VM: VMBinding>(
...
+                top_offset,
+            }
+        }
+
+        AllocatorSelector::Immix(index) => {
+            let base_offset = offset_of!(Mutator<VM>, allocators)
+                + offset_of!(Allocators<VM>, immix)
+                + size_of::<ImmixAllocator<VM>>() * index as usize;
+            let limit_offset = base_offset + offset_of!(ImmixAllocator<VM>, limit);
+            let top_offset = base_offset + offset_of!(ImmixAllocator<VM>, cursor);
```
wks, line 386: I still think the name `cursor` is better than `top` because `top` may be interpreted as a synonym of limit.

src/memory_manager.rs
playXE: Should I implement `From<AllocatorSelector>`?

qinsoon: Either looks good to me.

src/util/alloc/allocators.rs
playXE: Should not small limit still be checked? Or am I wrong here?

qinsoon: You don't have to check "small_limit". If the allocation can fit into the normal buffer (cursor/limit), it will be allocated in the buffer. However, the buffer can be as small as one line (depending on how fragment the heap is). So sometimes the allocation that is larger than one line can't fit into the normal buffer, then the large cursor/limit is used. But that is not in the fastpath path any more.

The actual code that is considered as Immix's fastpath is (which is as simple as a normal bump pointer allocation):
https://github.com/mmtk/mmtk-core/blob/4c1e58a1b293e0e710feebdb7bec876bd98866f8/src/util/alloc/immix_allocator.rs#L75-L78
and
https://github.com/mmtk/mmtk-core/blob/4c1e58a1b293e0e710feebdb7bec876bd98866f8/src/util/alloc/immix_allocator.rs#L90-L102

k-sareen: I remembered @wenyuzhao mentioning something about this in-person to me. Maybe he can chime in and clarify expectations.

src/memory_manager.rs
playXE: Should be fixed now

src/memory_manager.rs
playXE: Updated

src/util/alloc/mod.rs
playXE: Done

src/util/alloc/bumpallocator.rs
playXE: Should this be done in this PR as well?

wks: Yes, please.  This is the best chance to do it.

playXE: How this structure should be named?

wks: Good question.  I think we should discuss on Zulip and let everyone vote.

It should be defined in `allocator.rs` because it is common to many concrete allocators.

steveblackburn: 1. This PR seems a step in the right direction.  Thank you!
2. Placing cursor and limit in a struct seems sensible.
3. The cursor and limit should be called a bump pointer, because that is literally what it is.  I teach people how GCs are built and this is literally how I describe a bump pointer: a cursor and a limit.  Inventing a new term to describe this most basic of data structures is not a good idea.

Meta point: I don't think "voting" on conceptual points like this on Zulip is a great idea in general.

playXE:
So far what would be the next proper steps to get this PR merged? I am actually now all in for idea of bindings storing bump-pointers/free-lists on their side and syncing with MMTk when necessary. BDWGC works this way by providing support for inline allocation by allowing user-code to create free-lists.

cc @steveblackburn @qinsoon

qinsoon:
> So far what would be the next proper steps to get this PR merged? I am actually now all in for idea of bindings storing bump-pointers/free-lists on their side and syncing with MMTk when necessary. BDWGC works this way by providing support for inline allocation by allowing user-code to create free-lists.
> 
> cc @steveblackburn @qinsoon

It seems there are two options:
1. Expose the offset of cursor/limit (and similar fields for other allocators), or
2. Expose the values of cursor/limit.

My preference is Option 1. It covers what Option2 does, and is more flexible for the bindings. Option 2 enforces the binding to create their own fastpath data structures, and sync values with MMTk. I am not sure what others think about. @steveblackburn @wks

We will get this resolved no later than tomorrow (we have a meeting tomorrow and I will raise this if necessary).

qinsoon:
> > So far what would be the next proper steps to get this PR merged? I am actually now all in for idea of bindings storing bump-pointers/free-lists on their side and syncing with MMTk when necessary. BDWGC works this way by providing support for inline allocation by allowing user-code to create free-lists.
> > cc @steveblackburn @qinsoon
> 
> It seems there are two options:
> 
> 1. Expose the offset of cursor/limit (and similar fields for other allocators), or
> 2. Expose the values of cursor/limit.
> 
> My preference is Option 1. It covers what Option2 does, and is more flexible for the bindings. Option 2 enforces the binding to create their own fastpath data structures, and sync values with MMTk. I am not sure what others think about. @steveblackburn @wks
> 
> We will get this resolved no later than tomorrow (we have a meeting tomorrow and I will raise this if necessary).

@playXE 

We discussed this. Exposing field offsets to the binding is a good idea. It allows the JIT compiler to generate code, and allows the runtime to load/store the values.

A refinement on the current PR we plan to do is that we could have public fastpath structs with `#[repr(C)]`, and expose the offset to the struct instead of exposing offsets to separate fields. Such as

```rust
#[repr(C)]
pub struct BumpPointer {
  pub cursor: Address,
  pub limit: Address,
}

struct BumpAllocator {
  bump_pointer: BumpPointer
  ...
}

pub enum AllocatorInfo {
    BumpPointer(usize), // offset from mutator to the BumpPointer struct
    // FIXME: Add free-list fast-path
    Unimplemented,
    None,
}
```

This change does not have to be included in the PR. Just FYI, as we may change the actual interface in the future. But we will follow the general idea of exposing offsets to fields for bindings.

If this sounds good to you, you can make the PR ready, and we can proceed from there. Thanks for the PR.

Cargo.toml
```diff
@@ -19,6 +19,7 @@ crate-type = ["rlib"]
 doctest = false
 
 [dependencies]
+memoffset = "0.9"
```
k-sareen, line 22: Can you sort this alphabetically?

src/util/alloc/allocators.rs
```diff
@@ -175,3 +179,69 @@ impl Default for AllocatorSelector {
         AllocatorSelector::None
     }
 }
+
+/// This type describes allocator information. It is used to
+/// generate fast paths for the GC. All offset fields are relative to [Mutator](crate::Mutator).
```
k-sareen, line 184:
```suggestion
/// generate fast paths for the GC. All offset fields are relative to [`Mutator`](crate::Mutator).
```

src/util/alloc/immix_allocator.rs
```diff
@@ -28,9 +26,9 @@ pub struct ImmixAllocator<VM: VMBinding> {
     /// Is this a copy allocator?
     copy: bool,
     /// Bump pointer for large objects
-    large_cursor: Address,
+    pub(crate) large_cursor: Address,
```
k-sareen, line 29: You could coalesce the `large_{cursor,limit}` fields into one `large_bump_pointer` as well.

qinsoon:
I am trying to push some changes to the PR. Mainly about the new method `BumpPointer::alloc`. Though it is better to provide such a method that can be reused, it introduce an extra conditional check in the fastpath that the Rust compiler does not optimize away. I think it would be a better idea to revert that part.

qinsoon (approved):
LGTM

PR #892, opened 2023-08-08
Fix performance regression test scripts

qinsoon:
This PR updates to https://github.com/mmtk/ci-perf-kit/releases/tag/0.6.9. This allows us to run performance tests for each commit in `master`. The remaining broken tests are: 1. running performance tests per PR, and 2. running baseline performance for JikesRVM and OpenJDK weekly (low priority). They will be fixed separately.


wks (approved):
LGTM

Issue #853, opened 2023-06-08
Remove `cast_ref_to_mut` related to FreeList and PageResource


wks:
parent: https://github.com/mmtk/mmtk-core/issues/850

# Problem

## Atomicity of primitive FreeList operations

The `FreeList` implementation is ported from JikesRVM.  In JikesRVM, the `GenericFreeList` base class provided a free list implementation based on an abstract array of `int` (`i32` in Rust), but is agnostic to how the array is allocated.  It has the abstract methods `getEntry` and `setEntry`, and concrete table implementations, including `RawMemoryFreeList` and `IntArrayFreeList`, implement those methods to access the underlying array.  Particularly, in `IntArrayFreeList`, the freelist is backed by an `int[]`.

Primitive FreeList operations are not thread-safe.  `GenericFreeList` often separately updates the "low" and "high" bit-fields of each `int` (32-bit integer)  Complex operations (such as splitting a region, or merging two adjacent regions) take multiple updates to complete, and they never use any atomic read-modify-write operations.

In JikesRVM, `FreeListPageResource.allocPages` holds the `PageResource.lock`; `FreeListPageResource.freePage` is only executed by one thread.  This suggest that the `FreeList` itself is not thread-safe, but it is used with external synchronisation.

## Two-level FreeList structure

The `IntArrayFreeList` has a two level design where children share the same underlying array with the parent.  The parent free list has multiple "heads" (`1 + NUM_OF_CHILDREN`).  The parent uses one of the heads, and each child uses one head.

If I read the JikesRVM source code correctly, the "parent head" is owned by the global `Map32` instance, and is protected by `Map32.lock`.  Each "child head" is owned by a `FreeListPageResource`, and is protected by `PageResource.lock`.  A page resource may acquire pages from the global `Map32` instance, holding `Map32.lock`.  This will poll FreeList entries from the doubly-linked list of the "parent head".  Then it will insert those entries into the PageResource's "child head", while holding `PageResource.lock`.  In this way, the single global array can contain multiple doubly-linked lists starting with each head, and different doubly-linked lists have no overlapping entries.  Therefore, it is safe for two different threads to access two different doubly-linked lists using two different heads, even though their entries are allocated in one single contiguous global array.

# Solution

1.  Protect `CommonFreeListPageResource.free_list` with Mutex!  This will eliminate the need to call `table_mut` because a thread always gets `&mut FreeList` when it acquires the Mutex.
[...]

Issue #852, opened 2023-06-08
Remove `cast_ref_to_mut` related to Prepare and Release


wks:
Parent: https://github.com/mmtk/mmtk-core/issues/850

# Problem

There are periods in GC when we know some work packets have exclusive access to the `Plan` instance.  Examples include `Prepare`, `Release` and `EndOfGC`.  The problem is, *we know this fact as humans*, but the Rust programming language cannot check if that work packet has exclusive access to `Plan`, using either compile-time checking or run-time checking.  Therefore, we worked around this by casting `&Plan` to `&mut Plan`

# Possible solution

Because of the dynamic nature of MMTk-VM interaction and the work packet system, it is hard to check whether a thread is the only thread that can access a data structure.

But it is possible to do this dynamically using reference counting.  If we can count how many threads have access to a given data structure (for example, MMTk, or Plan), we can let the users "yield" their right to access to that data structure.  Once we are sure all read-only users yielded, we can grant another thread (the one running `Prepare` or other work packets) read-write access.  It is checked at run time.  If a thread attempts to get read-write access while the read-only user count is greater than 0, it shall panic (which is not as good as compile-time checking in which case unsafe code doesn't compile).

p.s. The same idea can be applied to the parking of GC worker threads.  If all workers parked, the last GC worker can gain read-write access and modify work buckets states.

`AtomicRefCell` is the closest tool for this need.  However, `AtomicRefCell` requires the "handle" instances (`Borrow` and `BorrowMut`) to live within the lifetime of the `AtomicRefCell` itself, with compile-time lifetime checking.  Because of this limitation, we can't send the `Borrow<T>` handle to another thread.

I once made a proof-of-concept project for "lending" some owned data to another thread, with run-time checking.  The main difference is, the compiler will not prevent the handles from being sent to other threads.  But if the owner of the object dies while it is still lent, it panics.  (Sorry. In such a dynamic environment, there is not much the compiler can do.)

wks:
In `Prepare` and `Release`, we call `Plan.prepare/release` which calls the `prepare/release` methods of each space.  All of those `prepare/release` methods have `&mut self` as parameters.  Obviously, they are intended to modify the internal states of those spaces.

# Case study of ImmixSpace

I had a look at `ImmixSpace`.

```rs
pub struct ImmixSpace<VM: VMBinding> {
    common: CommonSpace<VM>,
    pr: BlockPageResource<VM, Block>,
    pub chunk_map: ChunkMap,
[...]

PR #893, opened 2023-08-08
Remove cast ref to mut everywhere

playXE:
- MMTK type now stores `Plan` inside `UnsafeCell` to allow safe mutable access when necessary
- Various types that previously had `mut_self()` method now are separated into `Type` and `TypeInner`, second one is stored inside `Type` behind `UnsafeCell` which allows to obtain mutable access when necessary without involving UB.
- `&'static` references that were previously stored in some types are now converted into `NonNull<>`


playXE:
Note: it still requires `clippy::mut_from_ref` for using `&mut *UnsafeCell::get()`, I am not entirely sure how to transition everything to safe code there.

.github/scripts/ci-style.sh
```diff
@@ -1,6 +1,6 @@
 . $(dirname "$0")/ci-common.sh
 
-export RUSTFLAGS="-D warnings"
+export RUSTFLAGS="-D warnings -A unknown-lints"
```
playXE, line 3: Added this because in GitHub CI clippy version is older than on my PC, this is needed for `#[allow(clippy::needless_pass_by_ref_mut)]`

src/scheduler/gc_work.rs
```diff
@@ -101,23 +102,26 @@ impl<VM: VMBinding> GCWork<VM> for PrepareCollector {
...
 impl<C: GCWorkContext + 'static> GCWork<C::VM> for Release<C> {
     fn do_work(&mut self, worker: &mut GCWorker<C::VM>, mmtk: &'static MMTK<C::VM>) {
         trace!("Release Global");
 
-        self.plan.base().gc_trigger.policy.on_gc_release(mmtk);
-
+        unsafe {
+            (*self.plan).base().gc_trigger.policy.on_gc_release(mmtk);
+        }
         // We assume this is the only running work packet that accesses plan at the point of execution
```
k-sareen, line 123: We actually know this is false. See https://github.com/mmtk/mmtk-core/issues/824#issuecomment-1558661794. I think we shouldn't ignore this issue

k-sareen: @qinsoon Do you have any thoughts on the above?

playXE: I've fixed only UB problems in this PR so didn't know about this issue. So what should I do in that line? Simply remove this call?

qinsoon: The discussion on this topic is not related with the PR. I make this clear so it won't confuse @playXE . I think nothing needs to be done for the PR.

I listed assumptions in https://github.com/mmtk/mmtk-core/issues/717#issuecomment-1362097538. This line is correct if the assumptions are true. The issue in #824 is a case in mark sweep that violates the assumptions: mark sweep's mutator release and plan's release both modify the block list. We either rethink our assumptions, or fix the implementation in mark sweep.

src/util/heap/freelistpageresource.rs
```diff
@@ -262,52 +294,66 @@ impl<VM: VMBinding> FreeListPageResource<VM> {
     }
 
     fn allocate_contiguous_chunks(
-        &mut self,
+        &self,
```
qinsoon, line 297: I am concerned about changes like this in this PR. The current code (before this PR) gets a mutable reference when we acquire a lock. In this case, we can reason that getting the mutable reference is correct, as we have the lock. In this PR, many methods are changed from `&mut self` to `&self`, and they get mutable references internally. If the caller is not aware of this, this could easily lead to a race condition.

playXE: One way around this is mark these functions as unsafe, other way around is add mutating methods to `*Inner` types, what would be the best?

qinsoon: Maybe just mark them as unsafe for now. We hope to do a safe implementation at some point.

playXE: Marked these functions as unsafe now.

qinsoon:
Thanks for the PR. I will look at it tomorrow.

qinsoon:
Again thanks for the PR. This is a good cleanup for identifying and removing `cast_ref_to_mut` for the issue https://github.com/mmtk/mmtk-core/issues/850. 

As a summary, the changes in this PR are:
* `MMTK.plan` uses `UnsafeCell`.
* `Prepare/Release.plan` uses `*const C::PlanType` instead of `&'static C::PlanType` so we can cast it to a mutable pointer.
* Add an inner type for a few types and use `UnsafeCell<XInner>` to allow mutability
  * `FreeListPageResource`
  * `FragmentedMmapper`
  * `Map32`
  * `Map64`
* Add `Send/Sync` to types that uses unsafe pointers.
* Use `NonNull` instead of `&`:
  * `&'static CommonFreeListPageResource` in `Map32`
  * `&'static IntArrayFreeList` in `IntArrayFreeList`

Generally, undefined behavior < unsafe code < safe code. However, this PR adds a large number of unsafe code (see the following table) to remove the undefined behavior code (which is also counted as unsafe in the table). As reducing unsafe code is one of our goals for the project (https://github.com/mmtk/mmtk-core/issues/97), this makes it hard to argue that this PR is an improvement for the code base.

|unsafe|master|PR|
|---|---|---|
|functions|13|13|
|exprs|1414|1671|
|item_impls|14|28|
|item_traits|0|0|
|methods|40|50|

(The data is from `count-unsafe`)

The urgency for this PR is that from Rust 1.72, `cast_ref_to_mut` (which is removed by the PR) will be a hard error. That means without any change, our code base won't build with Rust 1.72 (which will be released on 24th August). 24th August falls in our next release cycle, and we won't increase our pinned Rust version before next release. This means we have at least 6 weeks to address this issue. 

I think the PR is good on its own, no change is needed. But it is a temporary solution for `cast_ref_to_mut` for Rust 1.72. Our options for the PR are:
1. We merge this PR, but we should replace the unsafe code around those places with safe code as much as we can before the next release. 
2. We put this PR on hold, refactor related code in `master` to safe code as much as we can before the next release, and merge this PR with whatever is left.

My current preference is 2, considering that we will still have some time before Rust 1.72 and the next MMTk release.

wks:
I am OK with either merging this PR in this release or the next release.  I am OK with doing some refactoring before merging this PR, too, because that will mean this PR will need to change fewer places.

But as a reminder, some of the unsafe code introduced in this PR can be very hard to refactor fully.  For example, fully fixing https://github.com/mmtk/mmtk-core/issues/852 requires refactoring the work packet scheduler, which should be part of the research topic of a PhD.  However, a shallow refactoring may be possible.  We can use `Atomic` and `UnsafeCell` to push the `unsafe` to specific fields.

Others should be easy to fix.  https://github.com/mmtk/mmtk-core/issues/853 I think the `FreeList` data structure is low-level, and it can legitimately use `unsafe`.

qinsoon:
> But as a reminder, some of the unsafe code introduced in this PR can be very hard to refactor fully. For example, fully fixing #852 requires refactoring the work packet scheduler, which should be part of the research topic of a PhD. However, a shallow refactoring may be possible. We can use `Atomic` and `UnsafeCell` to push the `unsafe` to specific fields.

We do not plan to change the scheduler (the design or the algorithm) for now (which is a research goal for @wenyuzhao), but we intend to allow mutable access to `Plan` in work packets (safe access if possible) which is a pure engineering goal. 

> Others should be easy to fix. #853 I think the `FreeList` data structure is low-level, and it can legitimately use `unsafe`.

Not necessarily so. The reason for the issue is that most of the related code is a direct port of Java MMTk which has no concept about mutable vs immutable. The unsafe code was just used to work around the issues Rust raised. What we want is to figure out mutability, and allow Rust to reason and check for us.

Another example is `RawMemoryFreeList`. It is necessary as JikesRVM is metacircular and it needs an implementation of free list without doing any runtime allocation. We do not have those restrictions now. I feel we may not even need `RawMemoryFreeList`.

qinsoon (approved):
LGTM

wks (approved):
LGTM, too.

Issue #867, opened 2023-07-18
Gracefully deal with unrealistically large allocation sizes


qinsoon:
When we request `alloc` with an unrealistically large allocation size (e.g. `2251799813685249 pages` or `9 223 372 000 GB` in the following example from GHC), this breaks our assumptions about virtual memory (how much virtual memory each space has, how we store VM map, etc). This causes all sorts of assertion failures. 

We would at least fail gracefully with a clear error message. It would be better we use our `out_of_memory` hook to let the binding deal with it.

See the discussion on Zulip: https://mmtk.zulipchat.com/#narrow/stream/315620-Porting/topic/mmtk_alloc

> The backtrace for reference:
```
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140737347598464) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140737347598464) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140737347598464, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff7c61476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff7c477f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x000000000094f907 in std::sys::unix::abort_internal () at library/std/src/sys/unix/mod.rs:350
#6  0x000000000094d0b2 in std::panicking::rust_panic () at library/std/src/panicking.rs:742
#7  0x000000000094cea7 in std::panicking::rust_panic_with_hook () at library/std/src/panicking.rs:710
#8  0x000000000094cba2 in std::panicking::begin_panic_handler::{closure#0} () at library/std/src/panicking.rs:577
#9  0x000000000094b62c in std::sys_common::backtrace::__rust_end_short_backtrace<std::panicking::begin_panic_handler::{closure_env#0}, !> () at library/std/src/sys_common/backtrace.rs:137
#10 0x000000000094c8f2 in std::panicking::begin_panic_handler () at library/std/src/panicking.rs:575
#11 0x0000000000413543 in core::panicking::panic_fmt () at library/core/src/panicking.rs:64
#12 0x00000000004135dd in core::panicking::panic () at library/core/src/panicking.rs:114
[...]

Issue #746, opened 2023-02-06
Runtime-specific panic/fail


qinsoon:
Currently MMTk panics using Rust's built-in panic mechanism. However, this may not work with some runtimes. @wks found that in Ruby, panicking in a Rust thread does not cause the process to crash. We probably need an API similar to `Collection::out_of_memory()` and allow the runtime to implement panic in their own way.
https://github.com/mmtk/mmtk-core/blob/a272c2399df39177361788449b1f96d060a86080/src/vm/collection.rs#L91

See https://doc.rust-lang.org/std/panic/fn.set_hook.html.

We need to be careful that we should only capture MMTk's panic. If the runtime has other component written in Rust, we should not capture their panic.

k-sareen:
I remember having a short conversation about this last year in the OOM PR: https://github.com/mmtk/mmtk-core/pull/535#issuecomment-1029596823

PR #896, opened 2023-08-10
Call Collection::out_of_memory if the allocation size is larger than max heap size

qinsoon:
This closes https://github.com/mmtk/mmtk-core/issues/867.


src/policy/space.rs
```diff
@@ -46,8 +46,42 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
...
+    /// memory in fixed size blocks do not need to call this method.
+    fn oom_size_check(&self, tls: VMThread, size: usize) {
+        let max_pages = self.get_gc_trigger().policy.get_max_heap_size_in_pages();
+        let requested_pages = size >> LOG_BYTES_IN_PAGE;
+        if requested_pages > max_pages {
+            VM::VMCollection::out_of_memory(
+                tls,
+                crate::util::alloc::AllocationError::HeapOutOfMemory,
+            );
+            panic!(
```
wks, line 68: If `panic!()` immediately follows `Collection::out_of_memory()`, it implies the only way not to reach this `panic!` is doing stack unwinding (or swap-stack or other advanced stack-manipulation techniques) within `out_of_memory`.  I don't know how a VM can do this reliably because unwinding through Rust frames is ABI-specific.  But given that existing code (such as `util::memory::handle_mmap_error`) also assumes this function never returns, we may leave this to another day.

wks (changes requested):
See in-line comments.  We probably can't let the VM recover from OOM for now.  For the test case, maybe `catch_unwind` can be used to prevent the lock from being poisoned.

src/policy/space.rs
qinsoon: We can just return `Address::ZERO` here without panic (maybe with a `warn!`), in the same way as other `out_of_memory()` uses.

vmbindings/dummyvm/src/tests/fixtures/mod.rs
```diff
@@ -57,7 +57,7 @@ impl<T: FixtureContent> SerialFixture<T> {
     }
 
     pub fn with_fixture<F: Fn(&T)>(&self, func: F) {
-        let mut c = self.content.lock().unwrap();
+        let mut c = self.content.lock().unwrap_or_else(|poisoned| poisoned.into_inner());
```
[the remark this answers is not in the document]
qinsoon: Done.

src/policy/space.rs
qinsoon: I changed the method name to `will_go_oom_on_acquire`, and returns a boolean. If the return value is true, the allocator will return `Address::ZERO`.

qinsoon:
There is an issue with this PR: if we try to do any address arithmetic in the allocation fastpath, we may have overflow in the computation, and we do not call `out_of_memory()` for that case. See the test below. I don't think we have a good way to deal with this, given that we do not want to introduce any check in the fastpath.
https://github.com/mmtk/mmtk-core/blob/04c0c48b3e0c6710bd2308c3477006cd070b4ef3/vmbindings/dummyvm/src/tests/issue867_allocate_unrealistically_large_object.rs#L27

wks (approved):
LGTM

k-sareen:
> There is an issue with this PR: if we try to do any address arithmetic in the allocation fastpath, we may have overflow in the computation, and we do not call `out_of_memory()` for that case. See the test below. I don't think we have a good way to deal with this, given that we do not want to introduce any check in the fastpath.
> 
> https://github.com/mmtk/mmtk-core/blob/04c0c48b3e0c6710bd2308c3477006cd070b4ef3/vmbindings/dummyvm/src/tests/issue867_allocate_unrealistically_large_object.rs#L27

One alternative, although I don't know how feasible it is, would be to register a custom panic handler and then call either `out_of_memory` or `panic_in_vm` (random function) or something to allow the VM binding to crash by itself. We discussed doing something similar in the OOM PR and https://github.com/mmtk/mmtk-core/issues/746

wks:
> There is an issue with this PR: if we try to do any address arithmetic in the allocation fastpath, we may have overflow in the computation, and we do not call `out_of_memory()` for that case. See the test below. I don't think we have a good way to deal with this, given that we do not want to introduce any check in the fastpath.
> 
> https://github.com/mmtk/mmtk-core/blob/04c0c48b3e0c6710bd2308c3477006cd070b4ef3/vmbindings/dummyvm/src/tests/issue867_allocate_unrealistically_large_object.rs#L27

I think doing checking in Rust is OK.  The "real" fast path should be implemented by the JIT compiler, or inlined to the concrete allocation sites.  In most of the cases, the object size should be a constant, and doesn't need checking anyway, given that a bump-pointer allocator always checks against a reaonably-sized `limit`.  If the object size is a variable, the VM should do the check anyway to avoid integer overflow.  If the VM simply calls MMTk API function for fast-path allocation, it means the VM has not started optimising for performance (like the status of the Ruby binding two months ago).  In that case, we should prefer correctness over performance.

k-sareen:
> If the VM simply calls MMTk API function for fast-path allocation, it means the VM has not started optimising for performance

In the case of a VM implemented in Rust, this is not necessarily true. It would be inlined by PGO

wks:
> > If the VM simply calls MMTk API function for fast-path allocation, it means the VM has not started optimising for performance
> 
> In the case of a VM implemented in Rust, this is not necessarily true. It would be inlined by PGO

That makes sense.  But if that's the case, the compiler can still eliminate the object size check if the supplied object size is a constant, although I am not 100% confident that the compiler will do it.

k-sareen:
Often a "checked" and "unchecked" API is presented to a developer. They are recommended to use the "checked" one always but they can use the "unchecked" one for more performance, but they have to guarantee that they have checked the assumptions beforehand

src/policy/space.rs
k-sareen: May I suggest `will_oom_on_acquire`?

qinsoon: Sounds good. Done.

k-sareen:
Should we merge this or discuss how to fix the bug Yi noticed?

qinsoon:
> Should we merge this or discuss how to fix the bug Yi noticed?

I think we can merge this. I will try out the idea of catching Rust panics in a separate PR (I cannot guarantee it would work out though).

PR #897, opened 2023-08-10
This PR enables transitively pinning (TP) objects from particular roots for Immix/StickyImmix

udesou:
- A new bucket `ImmovableClosure` has been created to trace and transitively pin roots, i.e., roots in which no object in its transitive closure is allowed to move.
- `RootsWorkFactory` has two new functions `create_process_tp_edge_roots_work` and `create_process_tp_node_roots_work` to create work to process these roots.
- `GCWorkContext` expects an `TPProcessEdges` type, which performs the transitively pinning trace (and is unimplemented for unsupported plans).
- `create_process_node_roots_work` creates work to process roots in the `NodeRootsTrace` bucket, which is executed after `TPClosure`, but any work derived from it is put into the regular `Closure` bucket, meaning that the binding shouldn't need to pin root nodes.
- For sticky immix, currently we only support non-moving nursery collections (`sticky_immix_non_moving_nursery`), but the solution here is to use the `TPClosure` on the modified buffer constructed from the write barrier (since during nursery collection we do not know if these objects should be transitively pinned or not).


src/scheduler/gc_work.rs
```diff
@@ -683,40 +709,73 @@ impl<VM: VMBinding> ProcessEdgesWork for SFTProcessEdges<VM> {
...
     fn create_process_node_roots_work(&mut self, nodes: Vec<ObjectReference>) {
-        // We want to use E::create_scan_work.
-        let process_edges_work = E::new(vec![], true, self.mmtk);
-        let work = process_edges_work.create_scan_work(nodes, true);
-        crate::memory_manager::add_work_packet(self.mmtk, WorkBucketStage::Closure, work);
+        crate::memory_manager::add_work_packet(
+            self.mmtk,
+            WorkBucketStage::Closure,
+            ProcessRootNode::<VM, I, E>::new(nodes, WorkBucketStage::Closure),
+        );
```
wks, line 751: The `create_process_node_roots_work` is supposed to list "black roots" which pin the objects in `nodes` but not their children.  The objects in `nodes` must not move.  However, if you put the `ProcessRootNode` work packet in `Closure`, other work packets that can move objects may be executed before this `ProcessRootNode` work packet.  When this `ProcessRootNode` work packet is executed, it may find some objects in `nodes` are already moved.

The right way is putting `ProcessRootNode` in a new `NTPinTrace` bucket (see comments in `work_bucket.rs`), but putting its derived `ScanObjects` and `ProcessEdgesWork` work packets into the regular `Closure` bucket.  In this way, it will guarantee that objects in `nodes` are visited before other non-pinning edges visit those objects.

```suggestion
        crate::memory_manager::add_work_packet(
            self.mmtk,
            WorkBucketStage::NTPinTrace,
            ProcessRootNode::<VM, I, E>::new(nodes, WorkBucketStage::Closure),
        );
```

src/scheduler/gc_work.rs
```diff
@@ -683,40 +709,73 @@ impl<VM: VMBinding> ProcessEdgesWork for SFTProcessEdges<VM> {
...
+            WorkBucketStage::ImmovableClosure,
+            I::new(edges, true, self.mmtk, WorkBucketStage::ImmovableClosure),
+        );
+    }
+
+    fn create_immovable_process_node_roots_work(&mut self, nodes: Vec<ObjectReference>) {
+        crate::memory_manager::add_work_packet(
+            self.mmtk,
+            WorkBucketStage::ImmovableClosure,
+            ProcessRootNode::<VM, I, I>::new(nodes, WorkBucketStage::ImmovableClosure),
```
wks, line 766: This is right.  Neither the packet itself nor its derived packets may move objects.

src/scheduler/work_bucket.rs
```diff
@@ -232,6 +232,8 @@ pub enum WorkBucketStage {
     /// Clear the VO bit metadata.  Mainly used by ImmixSpace.
     #[cfg(feature = "vo_bit")]
     ClearVOBits,
+    /// Compute the transtive closure following only strong references, making sure that objects do not move
+    ImmovableClosure,
```
wks, line 236: One `ImmovableClosure` may not be enough.  We need two.

```suggestion
    /// Compute the transtive closure starting from t-pin (transitively pinning) roots, following only strong references.  No objects in this transitive closure may move.
    TPinClosure,
    /// Trace nt-pin (non-transitively pinning) roots, but do not compute transitive closure.  Nt-pin roots may not move, but their children may move.
    NTPinTrace,
```

One reason is that black (nt-pin) roots must be traced before any other non-pinning edges.  Otherwise, the object may have been moved before we pin it.  If we put `ProcessRootNode` work packets for black roots into the `Closure` bucket, other non-pinning packets in the `Closure` buckets may be executed before this.  So `ProcessRootNode` work packets for black roots needs to be executed before the `Closure` bucket.

The other reason is that `ProcessRootNode` work packets for black roots must be executed after all objects in transitive closure of red (t-pin) roots are visited.

1.  An object may be pointed by both a black root and a red root.  Suppose we have two variables, one C local variable and a Julia local variable.  Then the C variable will be red and the Julia variable will be black. (Correct me if I am wrong.)  It is possible that both variables point to the same object. When we scan stacks, both variables will be visited, and they will be added to two `ProcessRootNode` work packets, one for red roots, and the other for black roots.
2.  If both work packets are in the same bucket, the black (nt-pin) packet may be executed first.  It marks the object.  Then the read packet is executed.  When visiting the same object, it finds the object is already marked, and will not enqueue its children into the red transitive closure.  That will cause the children of that object not to be pinned, even though it needs transitive pinning.

For those reasons, we need two buckets: `TPinClosure` and `NTPinTrace` before the `Closure` bucket.  The `NTPinTrace` bucket is not named `*****Closure` because the children of black (nt-pin) roots belong to the regular non-pinned transitive closure.

wks (changes requested):
There is a correctness issue regarding the processing of black (non-transitively pinning) roots.  See comments.

qinsoon:
I think we should stop using red/black roots in the PR, and in future discussion just to avoid any confusion. As we discussed, we will call it `TransitivelyPin`, `TPin`, or `TP`.

src/scheduler/gc_work.rs
qinsoon: For normal root nodes, we can put the `ProcessRootNode` work to the `TPinClosure`, but allow the packets generated from the work to  be put to the normal closure bucket. In this case, we do not need a third closure bucket other than TPin closure and normal closure.

wks: > For normal root nodes, we can put the `ProcessRootNode` work to the `TPinClosure`, but allow the packets generated from the work to be put to the normal closure bucket. In this case, we do not need a third closure bucket other than TPin closure and normal closure.

That won't work, for the same reason why black roots cannot be put into TPinClosure.  If an object is traced from a normal root before it is traced from a TPin (red) root, its children will be put into `Closure`.  When it is then traced from a TPin (red) root, it will find it already marked, and will not enqueue its children in `TPinClosure`.  As a result, its children will not be pinned, but they should be pinned because they are reachable from a TPin (red) root.

qinsoon: I see.

udesou: I've created a `NodeRootsTrace` to try to fix the issue (see the updated code). I called it NodeRoots because it is really specific to node roots, in which the root node shouldn't move, but there is no restriction about their transitive closure.

wks:
The `GCWorkContext` trait is just a collection of associated types.  The concrete types that implement it should never be instantiated, and is usually implemented with structs that do not have any fields (except `PhantomData`).  It doesn't make sense for its instances to implement `Send` and `'static`.  However, some work packets are parameterised with `C: GCWorkContext`, and Rust uses the type parameters to determine whether the work packet type automatically implement `Send`.

To workaround this, we can let the `GCWorkContext` trait require `Send + 'static`.  Because none of its concrete implementations have actual fields, they should automatically implement `Send + 'static`, too.  Then we can remove `+ Send + 'static` from `impl<C: GCWorkContext + Send + 'static> GCWork<C::VM> for SomeWorkPacketType<C>`

You can cherry-pick this commit: https://github.com/wks/mmtk-core/commit/6f79aea72fb91dd34008c4a20bf0bf750cba80e0

src/vm/scanning.rs
```diff
@@ -107,16 +107,32 @@ pub trait RootsWorkFactory<ES: Edge>: Clone + Send + 'static {
...
+    /// The work packet must not update the edges transitively found from any of the root edges.
+    ///
+    /// Arguments:
+    /// * `edges`: A vector of edges.
+    fn create_process_tpinned_edge_roots_work(&mut self, edges: Vec<ES>);
+
+    /// Create work packets to handle nodes pointed by root edges.
+    ///
+    /// Similar to `create_process_pinned_roots_work`, this work packet won't move root objects, but also will
+    /// not move any object in their transitive closure.
```
wks, line 131: Emphasise the kind of roots in the first sentence.

```suggestion
    /// Create work packets to handle transitively pinning (TP) roots.
    ///
    /// Similar to `create_process_pinning_roots_work`, this work packet will not move objects in `nodes`.
    /// Unlike ``create_process_pinning_roots_work`, no objects in the transitive closure of `nodes` will be moved, either.
```

wks (changes requested):
## The word "pinning" vs "pinned"

The verb "pin", when used in the root-scanning context, should be used as "a root pins an object".  So we should have "pinning roots" and "pinned objects".

## The phrase "node root" and "root node"

"Node root" and "root node" are not formal GC terms.  According to the GC Handbook, in an object graph, a node is an object, and an edge points from a source node or a root to a target node.  So roots are not nodes.  I felt a bit guilty for introducing the method `create_process_node_root_work` because it is a bit misleading.

We may use those terms (such as in `ProcessRootNode`) as a concise way to express "nodes pointed by root edges".  But that is an implementation detail.

When describing work bucket stages, it's better to say "this stage processes pinning roots" instead of "... processes node roots".

As for `RootsWorkFactory`, I suggest we just rename its methods as

-   `create_process_roots_work(edges: Vec<Edges>)`
-   `create_process_pinning_roots_work(nodes: Vec<ObjectReference>)`
-   `create_process_tpinning_roots_work(nodes: Vec<ObjectReference>)`

I think that's enough.  We don't need to mention edges or nodes because those roots can be expressed in either way, as edges, or as objects pointed by those edges.  We just make it clear in the comments.

qinsoon:
When all the review comments are resolved, we need to measure performance for the PR.

src/plan/generational/immix/gc_work.rs
```diff
@@ -9,6 +10,7 @@ impl<VM: VMBinding> crate::scheduler::GCWorkContext for GenImmixNurseryGCWorkCon
     type VM = VM;
     type PlanType = GenImmix<VM>;
     type ProcessEdgesWorkType = GenNurseryProcessEdges<VM, Self::PlanType>;
+    type TPProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType>;
```
qinsoon, line 13: This should be `UnsupportedProcessEdges`. Genimmix uses a copying nursery which cannot pin objects.

src/plan/generational/immix/gc_work.rs
```diff
@@ -20,4 +22,5 @@ impl<VM: VMBinding, const KIND: TraceKind> crate::scheduler::GCWorkContext
     type VM = VM;
     type PlanType = GenImmix<VM>;
     type ProcessEdgesWorkType = PlanProcessEdges<VM, GenImmix<VM>, KIND>;
+    type TPProcessEdges = PlanProcessEdges<VM, GenImmix<VM>, TRACE_KIND_TRANSITIVE_PIN>;
```
qinsoon, line 25: Same here. This should be `UnsupportedProcessEdges`.

src/scheduler/gc_work.rs
```diff
@@ -1030,3 +1078,121 @@ impl<E: ProcessEdgesWork, P: Plan<VM = E::VM> + PlanTraceObject<E::VM>> GCWork<E
         trace!("PlanScanObjects End");
     }
 }
+
+struct ProcessRootNode<VM: VMBinding, E1: ProcessEdgesWork<VM = VM>, E2: ProcessEdgesWork<VM = VM>>
```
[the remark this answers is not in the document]
qinsoon: Need some comments for the type as well: what does this type do, when is it used, why do we need it.

src/scheduler/gc_work.rs
```diff
@@ -1030,3 +1078,121 @@ impl<E: ProcessEdgesWork, P: Plan<VM = E::VM> + PlanTraceObject<E::VM>> GCWork<E
...
+
+        let process_edges_work = E2::new(vec![], false, mmtk, self.bucket);
+        let work = process_edges_work.create_scan_work(scanned_root_objects, false);
+        crate::memory_manager::add_work_packet(mmtk, self.bucket, work);
+
+        trace!("ProcessRootNode End");
+    }
+}
+
+// // create UnsupportedProcessEdge
```
qinsoon, line 1154:
```suggestion
/// A `ProcessEdgesWork` type that panics when any of its method is used.
/// This is currently used for plans that do not support transitively pinning.
```

src/policy/copyspace.rs
```diff
@@ -126,6 +127,10 @@ impl<VM: VMBinding> crate::policy::gc_work::PolicyTraceObject<VM> for CopySpace<
         copy: Option<CopySemantics>,
         worker: &mut GCWorker<VM>,
     ) -> ObjectReference {
+        assert!(
```
qinsoon, line 130: This is performance critical code. So use `debug_assert` or `static_assertions` instead.

src/policy/markcompactspace.rs
```diff
@@ -128,6 +128,10 @@ impl<VM: VMBinding> crate::policy::gc_work::PolicyTraceObject<VM> for MarkCompac
         _copy: Option<CopySemantics>,
         _worker: &mut GCWorker<VM>,
     ) -> ObjectReference {
+        assert!(
```
qinsoon, line 131: This is performance critical code. So use `debug_assert` or `static_assertions` instead.

src/scheduler/gc_work.rs
```diff
@@ -749,61 +810,25 @@ pub trait ScanObjectsWork<VM: VMBinding>: GCWork<VM> + Sized {
...
     /// The common code for ScanObjects and PlanScanObjects.
     fn do_work_common(
         &self,
         buffer: &[ObjectReference],
         worker: &mut GCWorker<<Self::E as ProcessEdgesWork>::VM>,
-        mmtk: &'static MMTK<<Self::E as ProcessEdgesWork>::VM>,
+        _mmtk: &'static MMTK<<Self::E as ProcessEdgesWork>::VM>,
     ) {
         let tls = worker.tls;
+        assert!(!self.roots());
```
qinsoon, line 823: I am not sure about performance implication here. But `debug_assert!` should be a safer option.

src/policy/immix/immixspace.rs
```diff
@@ -618,11 +620,11 @@ impl<VM: VMBinding> ImmixSpace<VM> {
...
-            debug_assert!(
-                nursery_collection || self.defrag.space_exhausted() || self.is_pinned(object),
-                "Forwarded object is the same as original object {} even though it should have been copied",
-                object,
-            );
+            // debug_assert!(
+            //     nursery_collection || self.defrag.space_exhausted() || self.is_pinned(object),
+            //     "Forwarded object is the same as original object {} even though it should have been copied",
+            //     object,
+            // );
```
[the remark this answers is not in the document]
udesou: I'm not sure I understand how to do this. This would trigger also after then TP closure (in fact, only after), when the objects have been marked already by the TP closure, which is the only reason why we're not moving the object. When we're doing the TP closure, we call into `trace_object_without_moving`.

udesou:
I've addressed the comments above. @wks and @qinsoon, feel free to have a look when you can.

src/policy/immix/immixspace.rs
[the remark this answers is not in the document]
qinsoon: You could add a parameter to the method (like the `nursery_collection` parameter) so we know whether this is in a TP closure and use that in the assertion. Probably discuss with Kunshan on this.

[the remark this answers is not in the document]
udesou: But this will run _after_ the TP closure, as far as I understand. The TP closure will mark things in place, this code will see that these objects have been marked by the TP closure and therefore have not been moved - which would fail the assertion.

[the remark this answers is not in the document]
udesou: IMO, the only way we could keep this assertion is when we're freely able to move objects opportunistically. However, as soon as we introduce pinning roots or transitively pinning roots, we cannot have the assertion, since the previous trace could have already marked the object in place (either the pinning or the transitively pinning one).

udesou:
> When all the review comments are resolved, we need to measure performance for the PR.

Assuming the tests succeed and we're only disagreeing on names/comments, @qinsoon would you mind doing a performance run for these changes to see if there's any regression?

qinsoon:
> > When all the review comments are resolved, we need to measure performance for the PR.
> 
> Assuming the tests succeed and we're only disagreeing on names/comments, @qinsoon would you mind doing a performance run for these changes to see if there's any regression?

Yeah, after the tests pass and binding tests pass, I will run some performance evaluation on this.

src/policy/immix/immixspace.rs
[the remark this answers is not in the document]
wks: I see.  Because the object may have been marked by a different worker, we no longer know why the object was pinned by that worker because that information was not recorded.  It was possible to do the assertion because the condition of when to move and when not to move was simple.  With TP closure introduced, it becomes a bit more dynamic.

I am OK with removing this assertion.

wks (approved):
LGTM.

qinsoon (changes requested):
All the bindings need to be updated for the API change.

udesou:
> All the bindings need to be updated for the API change.

That is:

- [x] ruby (Thanks @wks)
- [ ] Julia
- [ ] openJDK
- [ ] v8
- [ ] Jikes RVM 

Do we do anything about the others such as GCH or PyPy (notify their main developers at least?)

wks:
I have changed the Ruby binding for this PR.  https://github.com/mmtk/mmtk-ruby/pull/29

When doing binding test, use the following lines in the comment

RUBY_BINDING_REF=feature/rb-roots

qinsoon:
> Do we do anything about the others such as GHC or PyPy (notify their main developers at least?)

No, we don't need to.

> Edit: I think only Julia and Ruby are using API methods that were changed in this PR, but I've created PRs for the others just to make sure we don't get any errors with the changes that were made.

Yeah. I just realized the other bindings only use `create_process_roots_work` which is not changed in the PR.

qinsoon:
binding-refs
OPENJDK_BINDING_REPO=udesou/mmtk-openjdk
OPENJDK_BINDING_REF=fix/upstream-api-change-pr897
JIKESRVM_BINDING_REPO=udesou/mmtk-jikesrvm
JIKESRVM_BINDING_REF=fix/upstream-api-change-pr897
V8_BINDING_REPO=udesou/mmtk-v8
V8_BINDING_REF=fix/upstream-api-change-pr897
JULIA_BINDING_REPO=udesou/mmtk-julia
JULIA_BINDING_REF=fix/upstream-api-change-pr897
RUBY_BINDING_REF=feature/rb-roots

qinsoon:
The Julia tests failed, as the Julia PR did not include the change about its tests so it timed out. I don't know what happened for the Ruby tests.

udesou:
> The Julia tests failed, as the Julia PR did not include the change about its tests so it timed out. I don't know what happened for the Ruby tests.

@wks I'm also not familiar with that error. Do you have any idea why the ruby tests failed to run?

wks:
> @wks I'm also not familiar with that error. Do you have any idea why the ruby tests failed to run?

When running binding tests for mmtk-core, I used a Python script to replace the mmtk dependency with a local directory.  See: https://github.com/mmtk/mmtk-core/blob/master/.github/scripts/replace-mmtk-dep.py

This script works by deleting the `dependencies.mmtk.git` node and adding a `dependencies.mmtk.path` node.  That worked if the `Cargo.toml` is like what it is in the master branch of `mmtk-ruby`, i.e. the `dependencies.mmtk` node has exactly one child, namely `git`.

However, there is a problem.  In my `mmtk-ruby` PR that changes `mmtk-ruby` for this PR, I modified `Cargo.toml` to also include a `branch` node (See https://github.com/wks/mmtk-ruby/blob/feature/rb-roots/mmtk/Cargo.toml):

```toml
[dependencies.mmtk]
features = ["is_mmtk_object", "object_pinning"]

# Uncomment one of the following lines to choose where to find mmtk-core.
#git = "https://github.com/mmtk/mmtk-core.git"  # Use mmtk-core from the official repository.
#path = "../../mmtk-core"  # Use mmtk-core from a local repository.
git = "https://github.com/udesou/mmtk-core"  # Use mmtk-core with rb-roots feature
branch = "feature/rb-roots"
```

By removing the `git` node and adding a `path` node, it will look like

```toml
branch = "feature/rb-roots"
path = "/home/wks/projects/mmtk-github/mmtk-core"
```

`branch` and `path` cannot co-exist.  Therefore, Cargo complained that "key `branch` is ignored for dependency (mmtk)."

It is a bug in https://github.com/mmtk/mmtk-core/blob/master/.github/scripts/replace-mmtk-dep.py

udesou:
> > @wks I'm also not familiar with that error. Do you have any idea why the ruby tests failed to run?
> 
> When running binding tests for mmtk-core, I used a Python script to replace the mmtk dependency with a local directory. See: https://github.com/mmtk/mmtk-core/blob/master/.github/scripts/replace-mmtk-dep.py
> 
> This script works by deleting the `dependencies.mmtk.git` node and adding a `dependencies.mmtk.path` node. That worked if the `Cargo.toml` is like what it is in the master branch of `mmtk-ruby`, i.e. the `dependencies.mmtk` node has exactly one child, namely `git`.
> 
> However, there is a problem. In my `mmtk-ruby` PR that changes `mmtk-ruby` for this PR, I modified `Cargo.toml` to also include a `branch` node (See https://github.com/wks/mmtk-ruby/blob/feature/rb-roots/mmtk/Cargo.toml):
> 
> ```toml
> [dependencies.mmtk]
> features = ["is_mmtk_object", "object_pinning"]
> 
> # Uncomment one of the following lines to choose where to find mmtk-core.
> #git = "https://github.com/mmtk/mmtk-core.git"  # Use mmtk-core from the official repository.
> #path = "../../mmtk-core"  # Use mmtk-core from a local repository.
> git = "https://github.com/udesou/mmtk-core"  # Use mmtk-core with rb-roots feature
> branch = "feature/rb-roots"
> ```
> 
> By removing the `git` node and adding a `path` node, it will look like
> 
> ```toml
> branch = "feature/rb-roots"
> path = "/home/wks/projects/mmtk-github/mmtk-core"
> ```
> 
> `branch` and `path` cannot co-exist. Therefore, Cargo complained that "key `branch` is ignored for dependency (mmtk)."
> 
> It is a bug in https://github.com/mmtk/mmtk-core/blob/master/.github/scripts/replace-mmtk-dep.py

I see. But the local directory would take the `branch` node into consideration when rewriting from `git` to `path`? i.e., is this just a matter of deleting the `branch`node after the rewriting?

wks:
> I see. But the local directory would take the `branch` node into consideration when rewriting from `git` to `path`? i.e., is this just a matter of deleting the `branch`node after the rewriting?

Yes.  I made a PR to fix it. (https://github.com/mmtk/mmtk-core/pull/947) Now the script simply remove all of "git", "branch", "version" and "registry".

But it should be easier to just cherry-pick https://github.com/mmtk/mmtk-core/pull/947 into this PR and run the CI again.

udesou:
Done. Thank you @wks!

qinsoon:
> Details

```console
warning: spurious network error (3 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
warning: spurious network error (2 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
warning: spurious network error (1 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
```

Seems to be a network issue of the cargo registry. Probably just rerun it later.

qinsoon:
I ran benchmarks for the PR with OpenJDK/Immix/2.5x MC min heap ([plotty link](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2023-09-08-Fri-064233&benchmark^build^invocation^iteration&time^time.other^time.stw&|10&iteration^1^2&benchmark^2^batik&benchmark^2^jme|20&1^invocation|30&1&benchmark&build;jdk-mmtk-master|41&Histogram%20(with%20CI)^build^benchmark&)). It measures the impact of the code changes and extra buckets from the PR, but it does not measure the transitively pinning. The results show there is no slowdown in this PR for introducing transitively pinning.

Mutator time
pr897-mutator

GC time
pr897-gc

qinsoon (approved):
LGTM

PR #898, opened 2023-08-11
Add tracing tools and documentation

caizixian:
These tools utilize the tracepoints added in #883. The GC visualization tools requires some post-processing and is just a bit more complicated in general. That will be added in a separate PR.


wks:
Directory structure: I think this tracing tool will not be the only tool we provide.  I recommend adding one level of sub-directory, such as `mmtk-core/tools/tracing/`

tools/run.py
```diff
@@ -0,0 +1,57 @@
...
+        template = Template(prologue + fd.read() + epilogue)
+        with tempfile.NamedTemporaryFile(mode="w+t") as tmp:
+            content = template.safe_substitute(
+                MMTK=mmtk_bin, TMP_FILE=tmp.name)
+            if args.print_script:
+                print(content)
+            tmp.write(content)
+            tmp.flush()
+            os.execvp("sudo", ["sudo", args.bpftrace,
+                      "--unsafe", "-f", args.format, tmp.name])
```
wks, line 53: Python will automatically delete the `NamedTemporaryFile` when exiting the scope of `with`.  However, `execvp` never returns.  You can replace this with `subprocess.run` or `subprocess.Popen`.

An alternative to using temp file is using pipes.  I tried piping the script into `sudo bpftrace` and it works.  For example, I am currently using `erb` (from Ruby) for templating, and my command line looks like this:

```shell
erb record_every=50 so_path=/home/wks/projects/mmtk-github/openjdk/build/linux-x86_64-normal-server-release/jdk/lib/server/libmmtk_openjdk.so  gcvis.erb | sudo bpftrace - > openjdk-packet-size3-nopoll.log
```

In Python, [the `subprocess.run` function](https://docs.python.org/3/library/subprocess.html#subprocess.run) (or `popen.communicate`) allows you to specify the stdin as a pipe, so you can directly pipe the Python string `prologue + content + epilogue` into `sudo bpftrace`.

tools/epilogue.bt.fragment
```diff
@@ -0,0 +1,7 @@
+END {
+    system("rm $TMP_FILE");
+}
```
wks, line 3: We only need to delete the temp file if we use exec.  See the comment on `run.py`.

tools/README.md
```diff
@@ -0,0 +1,193 @@
...
+- `-H`: pass this flag is you want to only measure the timing iteration of a benchmark. By default, the tracing tools will measure the entire execution.
+- `-p`: print the entire tracing script before execution. This is mainly for debugging use.
+- `-f`: change the bpftrace output format. By default, it uses human-readable plain text output (`text`). You can set this to `json` for easy parsing.
+
+Please run the tracing tools **before** running the workload. If you use `-H`, the tracing tools will automatically end with `harness_end` is called. Otherwise, you will need to terminate the tools manually with `Ctrl-C`. These tools also have a timeout of 1200 seconds so not to stall unattended benchmark execution. 
+
+## Tracing tools
+### Measuring the time spend in allocation slow path (`alloc_slow`)
+This tool measures the distribution of the allocation slow path time.
+The time unit is ns / 400, so that we use the histogram bins with higher fidelity better.
```
wks, line 56: "The time unit is ns / 400" means "1 in the output means (1/400) ns, i.e. 0.0025 nanoseconds, i.e. 2.5 picoseconds".  That's probably not what the output really represents.  The next paragraph says it is 3.2 us to 6.4 us, i.e  8 * 400 ns to 16 * 400 ns.  So the unit should be "400 ns".

```suggestion
The time unit is 400 ns, so that we use the histogram bins with higher fidelity better.
```

tools/README.md
```diff
@@ -0,0 +1,193 @@
...
+#[cold]
+fn lock_contended(&self) {
+    <snip>
+}
+```
+
+
+MMTk uses Rust `Mutex`, e.g., in allocation slow paths for synchronization, and this tool can be useful to measure the contention in these parts of code.
+
+The time unit is ns / 256.
```
wks, line 114: The same unit problem.

tools/README.md
```diff
@@ -0,0 +1,193 @@
...
+[32K, 64K)            37 |                                                    |
+[64K, 128K)           19 |                                                    |
+[128K, 256K)           1 |                                                    |
+```
+
+In the above output, we can see that most allocation slow paths finish between 3.2us and 6.4us. However, there is a long tail, presumably due to GC pauses.
+
+### Measuring the time spend in different GC stages (`gc_stages`)
+This tool measures the time spent in different stages of GC: before `Closure`, during `Closure`, and after `Closure`.
+The time unit is ns.
```
wks, line 83: A general style problem: I find that you never uses newline characters within one paragraph.  Markdown allows this.

But here the text spans over three lines.  I think you intend to introduce one title and two paragraphs.  However, Markdown will interpret consecutive non-blank lines as one paragraph (except for titles (`### blah blah blah`)).  If you intend to put "The time unit is ns." in a single paragraph, you need to add an empty line before it.

I also recommend formatting the text so that the line width is no more than 100 characters long (or 80 or 72 if you prefer).  The `fmt` command line utility may come in handy.  If you are a Vim user, you can `set tw=100`, select one or more paragraphs and press `gq` to format them.  In this way, this text file will still be easy to read using text editors.

tools/README.md
```diff
@@ -0,0 +1,193 @@
...
+[64, 128)           2991 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@                     |
+[128, 256)          4846 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@  |
+[256, 512)          5013 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
+[512, 1K)           1203 |@@@@@@@@@@@@                                        |
+[1K, 2K)              34 |                                                    |
+[2K, 4K)              15 |                                                    |
+```
+
+In the above output, we can see that the lock instance (140637228007056, or 0x7fe8a8047e90) roughly has a bimodal distribution in terms of the time spent in lock contended code path. The first peak is around 512ns\~1024ns, and the second peak is around 66us\~131us.
+If you can't tell which lock instance is for which lock in MMTk, you can trace the allocation of the Mutex and record the stack trace (note that you might want to compile MMTk with `force-frame-pointers` to obtain better stack traces).
```
wks, line 134: If they are two paragraphs, there needs to be an empty line in between.

tools/run.py
```diff
@@ -0,0 +1,57 @@
...
+        print(f"Tracing script {str(bpftrace_script)} not found.")
+        sys.exit(1)
+    mmtk_bin = Path(args.mmtk)
+    if not mmtk_bin.exists():
+        print(f"MMTk binary {str(mmtk_bin)} not found.")
+        sys.exit(1)
+    prologue_file = here / \
+        "prologue_with_harness.bt.fragment" if args.harness else "prologue_without_harness.bt.fragment"
+    with prologue_file.open() as fd:
+        prologue = fd.read()
```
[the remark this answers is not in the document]
caizixian: Fixed

tools/run.py
```diff
@@ -0,0 +1,57 @@
...
+    with (here / "epilogue.bt.fragment").open() as fd:
+        epilogue = fd.read()
+    with bpftrace_script.open() as fd:
+        template = Template(prologue + fd.read() + epilogue)
+        with tempfile.NamedTemporaryFile(mode="w+t") as tmp:
```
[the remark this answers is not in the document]
caizixian: Fixed

tools/run.py
caizixian: That's a deliberate choice. I will add comments to explain this.

tools/epilogue.bt.fragment
caizixian: See above

tools/README.md
caizixian: Fixed

tools/README.md
caizixian: Fixed

tools/README.md
caizixian: Fixed.

tools/README.md
caizixian: Done

caizixian:
> Directory structure: I think this tracing tool will not be the only tool we provide. I recommend adding one level of sub-directory, such as `mmtk-core/tools/tracing/`

Yeah, that makes sense. Fixed.

steveblackburn:
I am confused about the time "units".   Do you really mean _units_ or do you mean _resolution_?  The text seems to use seconds (or  us / ns) as the _units_.

Other than that, this looks excellent.  Nicely documented and packaged.

caizixian:
> I am confused about the time "units". Do you really mean _units_ or do you mean _resolution_? The text seems to use seconds (or us / ns) as the _units_.
> 
> Other than that, this looks excellent. Nicely documented and packaged.

@steveblackburn Yes I meant units. Just to clarify, it is about how much is one time unit in the tracing output. We sometimes scale the numbers to have better resolution on the log2-based histogram, where the numbers on the histogram bins are in the time unit. This is different from the measuring resolution,

tools/tracing/run.py
```diff
@@ -0,0 +1,63 @@
...
+    bpftrace_script = here / f"{args.tool}.bt"
+    if not bpftrace_script.exists():
+        print(f"Tracing script {str(bpftrace_script)} not found.")
+        sys.exit(1)
+    mmtk_bin = Path(args.mmtk)
+    if not mmtk_bin.exists():
+        print(f"MMTk binary {str(mmtk_bin)} not found.")
+        sys.exit(1)
+    prologue_file = here / \
+        "prologue_with_harness.bt.fragment" if args.harness else "prologue_without_harness.bt.fragment"
```
wks, line 38: The bracket is necessary because the `/` operator has higher precedence than the `x if c then y` expression.  Otherwise if `args.harness` is false, `prologue_file` will be the string `"prologue_without_harness.bt.fragment"`.

```suggestion
        ("prologue_with_harness.bt.fragment" if args.harness else "prologue_without_harness.bt.fragment")
```

tools/tracing/lock_contended.bt
```diff
@@ -0,0 +1,11 @@
+uprobe:$MMTK:_ZN3std3sys4unix5locks11futex_mutex5Mutex14lock_contended17he467d89501d7157aE {
```
wks, line 1: The last part `17he467d89501d7157aE` is specific to the build.  In my local build, the same function is suffixed with `17h2a372fcb57338e4dE`.  Fortunately, `bpftrace` supports wildcard in function names.

```suggestion
uprobe:$MMTK:_ZN3std3sys4unix5locks11futex_mutex5Mutex14lock_contended* {
```

tools/tracing/lock_contended.bt
```diff
@@ -0,0 +1,11 @@
...
+    if (@stats_enabled) {
+        @lock_nsecs[tid] = (arg0, nsecs);
+    }
+}
+
+uretprobe:$MMTK:_ZN3std3sys4unix5locks11futex_mutex5Mutex14lock_contended17he467d89501d7157aE {
```
wks, line 7:
```suggestion
uretprobe:$MMTK:_ZN3std3sys4unix5locks11futex_mutex5Mutex14lock_contended* {
```

wks (changes requested):
Just a few more things to fix.

wks (approved):
LGTM

PR #899, opened 2023-08-14
Boot-time configurable heap constants

wenyuzhao:
This CL refactors the `vm_layout_constants` related code and changes all the related compile-time constants to run-time constants, so that VM bindings can configure them during boot time based on dynamic user inputs.

This is a necessary refactoring to support compressed pointers. For most of the VMs (e.g. OpenJDK), compressed pointers are enabled at run-time based on command line arguments, and VM bindings should be able to enable compressed pointer heap dynamically during the VM boot process. Such a process usually involves dynamically configuring heap ranges, space memory layout, and choosing a different `VMMap` implementation [1]. Thus all `vm_layout_constants`, including heap constants, should be made as boot-time configurable, instead of compile-time constants.

This CL also makes risc-v's 39-bit heap support easier, as it now should only involve adding a new set of `vm_layout_constants`.

[1] This is already done by https://github.com/mmtk/mmtk-core/pull/732

---

Benchmark results: Almost no performance impact

http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/NkK9YG


caizixian:
@wenyuzhao can you suppress some warnings on x86?

caizixian:
V8 binding tests will fail until #472 is merged

qinsoon:
Can you run performance tests for the PR? 

I am wondering if we would need a feature like `runtime_vm_layout`. Without the feature, we can just have a const `VMLayoutConstants`.

wenyuzhao:
> Can you run performance tests for the PR?

will do

> I am wondering if we would need a feature like `runtime_vm_layout`. Without the feature, we can just have a const `VMLayoutConstants`.

Yes, this is possible. Let's wait for the performance test results and see if it is worth doing so.

src/util/heap/layout/vm_layout_constants.rs
```diff
@@ -120,48 +34,175 @@ pub const LOG_PAGES_IN_SPACE64: usize = LOG_SPACE_SIZE_64 - LOG_BYTES_IN_PAGE as
...
+            space_size_64: 1 << 41,
+        }
+    }
+    /// 64-bit configuration with compressed pointers
+    pub fn new_64bit_with_pointer_compression(heap_size: usize) -> Self {
+        assert!(
+            heap_size <= (32usize << LOG_BYTES_IN_GBYTE),
+            "Heap size is larger than 32 GB"
+        );
+        let start = 0x4000_0000;
```
k-sareen, line 143: Ideally we allow a starting address to be configured instead of a hardcoded starting address

src/util/heap/layout/vm_layout_constants.rs
```diff
@@ -120,48 +34,175 @@ pub const LOG_PAGES_IN_SPACE64: usize = LOG_SPACE_SIZE_64 - LOG_BYTES_IN_PAGE as
...
+        let end = match start + heap_size {
+            end if end <= (4usize << 30) => 4usize << 30,
+            end if end <= (32usize << 30) => 32usize << 30,
+            _ => 0x4000_0000 + (32usize << 30),
+        };
+        Self {
+            log_address_space: 35,
```
k-sareen, line 150: Pointer compression doesn't necessarily have to be 35-bit address space. Android uses 32-bit address space. I was using an older version of this patch and set the address space to 32-bits here: https://github.com/k-sareen/mmtk-core/blob/b68292cbe8c875906cf19501c34641628c9fbc3c/src/util/heap/layout/vm_layout_constants.rs#L111

qinsoon:
> Do we expect the JikesRVM binding test to pass?

It may occasionally fail in mark sweep: https://github.com/mmtk/mmtk-jikesrvm/issues/108

src/util/heap/layout/vm_layout_constants.rs
wenyuzhao: Right. I also hardcoded a few OpenJDK-specific heap range optimizations here.

I moved the code to openjdk binding. Bindings can now call `VMLayoutConstants::set_custom_vm_layout_constants` to configure the values to anything they want.

src/util/heap/layout/vm_layout_constants.rs
wenyuzhao: The new `VMLayoutConstants::set_custom_vm_layout_constants` interface should solve this as well. Bindings can specify a different `log_address_space`

qinsoon: I don't think we want to expose the entire `VMLayoutConstants` to the binding. The struct includes many fields that we do not want the binding to customize, and we cannot guarantee MMTk to work with arbitrary values from the bindings. I suggest exposing limited options through our runtime `Option`, such as `use_compressed_pointer_bits`. We only need to expose options that we are supporting right now. If we haven't tested things like `heap_start`, we do not need to expose them at this point.

wenyuzhao: runtime `Option` is user configurable, this is not a good place for VM configurable values. For most bindings there is no need to use this function, as the default 32bit or 64bit values should be enough. For bindings that use this function, they should make sure the heap constants are correct.

Also, I believe make VMs be able to configure `heap_start` is necessary. You can't just assume our default fixed `heap_start` can work for all the VMs without any virtual space conflict. I actually did solve a virtual address conflict issue on GoogleJDK by manually modifying `heap_start` in mmtk-core. If VM bindings has no control to heap start address, this will happen again to other VMs.

qinsoon: > runtime `Option` is user configurable, this is not a good place for VM configurable values. For most bindings there is no need to use this function, as the default 32bit or 64bit values should be enough. For bindings that use this function, they should make sure the heap constants are correct.

If runtime option is not suitable, we can use constants on `VMBinding`. My point is we should only expose values that we want the bindings to customize.

> Also, I believe make VMs be able to configure `heap_start` is necessary. You can't just assume our default fixed `heap_start` can work for all the VMs without any virtual space conflict. I actually did solve a virtual address conflict issue on GoogleJDK by manually modifying `heap_start` in mmtk-core. If VM bindings have no control to heap start address, this will happen again to other VMs.

Yeah. We can expose `heap_start` if you tested that our code base works with an arbitrary heap start.

wenyuzhao: > if you tested that our code base works with an arbitrary heap start.

This can never be true. e.g. zero address can never work. I think it's VM binding's responsibility to ensure their configured heap range works with their virtual memory.

k-sareen: I agree with Wenyu here. It's up to the binding implementer to check that the heap start actually works for them. I had to mess around with the heap start for Android as well. But I think it makes sense to only expose a limited set of options/functions to the binding so that there is less chance of using stuff incorrectly.

src/util/heap/layout/vm_layout_constants.rs
```diff
@@ -120,48 +34,130 @@ pub const LOG_PAGES_IN_SPACE64: usize = LOG_SPACE_SIZE_64 - LOG_BYTES_IN_PAGE as
...
+    /// conditional definition.
+    pub space_mask_64: usize,
+    /// Size of each space in the 64-bit memory layout
+    /// We can't express this constant in a 32-bit environment, hence the
+    /// conditional definition.
+    /// FIXME: When Compiling for 32 bits this expression makes no sense
+    pub space_size_64: usize,
+    /// Should mmtk enable contiguous spaces and virtual memory for all spaces?
+    /// For normal 64-bit config, this should be set to true. Each space should own a contiguous piece of virtual memory.
+    /// FOr 32-bit or 64-bit compressed heap, we don't have enough virtual memory, so this should be set to false.
```
k-sareen, line 68:
```suggestion
    /// For 32-bit or 64-bit compressed heap, we don't have enough virtual memory, so this should be set to false.
```

wenyuzhao: Thanks. fixed.

src/util/heap/layout/vm_layout_constants.rs
```diff
@@ -120,48 +34,130 @@ pub const LOG_PAGES_IN_SPACE64: usize = LOG_SPACE_SIZE_64 - LOG_BYTES_IN_PAGE as
...
- *  subdivided into fields as follows
- *
- *    64                              0
- *    00...0SSSSSaaaaaaaaaaa...aaaaaaaa
- *
- * The field 'S' identifies the space to which the address points.
- */
+/// Runtime-initialized virtual memory constants
+#[derive(Clone)]
+pub struct VMLayoutConstants {
```
qinsoon, line 39: Consider renaming this. This is not a constant any more. `VMLayout` or `VMLayoutParameters` are good candidates.

src/memory_manager.rs
```diff
@@ -569,13 +568,13 @@ pub fn live_bytes_in_last_gc<VM: VMBinding>(mmtk: &MMTK<VM>) -> usize {
 /// Return the starting address of the heap. *Note that currently MMTk uses
 /// a fixed address range as heap.*
 pub fn starting_heap_address() -> Address {
-    HEAP_START
+    VM_LAYOUT_CONSTANTS.heap_start
```
qinsoon, line 571: This could accidentally get the `lazy_static` initialised. Consider adding an `&MMTK` reference so the binding won't call it before creating MMTk.

qinsoon:
I suggest adding some tests that only run on 64 bits architecture in `DummyVM` (vmbindings/dummyvm/src/tests): set a custom VM layout for the compressed pointer layout, initialize MMTk, create a mutator, and finish one allocation. The test should  run for all plans (Add `// GITHUB-CI: MMTK_PLAN=all` at the top of the test file)

src/memory_manager.rs
wenyuzhao: This breaks mmtk interface. Also I think a more generic and complete solution would be adding a check in `vm_layout()` getter and `set_custom_vm_layout`, to ensure that the setter can never be called after the first call of `vl_layout()`. This can also check and catch other uses of `vm_layout`, not only the ones here.

qinsoon: Sounds good to me. Those checks will have to be `debug_assert` if it is in the getter, right? I am okay with it.

wenyuzhao: Yes I put them within `cfg!(debug_assertions)`

src/util/heap/layout/vm_layout_constants.rs
```diff
@@ -120,48 +34,130 @@ pub const LOG_PAGES_IN_SPACE64: usize = LOG_SPACE_SIZE_64 - LOG_BYTES_IN_PAGE as
...
+            "Custom VM_LAYOUT_CONSTANTS can only be set once"
+        );
+        *guard = Some(constants);
+    }
+}
+
+static CUSTOM_CONSTANTS: Mutex<Option<VMLayoutConstants>> = Mutex::new(None);
+
+lazy_static! {
+    pub static ref VM_LAYOUT_CONSTANTS: VMLayoutConstants = {
```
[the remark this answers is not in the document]
wks: If it is determined at boot time, it is no longer constants.  So it always requires a memory load no matter if it is loaded from a global static struct or a field of MMTk.  I prefer calling them properties of a MMTk instance, and they should be stored in read-only fields in the MMTk instance, and propagated to sub-components if needed.

An alternative is making them part of the `VMBinding` trait so that they can really be constants.  If `MMTK<VM>` is parameterised by `<VM>`, the Rust compiler knows the constants in `VM` at compile time when specialising `MMTK` for the `VM`.  Even if those properties are expressed as methods in `VMBinding`, if the concrete implementation of those functions return constant values, the compiler can still inline those functions at compile time.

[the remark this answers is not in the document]
qinsoon: > If it is determined at boot time, it is no longer constants. So it always requires a memory load no matter if it is loaded from a global static struct or a field of MMTk. I prefer calling them properties of a MMTk instance, and they should be stored in read-only fields in the MMTk instance, and propagated to sub-components if needed.

The VM map depends on the VM layout. But our VM map is currently global. I don't think we can or we should make the layout as a field in an MMTk instance without changing the VM map. I would think we just leave the VM layout in the same way as the VM map, and we can refactor both in the future if we need to.

> An alternative is making them part of the `VMBinding` trait so that they can really be constants. If `MMTK<VM>` is parameterised by `<VM>`, the Rust compiler knows the constants in `VM` at compile time when specialising `MMTK` for the `VM`. Even if those properties are expressed as methods in `VMBinding`, if the concrete implementation of those functions return constant values, the compiler can still inline those functions at compile time.

Those are not constants. For OpenJDK, it is determined by a command line option `-XX:+UseCompressedOops`. I think we would see the performance results, then decide if runtime VM layout should be a feature or not. If the feature is enabled, we get the layout from the binding and store it as a variable (as it is now). If the feature is not enabled, the VM layout is a constant (either defined in MMTk core, or defined in the `VMBinding`).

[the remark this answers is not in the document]
k-sareen: Hm. I think that making this into a feature might be unnecessary. I would say it's pretty standard to want different starting heap addresses, address space sizes, etc.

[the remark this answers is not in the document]
qinsoon: > Hm. I think that making this into a feature might be unnecessary. I would say it's pretty standard to want different starting heap addresses, address space sizes, etc.

Yeah, we should allow bindings to set those info. The proposed feature is about whether we allow to set them at run-time (as opposed to compile-time). We should know if runtime VM layout introduces any overhead. If it introduces some overhead, it is reasonable to make it into a feature so it won't slow down bindings that do not need run-time VM layout.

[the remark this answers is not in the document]
k-sareen: Right. It would be good to test the performance of this change.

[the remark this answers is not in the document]
wenyuzhao: Just finished the benchmark. Looks like there is no noticeable slowdown

Immix: http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/NkK9YG
GenImmix: http://squirrel.anu.edu.au/plotty-public/wenyuz/v8/p/StQwU3

SS and GenSS are also included in the run, but they didn't finish all the benchmarks due to the small heap factor.

src/util/heap/space_descriptor.rs
```diff
@@ -36,12 +32,12 @@ impl SpaceDescriptor {
     pub const UNINITIALIZED: Self = SpaceDescriptor(0);
 
     pub fn create_descriptor_from_heap_range(start: Address, end: Address) -> SpaceDescriptor {
-        let top = end == vm_layout_constants::HEAP_END;
+        let top = end == VM_LAYOUT_CONSTANTS.heap_end;
         if cfg!(target_pointer_width = "64") {
```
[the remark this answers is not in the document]
wenyuzhao: Although this is not documented well, my understanding is that `create_descriptor_from_heap_range` can only be called on contiguous spaces (for both 32-bit and 64-bit architecture). The function implements two descriptor calculation policy, based on the architecture pointer width, i.e. `cfg!(target_pointer_width = "64")`. This is not depend on the VM layout constants.

src/util/heap/space_descriptor.rs
```diff
@@ -96,12 +92,21 @@ impl SpaceDescriptor {
 
     #[cfg(target_pointer_width = "64")]
```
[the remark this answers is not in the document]
wenyuzhao: Done

src/util/heap/space_descriptor.rs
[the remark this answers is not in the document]
wenyuzhao: `create_descriptor_from_heap_range` has only one caller, and the caller only call this for contiguous spaces.

src/util/heap/layout/vm_layout_constants.rs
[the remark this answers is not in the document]
k-sareen: Just checking that this is with or without PGO?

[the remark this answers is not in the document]
wenyuzhao: without pgo

src/util/heap/space_descriptor.rs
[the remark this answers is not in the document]
qinsoon: The current `master` code assumes "64 bits === 64 bits VM layout === contiguous space". But this assumption is no longer true with the changes in the PR.

In the case where we use the compressed pointer layout on a 64 bits machine, `space_shift_64` is not set for compressed pointer layout, and should not be used. But the code here still calculates the descriptor from `space_shift_64` which would be incorrect. I think we should use the code from L51 for the compressed pointer layout, right? This is the reason I suggested that it should depends on the layout rather than the pointer width.

src/policy/sft_map.rs
```diff
@@ -150,36 +154,37 @@ mod space_map {
         /// Currently our spaces are using address range 0x0000_0200_0000_0000 to 0x0000_2200_0000_0000 (with a maximum of 16 spaces).
         /// When masked with this constant, the index is 1 to 16. If we mask any arbitrary address with this mask, we will get 0 to 31 (32 entries).
         pub const ADDRESS_MASK: usize = 0x0000_3f00_0000_0000usize;
```
caizixian, line 156: We need to fix this as well right? An example is for sv39, we might still have the same number of spaces, but each space is smaller.

src/util/heap/layout/vm_layout.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub log_space_extent: usize,
+    /// Should mmtk enable contiguous spaces and virtual memory for all spaces?
+    /// For normal 64-bit config, this should be set to true. Each space should own a contiguous piece of virtual memory.
+    /// For 32-bit or 64-bit compressed heap, we don't have enough virtual memory, so this should be set to false.
+    pub force_use_contiguous_spaces: bool,
+}
+
+impl VMLayout {
+    #[cfg(target_pointer_width = "32")]
+    pub const LOG_ARCH_ADDRESS_SPACE: usize = 32;
```
caizixian, line 54: We should make this a part of `VMLayout`.

src/util/options.rs
```diff
@@ -716,7 +725,9 @@ options! {
     // Default to a fixed heap size of 0.5x physical memory.
     gc_trigger     :        GCTriggerSelector    [env_var: true, command_line: true] [|v: &GCTriggerSelector| v.validate()] = GCTriggerSelector::FixedHeapSize((crate::util::memory::get_system_total_memory() as f64 * 0.5f64) as usize),
     // Enable transparent hugepage support via madvise (only Linux is supported)
-    transparent_hugepages: bool                  [env_var: true, command_line: true]  [|v: &bool| !v || cfg!(target_os = "linux")] = false
+    transparent_hugepages: bool                  [env_var: true, command_line: true]  [|v: &bool| !v || cfg!(target_os = "linux")] = false,
+    // Enable 35-bit heap address space (normally for compressed pointers)
+    use_35bit_address_space: bool                [env_var: true, command_line: true]  [always_valid] = false
```
caizixian, line 730: Can we make this an enum to choose from different sets of vm constants

src/util/heap/space_descriptor.rs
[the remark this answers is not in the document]
wenyuzhao: Fixed. Changed this to check on vm_layout instead.

src/util/heap/layout/vm_layout.rs
qinsoon: Do you mean this should be configurable? Why? There is already a `log_address_space` in `VMLayout`. This is the max address space we may use in the architecture.

k-sareen: Would RISC-V lower it to 39-bits or would that only be `log_address_space`? I remember I have to lower this value for [Android](https://github.com/k-sareen/mmtk-core/commit/d5e75b9a1ce6016d5a4d4c843d1205d7f83ae081) because of some errors with side-metadata (see [here](https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/LOS.20local.20nursery-bit.20side.20metadata/near/364172330)).

qinsoon: > Would RISC-V lower it to 39-bits or would that only be `log_address_space`? I remember I have to lower this value for [Android](https://github.com/k-sareen/mmtk-core/commit/d5e75b9a1ce6016d5a4d4c843d1205d7f83ae081) because of some errors with side-metadata (see [here](https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/LOS.20local.20nursery-bit.20side.20metadata/near/364172330)).

If RISC-V only allows 39-bits, we can use conditional compilation to have different constant values based on the architecture. If I understand correct, this is the upper bound of the allowed virtual address space, while `log_address_space` is configurable and should be smaller than the upper bound. Having the upper bound as a constant is beneficial. Looks like some side metadata calculation is using the constant.

src/util/options.rs
qinsoon: This option is no longer used in the PR, and can be removed.

src/policy/sft_map.rs
qinsoon: Yes. This should be calculated from heap start and space extent.

src/util/heap/layout/vm_layout.rs
k-sareen: Right. But the side-metadata shouldn't be using this constant in its calculation then, right? The calculation then assumes you can have XYZ-bit address space which may not be true depending on the architecture/target environment (such as Android).

qinsoon: Better let Wenyu clarify this.

caizixian: The virtual address range is should ideally be run-time configurable. The same riscv64 binary should run on sv39, sv48, and sv57 machines, and exploit the address range available on each of the platform by detecting mmu type. IIRC, both OpenJDK and QEMU support this sort of run-time detection.

src/util/options.rs
wenyuzhao: Removed

src/util/heap/layout/vm_layout.rs
wenyuzhao: `LOG_ARCH_ADDRESS_SPACE` means the maximum available user-space virtual address size for a user process. `log_address_space` is the usable virtual address size for the Java heap.

For example, x64 openjdk with compressed pointers would set `LOG_ARCH_ADDRESS_SPACE` to 47 and `log_address_space` to 35.

But @caizixian is true, `LOG_ARCH_ADDRESS_SPACE` is not a constant as well. The available pointer bits are totally depends on your machine, even you are using the same build. This should be a runtime variable. However, changing this to a variable basically means a rewrite to the whole side metadata address calculation, layout and barrier fast-path code. This cannot be done in this PR. We need to do the re-design carefully before rewriting the side metadata

caizixian: Yeah, I agree.

src/util/heap/layout/vm_layout_constants.rs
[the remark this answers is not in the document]
qinsoon: The performance results look good.

qinsoon:
> I suggest adding some tests that only run on 64 bits architecture in `DummyVM` (vmbindings/dummyvm/src/tests): set a custom VM layout for the compressed pointer layout, initialize MMTk, create a mutator, and finish one allocation. The test should run for all plans (Add `// GITHUB-CI: MMTK_PLAN=all` at the top of the test file)

I added some unit tests for `VMLayout`. Some are failing. Feel free to modify the tests if they are wrong, or ignore some tests. I just added the tests to see what is working, and what is not.

src/policy/sft_map.rs
wenyuzhao: Done

src/mmtk.rs
```diff
@@ -65,6 +66,12 @@ impl MMTKBuilder {
         self.options.set_bulk_from_command_line(options)
     }
 
+    /// Custom VM layout constants. VM bindings may use this function for compressed or 39-bit heap support.
+    /// This function must be called before MMTk::new()
+    pub fn set_custom_vm_layout(&mut self, constants: VMLayout) {
```
k-sareen, line 71: Can we rename this to `set_vm_layout` to be consistent with the other function names? Or do others like `set_custom_vm_layout` more?

wenyuzhao: Done

wenyuzhao:
> > I suggest adding some tests that only run on 64 bits architecture in `DummyVM` (vmbindings/dummyvm/src/tests): set a custom VM layout for the compressed pointer layout, initialize MMTk, create a mutator, and finish one allocation. The test should run for all plans (Add `// GITHUB-CI: MMTK_PLAN=all` at the top of the test file)
> 
> I added some tests for `VMLayout`. Some are failing. Feel free to modify the tests if they are wrong, or ignore some tests. I just added the tests to see what is working, and what is not.

Interesting. I can't even pass all the dummyvm tests on the master branch

https://gist.github.com/wenyuzhao/ed9614dfec7b1fdac7bee8664b547e46

Is `MMTK_PLAN=NoGC cargo test` the correct command to run dummyvm tests?

qinsoon:
> > > I suggest adding some tests that only run on 64 bits architecture in `DummyVM` (vmbindings/dummyvm/src/tests): set a custom VM layout for the compressed pointer layout, initialize MMTk, create a mutator, and finish one allocation. The test should run for all plans (Add `// GITHUB-CI: MMTK_PLAN=all` at the top of the test file)
> > 
> > 
> > I added some tests for `VMLayout`. Some are failing. Feel free to modify the tests if they are wrong, or ignore some tests. I just added the tests to see what is working, and what is not.
> 
> Interesting. I can't even pass all the dummyvm tests on the master branch
> 
> https://gist.github.com/wenyuzhao/ed9614dfec7b1fdac7bee8664b547e46
> 
> Is `MMTK_PLAN=NoGC cargo test` the correct command to run dummyvm tests?

No. Each test module (file) is executed separately: https://github.com/mmtk/mmtk-core/blob/3cfdf75ef033eeb61e8c149babc9ca13ed5ca9cc/vmbindings/dummyvm/src/tests/mod.rs#L1

qinsoon:
Maybe try a different heap layout for the failing test on macos, or simply ignore that test on macOS. I think either is okay.

caizixian:
The address is too low for 64-bit macOS https://stackoverflow.com/questions/46916112/osx-ld-why-does-pagezero-size-default-to-4gb-on-64b-osx

wenyuzhao:
> Maybe try a different heap layout for the failing test on macos, or simply ignore that test on macOS. I think either is okay.

Switched to a different address on macos. Now all the checks are passed.

qinsoon (approved):
Looks good. Thanks for the PR. Let's merge this after the current release (0.19) which should be done today.

PR #907, opened 2023-08-17
Merge tutorial and porting guide into user guide

qinsoon:
This PR merges the current tutorial and the current porting guide into one `mdbook`: user guide. This PR fixes related links and CI scripts, but does not change the wording or the structure for the docs. Further PRs on improving the docs are welcome.


caizixian:
Should we also target this for the 0.19 release?

qinsoon:
> Should we also target this for the 0.19 release?

I don't know if it matters, as most people don't read the docs from the repo. Any reason we should or should not include this in the current 0.19 release? If there is no particular reason, we can be relaxed on when to merge this.

caizixian:
> > Should we also target this for the 0.19 release?
> 
> I don't know if it matters, as most people don't read the docs from the repo. Any reason we should or should not include this in the current 0.19 release? If there is no particular reason, we can be relaxed on when to merge this.

It's not super urgent, but I think because when we make a new release, the README of the repo is persisted on crates.io and docs.rs, so it would be nice if we can fix up the links.

caizixian (approved):
I think this mostly addresses our documentation discoverability problem. Once merged, we can update the mmtk website, and gradually add other documentation to the user guide, such as the tracing documentation.

.github/workflows/rustdoc.yml
```diff
@@ -24,22 +24,19 @@ jobs:
...
+        # mmtk public doc: /api/public-doc
+        # mmtk full doc: /api/full-doc
         run: |
-          mkdir -p to_publish/docs
-          cp -r target/mmtk-full/doc/* to_publish/docs/
-          mv to_publish/docs/mmtk to_publish/docs/full-doc
-          cp -r target/mmtk-public/doc/mmtk to_publish/docs/public-doc
-          cp -r docs/portingguide/book to_publish/docs/portingguide
-          cp -r docs/tutorial/book to_publish/docs/tutorial
+          cp -r docs/userguide/book to_publish
```
qinsoon, line 31: Currently the Github page is built from `/docs` in `gh-pages`. This is the reason we copy things to `docs`. Do you plan to change that setting as well?

caizixian: Yes. I will change it once this PR is merged.

PR #908, opened 2023-08-17
Bump version to v0.19

qinsoon:
This PR bumps MMTk core to v0.19. It updates the pinned Rust version to 1.71.1, and MSRV to 1.66.

Note: I am trying to use Github releases to generate a release note, categorize the items in the note quickly and use it as the changelog for the release. I think we would eventually automate the process for tagging a release.


wks (approved):
LGTM

wks:
By the way, is it possible to mark API-breaking changes in the change log, too?  (i.e. list commits that has the "Public API Check" CI failed)

qinsoon:
> By the way, is it possible to mark API-breaking changes in the change log, too? (i.e. list commits that has the "Public API Check" CI failed)

There is a section in the changelog about API. See [here](https://github.com/mmtk/mmtk-core/blob/c76d7c9aef8820ab14deaf0299fd90382202ea52/CHANGELOG.md?plain=1#L58). But it is not determined by "Public API Check". I manually categorize the commits. Do you think that is sufficient for now?

Also I think we would automate the changelog generation (for example including a tag for the category of the commit), and the release process if we can. I am trying with the auto generated release note from Github with this PR. 

About "Public API Check", I am not sure if we can or should completely rely on it now. It prompts us for any change it detects. The change may not affect users. For example, if you look at the "Public API Check" for this PR, it shows there are some changes (as we replace manual implementation of `Default` with `#[derive(Default)]`. It should not affect users, but the check still detects it as a change. Also there are some API changes that users can deal with straightforwardly. I think what we want is to let users know about the API changes that are subtle to fix. My preference is to have a migration guide at some point, rather than dumping all the breaking change commits to the users.

qinsoon:
binding-refs
OPENJDK_BINDING_REF=v0.19
JIKESRVM_BINDING_REF=v0.19

wks:
> There is a section in the changelog about API. See [here](https://github.com/mmtk/mmtk-core/blob/c76d7c9aef8820ab14deaf0299fd90382202ea52/CHANGELOG.md?plain=1#L58). But it is not determined by "Public API Check". I manually categorize the commits. Do you think that is sufficient for now?

That section contains notable API changes.  But other things, like "renaming alloc bits to VO bits" changes the API, too, and bindings are required to change accordingly.

> Also I think we would automate the changelog generation (for example including a tag for the category of the commit), and the release process if we can. I am trying with the auto generated release note from Github with this PR.
> 
> About "Public API Check", I am not sure if we can or should completely rely on it now. It prompts us for any change it detects. The change may not affect users. For example, if you look at the "Public API Check" for this PR, it shows there are some changes (as we replace manual implementation of `Default` with `#[derive(Default)]`. It should not affect users, but the check still detects it as a change. Also there are some API changes that users can deal with straightforwardly. I think what we want is to let users know about the API changes that are subtle to fix. My preference is to have a migration guide at some point, rather than dumping all the breaking change commits to the users.

Yes.  Automation will help us writing the change log.  As for the false positives of the "Public API Check" API, we can let the script list all commits but add a star symbol before API-breaking commits.  And then human can review those commits with starts and see if they really break API, or if they only changed things like `Default` trait.  Even though we can't completely rely on it, I think it can make human's work easier.

PR #910, opened 2023-08-17
Add a ready-to-merge check

qinsoon:
Github allows setting a list of required status before merging into a protected branch (in our case, `master`). However, it requires the checks to have a fixed name (no pattern matching allowed). As some of our checks are dynamically generated using a testing matrix, the names may change. It is not convenient to set them as required status.

This PR adds another check. It passes if all other checks pass (except ignored checks). We can use this new check as the required status for merging. Once we set up a proper required check, we can enable merge queue.

This PR also adds a `merge_group` trigger for pre code review checks.


qinsoon:
I think this is what is needed for enabling merge queue for us. For PRs in the merge queue, we only run pre code review checks, and the ready-to-merge check added by the PR. If the ready-to-merge check passes, the PR will be merged. 

Other checks like MSRV, public API and binding tests are not triggered by merge queue.
* MSRV and public API checks are not required to pass.
* Binding tests do not work well with MMTk core merge queue, as when PRs are automatically merged in MMTk core, we cannot guarantee the corresponding PRs are also merged in the binding. For example, we have MMTk core PRs, X and Y, and binding PRs, Xa/Xb and Ya/Yb. When we put X and Y into MMTk core's merge queue, we have X+Y in mmtk-core, but we do not have Xa+Ya or Xb+Yb in the bindings. 

Let me know if there is anything that I missed. @wks @caizixian

wks (approved):
LGTM

caizixian (approved):

PR #905, opened 2023-08-16
Implement transparent hugepage support

caizixian:

PR #913, opened 2023-08-17
Apply style check on auxiliary crates (macros and dummyvm)

caizixian:
This commit was originally backed out of #905. Add clippy check and cargo fmt check for auxiliary crates (`macros` and `dummyvm`).


qinsoon (approved):
LGTM

PR #919, opened 2023-08-24
Update doc comment of Scanning::process_weak_refs

wks:
Mention `ObjectReference::get_forwarded_object()` and recommend it for VM bindings that never resurrect objects.


src/vm/scanning.rs
```diff
@@ -242,20 +242,33 @@ pub trait Scanning<VM: VMBinding> {
...
-    /// The VM binding can use `tracer_factory` to get access to an `ObjectTracer`, and call
-    /// its `trace_object(object)` method to keep `object` and its decendents alive.
-    ///
-    /// The return value of `ObjectTracer::trace_object(object)` is the new address of the given
-    /// `object` if it is moved by the GC.
+    /// The VM binding can query if an object is currently reached by calling
+    /// `ObjectReference::is_reachable()`.
+    ///
+    /// If an object is already reached, the VM binding can get its new address by calling
+    /// `ObjectReference::get_forwarded_object()`.
```
qinsoon, line 253:
```suggestion
    /// If an object is already reached, the VM binding can get its new address by calling
    /// `ObjectReference::get_forwarded_object()` as the object may have been moved.
```

qinsoon (approved):
LGTM

src/vm/scanning.rs
wks: Done.

PR #923, opened 2023-08-25
Fix api-check CI

wks:
The `cargo-public-api` tool now requires a more recent Rust version.  We just use the stable version to compile it.


qinsoon (approved):
LGTM

PR #925, opened 2023-08-28
Let FreeListPageResource hold FreeList in sync

wks:
DRAFT: This is highly experimental, and currently it breaks 32-bit systems completely.

This PR does the following

-   Move the `FreeList` instance into `FreeListPageResource::sync`.
-   Eliminate the need to "bind_freelist" and "boot" for both 32-bit and 64-bit systems.
-   Eliminate the need to "finalize_static_space_map" for 64-bit systems.  Instead it eagerly handle the size of `RawMemoryFreeList` at the start of each space when creating `FreeListPageResource`.
-   Eliminated the need for `Map64::allocate_contiguous_chunks` to look up `RawMemoryFreeList` from address using `Map64::fl_map`.  Instead the caller will pass the `Some(&mut RawMemoryFreeList)` to `Map64::allocate_contiguous_chunks` if the space actually uses a `RawMemoryFreeList`.
-   Let plans enumerate spaces mutably, and let spaces enumerate its PageResource mutably (if there is any) so that on 32-bit systems, `Map32::finalize_static_space_map` can update their start addresses once the location of discontiguous spaces is determined.

This PR does not attempt to

-   Eliminate the unsafety due to the two-level structure of `IntArrayFreeList`.  It's relatively independent.  I'll leave it to a different PR.

# Description

## Moving `FreeList` in `sync`

`FreeList` instances are not thread safe, and always requires lock to access.  Whenever `FreeListPageResource` accesses the `FreeList`, it always holds the mutex of `FreeListPageResource::sync`.  This PR attempts to move the `FreeList` into `sync` to make the access safe.  By doing this, we eliminate the need to cast `&FreeListPageResource` to `&mut FreeListPageResource`.

We observe that the `limit` of `RawMemoryFreeList` is determined when it is created.  So we let it return the limit directly so that we don't need to update the starting address later.  The `FreeList::maybe_get_limit()` returns `Some(limit)` if it is `RawMemoryFreeList`, and `None` if it is `IntArrayFreeList`.  The caller can decide whether to bind the freelist.  But I plan to revamp the plan creation and space allocation process so that we won't need a separate `finalize_static_space_map`.

## Eliminating the need to register FreeListPageResources or RawMemoryFreeLists in global maps

The global maps `Map32::shared_fl_map` and `Map64::{fl_page_resources,fl_map}` were inherited from JikesRVM. They provided ways to enumerate FreeListPageResources and RawMemoryFreeList instances.  This is unsafe in Rust.

We found them unnecessary.  We can enumerate page resources from plans, and we can determine the size of `RawMemoryFreeList` eagerly.

# Related issues

https://github.com/mmtk/mmtk-core/issues/853


wks:
This PR tries to do too many things at the same time, for example, refactoring Map32 to properly synchronise its operations.  I am splitting this into multiple PRs:

-   Map32 synchronisation: https://github.com/mmtk/mmtk-core/pull/951
-   FreeListPageResource sync and removing CommomFreeListPageResource tables in Map32 and Map64: https://github.com/mmtk/mmtk-core/pull/953

PR #934, opened 2023-09-01
Refactor derive macros and add HasSpaces trait

wks:
The main part of this PR is adding a `HasSpaces` trait that enumerates spaces in a plan.  This trait can be automatically generated using derive macro and attributes, saving the effort to enumerate spaces by hand.

-   This PR adds a `HasSpaces` trait that enumerates spaces of a plan.  Plans (including NoGC) are now required to implement this trait.
-   This PR implements derived macro for `HasSpaces`
-   The `#[trace]` and `#[fallback_trace]` attributes are replaced by the `#[space]` and `#[parent]` attributes for generality.  A dedicated `#[copy_semantics(xxxxx)]` attribute is added to specify copy semantics for `trace_object`.
-   The manually written `Plan::get_spaces()` method is removed in favor for the automatically generated `HasSpace::for_each_space` method.
-   Added a `Plan::verify_side_metadata_sanity` that calls `Space::verify_side_metadata_sanity` for all spaces in the plan.  This replaces the manual invocations of `Space::verify_side_metadata_sanity` in the constructors of plans.

This PR also slightly refactors the mmtk-macros crate.
-   The `syn` dependency is one major version behind the latest.  This PR bumps the version and refactors the code to adapt to this change.
-   `mmtk-core/macros/src/lib.rs` is cleaned up so that it only contains functions with minimum function bodies.  Those functions for derive macros are required to be in the root module of the crate.  Their function bodies are off-loaded to other modules.

This PR is a stepping stone for https://github.com/mmtk/mmtk-core/pull/925.  This PR makes it possible to enumerate spaces of a plan without resorting to unsafe global maps.

### DIFF
--- a/docs/userguide/src/tutorial/code/mygc_semispace/gc_work.rs
+++ b/docs/userguide/src/tutorial/code/mygc_semispace/gc_work.rs
@@ -23,7 +23,7 @@ impl<VM: VMBinding> crate::scheduler::GCWorkContext for MyGCWorkContext2<VM> {
     type PlanType = MyGC<VM>;
     type ProcessEdgesWorkType = PlanProcessEdges<Self::VM, MyGC<VM>, DEFAULT_TRACE>;
 }
-// ANCHOR: workcontext_plan
+// ANCHOR_END: workcontext_plan
 
 use crate::util::ObjectReference;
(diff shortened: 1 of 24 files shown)


qinsoon (approved):
LGTM

Issue #929, opened 2023-08-30
Clippy bug in 1.72: forcing query with already existing `DepNode`


qinsoon:
This started to appear on Rust stable (1.72) after 66e8cb8ae7dc40b9ca504f1775d2bb236b087473 is merged. It failed when we tried to check style for all the features.

```console
++ for_all_features 'cargo clippy'
++ cargo clippy --features vm_space,ro_space,code_space,vo_bit,is_mmtk_object,object_pinning,immix_non_moving,immix_smaller_block,immix_zero_on_release,sanity,analysis,nogc_lock_free,nogc_no_zeroing,single_worker,extreme_assertions,nogc_multi_space,work_packet_stats,malloc_counted_size,count_live_bytes_in_gc
   Compiling mmtk v0.19.0 (/home/runner/work/mmtk-core/mmtk-core)
thread 'rustc' panicked at 'forcing query with already existing `DepNode`
- query-key: (ByRef { alloc: ConstAllocation { .. }, offset: Size(0 bytes) }, util::metadata::side_metadata::global::SideMetadataSpec)
- dep-node: try_destructure_mir_constant_for_diagnostics(8ee85db3da0b8d25-401d3d57adad2b0d)', /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/compiler/rustc_query_system/src/dep_graph/graph.rs:351:9
stack backtrace:
   0: 0xee2eda78 - std::backtrace_rs::backtrace::libunwind::trace::ha70eea129e7d2a5b
                       at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1: 0xee2eda78 - std::backtrace_rs::backtrace::trace_unsynchronized::hf2790414cd064fa3
                       at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2: 0xee2eda78 - std::sys_common::backtrace::_print_fmt::h20ee8058fc5e04f2
                       at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/sys_common/backtrace.rs:65:5
   3: 0xee2eda78 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h58d518f543159de3
[...]

PR #938, opened 2023-09-04
Add CARGO_INCREMENTAL=0 to work around clippy 1.72 bug

qinsoon:
This PR works around the issue https://github.com/mmtk/mmtk-core/issues/929, using the suggestion in https://github.com/rust-lang/rust/issues/83085#issuecomment-1504033603.


wks (approved):
LGTM

PR #939, opened 2023-09-04
Discontiguous mark compact space support

### LINKED ISSUE #930: Discontiguous MarkCompactSpace support
Currently, the MarkComapctSpace relies on `self.common.start`, which is only set for contiguous spaces: https://github.com/mmtk/mmtk-core/blob/master/src/policy/markcompactspace.rs#L322

This brings an implicit assumption that the MarkCompactSpace can work correctly only when being configured as contiguous. Although it's constructor actually supports discontiguous layout https://github.com/mmtk/mmtk-core/blob/master/src/policy/markcompactspace.rs#L202-L206.

This is a blocker for the openjdk compressed oops support, as it will enable discontiguous MarkCompactSpace for MarkCompact plan.

wenyuzhao:
This PR adds discontiguous heap support for mark compact GC, and rewrites the forwarding pointer calculation and compaction code to be compactable to both contiguous and discontiguous spaces.

Fixes #930

### DIFF
--- a/src/policy/markcompactspace.rs
+++ b/src/policy/markcompactspace.rs
@@ -325,6 +325,7 @@ impl<VM: VMBinding> MarkCompactSpace<VM> {
     fn linear_scan_objects(
         &self,
         range: Range<Address>,
+        to_be_compacted_only: bool,
         mut f: impl FnMut(ObjectReference, usize, usize, usize),
     ) {
         let linear_scan = crate::util::linear_scan::ObjectIterator::<
@@ -332,7 +333,7 @@ impl<VM: VMBinding> MarkCompactSpace<VM> {
             MarkCompactObjectSize<VM>,
             true,
         >::new(range.start, range.end);
-        for obj in linear_scan.filter(|obj| Self::to_be_compacted(*obj)) {
+        for obj in linear_scan.filter(|obj| !to_be_compacted_only || Self::to_be_compacted(*obj)) {
             let copied_size =
                 VM::VMObjectModel::get_size_when_copied(obj) + Self::HEADER_RESERVED_IN_BYTES;
             let align = VM::VMObjectModel::get_align_when_copied(obj);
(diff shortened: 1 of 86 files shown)


src/policy/markcompactspace.rs
```diff
@@ -318,69 +321,126 @@ impl<VM: VMBinding> MarkCompactSpace<VM> {
...
+                    VM::VMObjectModel::get_size_when_copied(obj) + Self::HEADER_RESERVED_IN_BYTES;
+                let align = VM::VMObjectModel::get_align_when_copied(obj);
+                let offset = VM::VMObjectModel::get_align_offset_when_copied(obj);
+                // move to_cursor to aliged start address
+                to_cursor = align_allocation_no_fill::<VM>(to_cursor, align, offset);
+                // move to next to-block if there is no sufficient memory in current region
+                if to_cursor + copied_size > to_end {
+                    (to_cursor, to_end) = to_iter.next().unwrap();
+                    to_cursor = align_allocation_no_fill::<VM>(to_cursor, align, offset);
+                    assert!(to_cursor + copied_size <= to_end);
```
tianleq, line 398: can make it debug_assert!

src/policy/markcompactspace.rs
```diff
@@ -318,69 +321,126 @@ impl<VM: VMBinding> MarkCompactSpace<VM> {
...
+        &self,
+        range: Range<Address>,
+        to_be_compacted_only: bool,
+    ) -> impl Iterator<Item = ObjectReference> {
+        let linear_scan = crate::util::linear_scan::ObjectIterator::<
+            VM,
+            MarkCompactObjectSize<VM>,
+            true,
+        >::new(range.start, range.end);
+        linear_scan.filter(move |obj| !to_be_compacted_only || Self::to_be_compacted(*obj))
```
qinsoon, line 335: I find the condition hard to understand. Maybe move the `filter` to the caller, and just return the `ObjectIterator` here -- that can make the code more intuitive.

src/policy/markcompactspace.rs
```diff
@@ -318,69 +321,126 @@ impl<VM: VMBinding> MarkCompactSpace<VM> {
...
+    }
+
+    /// Iterate over all contiguous memory regions in this space.
+    /// For contiguous space, this iterator should yield only once, and returning a contiguous memory region covering the whole space.
+    fn iterate_contiguous_regions(&self) -> impl Iterator<Item = (Address, Address)> + '_ {
```
qinsoon, line 340: I think this method can be moved to `MonotonePageResource`. Also there is a method that looks very similar in `MonotonePageResource`. Does `for_allocated_regions` work for this case?
https://github.com/mmtk/mmtk-core/blob/2e29b978272cee30be85cb1fedb7a303086aa02e/src/util/heap/monotonepageresource.rs#L298

If they are doing the same thing, I suggest we keep one implementation (either the old one or the new one).

src/policy/markcompactspace.rs
wenyuzhao: Fixed

src/policy/markcompactspace.rs
wenyuzhao: Fixed

src/policy/markcompactspace.rs
wenyuzhao: These are basically the same thing. However, within the `MarkCompactSpace::compact`, it has to be an iterator. The function needs to bump the to-space region iterator conditionally instead of blindly iterating over the whole heap.

I removed `for_allocated_regions` and rewrote that to a function returning an iterator, so it can work on both MarkCompactSpace and ImmortalSpace (which is the only caller of this function).

qinsoon (approved):
LGTM

tianleq (approved):

PR #942, opened 2023-09-05
Fix outdated Rust version in README

qinsoon:
This PR fixes some outdated Rust versions mentioned in README.
* We no longer mention a specific Rust version in README. Instead, it provides links to where we define the versions. 
* Add an MSRV badge.


wks (approved):
LGTM

PR #944, opened 2023-09-05
Make MarkCompact LOS support 2nd transitive closure

wenyuzhao:
#939 makes MarkCompact GC allocate large objects into LOS, instead of mark-compacting everything. However, LOS should be released and re-prepared properly, otherwise marking in the pointer forwarding trace will not behave correctly.

This should fix the OpenJDK binding test failures.

### DIFF
--- a/src/plan/markcompact/gc_work.rs
+++ b/src/plan/markcompact/gc_work.rs
@@ -32,8 +32,11 @@ impl<VM: VMBinding> CalculateForwardingAddress<VM> {
 /// create another round of root scanning work packets
 /// to update object references
 pub struct UpdateReferences<VM: VMBinding> {
+    plan: *const MarkCompact<VM>,
     p: PhantomData<VM>,
 }
+
+unsafe impl<VM: VMBinding> Send for UpdateReferences<VM> {}
 
 impl<VM: VMBinding> GCWork<VM> for UpdateReferences<VM> {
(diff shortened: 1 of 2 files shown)


qinsoon (dismissed):
LGTM. Thanks for the fix.

qinsoon:
I just realized I attempted a similar approach on this before: https://github.com/mmtk/mmtk-core/pull/527. @tianleq left a comment here https://github.com/mmtk/mmtk-core/pull/527#discussion_r789304003, suggesting there could be a memory leak. That comment is probably still relevant for this PR.

Given that the current tests are broken, getting the tests running again would be the priority. If we do not have a better solution, I suggest we merge this PR once we understand the consequence (e.g. we would have memory leak for LOS in mark compact).

wenyuzhao:
> I just realized I attempted a similar approach on this before: #527. @tianleq left a comment here [#527 (comment)](https://github.com/mmtk/mmtk-core/pull/527#discussion_r789304003), suggesting there could be a memory leak. That comment is probably still relevant for this PR.
> 
> Given that the current tests are broken, getting the tests running again would be the priority. If we do not have a better solution, I suggest we merge this PR once we understand the consequence (e.g. we would have memory leak for LOS in mark compact).

After discussing with @tianleq , releasing once before the second prepare call should have no memory leak issue. All dead objects are cleared before the second prepare.

qinsoon (approved):
LGTM

PR #946, opened 2023-09-06
Discontiguous PageProtect GC support 

wenyuzhao:
This CL fixes PageProtect's memory unprotect calls. When using discontiguous space,
* If the allocated pages are not in a new chunk, unprotect the pages as before.
* If the pages are in a new chunk, there can be two cases:
   * If the new chunk is not mapped, no need to unprotect
   * If the new chunk is mapped, this means that the chunk is allocated and released on previous GC epochs before. The GC should unprotect these pages.

### DIFF
--- a/src/util/heap/freelistpageresource.rs
+++ b/src/util/heap/freelistpageresource.rs
@@ -147,7 +147,7 @@ impl<VM: VMBinding> PageResource<VM> for FreeListPageResource<VM> {
         let rtn = self.start + conversions::pages_to_bytes(page_offset as _);
         // The meta-data portion of reserved Pages was committed above.
         self.commit_pages(reserved_pages, required_pages, tls);
-        if self.protect_memory_on_release && !new_chunk {
+        if self.protect_memory_on_release && (!self.common().contiguous || !new_chunk) {
             // This check is necessary to prevent us from mprotecting an address that is not yet mapped by mmapper.
             // See https://github.com/mmtk/mmtk-core/issues/400.
             // It is possible that one thread gets a new chunk, and returns from this function. However, the Space.acquire()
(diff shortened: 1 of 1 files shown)


qinsoon:
Can you explain a bit why we need this change? What is the bug before this change?

src/util/heap/freelistpageresource.rs
```diff
@@ -147,7 +147,7 @@ impl<VM: VMBinding> PageResource<VM> for FreeListPageResource<VM> {
         let rtn = self.start + conversions::pages_to_bytes(page_offset as _);
         // The meta-data portion of reserved Pages was committed above.
         self.commit_pages(reserved_pages, required_pages, tls);
-        if self.protect_memory_on_release && !new_chunk {
+        if self.protect_memory_on_release && (!self.common().contiguous || !new_chunk) {
```
qinsoon, line 150: By looking at the code, I guess `new_chunk` is false when we reuse a chunk in a discontiguous page resource. Right?

wenyuzhao: Yes. new_chunk is false if the page resource reuse some pages in a chunk it owns. However, if the chunk is released to the global vmmap, and reallocated afterwards, new_chunk will be true

wenyuzhao:
> Can you explain a bit why we need this change? What is the bug before this change?

This is a necessary fix to make page protect gc be able to run with compressed pointers, which uses discontiguous heap

src/util/heap/freelistpageresource.rs
qinsoon: The code can be refactored to this:
```rust
if self.protect_memory_on_release {
  if !new_chunk {
    // same as before -- to make sure the page is mapped, then unprotect
    while !MMAPPER.is_mapped_address(rtn) {}
    self.munprotect(rtn, ...);
  } else if !self.common().contiguous && new_chunk {
    // Added in this PR. See my question below.
    if MMAPPER.is_mapped_address(rtn) {
      self.munprotect(rtn, ...);
    }
  }
}
```

If we see this is a new chunk, is it always mapped? If it is really a new chunk, we should have mapped it in this thread. If it is a reused chunk, it should be mapped. If it is always mapped, basically the change is just to unconditionally unprotect any new chunk (no matter it is reused, or it is an actual new chunk). As we cannot tell which case it is, we just unconditionally unprotect a new chunk. I am okay with the change, just to make sure my understanding is correct. And I suggest using the refactored code above, which is easier to read.

wenyuzhao: > If we see this is a new chunk, is it always mapped?

No

> If it is really a new chunk, we should have mapped it in this thread.

If this is the first time the new chunk is allcoated, it is not mapped here. The current thread will map it afterwards shortly.
If the space released a chunk and allocated it again, new_chunk is still set to true. But the memory is already mapped.

> if it is always mapped, basically the change is just to unconditionally unprotect any new chunk (no matter it is reused, or it is an actual new chunk). As we cannot tell which case it is, we just unconditionally unprotect a new chunk.

It's not always mapped. Unconditionally unprotecting will not work. This is why the `if MMAPPER.is_mapped_address(rtn) {` check in your suggestion above is necessary, it skips unprotecting on unmapped new chunks.

wenyuzhao: The code is refactored

qinsoon: I see. The space calls mmapper afterwards.

qinsoon:
The JikesRVM failure should not be related with the PR. We can rerun the test.

qinsoon (approved):
LGTM

PR #947, opened 2023-09-07
Fix dependency replacing script

wks:
Remove all keys that may specify the location of a dependency since they may conflict.


wks:
This change has been merged by piggy-backing on https://github.com/mmtk/mmtk-core/pull/897

Closing...

PR #948, opened 2023-09-08
Fix vo-bit reset for discontiguous space

wenyuzhao:
This PR fixes `CpoySpace::reset_vo_bit` and vo-bit reset implementation for discontiguous space


wks (approved):
LGTM

Issue #945, opened 2023-09-06
Should we avoid using fat pointers for SFT?


qinsoon:

qinsoon:
As we use `portable_atomic` for 128 bits atomics, they list the support for different platforms: https://github.com/taiki-e/portable-atomic/blob/HEAD/src/imp/atomic128/README.md

* x86_64: 
  * Load/store requires `cmpxchg16b`, or `vmovdqa`. CAS/RMW requires `cmpxchg16b`. We currently only need load/store.
  * `cmpxchg16b` starts from `x86_64-v2` (roughly 2009) [source](https://en.wikipedia.org/wiki/X86-64).
  * `vmovdqa` is in AVX extension (starts in 2011) [source](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions)
* aarch64:
  * `LDXP/STXP` loop from ARMv8 (2011)
  * `CASP` (DWCAS) added as FEAT_LSE (mandatory from armv8.1-a) (2014)
  * `LDP/STP` (DW load/store) if FEAT_LSE2 (optional from armv8.2-a, mandatory from armv8.4-a) (2017)
  * `LDIAPP/STILP` (DW acquire-load/release-store) added as FEAT_LRCPC3 (optional from armv8.9-a/armv9.4-a)
  * `LDCLRP/LDSETP/SWPP` (DW RMW) added as FEAT_LSE128 (optional from armv9.4-a)
  * source: [1](https://github.com/taiki-e/portable-atomic/blob/a2ae132ba5ed5e4de48f14af2bedc1a89f640742/src/imp/atomic128/aarch64.rs), [2](https://en.wikipedia.org/wiki/AArch64)
* powerpc64, s390x: may have instructions for 128 bits atomics, but I did not explore further.
* Other platforms: use fallback implementation.

For x86_64 and aarch64, we should be fine. On platforms that use fallback implementation, we will have a performance issue.

wks:
> If we would like to provide efficient implementation for platforms that do not support 128 bits atomics, we should consider not using `&dyn SFT`. We need a way to represent an SFT entry as a normal pointer.

I think there are many possibilities.

For the generous heap layout where each space occupies a dedicated 2TB region, we can eagerly populate the SFTSpaceMap.  From then on, we only need read-only access for SFT entries.

[...]

PR #952, opened 2023-09-12
Remove descriptor_map

wks:
**DRAFT**: This PR has two problems:

1. The current implementation of `address_in_space` in this PR has noticeable performance overhead.  From benchmark results, it looks like the check of having SFT entry and the 128-bit load is a bottleneck.
2. The semantics of casting the 128-bit `*const dyn SFT` to `*const ()` is unclear, and may be unreliable.

So I decide to postpone this PR until two other things are done:

1.  Refactoring the `SFT_MAP` implementation to eliminate 128-bit atomic read.  See https://github.com/mmtk/mmtk-core/issues/945
2.  Merging the compressed pointer support in the OpenJDK binding so that we will have a more realistic use case to evaluate.  See https://github.com/mmtk/mmtk-openjdk/pull/235

# Description

This PR removes the `descriptor_map` from both `Map32` and `Map64`.

Currently, the `descriptor_map` is only used by `Space::address_in_space` when using `Map32`, and not used (except in some assertions about newly acquired pages) when using `Map64`.  With `descriptor_map` removed, `Space::address_in_space` will use the SFT to find the space of a given address.

Performance: The `Space::in_space` function (which calls `Space::address_in_space`) is used by the derive macro of the `trait PlanTraceObject`.  Therefore, this PR will affect the performance of tracing for plans that use `PlanTraceObject` when using `Map32`.  This needs to be tested.

Related PRs:
-   https://github.com/mmtk/mmtk-core/pull/951: If `descriptor_map` were moved into the `Mutex`, it would be inefficient.  This PR removes `descriptor_map`, instead, so that https://github.com/mmtk/mmtk-core/pull/951 can move other fields into the `Mutex`.

### DIFF
--- a/src/policy/space.rs
+++ b/src/policy/space.rs
@@ -180,11 +180,6 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
                         debug_assert_eq!(SFT_MAP.get_checked(res.start).name(), self.get_name());
                         // The start address is in our space.
                         debug_assert!(self.address_in_space(res.start));
-                        // The descriptor should be correct.
-                        debug_assert_eq!(
-                            self.common().vm_map().get_descriptor_for_address(res.start),
-                            self.common().descriptor
-                        );
 
                         // --- Assert the last byte in the allocated region ---
                         let last_byte = res.start + bytes - 1;
@@ -192,11 +187,6 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
                         debug_assert_eq!(SFT_MAP.get_checked(last_byte).name(), self.get_name());
                         // The last byte in the allocated memory should be in this space.
                         debug_assert!(self.address_in_space(last_byte));
-                        // The descriptor for the last byte should be correct.
-                        debug_assert_eq!(
-                            self.common().vm_map().get_descriptor_for_address(last_byte),
-                            self.common().descriptor
-                        );
                     }
 
                     debug!("Space.acquire(), returned = {}", res.start);
@@ -229,7 +219,16 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
 
     fn address_in_space(&self, start: Address) -> bool {
         if !self.common().descriptor.is_contiguous() {
-            self.common().vm_map().get_descriptor_for_address(start) == self.common().descriptor
+            // If the space is not contiguous, we use `SFT_MAP` to check.
+            //
+            // Note: When comparing references (`&T`) using `==`, it compares by values.
+            // To compare by identities, we need to compare pointers.
+            //
+            // TODO: For `DenseChunkMap`, it is possible to reduce one level of dereferencing by
+            // letting each space remember its index in `DenseChunkMap::index_map`.
+            let self_ptr = self as *const _ as *const ();
+            let other_ptr = SFT_MAP.get_checked(start) as *const _ as *const ();
+            self_ptr == other_ptr
         } else {
             start >= self.common().start && start < self.common().start + self.common().extent
         }
@@ -554,22 +553,6 @@ impl<VM: VMBinding> CommonSpace<VM> {
         // FIXME
         rtn.descriptor = SpaceDescriptor::create_descriptor_from_heap_range(start, start + extent);
         // VM.memory.setHeapRange(index, start, start.plus(extent));
-
-        // We only initialize our vm map if the range of the space is in our available heap range. For normally spaces,
-        // they are definitely in our heap range. But for VM space, a runtime could give us an arbitrary range. We only
-        // insert into our vm map if the range overlaps with our heap.
-        {
-            use crate::util::heap::layout;
-            let overlap =
-                Address::range_intersection(&(start..start + extent), &layout::available_range());
-            if !overlap.is_empty() {
-                args.plan_args.vm_map.insert(
-                    overlap.start,
-                    overlap.end - overlap.start,
-                    rtn.descriptor,
-                );
-            }
-        }
 
         // For contiguous space, we know its address range so we reserve metadata memory for its range.
         if rtn
--- a/src/util/heap/layout/map.rs
+++ b/src/util/heap/layout/map.rs
@@ -4,8 +4,6 @@ use crate::util::heap::space_descriptor::SpaceDescriptor;
 use crate::util::Address;
 
 pub trait VMMap: Sync {
-    fn insert(&self, start: Address, extent: usize, descriptor: SpaceDescriptor);
-
     /// Create a free-list for a discontiguous space. Must only be called at boot time.
     /// bind_freelist() must be called by the caller after this method.
     fn create_freelist(&self, start: Address) -> Box<dyn FreeList>;
@@ -60,8 +58,6 @@ pub trait VMMap: Sync {
     fn finalize_static_space_map(&self, from: Address, to: Address);
 
     fn is_finalized(&self) -> bool;
-
-    fn get_descriptor_for_address(&self, address: Address) -> SpaceDescriptor;
 
     fn add_to_cumulative_committed_pages(&self, pages: usize);
 }
--- a/src/util/heap/layout/map32.rs
+++ b/src/util/heap/layout/map32.rs
@@ -28,7 +28,6 @@ pub struct Map32Inner {
     shared_fl_map: Vec<Option<NonNull<CommonFreeListPageResource>>>,
     total_available_discontiguous_chunks: usize,
     finalized: bool,
-    descriptor_map: Vec<SpaceDescriptor>,
 
     // TODO: Is this the right place for this field?
     // This used to be a global variable. When we remove global states, this needs to be put somewhere.
@@ -53,7 +52,6 @@ impl Map32 {
                 shared_fl_map: vec![None; MAX_SPACES],
                 total_available_discontiguous_chunks: 0,
                 finalized: false,
-                descriptor_map: vec![SpaceDescriptor::UNINITIALIZED; max_chunks],
                 cumulative_committed_pages: AtomicUsize::new(0),
             }),
             sync: Mutex::new(()),
@@ -69,28 +67,6 @@ impl std::ops::Deref for Map32 {
 }
 
 impl VMMap for Map32 {
-    fn insert(&self, start: Address, extent: usize, descriptor: SpaceDescriptor) {
-        // Each space will call this on exclusive address ranges. It is fine to mutate the descriptor map,
-        // as each space will update different indices.
-        let self_mut: &mut Map32Inner = unsafe { self.mut_self() };
-        let mut e = 0;
-        while e < extent {
-            let index = (start + e).chunk_index();
-            assert!(
-                self.descriptor_map[index].is_empty(),
-                "Conflicting virtual address request"
-            );
-            debug!(
-                "Set descriptor {:?} for Chunk {}",
-                descriptor,
-                conversions::chunk_index_to_address(index)
-            );
-            self_mut.descriptor_map[index] = descriptor;
-            //   VM.barriers.objectArrayStoreNoGCBarrier(spaceMap, index, space);
-            e += BYTES_IN_CHUNK;
-        }
-    }
-
     fn create_freelist(&self, _start: Address) -> Box<dyn FreeList> {
         Box::new(IntArrayFreeList::from_parent(
             &self.global_page_map,
@@ -119,7 +95,7 @@ impl VMMap for Map32 {
 
     unsafe fn allocate_contiguous_chunks(
         &self,
-        descriptor: SpaceDescriptor,
+        _descriptor: SpaceDescriptor,
         chunks: usize,
         head: Address,
     ) -> Address {
@@ -131,7 +107,6 @@ impl VMMap for Map32 {
         }
         self_mut.total_available_discontiguous_chunks -= chunks;
         let rtn = conversions::chunk_index_to_address(chunk as _);
-        self.insert(rtn, chunks << LOG_BYTES_IN_CHUNK, descriptor);
         if head.is_zero() {
             debug_assert!(self.next_link[chunk as usize] == 0);
         } else {
@@ -260,11 +235,6 @@ impl VMMap for Map32 {
         self.finalized
     }
 
-    fn get_descriptor_for_address(&self, address: Address) -> SpaceDescriptor {
-        let index = address.chunk_index();
-        self.descriptor_map[index]
-    }
-
     fn add_to_cumulative_committed_pages(&self, pages: usize) {
         self.cumulative_committed_pages
             .fetch_add(pages, Ordering::Relaxed);
@@ -304,8 +274,6 @@ impl Map32 {
             for offset in 0..chunks {
                 let index = (chunk + offset) as usize;
                 let chunk_start = conversions::chunk_index_to_address(index);
-                debug!("Clear descriptor for Chunk {}", chunk_start);
-                self.mut_self().descriptor_map[index] = SpaceDescriptor::UNINITIALIZED;
                 SFT_MAP.clear(chunk_start);
             }
             chunks as _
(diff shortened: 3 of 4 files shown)


src/policy/space.rs
```diff
@@ -229,7 +219,16 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
...
         if !self.common().descriptor.is_contiguous() {
-            self.common().vm_map().get_descriptor_for_address(start) == self.common().descriptor
+            // If the space is not contiguous, we use `SFT_MAP` to check.
+            //
+            // Note: When comparing references (`&T`) using `==`, it compares by values.
+            // To compare by identities, we need to compare pointers.
+            //
+            // TODO: For `DenseChunkMap`, it is possible to reduce one level of dereferencing by
+            // letting each space remember its index in `DenseChunkMap::index_map`.
+            let self_ptr = self as *const _ as *const ();
```
qinsoon, line 229: We could add a method in `SFT` to return the space descriptor, and compare on that.

qinsoon (commented):
The refactoring looks good. We need some performance data before we can confidently merge the PR.

src/policy/space.rs
qinsoon: That would be less efficient. The current code is good.

wks:
I tested on `bobcat.moma`, comparing master and this PR, running `lusearch`, with the Immix plan, at 2.5x min heap size, 20 iterations.  The result shows less than 1% impact on total time, STW time, and mutator time.  This looks promising.  I'll retry this test with other plans running `lusearch`.

image

Plotty: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2023-09-12-Tue-093227&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark&build;build1|41&Histogram%20(with%20CI)^build^benchmark&

qinsoon:
> I tested on `bobcat.moma`, comparing master and this PR, running `lusearch`, with the Immix plan, at 2.5x min heap size, 20 iterations. The result shows less than 1% impact on total time, STW time, and mutator time. This looks promising. I'll retry this test with other plans running `lusearch`.
> 
> image
> 
> Plotty: [https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2023-09-12-Tue-093227&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark&build;build1|41&Histogram%20(with%20CI)^build^benchmark&](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Cbobcat-2023-09-12-Tue-093227&benchmark%5Ebuild%5Einvocation%5Eiteration%5Emmtk_gc&GC%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4%7C20&1%5Einvocation%7C30&1&benchmark&build;build1%7C41&Histogram%20(with%20CI)%5Ebuild%5Ebenchmark&)

Is this measured with a 64 bit build? If that's the case, the new code in `address_in_space` is not used.

wks:
> Is this measured with a 64 bit build? If that's the case, the new code in `address_in_space` is not used.

I built and ran it on x86-64.  But I used the following patch to force it to use `Map32`.

```diff
diff --git a/src/util/heap/layout/vm_layout.rs b/src/util/heap/layout/vm_layout.rs
index ddf4472a5..66a7ee083 100644
--- a/src/util/heap/layout/vm_layout.rs
+++ b/src/util/heap/layout/vm_layout.rs
@@ -178,14 +178,14 @@ impl std::default::Default for VMLayout {
 
     #[cfg(target_pointer_width = "64")]
     fn default() -> Self {
-        Self::new_64bit()
+        Self::new_32bit()
     }
 }
 
 #[cfg(target_pointer_width = "32")]
 static mut VM_LAYOUT: VMLayout = VMLayout::new_32bit();
 #[cfg(target_pointer_width = "64")]
-static mut VM_LAYOUT: VMLayout = VMLayout::new_64bit();
+static mut VM_LAYOUT: VMLayout = VMLayout::new_32bit();
 
 static VM_LAYOUT_FETCHED: AtomicBool = AtomicBool::new(false);
```

The log shows it is using the new code.

wks:
I ran again on bobcat.moma, 2.5x min heap size, 40 iterations, with several plans (MarkSweep and MarkCompact couldn't run with that heap size) and used the patch above to force it using Map32.

The result shows noticeable slowdown for all plans.  Immix has the smallest slowdown.  Others have about 1% slowdown in STW time.

image

Plotty: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2023-09-12-Tue-120535&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|40&Histogram%20(with%20CI)^build^mmtk_gc&

qinsoon:
Your build should be using `Map32` and `SFTSparseChunkMap`. Getting `SFT` from `SFTSparseChunkMap` should be as simple as indexing into a vec, and it should be no different than the previous `descriptor_map` code. Is it possible to use 
 `SFTMap::get_unchecked()` instead? It could be the extra check that caused the slowdown.

wks:
> Your build should be using `Map32` and `SFTSparseChunkMap`. Getting `SFT` from `SFTSparseChunkMap` should be as simple as indexing into a vec, and it should be no different than the previous `descriptor_map` code. Is it possible to use `SFTMap::get_unchecked()` instead? It could be the extra check that caused the slowdown.

We can't use `SFTMap::get_unchecked()` because the object may not be in any space in MMTk.  That's what `ActivePlan::vm_trace_object` handles.  But I'll try running `get_unchecked` on OpenJDK to see if it is causing the slowdown.

I guess the 128-bit atomic load may also be a reason for the slowdown.  I'll check that, too.

qinsoon:
The

> > Your build should be using `Map32` and `SFTSparseChunkMap`. Getting `SFT` from `SFTSparseChunkMap` should be as simple as indexing into a vec, and it should be no different than the previous `descriptor_map` code. Is it possible to use `SFTMap::get_unchecked()` instead? It could be the extra check that caused the slowdown.
> 
> We can't use `SFTMap::get_unchecked()` because the object may not be in any space in MMTk. That's what `ActivePlan::vm_trace_object` handles. But I'll try running `get_unchecked` on OpenJDK to see if it is causing the slowdown.
> 
> I guess the 128-bit atomic load may also be a reason for the slowdown. I'll check that, too.

The address does not have to be in an MMTk space. It just needs an entry in SFT (which could be mapped to `SFT_EMPTY_SPACE`). For `SFTSpaceMap` and `SFTSparseChunkMap`, we should be fine, as we map all the address space we may use into SFT map. But for `SFTDenseChunkMap` where we use side metadata, we cannot guarantee the side metadata is available for the entire address space.

wks:
This time I added build3 and build4.  Build3 uses `SFT_MAP::get_unchecked`.  Build4 does not check either, but also only load the lowest 64 bits of the `SFTSparseChunkMap` using unsafe non-atomic 64-bit load.  In the current Rust version on x86_64, the lower 64 bits of a `&dyn` is the object pointer.

From the plot, build3 is noticeably faster than build2 in STW time.  Build4 is about the same as build3.  It may indicate that the "check" is the bottleneck.

But since the number is still a bit noisy, I'll do some extra experiments with micro-benchmarks.

https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2023-09-13-Wed-061313&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

wks:
# lusearch

This is the same setting but with added tests for other plans, and increased the number of invocations to 40.

image

Plotty: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2023-09-13-Wed-144053&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

From this plot, we can see

-   Build2 (this PR) is consistently slower than build1 w.r.t. STW time.
-   Build3 (this PR but use `get_unchecked`) also has some overhead, but not as great as build2.  For Immix, the mean STW time is the same as build1's
-   Build4 exhibits speed-up in STW time over build1 in GenCopy, GenImmix and SemiSpace, but slowdown in Immix and StickyImmix.

# Microbenchmark

I also tested with a microbenchmark.  It is similar to GCBench, but
1.  it only has one long-lived tree, and
2.  it only triggers GC and do nothing else, and
3.  it prints out the time of **each** GC (not total STW time) in the end.

```java
class Node {
    int n;
    Node left;
    Node right;

    Node(int n, Node left, Node right) {
        this.n = n;
        this.left = left;
        this.right = right;
    }

    static Node makeTree(int depth) {
        if (depth == 0) {
            return null;
        } else {
            return new Node(depth, makeTree(depth - 1), makeTree(depth - 1));
        }
    }
}

public class TraceTest {
        int depth = Integer.parseInt(args[0]);
        int iterations = Integer.parseInt(args[1]);
        int warmups = Integer.parseInt(args[2]);
        long[] gctimes = new long[iterations];

        Node tree = Node.makeTree(depth);

        for (int i = 0; i < warmups; i++) {
            System.gc();
        }

        for (int i = 0; i < iterations; i++) {
            long time1 = System.nanoTime();
            System.gc();
            long time2 = System.nanoTime();

            gctimes[i] = time2 - time1;
        }

        for (long gctime: gctimes) {
            System.out.println(gctime);
        }
    }
}
```

I ran it on bobcat.moma with the following script:

```bash
for plan in SemiSpace GenCopy GenImmix StickyImmix Immix; do
        for j in {1..5}; do
                for i in {1..4}; do
                        echo $plan build$i iter$j
                        MMTK_THREADS=1 MMTK_PLAN=${plan} ~/compare/build${i}/openjdk/build/linux-x86_64-normal-server-release/images/jdk/bin/java -XX:+UseThirdPartyHeap -server -XX:ParallelGCThreads=1 -XX:MetaspaceSize=100M -Xm{s,x}500M TraceTest 22 100 10 > out/result-${plan}-build${i}-iter${j}.txt
                done
        done
done
```

The number of GC workers is set to 1.  For each plan-build pair, it runs 5 times.  Each time it creates a 22-level tree, trigger GC 10 times for warm-up, and trigger GC 100 more times, recording the time of each GC.

The results are plotted in the following violin plot + scattered point plot.  Each cell corresponds to a plan-build pair, and it contains 5 bars, each correspond to one of the 5 iteration.  The horizontal dash "-" in the middle of each "violin" is the median.

image

With outliers (zscore >= 3) removed, the result is:

image

The GC times exhibit an interesting bi-modal distribution in SemiSpace and StickyImmix.

For the two non-generational plans, namely SemiSpace and Immix, the median of build2 is significant greater than build1.  build3 is slightly faster than build2 but still noticeably slower than build1.  Build4 is similar to build1 in both SemiSpace and Immix, but build4 is slightly faster than build1 in SemiSpace, but slightly slower than build1 in Immix.

For the two GenXxxxx plans, namely GenCopy and GenImmix, the plot doesn't show significant differences in GC time.  The result varies in each run, and the noise is more significant than the medians.

StickyImmix is a bit interesting.  The bi-modal distribution disappeared in build3 (like this PR but using `get_unchecked`).  Since the script runs each build in turn (build1, build2, build3, build4 then build1 and buid2 again...), the difference should be intrinsic to build3.

This result is hard to interpret, but it looks like the cost of the 128-bit load is significant, and the check in `get_checked` also has a minor contribution to the overhead.

src/policy/space.rs
```diff
@@ -229,7 +219,16 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
...
+            let other_ptr = SFT_MAP.get_checked(start) as *const _ as *const ();
```
wenyuzhao, line 230: Better to use `as *const Self` instead of `*const _`. We had a bug https://github.com/mmtk/mmtk-core/pull/750 where if the return type of `get_checked` is changed, the `*const _` cast may still compile but can sliently fail the pointer comparison later.

Same to L229 as well.

wks: Yes.  We can do that on line 229.  But on L230,  the returned SFT instance may not have the same type as Self.

But I think a deeper problem is that `SFT_MAP.get_checked(start) as *const _` is a `*const dyn SFT`, which is 16 bytes long.  But `*const ()` is 8 bytes long. Embarrassingly, neither the Rust reference nor the Rustonomicon specify the semantics of casting pointer to pointer.  This means casting pointers this way is really no better than loading only half of the 128-bit `&dyn` from the SFT, as what build4 did.

k-sareen:
Could you run a smaller experiment with a different machine? bobcat is asymmetrical so scheduling may affect the results. (Or run just on the performance cores on bobcat)

src/policy/space.rs
k-sareen: > Embarrassingly, neither the Rust reference nor the Rustonomicon specify the semantics of casting pointer to pointer.

Not sure I follow. Doesn't the first rule [here](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/casting-between-types.html#pointer-casts) describe when it's allowed? Or do you mean the semantics of then deferencing the pointer, which is most likely undefined

wks: > Not sure I follow. Doesn't the first rule [here](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/casting-between-types.html#pointer-casts) describe when it's allowed? Or do you mean the semantics of then deferencing the pointer, which is most likely undefined

The link you provided points to an ancient version of the Rust Book for v1.25.  The newest version of the Rust Book does not contain that section.

I did mean the casting itself.  It is not an no-op bits-preserving cast like `transmute`, as it changes the number of bits.

k-sareen: Ah right. That was the first link on Google for me when I searched for "raw pointer casting Rust".

wks:
To further investigate the bi-modal distribution in my microbenchmark, I plotted the GC time of each GC in the first SemiSpace run with build2.

image

The GC time is jumping back and forth between two values.  I suspect the difference is the cost of a failed `in_space` test.  When running SemiSpace and `PlanTraceObject`, the macro-generated `trace_object` checks if an object is in `copyspace1` before checking if it is in `copyspace2`.  So in odd GCs, `copyspace1` is the from space, while in even GCs, `copyspace2` is the from space.  In my microbenchmark, the data structure is a large binary tree, so all from-space objects are visited by exactly one edge (i.e. it will not see edges pointing to objects that are already forwarded).

-   When `copyspace1` is the from-space, most objects will be in `copyspace1`, and the first `copyspace1.in_space(object)` check will usuallly succeed.
-   When `copyspace2` is the from-space, most objects will be in `copyspace2`.  But the generated `trace_object` function will still call `copyspace1.in_space(object)` first, which will return `false`.  Then it will call `copyspace2.in_space(object)` which will return `true`.

So it will call `in_space` once in odd GCs, but twice in even GCs.  That's probably the reason behind the bimodal distribution.

The curve for StickyImmix is different

image

This may indicate that the bimodal distribution is caused by something else that is periodic.

wks:
I ran the microbenchmark again on `fisher.moma`.

The overall result is consistent with `bobcat.moma`.  For non-generational plans, build2 is obviously slower; build3 is slightly less slower; build4 is almost as good as build1.   SemiSpace still exhibit the bi-modal distribution, but StickyImmix no longer does.

image

Semispace build2 run2:

image

StickyImmix build2 run2:

image

wks:
From our previous discussion, although the `descriptor_map` is not used in Map64 now, it is still useful, and one possible use case is identifying which space an object is in, and enqueue edges into specialised work packets for the space it points into.  So we shouldn't simply remove `descriptor_map`.  I am closing this PR without merging.

Issue #932, opened 2023-08-31
Refactor VMMap and introduce ChunkResource


wks:
# Status quo

A Space allocates objects.  A Space relies on a PageResource to allocates contiguous pages.  Currently, a PageResource asks VMMap for "contiguous chunks" if needed.

`VMMap` has two implementations: `Map32` and `Map64`.  They are named after the word size (32 and 64).  I think it is misleading.  They actually implement two completely different strategies.

-   `Map32` implements a "free-list-based discontiguous chunk allocator".
-   `Map64` implements a "monotonic contiguous chunk allocator".

## Map32: free-list-based, discontiguous

On 32-bit machines, the address space is precious.  Only a few spaces (Currently only the Gen nursery.  Previously the VM space, too.) occupy contiguous regions.  Other spaces share the remaining region of memory.  All free chunks in this region are owned by a free list `Map32::region_map`.  All discontiguous spaces can ask for "contiguous chunks" from this central free list.

When allocating a "contiguous chunks", 

1.  It grabs chunks from `region_map: IntArrayFreeList`.
2.  Then it links the "contiguous chunks" in a per-space linked list.  The head of the linked list is ield in `CommonPageResource::head_discontiguous_region` on behalf of the space, and the links are held in `prev_link: Vec<i32>` and `next_link: Vec<i32>` indexed by chunk index.
3.  Then it writes to `descriptor_map: Vec<SpaceDescriptor>` which is indexed by chunk.  For each chunk the "contiguous chunks" cover, it writes the SpaceDescriptor of it into this `descriptor_map`.  It describes which space owns this chunk.

Map32 is designed this way so that multiple discontiguous spaces can grab chunk from a freelist (`region_map`) which manages the shared region.  And Map32 keeps track of "which space owns which chunk".

Note that on 32-bit systems, both MonotonicPageResource and FreeListPageResource can use this discontiguous chunk allocation strategy.  In SemiSpace, for example, both CopySpace instances are discontiguous, but use MonotonicPageResource.  MonotonicPageResource, when created as "discontiguous", will gradually ask for "contiguous chunks" from Map32 (in `grow_discontiguous_space`), and then adjust its cursor and sentinel to the newly obtained "contiguous chunks".  In this way, the MonotonicPageResource allocates pages monotonically, while the underlying "chunk resource" is discontiguous.

[...]

PR #953, opened 2023-09-12
Remove fl_map and refactor FreeListPageResoure

wks:
This PR removes the `shared_fl_map` field from `Map32`, and removes `fl_map` and `fl_page_resource` fields from `Map64`.  These global maps were used to enumerate and update the starting address of spaces (stored in `CommonFreeListPageResource` instances), for several different purposes.

1.  `Map32` can only decide the starting address of discontiguous spaces after all contiguous spaces are placed, therefore it needs to modify the start address of all discontiguous `FreeListPageResource` instances.
    -   `shared_fl_map` was used to locate `CommonFreeListPageResource` instances by the space ordinal.
    -   With this PR, we use `Plan::for_each_space_mut` to enumerate spaces, and use the newly added `Space::maybe_get_page_resource_mut` method to get the page resource (if it has one) in order to update their starting addresses.
2.  `Map64` allocates `RawMemoryFreeList` in the beginning of each space that uses `FreeListPageResource`.  It needs to update the start address of each space to take the address range occupied by `RawMemoryFreeList` into account. 
    -   `fl_page_resource` was used to locate `CommonFreeListPageResource` instances by the space ordinal.
    -   `fl_map` was used to locate the underlying `RawMemoryFreeList` instances of the corresponding `FreeListPageResource` by the space ordinal.
    -   With this PR, we initialize the start address of the `FreeListPageResource` immediately after a `RawMemoryFreeList` is created, taking the limit of `RawMemoryFreeList` into account.  Therefore, it is no longer needed to update the starting address later.

Because those global maps are removed, several changes are made to `VMMap` and its implementations `Map32` and `Map64`.

-   The `VMMap::boot` and the `VMMap::bind_freelist` no longer serve any purpose, and are removed.
-   `Map64::finalize_static_space_map` now does nothing but setting `finalized` to true.

The `CommonFreeListPageResource` data structure was introduced for the sole purpose to be referenced by those global maps.  This PR removes `CommonFreeListPageResource` and `FreeListPageResourceInner`, and relocates their fields.

-   `common: CommonPageResource` is now a field of `FreeListPageResource`.
-   The `free_list: Box<dyn FreeList>` and `start: Address` are moved into `FreeListPageResourceSync` because they have always been accessed while holding the mutex `FreeListPageResource::sync`.
    -   Note that the method `FreeListPageResource::release_pages` used to call `free_list.size()` without holding the `sync` mutex, and subsequently calls `free_list.free()` with the `sync` mutex.  The algorithm of `FreeList` guarantees `free()` will not race with `size()`.  But the `Mutex` forbids calling the read-only operation `free()` without holding the lock because in general it is possible that a read-write operations may race with read-only operations, too.  For now, we extended the range of the mutex to cover the whole function.

After the changes above, the `FreeListPageResource` no longer needs `UnsafeCell`.

This PR is one step of the greater goal of removing unsafe operations related to FreeList, PageResource and VMMap.  See: https://github.com/mmtk/mmtk-core/issues/853

Related PRs:
-   https://github.com/mmtk/mmtk-core/pull/925: This PR is a subset of that PR.


wks:
I ran lusearch from DaCapo Chopin on vole.moma, 3x min heap size w.r.t. G1, 40 invocations, 5 iterations each, comparing master against this PR.

image

[plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2024-05-06-Mon-143938&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&)

To my surprise, the STW time increased for all plans except MarkSweep.  I thought the increased range of mutex in `FreeListPageResource::release_pages` would negatively impact MarkSweep, and other plans should not be affected by this change.  I'll investigate why it became slower.

src/mmtk.rs
```diff
@@ -188,9 +188,18 @@ impl<VM: VMBinding> MMTK<VM> {
         }
 
         // TODO: This probably does not work if we have multiple MMTk instances.
-        VM_MAP.boot();
         // This needs to be called after we create Plan. It needs to use HeapMeta, which is gradually built when we create spaces.
-        VM_MAP.finalize_static_space_map(heap.get_discontig_start(), heap.get_discontig_end());
+        VM_MAP.finalize_static_space_map(
+            heap.get_discontig_start(),
+            heap.get_discontig_end(),
+            &mut |start_address| {
```
qinsoon, line 195: The use of callback here is not quite intuitive. Some comments are needed. When does this closure get called? What is the argument? What does the closure do?

src/util/heap/freelistpageresource.rs
```diff
@@ -181,27 +144,21 @@ impl<VM: VMBinding> PageResource<VM> for FreeListPageResource<VM> {
...
-            });
-            // `CommonFreeListPageResource` lives as a member in space instances.
-            // Since `Space` instances are always stored as global variables, so it is okay here
-            // to turn `&CommonFreeListPageResource` into `&'static CommonFreeListPageResource`
-            unsafe {
-                vm_map.bind_freelist(common_flpr.as_ref() as *const _);
-            }
-            common_flpr
+        let free_list = vm_map.create_parent_freelist(start, pages, PAGES_IN_REGION as _);
+        let actual_start = if let Some(rmfl) = free_list.downcast_ref::<RawMemoryFreeList>() {
```
qinsoon, line 148: We could let `dyn FreeList` to return a value based on the implementation rather than downcasting the type here. The `FreeList` implementation is internal to the `VMMap`, and we should not break the abstraction and let `PageResource` do things based on the `FreeList` implementation.

src/util/heap/freelistpageresource.rs
```diff
@@ -210,42 +167,25 @@ impl<VM: VMBinding> FreeListPageResource<VM> {
...
-            // to turn `&CommonFreeListPageResource` into `&'static CommonFreeListPageResource`
-            unsafe {
-                vm_map.bind_freelist(common_flpr.as_ref() as *const _);
-            }
-            common_flpr
-        };
+        let start = vm_layout().available_start();
+        let free_list = vm_map.create_freelist(start);
+        debug_assert!(
+            free_list.downcast_ref::<RawMemoryFreeList>().is_none(),
```
qinsoon, line 173: Same here. We could let `dyn FreeList` to tell us if discontiguous space is allowed.

src/util/heap/freelistpageresource.rs
```diff
@@ -311,29 +251,25 @@ impl<VM: VMBinding> FreeListPageResource<VM> {
...
         let mut rtn = freelist::FAILURE;
         let required_chunks = crate::policy::space::required_chunks(pages);
-        let region = self
-            .inner()
-            .common
-            .grow_discontiguous_space(space_descriptor, required_chunks);
+        let maybe_rmfl = sync.free_list.downcast_mut::<RawMemoryFreeList>();
+        let region =
+            self.common
+                .grow_discontiguous_space(space_descriptor, required_chunks, maybe_rmfl);
```
qinsoon, line 257: We could use `Option<&mut dyn FreeList>` instead of `Option<&mut RawMemoryFreeList>` for `grow_discontiguous_space` and `allocate_contiguous_chunks`. In that case, the downcast happens in `VMMap`.

src/util/heap/freelistpageresource.rs
wks: I don't think `FreeList` is an implementation detail of `VMMap`.  It is a very general free list abstraction, and has two implementations, namely `IntArrayFreeList` and `RawMemoryFreeList`.  In this sense, I don't think `trait FreeList` should contain a method that returns the number of bytes it occupies in the space, because a `FreeList` is not necessarily used by a space.  It can be used to allocate any kind of resources.

But it is true that `VMMap` *selects* the implementation.  It is `VMMap::create_freelist` that decide whether to reserve a proportion of memory at the beginning of the space for the `RawMemoryFreeList`.  I think it is natural for `VMMap::create_freelist` to return the number of bytes reserved for the `FreeList` implementation alongside the `dyn FreeList`, i.e. it should return a tuple.

In my previous version of the PR, I introduced the [CreateFreeListResult](https://github.com/mmtk/mmtk-core/pull/925/files#diff-33c211ce4d7709987021ccf3ac97406730facd09ab61c5fae0e8c8a15a284489R7) type:

```rust
pub struct CreateFreeListResult {
    pub free_list: Box<dyn FreeList>,
    pub space_displacement: usize,
}
```

I think I can move it to this PR, too.  If that's too verbose, I can replace it with a simple tuple instead.

src/util/heap/freelistpageresource.rs
wks: Like I said, I don't think `FreeList` should be aware of whether it is used by one kind of space or another.

Our current implementation happens to disallow discontiguous spaces when using `Map64`, and `Map64` happens to select `RawMemoryFreeList`.  I think we can remove this assertion because it's only the status quo, and there is no reason why we can't use `RawMemoryFreeList` in discontiguous spaces.

wks: After a second thought, I think we should leave the comment as is, because it is an implementation detail of the current `FreeListPageResource`.  The way we deal with discontiguous space is that the `start` variable here is meaningless.  But when the address range of the discontiguous space is determined, we re-assign the `address` of all discontiguous page resources.  (That's the callback you saw in `mmtk.rs`.)  If `start` needs an offset to make room for the `RawMemoryFreeList`, it will be overwritten.

We could adapt to that possibility by requiring `start` to be initially zero (when using `IntArrayFreeList`) or the offset (when using `RawMemoryFreeList`) for discontiguous spaces.  When the discontiguous range is determined, we *add* the starting address to the discontiguous range to the `start` variable instead of overwriting it.  But we have no way to test that because we never have that use case, and I already proposed to revamp the way we create spaces (see [here](https://github.com/mmtk/mmtk-core/issues/932) and [here](https://github.com/mmtk/mmtk-core/issues/984)).  So we can leave this assertion here to reflect the status quo.  `downcast_ref` is not as elegant as virtual methods from OOP's point of view, but we are just inspecting an implementation detail, and it is in an assertion.  I'll add comment to clarify that this is only the status quo, not something that has to be true in general.

src/util/heap/freelistpageresource.rs
qinsoon: Returning an offset from the create functions sounds good.

wks:
> [plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Cvole-2024-05-06-Mon-143938&benchmark%5Ebuild%5Einvocation%5Eiteration%5Emmtk_gc&GC%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4%7C20&1%5Einvocation%7C30&1&benchmark%5Emmtk_gc&build;build1%7C41&Histogram%20(with%20CI)%5Ebuild%5Emmtk_gc&)
> 
> To my surprise, the STW time increased for all plans except MarkSweep. I thought the increased range of mutex in `FreeListPageResource::release_pages` would negatively impact MarkSweep, and other plans should not be affected by this change. I'll investigate why it became slower.

The slow-down can be easily reproduced on my personal computer.  Just compile MMTk and OpenJDK as usual and we can observe the performance difference in a microbenchmark that triggers GC many times.

The performance difference is due to some code change that subtly changed the inlining decision of the Rust compiler.  `perf` shows that after this PR, one of the compare-exchange operation in `object_forwarding::attempt_to_forward` is not inlined, and is taking a significant amount time.  The inlining decisions on the hot path, including the `do_work` of `ProcessEdgesWork` implementations and the `trace_object` of the spaces, are different in the two builds.

The slowdown is only reproducible by setting `profile.release.opt-level=2` or above in `Cargo.toml`.  If I set it to 1 or lower, there will be no performance difference regardless whether this PR is applied or not.

Compile both versions using PGO and the difference will go away.

I'll run the benchmarks again with PGO after addressing the review comments.

k-sareen:
You can mark the function as `#[hot]` then.

wks:
> You can mark the function as `#[hot]` then.

But that goes against our policy of not interfering with the compiler's inlining decision and relying on PGO to make more informed decision.

k-sareen:
The policy is not to never add them, but to add them judiciously and for proper reasons, specially if they make the PGO build faster itself. PGO is not always perfect.

src/mmtk.rs
wks: I added comments to `finalize_static_space_map` and this closure.

src/util/heap/freelistpageresource.rs
wks: I made `create_freelist` return `CreateFreeListResult` so that `FreeListPageResource` no longer performs downcast.

src/util/heap/freelistpageresource.rs
wks: I added comments to this assertion.

src/util/heap/freelistpageresource.rs
wks: I made the change as you suggested.  It makes sense because if the page resource uses free list, and Map64 can grow the space, they always need to grow the free list anyway.  It just requires `Map64` to choose a growable `FreeList` implementation.

wks:
> The policy is not to never add them, but to add them judiciously and for proper reasons, specially if they make the PGO build faster itself. PGO is not always perfect.

To be clear, when I observed the performance difference, I was not using PGO.  I always worried that PGO may be another source of non-determinism.  But this time, PGO is working properly and eliminated the performance difference.

qinsoon (approved):
LGTM

wks:
I am running the benchmark again on vole.moma, with PGO enabled.

wks:
I re-run lusearch from DaCapo Chopin on vole.moma.  40 invocations, 5 iterations.  This time I compiled mmtk-openjdk with PGO (training with fop).

3x min heap size w.r.t. G1 ([plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2024-05-08-Wed-081153&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&))

image

4.4x min heap size w.r.t. G1 ([plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2024-05-08-Wed-165640&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&))

image

The total time and STW time still get worse.  The difference is bigger with larger heap size.  Although the confidence interval is large, the difference is still obvious.

qinsoon:
Does this PR affect performance for `SemiSpace` and `GenCopy`? They mostly use `MonotonePageResource` which seems unchanged in this PR.

wks:
I ran it again.  This time I trained the PGO using only StickyImmix, and not using stress GC, and re-ran lusearch with StickyImmix.  4.4x min heap size.  ([plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2024-05-09-Thu-074303&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|40&Histogram%20(with%20CI)^build^mmtk_gc&))

image

My conclusion is that this thing is extremely sensitive to the profile.  Previously, I did PGO using MarkSweep, SemiSpace, MarkCompact, Immix, GenImmix, StickyImmix, and set a stress factor of 4M.  Maybe that made the Rust compiler think some functions on the hot paths are not worth inlining, or maybe the stress GC made it miss some hot paths not executed during non-stress GC.

I also tried locally and found that when I set the optimization level to 2, it failed to inline some function and caused slowdown.  This also shows the sensitivity to optimisation.

qinsoon:
> From the result, this PR is faster than the master branch.

I don't think we can say it is faster. The difference is so tiny and could be just noise. Probably run all the benchmarks for the plans that may be affected with this PR.

wks:
The same binary as [the last one](https://github.com/mmtk/mmtk-core/pull/953#issuecomment-2102202489), the same setting, except running other benchmarks, too.  (I omitted some benchmarks as before since their times increase as the number of iterations go up, probably due to the lack of reference processing.)

Results:  ([plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2024-05-09-Thu-142934&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark&build;build1|41&Histogram%20(with%20CI)^build^benchmark&))

image

The means are between -1.5% and +0.6% of the master, but the error bars are larger than the difference.  That means the performance difference is insignificant.

The STW time of graphchi is reduced from 57 ms to 54 ms (see the [un-normalized version](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|vole-2024-05-09-Thu-142934&benchmark^build^invocation^iteration^mmtk_gc&GC^majorGC^time^time.other^time.stw&majorGC%20/%20GC|10&iteration^1^4|20&1^invocation|41&Histogram%20(with%20CI)^build^benchmark&)), a 5% reduction, and the error bars are narrow.  Similar is true for kafka.

qinsoon:
@wks You can merge the PR when you think it is ready.

Issue #636, opened 2022-08-03
VM overriding default Options


wks:
The VM may want to provide a different set of default values for MMTk options than those defined in `src/util/options.rs`.  For example,
-   In `mmtk-core`, the default `plan` is `NoGC`, and the default `heap_size` is ~~512MiB~~ 50% of the physical memory.
-   The Ruby binding wants to set the default plan to `MarkSweep`, and the default `heap_size` to ~~80% the physical memory~~ (since the GCTrigger feature is available) dynamically resizing between 1MiB to 80% of the physical memory.

But the Ruby binding still allows the user to set those options via command line options or environment variables.  For example,

-   If the user specified `--mmtk-plan=Immix` or set env var `MMTK_PLAN=Immix`, MMTk should use the Immix plan; and
-   if the user specified `--mmtk-max-heap=16MiB` or set env var `THIRD_PARTY_HEAP_LIMIT=16777216`, the heap size should be 16MiB.

So we shall have the following 4 ways to set MMTK options, with increasing priorities:

**mmtk-core defaults < Ruby defaults < environment variables < command line arguments**

The current MMTk core does not allow us to put Ruby defaults between mmtk-core defaults and environment variables, because when the `Options` struct is created using `Options::default()`, it populates the instance with mmtk-core defaults, and immediately look for environment variables, leaving no time in between.

One way to solve this problem is provide two methods:
1.  One method creates an `Options` struct, and populates it with the default values defined in `src/util/options.rs`, and
2.  The other method, possibly an instance method on `Options`, applies `MMTK_*` env vars to the `Options` instance.

By doing this, Ruby can inject its own defaults between (1) and (2).

# Example

Let's use Ruby as an example.  Ruby wants to override the default *plan* and the default *gc trigger* to dynamic resizing.  The mmtk-ruby binding should do the following:

1.  The mmtk-ruby binding instantiates an `mmtk::util::Options` structure filled with default values.
2.  The mmtk-ruby binding overrides the default values:
    -   sets `Options::plan` to `MarkSweep`, and 
    -   sets `Option::gc_trigger` to `Dynamic`
3.  The mmtk-ruby calls a function provided by mmtk-core (for example, `Options::parse_from_env_var`) to parse environment variables and update the fields of `Options`
    -   For example, if the env var `MMTK_PLAN=Immix` is set, it changes `Options::plan` to Immix; if not, it remains to be `MarkSweep`.
4.  The mmtk-ruby parses its own command line arguments and sets options
[...]

PR #955, opened 2023-09-13
Let VM control when or if to read env var options

wks:
Allow Options to be created with MMTK's built-in defaults without reading options from environment variables, and let the VM decide when to read options from environment variables.

This will allow VMs to provide default options, override MMTK's default options, but can be overridden by environment variables.  It will also allow VMs to completely disable the "MMTK_*" variables.

For compatibility reasons, this PR does not change the default behavior of MMTKBuilder.  MMTKBuilder::new will still read from environment variables, but the new constructor MMTKBuilder::new_no_env_vars will skip environment variables.

Fixes: https://github.com/mmtk/mmtk-core/issues/636

### DIFF
--- a/src/util/options.rs
+++ b/src/util/options.rs
@@ -237,7 +237,7 @@ macro_rules! options {
             }
 
             /// Create an `Options` instance with built-in default settings.
-            pub fn new() -> Self {
+            fn new() -> Self {
                 Options {
                     $($name: MMTKOption::new($default, $validator, $env_var, $command_line)),*
                 }
@@ -263,6 +263,7 @@ macro_rules! options {
         }
 
         impl Default for Options {
+            /// By default, `Options` instance is created with built-in default settings.
             fn default() -> Self {
                 Self::new()
             }
--- a/src/util/options.rs
+++ b/src/util/options.rs
@@ -235,27 +235,36 @@ macro_rules! options {
                     _ => panic!("Invalid Options key: {}", s)
                 }
             }
-        }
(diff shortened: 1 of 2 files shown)


src/util/options.rs
```diff
@@ -235,27 +235,36 @@ macro_rules! options {
...
-        impl Default for Options {
-            fn default() -> Self {
-                let mut options = Options {
-                    $($name: MMTKOption::new($default, $validator, $env_var,$command_line)),*
-                };
 
-                // If we have env vars that start with MMTK_ and match any option (such as MMTK_STRESS_FACTOR),
-                // we set the option to its value (if it is a valid value). Otherwise, use the default value.
+            /// Create an `Options` instance with built-in default settings.
+            pub fn new() -> Self {
```
qinsoon, line 240: The public API changes include this function.
```console
Added items to the public API
=============================
+pub fn mmtk::util::options::Options::new() -> Self
+pub fn mmtk::util::options::Options::read_env_var_settings(&mut self)
+pub fn mmtk::MMTKBuilder::new_no_env_vars() -> Self
```

But this function does not need to public. The other two are fine.

wks: Done.  I have made `new()` private.

`Options` still implements `Default`, so VMs can still instantiate it, although `Options` should usually be created by the `MMTKBuilder`.

qinsoon (approved):
LGTM

wenyuzhao (approved):

PR #956, opened 2023-09-18
Fix length of Map64::descriptor_map

wks:
Its length should be MAX_SPACES instead of MAX_CHUNKS.

The size was `MAX_SPACES` in the JikesRVM MMTk.  It may have been a typo when porting JikesRVM MMTk to Rust.

The new `Map64::descriptor_map` does not deserve the unsafe `new_zeroed_vec` function which was introduced to solve the slow start-up problem.  We keep `new_zeroed_vec` just in case we may still need it in the future.

### DIFF
--- a/src/util/heap/layout/map64.rs
+++ b/src/util/heap/layout/map64.rs
@@ -8,7 +8,6 @@ use crate::util::heap::layout::vm_layout::*;
 use crate::util::heap::space_descriptor::SpaceDescriptor;
 use crate::util::memory::MmapStrategy;
 use crate::util::raw_memory_freelist::RawMemoryFreeList;
-use crate::util::rust_util::zeroed_alloc::new_zeroed_vec;
 use crate::util::Address;
 use std::cell::UnsafeCell;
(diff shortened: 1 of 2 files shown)


wks:
@wenyuzhao  We used `vm_layout()::max_chunks()` to find the number of chunks, but I didn't find a method in `VMLayout` that returns the number of spaces.  I assume we should still use `MAX_SPACES`, shouldn't we?  Did I miss anything?

qinsoon (approved):
LGTM

wenyuzhao:
> @wenyuzhao We used `vm_layout()::max_chunks()` to find the number of chunks, but I didn't find a method in `VMLayout` that returns the number of spaces. I assume we should still use `MAX_SPACES`, shouldn't we? Did I miss anything?

Yes I think `MAX_SPACES` is the correct one to use. Unlike `max_chunks` which is related to the vm layout defined at runtime, max spaces is a hardcoded constant.

wenyuzhao (approved):

Issue #959, opened 2023-09-24
Why is ImmixAllocator reset twice?


wks:
In `src/plan/immix/mutator.rs`:

```rust

pub fn immix_mutator_prepare<VM: VMBinding>(mutator: &mut Mutator<VM>, _tls: VMWorkerThread) {
    let immix_allocator = unsafe {
        mutator
            .allocators
            .get_allocator_mut(mutator.config.allocator_mapping[AllocationSemantics::Default])
    }
    .downcast_mut::<ImmixAllocator<VM>>()
    .unwrap();
    immix_allocator.reset();
}

pub fn immix_mutator_release<VM: VMBinding>(mutator: &mut Mutator<VM>, _tls: VMWorkerThread) {
    let immix_allocator = unsafe {
        mutator
            .allocators
            .get_allocator_mut(mutator.config.allocator_mapping[AllocationSemantics::Default])
    }
    .downcast_mut::<ImmixAllocator<VM>>()
    .unwrap();
    immix_allocator.reset();
}
```

The `immix_allocator.reset()` invocation does the following:

```rust
impl<VM: VMBinding> ImmixAllocator<VM> {
    pub fn reset(&mut self) {
        self.bump_pointer.reset(Address::ZERO, Address::ZERO);
        self.large_bump_pointer.reset(Address::ZERO, Address::ZERO);
        self.request_for_large = false;
        self.line = None;
    }
}
```

This method is idempotent, and we don't touch the allocators during GC.  Why do we need to reset the `ImmixAllocator` in both `prepare_func` and `release_func`?

p.s. This behaviour exists in JikesRVM, and it has behaved like this ever since Immix was introduced into JikesRVM MMTk.

wks:
One possible explanation is that it intended to reset `ImmixCollector.copy` (the counterpart of `ImmixCopyContext::copy_allocator` in Rust MMTk) during the PREPARE phase because it is used for evacuating objects.  But it mistakenly reset the `ImmixAllocator` of the mutator, instead.

PR #962, opened 2023-09-25
Do not reset ImmixAllocator when preparing mutator

wks:
It is unnecessary to reset the ImmixAllocator of a mutator when preparing mutator because the mutator's ImmixAllocator is not used during GC.  When a worker defragments the heap and promotes young objects, it uses the ImmixAllocator in ImmixCopyContext or ImmixHybridCopyContext.

The ImmixAllocator of a mutator still needs to be reset after GC (when releasing mutators) because the blocks it was allocating into could have been defragmented or recycled.  So it needs to reset the bump pointers.

Fixes: https://github.com/mmtk/mmtk-core/issues/959

### DIFF
--- a/src/plan/immix/mutator.rs
+++ b/src/plan/immix/mutator.rs
@@ -15,16 +15,7 @@ use crate::{
 };
 use enum_map::EnumMap;
 
-pub fn immix_mutator_prepare<VM: VMBinding>(mutator: &mut Mutator<VM>, _tls: VMWorkerThread) {
-    let immix_allocator = unsafe {
-        mutator
-            .allocators
-            .get_allocator_mut(mutator.config.allocator_mapping[AllocationSemantics::Default])
-    }
-    .downcast_mut::<ImmixAllocator<VM>>()
-    .unwrap();
-    immix_allocator.reset();
-}
+pub fn immix_mutator_prepare<VM: VMBinding>(_mutator: &mut Mutator<VM>, _tls: VMWorkerThread) {}
 
 pub fn immix_mutator_release<VM: VMBinding>(mutator: &mut Mutator<VM>, _tls: VMWorkerThread) {
     let immix_allocator = unsafe {


wks:
The behaviour of 'resetting ImmixAllocator of mutators in PREPARE' can be traced back to the commit in JikesRVM when Immix was first introduced into MMTk (see [this](https://github.com/JikesRVM/JikesRVM/blob/f10ba059ebdb801cc66e2cb6552a6f4a3a0f193d/MMTk/src/org/mmtk/policy/immix/MutatorLocal.java#L46)), and it still behaves this way in the latest version of JikesRVM (see [this](https://github.com/JikesRVM/JikesRVM/blob/5072f19761115d987b6ee162f49a03522d36c697/MMTk/src/org/mmtk/policy/immix/MutatorLocal.java#L44)).  During today's meeting, @wenyuzhao reported that in LXR we only need to clear the ImmixAllocator once during release.  I think the `reset()` in `prepare()` in the original code was intended to reset the ImmixAllocator used for defragmentation, but mistakenly reset the mutator's ImmixAllocator too.

@steveblackburn Could you have a look at the code, too, to see if I missed anything?

src/plan/immix/mutator.rs
```diff
@@ -15,16 +15,7 @@ use crate::{
...
-    let immix_allocator = unsafe {
-        mutator
-            .allocators
-            .get_allocator_mut(mutator.config.allocator_mapping[AllocationSemantics::Default])
-    }
-    .downcast_mut::<ImmixAllocator<VM>>()
-    .unwrap();
-    immix_allocator.reset();
-}
+pub fn immix_mutator_prepare<VM: VMBinding>(_mutator: &mut Mutator<VM>, _tls: VMWorkerThread) {}
```
wenyuzhao, line 18: Is it possible to not call this function, instead of making it a no-op?

wks: This function is called in the `PrepareMutator` work packet.  `PrepareMutator` simply calls `mutator.prepare` which calls `(*self.config.prepare_func)(self, tls)`.  We can avoid calling `prepare_func` by changing `prepare_func` to an `Option<...>` and test if it is `Some(...)` before calling.  Alternatively, we can assign a dedicated no-op function to `prepare_func` and test if it points to the no-op function before calling.  But I think both methods are too complicated.  Because `MutatorConfig::prepare_func` is only called once per GC, it is not worth eliminating this function call.

wenyuzhao: > Because MutatorConfig::prepare_func is only called once per GC, it is not worth eliminating this function call.

This is what I was worrying about. Some Google internal workloads can have over 1000 mutator threads...

qinsoon: > > Because MutatorConfig::prepare_func is only called once per GC, it is not worth eliminating this function call.
> 
> This is what I was worrying about. Some Google internal workloads can have over 1000 mutator threads...

Did you find `Release` and `Prepare` takes up a considerable proportion of the total time spent in GC in those cases?

wks: > This is what I was worrying about. Some Google internal workloads can have over 1000 mutator threads...

Honestly they should use OpenJDK 21 and Virtual Threads.  But if they insist on using native threads, we may try to eliminate the need to create the `PrepareMutator` work packet.  We can add a per-plan constant, or a field in `PlanConstraint`, like `need_prepare_mutator: bool`.  If it is false, we won't create the `PrepareMutator` work packet in the first place.

qinsoon: > > This is what I was worrying about. Some Google internal workloads can have over 1000 mutator threads...
> 
> Honestly they should use OpenJDK 21 and Virtual Threads. But if they insist on using native threads, we may try to eliminate the need to create the `PrepareMutator` work packet. We can add a per-plan constant, or a field in `PlanConstraint`, like `need_prepare_mutator: bool`. If it is false, we won't create the `PrepareMutator` work packet in the first place.

Having a constant in `PlanConstraint` and avoiding the work packet creation sounds like a good idea.

wks:
I have created a more aggressive PR https://github.com/mmtk/mmtk-core/pull/964 that allows plans to set `PlanConstraints::need_prepare_mutator` to `false` and disable the creation of `PrepareMutator`.  It removes the `prepare_func` of most plans, including `Immix` and `StickyImmix`.  That PR will subsume this PR.

wks:
Superseded by https://github.com/mmtk/mmtk-core/pull/964

PR #964, opened 2023-09-27
Disabling PrepareMutator from PlanConstraints

wks:
Some plans do nothing when preparing mutators.  We add a boolean flag so that we do not create the PrepareMutator work packets in the first place.

We also remove the function that prepares Mutator for Immix.  It is unnecessary to reset the ImmixAllocator of a mutator when preparing mutator because the mutator's ImmixAllocator is not used during GC. When a GC worker defragments the heap and promotes young objects, it uses the ImmixAllocator instances in ImmixCopyContext or ImmixHybridCopyContext.

Fixes: https://github.com/mmtk/mmtk-core/issues/959


qinsoon (approved):
LGTM

wenyuzhao (approved):

PR #968, opened 2023-10-05
Log space usage for a plan

qinsoon:


caizixian:
I think we should also make this a tracepoint.

PR #969, opened 2023-10-06
Immix: disable is_defrag_source check in get_forwarded_object

playXE:
This PR removes an incorrect check in `ImmixSpace::get_forwarded_object()`. The problem for this check is that defrag is not the only reason that an Immix object may gets moved. For example, in sticky immix, a nursery object may be moved in a nursery GC. In such cases, the block is not a defrag source and this check would return `None` suggesting there is no forwarded object. But the object may be actually moved.

### DIFF
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -86,10 +86,6 @@ impl<VM: VMBinding> SFT for ImmixSpace<VM> {
     }
 
     fn get_forwarded_object(&self, object: ObjectReference) -> Option<ObjectReference> {
-        if !Block::containing::<VM>(object).is_defrag_source() {
-            return None;
-        }
-
         if ForwardingWord::is_forwarded::<VM>(object) {
             Some(ForwardingWord::read_forwarding_pointer::<VM>(object))
         } else {


wks:
The reason why the 'is defrag source' property is check in the first place is that the forwarding bits are cleared lazily.  In this case, if an object is not reachable in this GC, but there is a stale forwarding bit at its address left over from the previous GC, then the `is_live` method will think it is alive because it seems 'forwarded', and the `get_forwarded_object` method will erroneously attempt to load its forwarding pointer because it thinks it is forwarded, too, while it is actually not visited at all.

Since defrag is not the only reason why an object can be moved, we need to think about the clearing of forwarding bits carefully.

For correctness, we may choose to clear the forwarding bits eagerly, during Prepare or Release.  By doing this, the VM binding can ask the question 'Is this object forwarded?' during `Scanning::process_weak_refs` (handling weak references) and `Collection::post_forwarding` (when enqueuing finalizable objects) which are currently the only two places where the `get_forwarded_object` method makes sense for the binding.

For performance, we should clear as few bits as possible because Immix (including StickyImmix) seldom moves any objects (if at all) during a GC.  For Immix and GenImmix, because we only move objects in the Immix space (not nursery) during defrag GC, we can just clear defrag sources.  For StickyImmix, we may choose to clear the forwarding bits for lines that contain young objects (or at a coarser grain, clear the forwarding bits of blocks that contain young objects).  We may clear forwarding bits the same way we clear VO bits in the `ClearAndReconstruct` strategy.

We may also evaluate how much the actual cost of clearing all forwarding bits is, because we are clearing all marking bits in major GCs anyway.  Although it is fast to clear bits in bulk, it may hurt performance if the heap is large.

qinsoon:
I think we probably want to clear forwarding bits when we release the blocks (unless we see a measurable slowdown). Essentially freeing those blocks means there is no forwarded object in them any more.

qinsoon:
This issue probably also applies to the similar check in `is_live()`:
https://github.com/mmtk/mmtk-core/blob/9a676e68dd41272b57e6812dc404ffa6391668e2/src/policy/immix/immixspace.rs#L112

playXE:
> I think we probably want to clear forwarding bits when we release the blocks (unless we see a measurable slowdown). Essentially freeing those blocks means there is no forwarded object in them any more.

Would that be done here? https://github.com/mmtk/mmtk-core/blob/master/src/policy/immix/immixspace.rs#L450

PR #971, opened 2023-10-08
Fix performance baseline workflow

qinsoon:
This PR fixes our workflow to run performance tests for JikesRVM's and OpenJDK's stock GCs. It combines the two workflows into one. The workflow is now triggered once per month, rather than once per week.


.github/workflows/perf-baseline.yml
```diff
@@ -0,0 +1,99 @@
+name: Performance Baseline
+
+on:
+  # Manual dispatch
+  workflow_dispatch:
+  # Or on a sunday 1200am UTC between 1st and 7th each month
+  schedule:
+    - cron: '0 0 1-7 * 0'
```
caizixian, line 8: This doesn't look right.
https://crontab.guru/#0_0_1-7_*_0

caizixian (changes requested):

.github/workflows/perf-baseline.yml
caizixian: https://stackoverflow.com/questions/3241086/how-to-schedule-to-run-first-sunday-of-every-month

qinsoon: Right. Thanks for finding the issue. There could be ways to get it work for the first Sunday in a month. But I just change it to run on the first day of a month to make things simple.

caizixian (approved):

Issue #972, opened 2023-10-09
Side forwarding bits are not cleared properly


wks:
When the forwarding bits metadata is on the side, we use an optimization which only clears the forwarding bits for blocks that are "defrag sources".  This approach is problematic.

-   For StickyImmix, young objects can be moved, too, if Immix is configured to prefer evacuating young objects in minor GCs.
-   In theory, Immix can opportunistically move any objects at any time, even if the current GC is not a defrag GC.  Although we are currently not doing this (the Immix plan only moves objects if the current GC is defrag GC), we should prepare for this possibility.

It is better to ensure the on-the-side forwarding bits are cleared after each GC so that when GC starts, forwarding bits will always be zero.

It is easier (and more general) to clear the forwarding bits for all live blocks, like how we clear the on-the-side mark bits.  We can do this as long as the performance overhead is minimal.  For performance, we may choose to only clear forwarding bits for blocks or lines where objects may actually be moved.

*Note: The best solution is always **using in-header forwarding bits** instead of on-the-side forwarding bits.  When a new object is allocated, the entire object will be zeroed, which implicitly clears the in-header forwarding bits.  The forwarding bits can be placed anywhere in the object as long as a given bit is guaranteed to be zero when GC happens.  That's because when an object is forwarded, its from-space copy is "condemned" which means its contents will never be accessed again.  However, we allow on-the-side forwarding bits because some VMs do not have any extra bits to spare in the object headers.*

playXE:
So I have a question regarding this. My binding uses in-header forwarding bits yet the bug occurs with StickyImmix only. Why would it happen in the first place if not this defrag check in `get_forwarded_object` and `is_live`?

wks:
> So I have a question regarding this. My binding uses in-header forwarding bits yet the bug occurs with StickyImmix only. Why would it happen in the first place if not this defrag check in `get_forwarded_object` and `is_live`?

I think that is a bug.  I implemented `get_forwarded_object` for ImmixSpace when developing the Ruby binding which uses on-the-side forwarding bits.  Ruby currently only supports Immix, but not StickyImmix, and that is why I didn't test it thoroughly with StickyImmix.

It didn't break other VM bindings because `get_forwarded_object` is only used during weak reference processing.
-   Only Ruby is actively using `get_forwarded_object` in weak reference processing.
[...]

PR #974, opened 2023-10-10
Clear all side forwarding bits

wks:
This reverts an optimization which was intended to reduce the number of side forwarding bits cleared for ImmixSpace by clearing the forwarding bits metadata of defrag sources, only.  This approach does not work for StickyImmix because it may move young objects, which is not "defrag".

Now ImmixSpace clears the on-the-side forwarding bits for all chunks just like how it clears the on-the-side marking bits.

This PR prefers correctness over performance, and may degrade the performance of VMs that use on-the-side forwarding bits.  Currently it only affects the Ruby binding, but we plan to let the Ruby binding use in-header forwarding states, too.  The performance of VMs that use in-header forwarding bits is not affected by this PR.

Fixes: https://github.com/mmtk/mmtk-core/issues/972

### DIFF
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -86,7 +86,8 @@ impl<VM: VMBinding> SFT for ImmixSpace<VM> {
     }
 
     fn get_forwarded_object(&self, object: ObjectReference) -> Option<ObjectReference> {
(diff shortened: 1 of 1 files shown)


qinsoon (approved):
LGTM

PR #975, opened 2023-10-12
Let the VM customize object forwarding implementation

wks:
**DRAFT: This is still a work in progress.**

Some VMs, such as Ruby, do not have sufficient amount of bits in the header for forwarding bits.  However, they can use the type tag to indicate the object is "condemned" and is forwarded to a different place.  This requires the VM to implement some object forwarding primitives for mmtk-core, as the representation of forwarding states (i.e. "not forwarded", "being forwarded", and "forwarded") is VM-specific.

### DIFF
--- a/src/policy/copyspace.rs
+++ b/src/policy/copyspace.rs
@@ -233,30 +233,25 @@ impl<VM: VMBinding> CopySpace<VM> {
         );
 
         trace!("attempting to forward");
-        let forwarding_status = object_forwarding::attempt_to_forward::<VM>(object);
+        match object_forwarding::ForwardingAttempt::<VM>::attempt(object) {
+            object_forwarding::ForwardingAttempt::Lost(lost) => {
+                trace!("... Lost the forwarding race.  Another thread is forwarding the object");
+                let new_object = lost.spin_and_get_forwarded_object();
+                trace!("... {} is already forwarded to {}", object, new_object);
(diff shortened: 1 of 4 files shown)


wks:
Preliminary experiments show that this change allows the Ruby binding to trigger fewer GCs (and have lower STW time) at tighter heap sizes, likely because of the heap space saved by eliminating the on-the-side forwarding bits.  At higher heap sizes (3x min heap size or larger), the difference between in-header and on-the-side forwarding bits becomes less significant.

wks:
I ran lusearch on shrew.moma, comparing build1 (master) and build2 (this PR), with no change in mmtk-openjdk, with 2.5x min heap size, 40 invocations.  The result is, the overhead in STW time is low (less than 1% for most plans except Immix), but not zero.  I think this abstraction should not have any overhead, and I'll look into it later.

Since we are not prioritising refactorings like this for now, I'll postpone this PR until later.

Plotty link: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|shrew-2023-11-01-Wed-081208&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

image

PR #976, opened 2023-10-12
Stop using the name "ForwardingWord"

wks:
The name is likely inherited from JikesRVM.  But since we changed its name, we should just use the module name "object_forwarding" directly.

This cleans up some naming issues, and prepares for subsequent refactoring that intends to change the logic of object forwarding in ImmixSpace.  At that time, I intend to fix this naming issue anyway.


qinsoon (approved):
LGTM

PR #987, opened 2023-10-17
Add a section about debugging tips in user guide

qinsoon:


docs/userguide/src/portingguide/debugging/prefix.md
```diff
@@ -0,0 +1,9 @@
+# Debugging Tips
+
+In this section, we offer insights into debugging strategies tailored for identifying issues
+in a binding implementation. Our emphasis is on the unique debugging tools and methods in MMTk.
```
k-sareen, line 4: I'm not sure I follow. Is this section about debugging tools that MMTk provides? Or general debugging tools/tips when working on an MMTk port? If it's the latter, then tracing (using eBPF) and `rr` should be mentioned. If the former, then tracing should still be mentioned since it is also an excellent debugging tool.

qinsoon: This section is about anything that is specific to MMTk and could be useful in debugging. For example, we probably won't discuss how to use `rr` or `gdb` in general, but if MMTk has functions that can be good breakpoints for `rr` or `gdb`, this is a good place to tell people about it.

docs/userguide/src/portingguide/debugging/assertions.md
```diff
@@ -0,0 +1,15 @@
+# Enabling debug assertions
+
+MMTk is implemented with an extensive amount of assertions to ensure the correctness.
+We strongly recommend using a debug build of MMTk that includes all the debugging assertions
+when one is developing on a MMTk binding. The assertions are normal Rust `debug_assert!`,
+and they can be turned on in a release build with Rust flags (https://doc.rust-lang.org/cargo/reference/profiles.html#debug-assertions).
+
+## Extreme debugging assertions
+
+Other than the normal debugging assertions, MMTk also has a set of
```
wks, line 10: 'Other than' is a bit ambiguous about whether 'normal debugging assertions' are included or not.

```suggestion
In addition to the normal debugging assertions, MMTk also has a set of
```

'Besides' should also be OK.

docs/userguide/src/portingguide/debugging/prefix.md
wks: But I would be more liberal about what we include here.  For example, the eBPF tracing tools are not specific to MMTk, (Perfetteo is also not specific to MMTk, either), but that combination is very effective in debugging performance bottlenecks in the work packet system.  I think we can include anything we find useful for debugging MMTk.  `rr`, for example, is a very useful tool, because conventional logging and breakpoints are not very effective when debugging GC bugs.  And not everyone know `rr`.  It doesn't seem to be as well-known as GDB.  I would prefer including `rr` somewhere in our documentation.

qinsoon: eBPF and perfetto matches what I said. We provides tracepoints that the users should know about, and we provides scripts to produce traces that need perfetto to view. It is a good example of things that we should include in our guide.

About `rr`, we do not want to provide a tutorial on how to use `rr`. But we could provide discussion and examples about how to use `rr` for certain scenarios in MMTk, like how to trace each step in a transitive closure, etc.

I will rephrase the paragraph to make it clear what the 'debugging tips' section is about. But I don't plan to add other discussions in this PR.

qinsoon: I have made some changes to this paragraph.

qinsoon:
Is there any other change required for the PR? @wks

wks (approved):
LGTM

Issue #205, opened 2021-01-04
Automating merging PRs from bindings for mmtk-core updates


qinsoon:
For breaking changes in mmtk-core, we always make sure that the bindings we maintain work with the new changes. To achieve this, we will need to have pull requests in each binding, and they should be merged as soon as the PR in mmtk-core is merged. Currently we manually do this merging process. It would be good if we can automate this. 

The automation should do the following:
1. We should specify which PRs/branches in the binding needs to be merged (possibly we can use the same specification as for the binding tests - there is no need to specify it twice).
2. We merge the mmtk-core PR.
3. We update each binding PR to refer to the latest mmtk-core commit (the merge commit in master)
4. We update `Cargo.lock` as the mmtk dependency is changed in the last step.
5. Commit and merge each binding PR.

PR #988, opened 2023-10-18
Allow auto merge binding PRs

qinsoon:
This PR allows auto merging related binding PRs once the mmtk-core PR is merged. It also adds some docs for MMTk team about merging process. The PR closes https://github.com/mmtk/mmtk-core/issues/205.

### DIFF
--- a/.github/scripts/replace-mmtk-dep.py
+++ b/.github/scripts/replace-mmtk-dep.py
@@ -2,6 +2,7 @@
 
 import argparse
 import os.path
+import sys
 import tomlkit
 
 parser = argparse.ArgumentParser(
@@ -9,7 +10,12 @@ parser = argparse.ArgumentParser(
         )
 
 parser.add_argument('toml_path', help='Path to Cargo.toml')
-parser.add_argument('mmtk_core_path', help='Path to the mmtk_core repo')
+# The following arguments are exclusive. Use either. If both are supplied, we use the local path.
+# 1. Point to a local path
+parser.add_argument('--mmtk_core_path', help='Path to the mmtk_core repo.')
+# 2. Point to a remote repo
+parser.add_argument('--mmtk_core_git', help='URL to the mmtk_core repo.')
+parser.add_argument('--mmtk_core_rev', help='Revision to use')
 
 args = parser.parse_args()
 
@@ -17,20 +23,33 @@ print("Reading TOML from '{}'".format(args.toml_path))
 with open(args.toml_path, "rt") as f:
     toml_data = tomlkit.load(f)
 
+if "mmtk" not in toml_data["dependencies"]:
+    print("Cannot find the mmtk dependency in {}".format(args.toml_path))
+    sys.exit(1)
+
 mmtk_node = toml_data["dependencies"]["mmtk"]
 
-# These keys may specify the locations of the dependency. Remove them.
-for key in ["git", "branch", "version", "registry"]:
+# Remove anything in the mmtk node.
+for key in ["git", "branch", "version", "registry", "rev"]:
     if key in mmtk_node:
         print("Deleting dependencies.mmtk.{}".format(key))
         del mmtk_node[key]
-    else:
-        print("Key dependencies.mmtk.{} does not exist.  Ignored.".format(key))
 
-# Use mmtk-core from the specified local directory.
-mmtk_repo_path = os.path.realpath(args.mmtk_core_path)
-print("Setting dependencies.mmtk.path to {}".format(mmtk_repo_path))
-mmtk_node["path"] = mmtk_repo_path
+# Construct the new mmtk node
+if args.mmtk_core_path is not None:
+    # Use mmtk-core from the specified local directory.
+    mmtk_repo_path = os.path.realpath(args.mmtk_core_path)
+    print("Setting dependencies.mmtk.path to {}".format(mmtk_repo_path))
+    mmtk_node["path"] = mmtk_repo_path
+elif args.mmtk_core_git is not None and args.mmtk_core_rev is not None:
+    mmtk_node["git"] = args.mmtk_core_git
+    mmtk_node["rev"] = args.mmtk_core_rev
+else:
+    print("No path or git/rev is supplied. We cannot update the toml")
+    sys.exit(1)
+
+# Store the mmtk node
+toml_data["dependencies"]["mmtk"] = mmtk_node
 
 print("Writing TOML to '{}'".format(args.toml_path))
 with open(args.toml_path, "wt") as f:
--- a/.github/scripts/replace-mmtk-dep.py
+++ b/.github/scripts/replace-mmtk-dep.py
@@ -27,13 +27,8 @@ if "mmtk" not in toml_data["dependencies"]:
     print("Cannot find the mmtk dependency in {}".format(args.toml_path))
     sys.exit(1)
 
-mmtk_node = toml_data["dependencies"]["mmtk"]
-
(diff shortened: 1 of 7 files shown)


qinsoon:
binding-prs
OPENJDK_BINDING_REF=test-auto-merging

qinsoon:
Most of the workflow is tested, and works fine. See https://github.com/mmtk/mmtk-core/actions/runs/6556991187. The workflows can update the mmtk dependency and turn on auto merging in the testing PR in OpenJDK (https://github.com/mmtk/mmtk-openjdk/pull/254).

One thing that is not tested yet is the merge commit hash. It is acquired with `github.event.pull_request.merge_commit_sha`. According to the documents (https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request), the `merge_commit_sha` should be "the SHA of the squashed commit on the base branch". We will only know if it works fine after we actually merge a pull request.

qinsoon:
I think the current approach may not work if a mmtk-core PR is opened from a fork. The workflow needs to access `secrets.CI_ACCESS_TOKEN` which is probably not available if the PR is opened from a fork. I can try change the trigger from "closed pull_request" to "push to master". In that case, the workflow is always triggered from the upstream repo.

.github/scripts/replace-mmtk-dep.py
```diff
@@ -2,35 +2,74 @@
...
+    # Use mmtk-core from the specified local directory.
+    mmtk_repo_path = os.path.realpath(args.mmtk_core_path)
+    print("Setting dependencies.mmtk.path to {}".format(mmtk_repo_path))
+    mmtk_node["path"] = mmtk_repo_path
+elif how == "point_to_repo":
+    # Remove any existing key about local path
+    remove_keys(mmtk_node, ["path"])
+    # Update git/rev
+    mmtk_node["git"] = args.mmtk_core_git
+    mmtk_node["rev"] = args.mmtk_core_rev
```
wks, line 69: I think we should remove all of the keys we know, regardless weather it is "point_to_local" or "point_to_repo", just in case any of them is defined and we are not aware of it.  For example, in the case of "point_to_repo", it does not make sense if more than one of "branch", "tag" and "rev" are specified.

```suggestion
remove_keys(mmtk_node, ["git", "rev", "tag", "branch", "path", "registry", "version"])
if how == "point_to_local":
    # Use mmtk-core from the specified local directory.
    mmtk_repo_path = os.path.realpath(args.mmtk_core_path)
    print("Setting dependencies.mmtk.path to {}".format(mmtk_repo_path))
    mmtk_node["path"] = mmtk_repo_path
elif how == "point_to_repo":
    # Update git/rev
    mmtk_node["git"] = args.mmtk_core_git
    mmtk_node["rev"] = args.mmtk_core_rev
```

.github/scripts/replace-mmtk-dep.py
```diff
@@ -2,35 +2,74 @@
...
+
+# Store the mmtk node
+toml_data["dependencies"]["mmtk"] = mmtk_node
```
wks, line 72: In Python, variables have reference semantics.  From the assignment `mmtk_node = toml_data["dependencies"]["mmtk"]` on line 46 up to this line, the variable `mmtk_node` has been referring to exactly the same object as `toml_data["dependencies"]["mmtk"]`.  So this line is strictly unnecessary.

wks (changes requested):
There are some problems in the Python script `replace-mmtk-dep.py` that should be changed.

.github/scripts/replace-mmtk-dep.py
wks: I just realised that this script is used for the purpose of making permanent changes to the binding's `Cargo.toml` instead of temporarily changing the dependency to test against a specific version of mmtk-core.

In this case, formatting matters.  There may be no general way to change `Cargo.toml` "back", because the style of the original `Cargo.toml` before setting `git` and `rev` to a specific repo and a specific rev may still vary from binding to binding.

mmtk-openjdk has this:

```toml
# But other changes including adding/removing whitespaces in commented lines may break the CI.
mmtk = { git = "https://github.com/mmtk/mmtk-core.git", rev = "9a676e68dd41272b57e6812dc404ffa6391668e2" }
# Uncomment the following to build locally
# mmtk = { path = "../repos/mmtk-core" }
```

but mmtk-ruby has this:

```toml
[dependencies.mmtk]
features = ["is_mmtk_object", "object_pinning"]

# Uncomment one of the following lines to choose where to find mmtk-core.
git = "https://github.com/mmtk/mmtk-core.git"  # Use mmtk-core from the official repository.
#path = "../../mmtk-core"  # Use mmtk-core from a local repository.
```

I think it is OK to assume our officially supported bindings have `git` and `rev` set so that we only need to update those.  `mmtk-ruby` currently doesn't have the `rev` field, but when running `cargo build`, it still fetches the revision specified in `Cargo.lock`.  I would argue that having the revision in `Cargo.lock` is sufficient for tracking the version we are testing against, but it is harmless to let `mmtk-ruby` have the `rev` field in the `mmtk-core` dependency, too.

.github/scripts/replace-mmtk-dep.py
```diff
@@ -2,35 +2,74 @@
...
         )
 
 parser.add_argument('toml_path', help='Path to Cargo.toml')
-parser.add_argument('mmtk_core_path', help='Path to the mmtk_core repo')
+# The following arguments are exclusive. Use either. If both are supplied, we use the local path.
+# 1. Point to a local path
+parser.add_argument('--mmtk_core_path', help='Path to the mmtk_core repo.')
+# 2. Point to a remote repo
+parser.add_argument('--mmtk_core_git', help='URL to the mmtk_core repo.')
+parser.add_argument('--mmtk_core_rev', help='Revision to use')
```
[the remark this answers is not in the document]
qinsoon: Done.

.github/scripts/replace-mmtk-dep.py
qinsoon: If I understand you right:
* If we are setting a path, we remove all the known keys about versioning, and add `path`. This will mess up the formatting. But the change is not persistent, so it doesn't matter.
* If we are updating `git` and `rev`, the change will be persistent and we want to preserve formatting. So we do not remove keys, and just update those two fields. This preserves the formatting.

I made changes as above. Let me know if I misunderstood any part.

.github/scripts/replace-mmtk-dep.py
qinsoon: I removed the extra store. Thanks.

qinsoon:
> I think the current approach may not work if a mmtk-core PR is opened from a fork. The workflow needs to access `secrets.CI_ACCESS_TOKEN` which is probably not available if the PR is opened from a fork. I can try change the trigger from "closed pull_request" to "push to master". In that case, the workflow is always triggered from the upstream repo.

I changed the workflow trigger. It should be fine now.

.github/scripts/replace-mmtk-dep.py
wks: Yes.  That's what I meant.

wks (approved):
LGTM

qinsoon:
I changed the way how to update the lock file. It now uses `cargo build` instead of `cargo generate-lockfile`, as `genearte-lockfile` may update the version of dependencies. If a dependency does not comply with semantic versioning, we may end up with a broken build. Use `cargo build` won't have such an issue. 

I don't plan to make any further changes to the PR unless requested.

PR #990, opened 2023-10-19
Allow overriding ScanObjects packet size

wks:
This works around a load balancing problem where the default packet size 4096 is too large for some work loads, resulting in a low level of parallelism.

In the long run, we should support work stealing to properly fix the problem.


qinsoon:
I don't know how useful it would be to allow bindings to have a *globally defined* capacity for scan objects work. The right capacity should highly depend on the workload and what type of objects that are in the packets. It should not depend on what language it is.

PR #991, opened 2023-10-20
Synchronize mmtk-core issue labels to other repos

qinsoon:


wks:
Do we want to synchronise all labels we create in the future? I have a feeling that some labels may not make sense for bindings.  If we can synchronise labels using the command line utility, can we simply run the same command locally to do the synchronisation?  I think that's easy enough.

qinsoon:
I just think it might be a good idea to sync the labels to all the repos when I was creating those priority labels. If it is not desirable to sync the labels, we can close this PR.

PR #993, opened 2023-10-22
Use BumpPointer::default()

qinsoon:
This PR changes the code snippet in our doc to use `BumpPointer::default()` to create a new zero-initialized bump pointer struct. It also changes our internal implementation to use `BumpPointer::default()` instead of `BumpPointer::new(zero, zero)`.

### DIFF
--- a/vmbindings/dummyvm/src/tests/doc_mutator_storage.rs
+++ b/vmbindings/dummyvm/src/tests/doc_mutator_storage.rs
@@ -76,19 +76,9 @@ pub fn embed_fastpath_struct() {
 
         // Bind an MMTk mutator
         let mutator = mmtk::memory_manager::bind_mutator(&fixture.mmtk, tls_opaque_pointer);
-        // Create a fastpath BumpPointer
-        let default_bump_pointer = {
-            // First find the allocator
-            let selector = mmtk::memory_manager::get_allocator_mapping(
-                &fixture.mmtk,
-                AllocationSemantics::Default,
-            );
-            let default_allocator = unsafe {
-                mutator.allocator_impl::<mmtk::util::alloc::BumpAllocator<DummyVM>>(selector)
-            };
-            // Copy the bump pointer struct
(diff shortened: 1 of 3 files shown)


qinsoon:
Hopefully this captures what we discussed on Friday. Feel free to update the PR if I missed anything. @wks

vmbindings/dummyvm/src/tests/doc_mutator_storage.rs
```diff
@@ -76,19 +76,9 @@ pub fn embed_fastpath_struct() {
...
-                &fixture.mmtk,
-                AllocationSemantics::Default,
-            );
-            let default_allocator = unsafe {
-                mutator.allocator_impl::<mmtk::util::alloc::BumpAllocator<DummyVM>>(selector)
-            };
-            // Copy the bump pointer struct
-            default_allocator.bump_pointer
-        };
+        // Create a fastpath BumpPointer with default(). The BumpPointer from default() will guarantee to fail on the first allocation
```
wks, line 79: Since we let the binding create `BumpPointer` with `default()`, we should document the behaviour of the `default()` method, noting that we initialise it with 0 and 0 and it will be guaranteed to fail in fast path.

Currently there is a comment inside the `BumpPointer::default()` method.

```rust
    fn default() -> Self {
        // Defaults to 0,0. In this case, the first
        // allocation would naturally fail the check
        // `cursor + size < limit`, and go to the slowpath.
        BumpPointer {
            cursor: Address::ZERO,
            limit: Address::ZERO,
        }
    }
```

I think we can move the comment into the doc comment (`///`) of the `default()` method so that the user knows how the `default()` method is intended to be used.

It is even better to mention this behaviour in the type-level doc comment of the `BumpPointer` type, too.

qinsoon: I have made changes as suggested.

wks (approved):
LGTM

PR #994, opened 2023-10-23
Support header mark bits for Immix

k-sareen:


k-sareen:
This PR is currently broken for OpenJDK. I haven't investigated it yet. But I'm currently using PR this for ART since it has a header mark-bit.

wks:
> This PR is currently broken for OpenJDK. I haven't investigated it yet. But I'm currently using PR this for ART since it has a header mark-bit.

IIRC it has something to do with the lock word and lock inflation.

PR #996, opened 2023-10-24
Use `cargo generate-lockfile` to update JikesRVM's Cargo.lock

qinsoon:
`auto-merge.yml` uses `cargo build` to generate lockfile for most bindings. But this does not work for JikesRVM:
1. JikesRVM uses `--target i686-unknown-linux-gnu` which is not installed automatically when running `cargo build`.
2. JikesRVM's cargo project uses some Rust source files that are generated during building JikesRVM. So we cannot build the cargo project alone from a fresh repo clone. We have to do a full build.

As a workaround, we use `cargo generate-lockfile` for JikesRVM. `cargo generate-lockfile` will update the version of dependencies. But we do not have a better option for JikesRVM

### DIFF
--- a/.github/workflows/auto-merge-inner.yml
+++ b/.github/workflows/auto-merge-inner.yml
@@ -68,6 +68,11 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Setup
+        if: steps.check-input.outputs.skip == 'false'
+        working-directory: ${{ env.BINDING_WORK_DIR }}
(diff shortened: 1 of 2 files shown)


qinsoon:
binding-refs
JIKESRVM_BINDING_REF=update-mmtk-core-pr-949

qinsoon:
The previous `auto-merge.yml` failed for JikesRVM: https://github.com/mmtk/mmtk-core/actions/runs/6621479194/job/17985509918

wks (approved):
LGTM

PR #1000, opened 2023-10-25
Update doc comments

wks:
We removed the feature for letting the coordinator call both `stop_all_mutators` and `resume_mutators` since https://github.com/mmtk/mmtk-core/pull/794 and https://github.com/mmtk/mmtk-core/pull/814.  The documentation should reflect that.

### DIFF
--- a/src/vm/collection.rs
+++ b/src/vm/collection.rs
@@ -12,14 +12,13 @@ pub enum GCThreadContext<VM: VMBinding> {
 /// VM-specific methods for garbage collection.
 pub trait Collection<VM: VMBinding> {
     /// Stop all the mutator threads. MMTk calls this method when it requires all the mutator to yield for a GC.
-    /// This method is called by a single thread in MMTk (the GC controller).
     /// This method should not return until all the threads are yielded.
     /// The actual thread synchronization mechanism is up to the VM, and MMTk does not make assumptions on that.
     /// MMTk provides a callback function and expects the binding to use the callback for each mutator when it
     /// is ready for stack scanning. Usually a stack can be scanned as soon as the thread stops in the yieldpoint.
     ///
     /// Arguments:
-    /// * `tls`: The thread pointer for the GC controller/coordinator.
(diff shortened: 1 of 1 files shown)


src/vm/collection.rs
```diff
@@ -12,23 +12,25 @@ pub enum GCThreadContext<VM: VMBinding> {
...
-    /// * `tls`: The thread pointer for the GC controller/coordinator.
+    /// * `tls`: The thread pointer for the GC worker.
     /// * `mutator_visitor`: A callback.  Call it with a mutator as argument to notify MMTk that the mutator is ready to be scanned.
     fn stop_all_mutators<F>(tls: VMWorkerThread, mutator_visitor: F)
     where
         F: FnMut(&'static mut Mutator<VM>);
 
     /// Resume all the mutator threads, the opposite of the above. When a GC is finished, MMTk calls this method.
     ///
+    /// This method may be called by any GC worker thread, and may not be the same GC worker thread that called
```
k-sareen, line 29: That is not correct. At least as of now. Currently the GC controller thread calls `resume_mutators` in [`EndOfGC`](https://github.com/mmtk/mmtk-core/blob/abd38ec31064e7d52ae2b492a75b4bcc7f7cc554/src/scheduler/gc_work.rs#L262C50-L262C50)

k-sareen: See here as well: https://github.com/mmtk/mmtk-core/blob/abd38ec31064e7d52ae2b492a75b4bcc7f7cc554/src/scheduler/controller.rs#L140-L143

wks: OK. That's true, although it is not intended.  I'll adjust the comment and not mention "worker".

k-sareen: Random aside, I feel like we should rename `stop_all_mutators` to `suspend_mutators` to be in line with `resume_mutators`.

k-sareen (approved):
LGTM. Thanks!

Issue #314, opened 2021-05-14
LockFreeImmortalSpace should not assume it can use the whole heap range


qinsoon:
`LockFreeImmortalSpace` now uses the address range between `AVAILABLE_START` and `AVAILABLE_START + total_bytes`.  This means it won't work if we use other spaces along with this space. 

The plan `nogc_lock_free` currently is only used by OpenJDK, and OpenJDK does not allocate into any other space other than the space bound with the default allocator (i.e. the LockFreeImmortalSpace for nogc_lock_free). So we are not seeing an issue.

PR #1003, opened 2023-10-30
Implement get_gc_trigger() for LockFreeImmortalSpace

qinsoon:
This PR fixes two issues in `LockFreeImmortalSpace` and adds a test.
* Implemented `get_gc_trigger()`. As `LockFreeImmortalSpace` does not have `CommonSpace`, we cannot use the default implementation.
* Properly reserve space with `HeapMeta`, so the address range of `LockFreeImmortalSpace` won't clash with other spaces. This closes https://github.com/mmtk/mmtk-core/issues/314.

### DIFF
--- a/src/policy/lockfreeimmortalspace.rs
+++ b/src/policy/lockfreeimmortalspace.rs
@@ -132,11 +132,9 @@ impl<VM: VMBinding> Space<VM> for LockFreeImmortalSpace<VM> {
         if start + bytes > self.limit {
             panic!("OutOfMemory")
         }
-        trace!("Going to zero");
         if self.slow_path_zeroing {
             crate::util::memory::zero(start, bytes);
         }
-        trace!("Going to zero - done");
         start
     }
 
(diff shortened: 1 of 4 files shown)


src/policy/lockfreeimmortalspace.rs
```diff
@@ -34,11 +36,11 @@ pub struct LockFreeImmortalSpace<VM: VMBinding> {
     /// start of this space
     start: Address,
     /// Total bytes for the space
-    extent: usize,
+    total_bytes: usize,
```
wks, line 39: What was wrong with "extent"?  I think "extent" and "total bytes" are synonyms.

qinsoon: In `let VMRequest::Extent{ extent, top } = vmrequest else { unreachable!() };`, the `extent` is the extent for the virtual memory range. We use it for `args.heap.reserve(extent, top);`.

The field `extent` in the space is not virtual memory range. It is the actual size we may use. We use it for mmapping side metadata and initializing SFT. I was concerned that it may cause confusion to call it `extent`.

wks (approved):
LGTM

PR #1006, opened 2023-10-31
# of objects traced / moved

udesou:
This PR adds the option to count the number of objects traced and moved in Immix GCs. In combination with `STRESS_DEFRAG` and `DEFRAG_EVERY_BLOCK` it can give an indication of how many objects are being (transitively) pinned and therefore cannot be copied.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,6 +110,8 @@ immix_non_moving = []
 # if `immix_non_moving` is in use.
 sticky_immix_non_moving_nursery = []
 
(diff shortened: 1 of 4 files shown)


wks:
> ... In combination with `STRESS_DEFRAG` and `DEFRAG_EVERY_BLOCK` ...

You also need to set `Defrag::DEFRAG_HEADROOM_PERCENT = 50` so that there is enough room for defragmentation.  Otherwise it will copy 2% of the heap size and then stop copying.

PR #1007, opened 2023-10-31
Add document for release

qinsoon:


docs/team/pull_request.md
```diff
@@ -43,6 +43,7 @@ when an mmtk-core PR is merged and the `binding-refs` comment is present.
 
 1. Make sure there is no other PR in this merging process. If so, resolve those first.
 1. Make sure all the PRs (the mmtk-core PR, the binding PRs, and the associated PRs in the VM repo if any) are ready to merge.
+1. Make sure there is a comment that provides `binding-refs` for all the binding PRs.
```
wks, line 46: We often read these documents in the text form (for example, from a text editor).  Although Markdown can automatically convert those `1.` into the right number, it is confusing for those who read them as plain text, especially when we mention "Step 3 is optional" in later paragraphs.  I suggest we simply number those items with 1, 2, 3, ...

wks (changes requested):

docs/team/pull_request.md
qinsoon: Sounds reasonable. I have made changes.

wks (approved):
LGTM

PR #1008, opened 2023-10-31
Move constants and add comments for stress copying

wks:
Move the constant `DEFRAG_HEADROOM_PERCENT` to `crate::policy::immix` so that all constants needed to be modified to stress copying are adjacent.

Add a comment about how to change those constants and the reasons behind.


qinsoon (approved):
LGTM

PR #1014, opened 2023-11-04
Fix multiple issues in recent CI changes

qinsoon:
There are some issues with the recent CI changes. This PR bumps ci-perf-kit to 0.7.2, which fixes some of them, and separate issues are created for the issue(s) that is not fixed in the PR.

* ci-perf-kit 0.7.1 allows drawing notes on the graph to show the date where there are known CI environment changes. But it has some issues:
   * Each note is by mistake plotted only once for all the graphs (which means only one graph will have the note). In the Python script, we should copy the note list before processing it.
   * In the case that there is only one run for several notes. (e.g. we have notes on 1101 and 1102, and we have runs on 1103 and 1104. Both notes for 1101 and 1102 should show on the data point for 1103. The current CI plots one note on 1103 and one note on 1104. We should combine the notes on 1101 and 1102, and plot them for the run 1103.
* 0.7.1 sets the heap size to 8g (for OpenJDK NoGC runs, and mutator runs), but by mistake uses a modifier `heap30g`. This causes confusion. It should be name as `heap8g`.
* 0.7.1 by mistake sets the plan to `gencopy` in a GenImmix config.

ci-perf-kit 0.7.2 (https://github.com/mmtk/ci-perf-kit/releases/tag/0.7.2) fixes those issues.

### DIFF
--- a/.github/workflows/micro-bm.yml
+++ b/.github/workflows/micro-bm.yml
@@ -66,7 +66,7 @@ jobs:
         with:
           repository: mmtk/ci-perf-kit
           token: ${{ secrets.GITHUB_TOKEN }}
-          ref: "0.7.3"
+          ref: "0.7.2"
           path: ci-perf-kit
           submodules: true
       # Use rust-toolchain in the trunk (it doesnt matter much - if the toolchains defined in the trunk and the branch are different, we cant run anyway)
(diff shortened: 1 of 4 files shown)


wks (approved):
LGTM

PR #1016, opened 2023-11-06
WIP: Add steps to implement GC to porting guide

k-sareen:


k-sareen:
@k-sareen: Reminder to talk about flushing mutator buffers before GC (including destroyed mutators)

qinsoon:
> @k-sareen: Reminder to talk about flushing mutator buffers before GC (including destroyed mutators)

https://github.com/mmtk/mmtk-core/issues/1047 suggested that the binding should not need to flush mutator explicitly, i.e. the binding should call `destroy_mutator` but not `flush_mutator`. The contract is that the binding should let MMTk know when a mutator thread dies, and MMTk deals with whatever needs to be done with the mutator (e.g. flushing thread local states).

k-sareen:
@qinsoon This is the PR I was talking about. It's a bit outdated now I think it we don't have the coordinator thread anymore, for example.

PR #1018, opened 2023-11-09
Mention revoking TLABs of all mutators after a GC

k-sareen:


docs/userguide/src/portingguide/perf_tuning/alloc.md
```diff
@@ -47,37 +47,55 @@ have an assertion to ensure that the native type has the exact same layout as th
...
+```
 
+And pseudo-code for how you would reset the `BumpPointer`s for all mutators in `resume_mutators`. Note that these mutators are the runtime's actual mutator threads (i.e. where the cached bump pointers are stored) and are different from MMTk's `Mutator` struct.
 ```rust
-{{#include ../../../../../vmbindings/dummyvm/src/tests/doc_mutator_storage.rs:mutator_storage_embed_fastpath_struct}}
+impl Collection<RtName> for RtNameCollection {
+  ...
+  fn resume_mutators(tls: VMWorkerThread) {
+    // Reset the cached bump pointers of each mutator to 0x0 after a GC to
+    // ensure that the VM sees a cohesive state
```
wks, line 90: A BumpPointer has a cursor and a limit.  It is not clear what 'reset bump pointer to 0' means.

```suggestion
    // Reset the cached bump pointers of each mutator (setting both cursor and limit to 0) after a GC to
    // ensure that the VM sees a cohesive state
```

wks (approved):
LGTM

PR #1019, opened 2023-11-10
Fix CI for OpenJDK nogc config

qinsoon:
Current perf CI failed to run OpenJDK NoGC, as there is a typo in the config file in the run. It is fixed in `ci-perf-kit` 0.7.3 (https://github.com/mmtk/ci-perf-kit/releases/tag/0.7.3). This PR updates `ci-perf-kit` to 0.7.3.

The failed run was here: https://github.com/mmtk/mmtk-core/actions/runs/6765511244/job/18385256502. The NoGC failed to run, and has no result. This caused the plotting to fail.


wks (approved):
LGTM

Issue #309, opened 2021-05-11
Missing documents/comments for public items


qinsoon:
For documentation, a first step we can do is to add missing documents for our public types and functions. We can enforce documents with `rustdoc`'s `missing_docs` lints, which requires all public items to be documented. Once https://github.com/mmtk/mmtk-core/issues/241 is done, we should then only expose a limited number of public items.

qinsoon:
We currently have 368 public items that do not have comments. I plan to add comments for them, and also check if those items really should be public. I probably will submit one PR for every 50 changes or so. The PRs should be removing `pub` for some items, and adding missing comments.

PR #1024, opened 2023-11-14
Add missing docs for some util modules

qinsoon:
This PR is a step towards https://github.com/mmtk/mmtk-core/issues/309.
* Add missing docs for some `util` modules.
* Make `util::metadata::side_metadata::helpers/helpers_32` not public.
* Make `util::reference_processor` not public.
* Allow adding docs for options defined by the `options!` macro.

### DIFF
--- a/.github/scripts/ci-test.sh
+++ b/.github/scripts/ci-test.sh
@@ -13,7 +13,7 @@ fi
 
 ./examples/build.py
 
-ALL_PLANS=$(sed -n '/enum PlanSelector/,/}/p' src/util/options.rs | xargs | grep -o '{.*}' | grep -o '\w\+')
+ALL_PLANS=$(sed -n '/enum PlanSelector/,/}/p' src/util/options.rs | sed -e 's/\/\/.*//g' -e '/^$/d' -e 's/,//g' | xargs | grep -o '{.*}' | grep -o '\w\+')
 
 # Test with DummyVM (each test in a separate run)
 cd vmbindings/dummyvm
--- a/.github/scripts/ci-test.sh
+++ b/.github/scripts/ci-test.sh
@@ -13,7 +13,7 @@ fi
 
 ./examples/build.py
 
-ALL_PLANS=$(sed -n '/enum PlanSelector/,/}/p' src/util/options.rs | sed -e 's/\/\/.*//g' -e '/^$/d' -e 's/,//g' | xargs | grep -o '{.*}' | grep -o '\w\+')
+ALL_PLANS=$(sed -n '/enum PlanSelector/,/}/p' src/util/options.rs | sed -e 's;//.*;;g' -e '/^$/d' -e 's/,//g' | xargs | grep -o '{.*}' | grep -o '\w\+')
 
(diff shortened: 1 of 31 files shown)


.github/scripts/ci-test.sh
```diff
@@ -13,7 +13,7 @@ fi
 
 ./examples/build.py
 
-ALL_PLANS=$(sed -n '/enum PlanSelector/,/}/p' src/util/options.rs | xargs | grep -o '{.*}' | grep -o '\w\+')
+ALL_PLANS=$(sed -n '/enum PlanSelector/,/}/p' src/util/options.rs | sed -e 's/\/\/.*//g' -e '/^$/d' -e 's/,//g' | xargs | grep -o '{.*}' | grep -o '\w\+')
```
wks, line 16: The `s` command of sed and vim also accept other separators in addition to `/`.  For example, it will be more readable if we write `s;//.*;;g` instead.

src/util/options.rs
```diff
@@ -672,88 +700,85 @@ mod gc_trigger_tests {
 // Currently we allow all the options to be set by env var for the sake of convenience.
 // At some point, we may disallow this and all the options can only be set by command line.
 options! {
-    // The plan to use.
+    #[doc = "The plan to use."]
```
wks, line 703: Like the other PR, we should be able to use `///` here, too.

wks (changes requested):
Mostly OK.  There are some aesthetic problems that can be fixed.

src/util/options.rs
qinsoon: `///` may not work here. We can use `#[doc]` as we capture the pattern `$(#[$outer:meta])*` in the macro. We should be able to use the same pattern to capture things like `#[cfg(feature = "a")]`. This seems to be a common way to add attributes for items that are generated by a macro.

qinsoon: Just to clarify, I was also using `#[doc]` to pass attributes (including docs) to macros in the previous PR.

.github/scripts/ci-test.sh
qinsoon: I have made changes as suggested.

src/util/options.rs
wks: Actually it works.  I changed it to 

```rust
    /// The plan to use. Blah blah. (to see if `///` works)
    plan:                  PlanSelector         [env_var: true, command_line: true] [always_valid] = PlanSelector::GenImmix,
```

Type `cargo doc` and the generated HTML contains the part in `///`.

FYI: https://stackoverflow.com/questions/33999341/generating-documentation-in-macros

qinsoon: Yeah. Interesting. It works. Rust did not allow `///` in `macro_rules` and that was why we have `//` in the `options!` macro. Probably Rust started to support this at some point. This is great. I will make these changes.

qinsoon: Oh I see. `#[$outer:meta]*` matches `///`.

wks (approved):
LGTM

PR #1026, opened 2023-11-16
Add missing docs for the rest of the util module

qinsoon:
This PR is a step towards https://github.com/mmtk/mmtk-core/issues/309.
* Deny `missing_docs` for the `util` module and the `vm` module.
* Change some items from `pub` to `pub(crate)`.
* Remove some unused constants.

### DIFF
--- a/src/util/alloc/bumpallocator.rs
+++ b/src/util/alloc/bumpallocator.rs
@@ -34,7 +34,7 @@ pub struct BumpAllocator<VM: VMBinding> {
 #[repr(C)]
 #[derive(Copy, Clone)]
 pub struct BumpPointer {
-    /// The current address of the allocation buffer. This is what will be allocated next.
+    /// The cursor inside the allocation buffer where the next object will be allocated.
     pub cursor: Address,
     /// The upperbound of the allocation buffer.
     pub limit: Address,
--- a/src/util/alloc/bumpallocator.rs
+++ b/src/util/alloc/bumpallocator.rs
@@ -5,7 +5,7 @@ use crate::util::Address;
 use crate::util::alloc::Allocator;
 
 use crate::policy::space::Space;
-use crate::util::conversions::bytes_to_pages;
+use crate::util::conversions::bytes_to_pages_up;
 use crate::util::opaque_pointer::*;
 use crate::vm::VMBinding;
 
@@ -199,7 +199,7 @@ impl<VM: VMBinding> BumpAllocator<VM> {
         }
 
         let block_size = (size + BLOCK_MASK) & (!BLOCK_MASK);
-        let acquired_start = self.space.acquire(self.tls, bytes_to_pages(block_size));
+        let acquired_start = self.space.acquire(self.tls, bytes_to_pages_up(block_size));
         if acquired_start.is_zero() {
             trace!("Failed to acquire a new block");
             acquired_start
--- a/src/util/alloc/bumpallocator.rs
+++ b/src/util/alloc/bumpallocator.rs
@@ -13,6 +13,8 @@ const BYTES_IN_PAGE: usize = 1 << 12;
 const BLOCK_SIZE: usize = 8 * BYTES_IN_PAGE;
 const BLOCK_MASK: usize = BLOCK_SIZE - 1;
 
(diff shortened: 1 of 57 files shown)


qinsoon:
binding-refs
OPENJDK_BINDING_REF=update-mmtk-core-pr-1026

src/util/alloc/bumpallocator.rs
```diff
@@ -32,11 +34,14 @@ pub struct BumpAllocator<VM: VMBinding> {
 #[repr(C)]
 #[derive(Copy, Clone)]
 pub struct BumpPointer {
+    /// The current address of the allocation buffer. This is what will be allocated next.
```
wks, line 37: I would not say the `cursor` *is* the current address of the allocation buffer itself.  Intuitively, the buffer doesn't move, but the cursor moves.  The buffer may be a whole block or a consecutive range of lines in an Immix block, but it is chosen in the slow path as the range from the *initial* `cursor` to the `limit`.  We don't keep track of the starting point of the buffer, but conceptually the cursor is not the "address of the allocation buffer".

```suggestion
    /// The cursor inside the allocation buffer where the next object will be allocated.
```

src/util/conversions.rs
```diff
@@ -70,6 +78,7 @@ pub fn bytes_to_pages_up(bytes: usize) -> usize {
     raw_align_up(bytes, BYTES_IN_PAGE) >> LOG_BYTES_IN_PAGE
 }
 
+/// Convert size in bytes to the number of pages (aligned down)
```
wks, line 81: The doc comment says the result is aligned *down*, but it calls `bytes_to_pages_up`.  Is this a mistake?

src/util/conversions.rs
```diff
@@ -2,65 +2,73 @@ use crate::util::constants::*;
...
     unsafe { Address::from_usize(chunk << LOG_BYTES_IN_CHUNK) }
 }
 
+/// Align up an integer to the given alignment.
 pub const fn raw_align_up(val: usize, align: usize) -> usize {
     // See https://github.com/rust-lang/rust/blob/e620d0f337d0643c757bab791fc7d88d63217704/src/libcore/alloc.rs#L192
     val.wrapping_add(align).wrapping_sub(1) & !align.wrapping_sub(1)
 }
 
+/// Align down an integer to the given alignment.
```
wks, line 61: Need to mention that `align` must be a power of two, otherwise it won't work.

src/util/conversions.rs
```diff
@@ -2,65 +2,73 @@ use crate::util::constants::*;
...
 pub const fn raw_align_down(val: usize, align: usize) -> usize {
     val & !align.wrapping_sub(1)
 }
 
+/// Is the integer aligned to the given alignment?
```
wks, line 66: Ditto.

src/util/mod.rs
```diff
@@ -39,6 +39,7 @@ pub(crate) mod erase_vm;
 pub(crate) mod finalizable_processor;
 /// Heap implementation, including page resource, mmapper, etc.
 pub mod heap;
+/// Checking if an address is an valid MMTk object.
```
wks, line 42: I am considering moving those doc comments into their respective modules in the form of `//!` at the beginning of the file so that we can have more detailed documentation for each module.  But we may do it in a separate PR.

wks (changes requested):

src/util/conversions.rs
qinsoon: Yes, it was a mistake when I added the comment. I thought it was 'align down', as we have the other function for aligning up. 

I think it is a bit confusing to have two functions for the same thing but with different names. I removed this function, and use `bytes_to_pages_up` instead.

src/util/mod.rs
qinsoon: I added the comments in a format that is consistent with the other comments in the same file. If we want to move the docs to the modules, we should systematically do so for the code base. I don't plan to do that in the PR.

qinsoon:
@wks  I have made most of the changes as suggested, and replied to the other comments.

wks (approved):
LGTM

PR #1028, opened 2023-11-20
Add missing docs for the rest of the code base (merge after #1026)

qinsoon:
This is the last PR for adding missing docs, and it closes https://github.com/mmtk/mmtk-core/issues/309. All the public items should be documented properly now, and `#![deny(missing_docs)]` is used so any new public item that does not have rustdoc will cause the build to fail.


qinsoon:
binding-refs
JIKESRVM_BINDING_REF=update-mmtk-core-pr-1028

qinsoon:
`#![deny(missing_docs)]` could be a bit annoying for prototyping and developing, as the build will fail. I will remove it, and only check and deny `missing_docs` in the CI script (`ci-doc.sh`).

src/scheduler/worker.rs
```diff
@@ -327,22 +327,36 @@ impl<VM: VMBinding> GCWorker<VM> {
...
     pub fn do_work(&'static mut self, mut work: impl GCWork<VM>) {
         work.do_work(self, self.mmtk);
     }
 
+    /// Explicitly execute a boxed work packet in the current context.
+    /// See [`GCWorker::do_work`] for more information.
+    pub fn do_boxed_work(&'static mut self, mut work: Box<dyn GCWork<VM>>) {
+        work.do_work(self, self.mmtk);
+    }
+
```
wks, line 359: Similar to `do_work`, but `do_boxed_work` is not used in mmtk-core at all.  We may remove this function, too.

wks (changes requested):

src/plan/plan_constraints.rs
```diff
@@ -10,10 +10,14 @@ use crate::util::constants::*;
 pub struct PlanConstraints {
     /// Does the plan collect garbage? Obviously most plans do, but NoGC does not collect.
     pub collects_garbage: bool,
+    /// True if the plan moves objects.
     pub moves_objects: bool,
+    /// The number of header bits that are required. This is deprecated. See [`crate::vm::ObjectModel`]
+    /// for the GC metadata that MMTk uses.
```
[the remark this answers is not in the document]
qinsoon: Sounds good. Actually I removed the two fields. I think only JikesRVM is using them. I will deal with them in the same way as `num_specialized_scans` in the binding.

src/scheduler/worker.rs
```diff
@@ -327,22 +327,36 @@ impl<VM: VMBinding> GCWorker<VM> {
...
+    /// Get a mutable reference of the copy context for this worker.
     pub fn get_copy_context_mut(&mut self) -> &mut GCWorkerCopyContext<VM> {
         &mut self.copy
     }
 
+    /// Explicitly execute a work packet in the current context (the given worker, and the current work packet).
+    /// The statistics collected for executing the work are counted into the current context.
+    /// This is NOT the normal way to execute a work packet: most work packets are polled and executed in
+    /// the worker's main loop ([`GCWorker::run`]). This method is only intended for cases where you would like to
+    /// explicitly execute a work packet under the current context.
```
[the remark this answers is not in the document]
qinsoon: I removed both methods.

src/scheduler/work.rs
```diff
@@ -4,9 +4,21 @@ use crate::vm::VMBinding;
...
+/// enough to ensure that the costs of managing the work packets do not dominate, and the packet must be
+/// small enough that good load balancing is achieved.
 pub trait GCWork<VM: VMBinding>: 'static + Send {
     /// Define the work for this packet. However, this is not supposed to be called directly.
     /// Usually `do_work_with_stat()` should be used.
+    /// Only call this method if you would like to explicitly execute a work packet in the current context
+    /// (the given worker, and the current work packet). The statistics collected for executing the work are
+    /// counted into the current context. This is NOT the normal way to execute a work packet: most work packets
+    /// are polled and executed in the worker's main loop ([`GCWorker::run`]) using `do_work_with_stat`.
+    /// This method should only be called directly if you would like to explicitly execute a work packet under the current context.
```
wks, line 21: We never used the term "context" to refer to the worker and the work packet in other places, so it should be easier to understand if we don't use that word, or be more specific.

```suggestion
    ///
    /// Most work packets are polled and executed in the worker's main loop ([`GCWorker::run`])
    /// using `do_work_with_stat`.  If `do_work` is called directly during the execution of another
    /// work packet, bypassing `do_work_with_stat()`, this work packet will not be counted into the
    /// number of work packets executed, and the execution time of this work packet will be counted
    /// as part of the execution time of the other work packet.  Only call this method directly if
    /// this is what you intended.  But you should always consider putting the other work packet
    /// into a bucket so that other GC workers can execute it in parallel, unless the context-
    /// switching overhead is a problem.
```

qinsoon: I updated the comments, and used most of your suggestions.

wks (approved):
LGTM

PR #1036, opened 2023-12-04
Update some stale comments

k-sareen:


k-sareen:
Weird. I'm not sure. There's no `pub use barriers::Barrier` anywhere.

Issue #1033, opened 2023-11-30
`Edge::update`, one method to update a slot (`Edge`)


wks:
Note: Ruby currently doesn't require this change in order to work with slot (Edge) enqueuing.  But V8 will need this change to efficiently forward references while preserving tags.

# Proposal

I propose a new method for the `Edge` trait: `Edge::update`.

```rust
pub trait Edge: Copy + Send + Debug + PartialEq + Eq + Hash {
    fn update<F>(&self, updater: F)
    where F: FnOnce(ObjectReference) -> Option<ObjectReference>;
}
```

The semantics of `Edge::update` is:
-   If the slot does not contain any object reference, return immediately.  Note that if the slot contains NULL, it is not considered containing any object reference.  The same is true for tagged non-reference values, such as small integers.
-   Otherwise, load the object reference from the slot, and call `updater`.
    -   If the updater returns `Some(new_object)`, update the slot so that it holds a reference to `new_object`.
    -   If the updater returns `None`, do nothing.

The `updater` is usually implemented by `ProcessEdgesWork::process_edge` to call `trace_object` and forward the slot.

```rust
    fn process_edge(&mut self, slot: EdgeOf<Self>) {
        slot.update(|object| {
            debug_assert!(!object.is_null()); // If the updater is called, it is guaranteed not to be null.
            let new_object = self.trace_object(object);
            if Self::OVERWRITE_REFERENCE {
                Some(new_object) // Let the VM binding update the slot
            } else {
                None // Do not update the slot
            }
        });
    }
```

# Rationale

## Supporting tagged union of references and values

https://github.com/mmtk/mmtk-core/issues/626 described the need to support slots that hold non-ref values in addition to null (such as tagged values including small integers).  By letting `Edge::update` decide whether to call the `updater`, the VM binding will have a chance to decode the tagged pointer, and choose not to call the `updater` (which calls `trace_object`) if the slot holds a small integer or special non-reference values such as `true`, `false`, `nil`, etc.

```rust
impl Edge for RubyEdge {
    fn update<F>(&self, updater: F)
    where F: FnOnce(ObjectReference) -> Option<ObjectReference> {
        let value = self.addr.load::<VALUE>();
        if !value.is_special_const() { // If it is not special values like small integers, true, false, nil, etc.
            if let Some(new_object) = updater(value.to_ref()) { // Call updater
                self.addr.store(new_object); // Update the slot
            }
        }
    }
}
```

## Supporting object references with tags.

[...]

PR #1037, opened 2023-12-05
Implement Edge::update_for_forwarding

wks:
This PR implements the single-method updating operation described in https://github.com/mmtk/mmtk-core/issues/1033

### DIFF
--- a/src/vm/edge_shape.rs
+++ b/src/vm/edge_shape.rs
@@ -41,9 +41,35 @@ use crate::util::{Address, ObjectReference};
 /// in a word as a pointer, it can also use `SimpleEdge` for weak reference fields.
 pub trait Edge: Copy + Send + Debug + PartialEq + Eq + Hash {
     /// Load object reference from the edge.
+    ///
+    /// If the slot is not holding an object reference, it should return `ObjectReference::NULL`.
+    /// Specifically,
+    ///
+    /// -   If the langauge has the concept of "null pointer" which does not point to any object in
+    ///     the heap, this method should return `ObjectReference::NULL` regardless how a null
+    ///     pointer is encoded in the VM.  However, if the VM uses a special object in the heap to
+    ///     represent a null value, such as the `None` object of `NoneType` in Python, this method
+    ///     should still return the object reference to such `None` objects so that they are
+    ///     properly traced, kept alive, and they have their references forwarded.
+    /// -   If, in a VM, the data type a slot can hold is a union of references and non-reference
+    ///     values, and the slot is currently holding a non-reference value, such as a small
+    ///     integer, a floating-point number, or any special value such as `true`, `false` or `nil`
+    ///     that do not point to any object, the slot is considered not holding an reference.  This
+    ///     method should return `ObjectReference::NULL` in such cases.
+    ///
+    /// If the slot holds an object reference with tag bits, the returned value shall be the object
+    /// reference with the tag bits removed.
     fn load(&self) -> ObjectReference;
 
     /// Store the object reference `object` into the edge.
+    ///
+    /// If the slot holds an object reference with tag bits, this method must preserve the tag
+    /// bits while updating the object reference so that it points to the forwarded object given by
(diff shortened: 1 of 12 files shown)


src/vm/edge_shape.rs
```diff
@@ -41,11 +41,72 @@ use crate::util::{Address, ObjectReference};
...
+    fn update_for_forwarding<F>(&self, updater: F)
+    where
+        F: FnOnce(ObjectReference) -> ObjectReference,
+    {
+        let object = self.load();
+        if object.is_null() {
+            return;
+        }
+        let new_object = updater(object);
+        if new_object.is_null() {
```
k-sareen, line 104: Can you do `new_object.is_null() && object == new_object`?

wks: I am expecting the `updater` (usually implemented by `trace_object`) to do that test inside (if necessary at all) so that we don't need to check `object == new_object` here.  One example is the nursery (for GenCopy and GenImmix).  It always moves the object, so it doesn't need to check.  It actually does `debug_assert!(!self.plan.is_object_in_nursery(new_object));` which implies `new_object` cannot be equal to `object`.

k-sareen: Hm? I don't follow. `trace_object` can't store the new object (if any) into the slot. It will only return an object reference

wks: That's right.  `trace_object` doesn't do the storing.  However, the contract of the `updater` closure is, if it returns `ObjectReference::NULL`, then `Edge::update_for_forwarding` will not do the storing.

`process_edge` implements the `updater` closure.  After executing `let new_object = trace_object(object)`, it should check if `new_object == object` and, if they are equal, return `ObjectReference::NULL` from the `updater` closure.  Currently `ProcessEdgesWork::process_edge` is not doing this because I don't want to change the semantics of the existing `ProcessEdgesWork::process_edge`.  We may give it a try by adding `if new_object == object { return ObjectReference::NULL; } else { return new_object; }` there.  IIRC, @qinsoon once tried doing this check after `trace_object`, and it is not always profitable.  We may give it another try, but at this moment, I want to make sure the closure itself doesn't introduce extra cost.

k-sareen: I believe Yi's work made it an `Option<ObjectReference>`. That has more overhead since it doesn't condense into a `usize` (given `ObjectReference::NULL` exists). I don't think we should retain the current semantics since we know the current semantics are broken. There is no need to store the same object again into the slot

wks: Related issues:
-   https://github.com/mmtk/mmtk-core/issues/574
-   https://github.com/mmtk/mmtk-core/issues/559#issuecomment-1128928487

My previous experiment showed that the performance impact is negligible.  So we have reasons to replace the constant `OVERWRITE_REFERENCE` and the invocation `P::may_move_objects::<KIND>()` with an actual `new_object == object` check. But we'll probably do it in a separate PR to address https://github.com/mmtk/mmtk-core/issues/574 specifically.

k-sareen: Sure. Alternatively, we can make a PR for fixing that right now (since it should be simple to fix) and then merge that first, given this PR is a "design" PR and I'm not sure if we've decided what's the best way to review them yet

wks: > Sure. Alternatively, we can make a PR for fixing that right now (since it should be simple to fix) and then merge that first

Yes. That's a good idea.

> given this PR is a "design" PR and I'm not sure if we've decided what's the best way to review them yet

I think this can be an "MEP", too, but this may not be as controversal as removing `ObjectReference::NULL`.  Since I already made this PR, I think I will test the performance with the OpenJDK binding and, if it performas well, I'll write up an "MEP" for this, too, and welcome everyone to discuss.  Actually I still have some design questions about this.  One is whether we need a way to update slots atomically.  The LXR branch sometimes does "compare exchange" on slots.

PR #1045, opened 2023-12-12
Flush mutator buffers in `destroy_mutator`

k-sareen:


wks (approved):
LGTM.

I think a deeper problem is that the semantics of `flush_mutator` is unclear, and it should be removed from the API for the VM binding.

See: https://github.com/mmtk/mmtk-core/issues/1047

And this PR LGTM because it makes the semantics of `destroy_mutator` clear.  We should remove `flush_mutator` and fix the OpenJDK binding in another PR.

qinsoon:
The Ruby tests failed, but it could be unrelated with the PR: https://github.com/mmtk/mmtk-core/actions/runs/7192554303/job/19589222684. @wks Can you check this before merging?

wks:
> The Ruby tests failed, but it could be unrelated with the PR: https://github.com/mmtk/mmtk-core/actions/runs/7192554303/job/19589222684. @wks Can you check this before merging?

Very strange.  It is a segmentation fault without any further message.

Actually I am about to make an update to the MMTk-ruby binding.  I'll re-run that after updating the Ruby binding, first.

qinsoon:
> > The Ruby tests failed, but it could be unrelated with the PR: https://github.com/mmtk/mmtk-core/actions/runs/7192554303/job/19589222684. @wks Can you check this before merging?
> 
> Very strange. It is a segmentation fault without any further message.
> 
> Actually I am about to make an update to the MMTk-ruby binding. I'll re-run that after updating the Ruby binding, first.

 There was an error message in the first attemp:
```console
  ERROR: An MMTk GC thread panicked.  This is a bug.
  panicked at 'Object 0x200fff78718 does not have HAS_MOVED_GIVTBL flag or original givtbl', src/abi.rs:111:17
  Backtrace is disabled.
  run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

wks:
> > > The Ruby tests failed, but it could be unrelated with the PR: https://github.com/mmtk/mmtk-core/actions/runs/7192554303/job/19589222684. @wks Can you check this before merging?
> > 
> > 
> > Very strange. It is a segmentation fault without any further message.
> > Actually I am about to make an update to the MMTk-ruby binding. I'll re-run that after updating the Ruby binding, first.
> 
> There was an error message in the first attemp:
> 
> ```
>   ERROR: An MMTk GC thread panicked.  This is a bug.
>   panicked at 'Object 0x200fff78718 does not have HAS_MOVED_GIVTBL flag or original givtbl', src/abi.rs:111:17
>   Backtrace is disabled.
>   run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

This is reproducible locally.  I think there is a bug.

Another bug is that the CI script in mmtk-core does not look at `Cargo.toml` in `mmtk-ruby` for the chosen revision of repo `ruby`, so it automatically checked out the latest revision of the `ruby` repo which I am just about to run CI scripts against (which failed, too).  I have made a force push to the `ruby` repo to roll back to the previous head.  I'll fix the CI script for mmtk-core later.

k-sareen:
You should also set `RUST_BACKTRACE=1` in the CI if you haven't already.

PR #1048, opened 2023-12-14
Fix Ruby binding test repo checkout

wks:
The Ruby binding test now checks out the revision of the `ruby` repo selected in `mmtk-ruby/mmtk/Cargo.toml` instead of the latest revision.


k-sareen (approved):
LGTM. Thanks!

tianleq (approved):

Issue #99, opened 2020-06-05
MockVM


qinsoon:
For testing, we include a `DummyVM` in the mmtk-core repo. We should rename it to `MockVM` to properly reflect its usage. And we should allow proper mocking for testing.

qinsoon:
We have gradually added some tests to `DummyVM` to MMTk core: https://github.com/mmtk/mmtk-core/tree/master/vmbindings/dummyvm/src/tests

We would at some point refactor `DummyVM` into a proper `MockVM`. Specifically, we need to be able to define the MockVM's behavior per each test. 

For example, currently, when a GC is triggered for a `DummyVM`, we simply panic.
https://github.com/mmtk/mmtk-core/blob/0babba20290d3c4e4cdb2a83284aa7204c9a23cc/vmbindings/dummyvm/src/collection.rs#L20
In the test in which we expect a GC to be triggered, we check if the test panics
https://github.com/mmtk/mmtk-core/blob/0babba20290d3c4e4cdb2a83284aa7204c9a23cc/vmbindings/dummyvm/src/tests/allocate_with_re_enable_collection.rs#L8
and in the test in which we expect no GC to be triggered, we simply see if the test passes without panic
https://github.com/mmtk/mmtk-core/blob/0babba20290d3c4e4cdb2a83284aa7204c9a23cc/vmbindings/dummyvm/src/tests/allocate_with_disable_collection.rs#L8

The panic behavior of `block_for_gc()` makes it impossible for us to use the binding for any GC test, as it will panic whenever a GC is triggered. We would expect that `MockVM` would allow us to define `block_for_gc()` per each test case.

The simplest approach is to use function pointers. `MockVM` forwards each call to a separate function pointer, and the function pointer initially points to a default implementation. Each test can create their own function, and update the function pointer for certain API calls.

@wks @KipHamiltons

wks:
I have another example.

I want my `is_mmtk_object` feature to work for different VMs.  The offsets from the allocated address and the object reference is different among VMs, So I want my test cases to control the behaviour of `ObjectModel::object_start_ref` of a VM.

PR #1049, opened 2023-12-14
Introduce MockVM

qinsoon:
This PR introduces `MockVM`, a type that implements `VMBinding` and allows users to control the behavior of each method for testing. This PR also moves all the tests in the current `DummyVM` to `MockVM`, and removes the current `DummyVM`. 

This PR closes https://github.com/mmtk/mmtk-core/issues/99. Note that the current `MockVM` implementation does not allow changing constants or associated types in `VMBinding` -- I would suggest we create another issue to track this problem.

Changes:
* Introduce `MockVM`, and write all the current `DummyVM` tests with `MockVM`.
* Remove `DummyVM`, and remove `./examples` which uses `DummyVM`.
* Change CI scripts to run tests with `MockVM`.
* Remove `pub` visibility for some modules. Those modules were exposed so we can test them from `DummyVM`. As now `MockVM` is a part of the main crate, we can test private items. We no longer need `pub` for those modules.


src/util/test_util/mock_vm.rs
```diff
@@ -0,0 +1,597 @@
...
+pub fn read_mockvm<F, R>(func: F) -> R
+where
+    F: FnOnce(&MockVM) -> R,
+{
+    let lock = MOCK_VM_INSTANCE
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner());
+    func(&lock)
+}
+/// Write to the staitc MockVM instance. It deals with the case of a poisoned lock.
```
wks, line 70:
```suggestion
/// Write to the static MockVM instance. It deals with the case of a poisoned lock.
```

src/util/test_util/mock_vm.rs
```diff
@@ -0,0 +1,597 @@
...
+                    VMWorkerThread,
+                    &'static mut Mutator<MockVM>,
+                    ProcessEdgesWorkRootsWorkFactory<
+                        MockVM,
+                        SFTProcessEdges<MockVM>,
+                        SFTProcessEdges<MockVM>,
+                    >,
+                ),
+                (),
+            >::new_unimplemented()),
```
wks, line 319: This is one example of API methods that have deep call-back structures.  The call sequence of a normal VM is:

1.  MMTk core wants to scan the roots of a Mutator.  It calls `scan_roots_in_mutator_thread`, passing an `impl RootsWorkFactory` instance.
2.  The VM binding implements this method.  It calls methods of the `RootsWorkFactory` to deliver roots.  For example, it calls `create_process_edge_roots_work` to deliver a list of slots (`Vec<impl Edge>`)

The testing script needs to attach a closure to `scan_roots_in_mutator_thread` when it is testing, for example, the `ScanMutatorRoots<C>` work packet.  However, the `ScanMutatorRoots<C>` work packet has a type parameter `<C>` and will determine the concrete `RootsWorkFactory` implementation.  Therefore, a hard-coded type

```
ProcessEdgesWorkRootsWorkFactory<
                        MockVM,
                        SFTProcessEdges<MockVM>,
                        SFTProcessEdges<MockVM>,
                    >
```

will simply not work.  The `MockAny` trait hides the concrete type, but the downcast will be guaranteed to fail (unless the chosen `Plan` actually uses `SFTProcessEdges` for both the `E` and the `I` type arguments of `RootsWorkFactory`, but we use `PlanProcessEdges` most of the time).

To solve the problem, we should look at what the `RootsWorkFactory` really is. It is really just a fancy way for the VM binding to deliver lists of roots to MMTk core.  So when we design the mocking facility, we can be more straightforward about this -- just let the closure deliver the roots.

Add a new enum type:

```rs
pub enum RootsWorkFactoryCall {
    Slots(Vec<MockVMEdge>),
    PinNodes(Vec<ObjectReference>),
    TPinNodes(Vec<ObjectReference>),
}
```

Change the `scan_roots_in_mutator_thread` and `scan_vm_specific_roots` fields to a `MockMethod`:

```rs
    pub scan_roots_in_mutator_thread: MockMethod<(VMWorkerThread, &'static mut Mutator<MockVM>), Vec<RootsWorkFactoryCall>>,
    pub scan_vm_specific_roots: MockMethod<VMWorkerThread, Vec<RootsWorkFactoryCall>>,
```

Note that we remove the `RootsWorkFactory` from the arguments, but add a `Vec<RootsWorkFactoryCall>` as the return value.

Then when implementing those methods for `Scanning`, we call the closure, and then call the actual factory according to the return values:

```rs
    fn scan_roots_in_mutator_thread(
        tls: VMWorkerThread,
        mutator: &'static mut Mutator<Self>,
        mut factory: impl RootsWorkFactory<<MockVM as VMBinding>::VMEdge>,
    ) {
        let results = mock!(scan_roots_in_mutator_thread(tls, mutator));
        for result in results {
            match result {
                RootsWorkFactoryCall::Slots(edges) => factory.create_process_edge_roots_work(edges),
                RootsWorkFactoryCall::PinNodes(nodes) => factory.create_process_pinning_roots_work(nodes),
                RootsWorkFactoryCall::TPinNodes(nodes) => factory.create_process_tpinning_roots_work(nodes),
            }
        }
    }
    fn scan_vm_specific_roots(
        tls: VMWorkerThread,
        mut factory: impl RootsWorkFactory<<MockVM as VMBinding>::VMEdge>,
    ) {
        let results = mock!(scan_vm_specific_roots(tls));
        for result in results {
            match result {
                RootsWorkFactoryCall::Slots(edges) => factory.create_process_edge_roots_work(edges),
                RootsWorkFactoryCall::PinNodes(nodes) => factory.create_process_pinning_roots_work(nodes),
                RootsWorkFactoryCall::TPinNodes(nodes) => factory.create_process_tpinning_roots_work(nodes),
            }
        }
    }
```

src/util/test_util/mock_vm.rs
```diff
@@ -0,0 +1,597 @@
...
+            notify_initial_thread_scan_complete: MockMethod::new_unimplemented(),
+            supports_return_barrier: MockMethod::new_unimplemented(),
+            prepare_for_roots_re_scanning: MockMethod::new_unimplemented(),
+            process_weak_refs: Box::new(MockMethod::<
+                (
+                    &'static mut GCWorker<Self>,
+                    ProcessEdgesWorkTracerContext<SFTProcessEdges<MockVM>>,
+                ),
+                bool,
+            >::new_unimplemented()),
```
wks, line 340: This method should be similar to the case of `scan_roots_in_mutator_thread`.  It may use the `TracerContext` to create an `ObjectTracer`, and use it to trace several objects.  One way to do mock testing is simply let the closure return a list of `ObjectReference` to be traced.

wks (commented):
I think the MockAny may not work because the original closures are supposed to accept objects that are implemented with different type arguments.  But this fact still doesn't prevent us from doing mock tests for those methods.  I have given concrete examples in in-line comments.  For the root-scanning methods, we simply let the closure provide the list of roots instead of making a `dyn` callback objects from it to call.  Similarly, for the weak reference processing methods, we just let the closure list the object references to be traced.

However, I tried to write a mock test case for testing the `ScanVMSpecificRoots` work packet, but failed because I found it hard to instantiate a concrete `ProcessEdgesWorkRootWorkFactory` without knowing the plan.

src/util/test_util/mock_vm.rs
qinsoon: > Therefore, a hard-coded type will simply not work. The `MockAny` trait hides the concrete type, but the downcast will be guaranteed to fail (unless the chosen `Plan` actually uses `SFTProcessEdges` for both the `E` and the `I` type arguments of `RootsWorkFactory`, but we use `PlanProcessEdges` most of the time).

Right. The default `MockMethod` is just a place holder and it won't work, as it carries some type information that won't fit in some cases. I used `SFTProcessEdges` in the place holder as it is simple to write. I will probably make this clearer. If someone wants to mock this method, they should create a `MockMethod` for their tests. For example,
```rust
MockVM {
  scan_roots_in_mutator_thread: Box::new(MockMethod::<..., ProcessEdgesWorkRootsWorkFactory<MockVM, PlanProcessEdges<MockVM, Immix, DEFAULT_TRACE>, PlanProcessEdges<MockVM, Immix, DEFAULT_TRACE>>, ()>,
  ..MockVM::default()
}
```
Do you think this would work?

But as you said, it only works for one plan, which means the user needs to make sure the tests only run for the plan. However, that is how our code works right now, rather than an issue of the mocking.

> Change the `scan_roots_in_mutator_thread` and `scan_vm_specific_roots` fields to a `MockMethod`:
> 
> ```rust
>     pub scan_roots_in_mutator_thread: MockMethod<(VMWorkerThread, &'static mut Mutator<MockVM>), Vec<RootsWorkFactoryCall>>,
>     pub scan_vm_specific_roots: MockMethod<VMWorkerThread, Vec<RootsWorkFactoryCall>>,
> ```

I don't like to transform some methods for mocking in an ad-hoc way. I hope we can mock any method (including new methods) by following some simple rules (like what I wrote down in the comments -- we just need to get rid of generic lifetime and type parameters). If we have to deal with some methods in an ad-hoc way, it mostly means the mocking cannot handle some cases, and there is a chance that we cannot mock some methods at all.

qinsoon: I have fixed other issues in the comments. Can you take a look at this? @wks I will try get it resolved.

wks: > Do you think this would work?

Probably.  I understand your point that the user of the MockVM can set the MockMethod on a per-instance basis, and they can always set it to the proper MockMethod that works for their particular test.

wks: I suggest adding a comment here, letting the reader know your point in the following paragraph

> Right. The default MockMethod is just a place holder and it won't work, as it carries some type information that won't fit in some cases. I used SFTProcessEdges in the place holder as it is simple to write. I will probably make this clearer. If someone wants to mock this method, they should create a MockMethod for their tests. For example,

wks (changes requested):

src/util/metadata/header_metadata.rs
```diff
@@ -454,7 +454,7 @@ impl fmt::Debug for HeaderMetadataSpec {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, debug_assertions))]
```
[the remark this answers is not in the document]
qinsoon: The tests use `HeaderMetadataSpec::assert_spec()` which is conditionally compiled when `cfg(debug_assertions)` is set. Before this PR, we only run `cargo test`, which sets both `cfg(test)` and `cfg(debug_assertions)`. In this PR, the benches are moved to `mmtk-core`. When we run `cargo bench`, it builds MMTk with `cfg(test)` but not `cfg(debug_assertions)`. Under this configuration, these tests fail to build, as `assert_spec` is not found.

benches/alloc.rs
```diff
@@ -0,0 +1,18 @@
+use criterion::criterion_group;
```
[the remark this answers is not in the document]
qinsoon: Done. I added `cargo clippy --features mock_test --tests/benches` to `ci-style.sh` to check these.

docs/userguide/src/portingguide/perf_tuning/alloc.md
```diff
@@ -78,7 +78,7 @@ which includes (but not limited to) `NoGC`, `SemiSpace`, `Immix`, generational p
 If a plan does not do bump-pointer allocation, we may still implement fast-paths, but we need to embed different data structures instead of `BumpPointer`.
 
 ```rust
-{{#include ../../../../../vmbindings/dummyvm/src/tests/doc_mutator_storage.rs:mutator_storage_embed_fast-path_struct}}
+{{#include ../../../../../src/vm/tests/mock_tests/mock_test_doc_mutator_storage.rs:mutator_storage_embed_fast-path_struct}}
```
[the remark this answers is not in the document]
qinsoon: Done. However, `mdbook` does not give a waring on this, and I did not find a way to let mdbook check this. I tried `mdbook-linkcheck`, and it did not warn on this.

src/util/test_util/mock_vm.rs
qinsoon: Done. I added some comments for `MockAny`, and `MockVM` in https://github.com/mmtk/mmtk-core/pull/1049/commits/9fe32eee86ce9475b5adf0302b52d3e4f658d6be. I further added some comments when we create the `MockMethod` for `MockAny` in https://github.com/mmtk/mmtk-core/pull/1049/commits/85227dfd02188f72c1d6491f52f7d015bfeaf84e.

wks (approved):
LGTM

Issue #918, opened 2023-08-23
Incorrect collection_pages calculation in mem balancer


qinsoon:
The mem balancer paper uses a metric called 'bytes collected during GC'. I interpreted as 'bytes reclaimed'. But it turned out to be incorrect. It should actually be 'bytes traversed during GC'.

Basically our calculation for the following field is wrong. We should fix this.
https://github.com/mmtk/mmtk-core/blob/9175e68a92484ecc6b4fac9db0a58f149f3386fb/src/util/heap/gc_trigger.rs#L188

As we use pages for our calculation (such as live pages, and allocation pages), the question for us is that how we can get the number of pages traversed by the GC. Earlier we added a way to count traversed bytes (https://github.com/mmtk/mmtk-core/pull/768), but it counts bytes not pages. We probably can calculate traversed pages from live pages and promoted pages.

wks:
Alternatively, we can count non-empty pages during the release stage.  This can avoid introducing cost during the transitive closure.

qinsoon:
We have reserved pages, which is non-empty pages. But that does not always equal to pages we traversed in a GC. There are memory that is alive and reserved, but is not traversed in a GC. Like unreachable objects in immortal spaces, or mature objects in a nursery GC.

k-sareen:
Can you point to the page in the paper which mentions it's traversed pages and not collected pages?

qinsoon:
In Fig 4 of the paper, when it refers to 'bytes collected during collection', it means 'bytes traversed by the GC' rather than 'bytes freed by the GC' (we checked with the author).

k-sareen:
Ah right. This is just a heuristic anyway, so maybe it's fine to just use live bytes instead of traversed bytes.

PR #1050, opened 2023-12-15
Use live memory to estimate collection pages in mem balancer

qinsoon:
This PR fixes https://github.com/mmtk/mmtk-core/issues/918. We misinterpreted 'bytes collected during collection' in the paper, and used reclaimed pages for it. It actually means 'bytes that are traversed during collection'. We now use live pages to estimate the value.


wks (approved):
LGTM.

PR #1056, opened 2023-12-21
MMTk Enhancement Proposal

wks:
This PR adds a document that describes the MMTk Enhancement Proposal (MEP).


docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+-   Bug fixes, including performance bug fixes, that require change to a major part of the MMTk
+    core.
+-   Changes to the MMTk core to implement a feature demanded by bindings.
+-   Major refactoring to the MMTk core.
+
+One purpose of MEP is reducing the risks to the future development of MMTk.  Large-scale changes and
+public API changes usually indicate such risks, but these are only indicators, not criteria.  The
+assessment of risks is subjective, and we need to discuss in order to reach consensus.
+
+Note: JEP is also required for things that "require two or more weeks of engineering effort" and/or
```
qinsoon, line 25: I would suggest not mentioning too much about JEP, especially when it is irrelevant. 1. As you said, MEP is clearly different from JEP. 2. Not many people are that familiar with JEP so it is hard for them to draw connection between the two.

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+"are in high demand by developers or customers".  We don't judge whether we need MEP based on
+engineering effort or public demand.  Many PRs for MMTk require multiple weeks of work and rigorous
+testing, but most of them can be settled with regular issues and PRs.  We label priorities using
+GitHub issue tags, such as `P-normal`, `P-high`, etc.  If a feature is requested often, and we have
```
qinsoon, line 29: I don't see why we discuss about priority tags here.

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+man power for that, we can raise the priority.
+
+# Format
+
+A MEP will be posted as a GitHub issue in the `mmtk-core` repository.  It should contain certain
```
qinsoon, line 34: The only thing that is required is the `MEP` tag. The others are just our normal convention about Github issues -- we don't need to discuss too much here. Like the `A-` tags and `C-` tags, we are not requiring the tags for issues (many issues are not tagged with those). `G-` tags only include `G-safety` and `G-performance`, which is clearly not covering all the issues. Thus `G-` tags should not be required as well.

qinsoon: We may want a separate doc about our policies about Github issues. Some of the discussions about tags should go into that doc (which does not need to be a part of this PR and this doc).

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+-   Risks
+    -   Long Term Performance Risks
+    -   Long Term Software Engineering Risks
+    -   Impact on API
+-   Testing
+-   Alternatives (optional)
+-   Risks and Assumptions (optional)
+-   Related Issues (optional)
+
+Note: Sections in JEP but not in MEP:
```
qinsoon, line 70: This is about JEP, and is irrelevant for MEP.

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+    -   For example, `C-refactoring`.
+    -   Note that not all MEP are "enhancement" in the sense of `C-enhancement`.  Some MEPs may
+        simply be intended for fixing long-standing hard-to-fix bugs by making non-trivial changes.
+-   **Goal (`G-*`)**
+    -   For example, `G-safety`.
+
+We use the format of JEP (https://openjdk.org/jeps/2) as a frame of reference, but deviate from it
+when needed.
+
+A MEP should have the following sections:
```
qinsoon, line 51: We should have a link to the Github issue template for `MEP`.

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+team. If an MEP is accepted, A PR may follow the MEP and will be reviewed with the normal PR review
+process.
+
+If an MEP is rejected, future related MEPs may not be reviewed again unless they are substantially
+different. We encourage people to get involved in the review discussion, and refine the proposal so
+it will be accepted.
+
+## Communication
+
+After an MEP is implemented, the MMTk team shall announce the significant changes as the results of
```
qinsoon, line 259: I don't think we need this section. Most of the review process is public and the outcome for the review process is public as well. I don't feel we should be obligated to do anything extra to communicate/announce to others.

> We shall communicate with our sponsors about such changes, too, in our regular meetings.

It would actually feel strange if we have to communicate MEP changes with our sponsors when those changes are irrelevant to them.

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+the MEP to make them known to the community.  We shall communicate with our sponsors about such
+changes, too, in our regular meetings.
+
+<!--
```
qinsoon, line 263: This should be omitted.

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+
+Note: Sections in JEP but not in MEP:
+
+-   Dependencies: We have the *Related Issues* section, instead.
+
+# Sections
+
+## TL;DR
+
+This section should use about one to three sentences to summarize the MEP.  JEP calls it "Summary",
```
qinsoon, line 78: Mentioning JEP is not quite helpful here. 'Summary' has no special meaning in JEP, and we are not calling it summary.

Also we stated that we only deviate from JEP when needed. Calling it 'TLDR' instead of 'summary' does not sound necessary (which contradicts to what we said about not deviating from JEP), and 'TLDR' is a bit informal for a formal process. But anyway, I am okay with the term. Just I think we should mention less about JEP.

qinsoon (changes requested):
We probably want to put this doc in a separate folder under `docs`. This doc is not only for the team, but also for any developers. We could have a folder `docs/contribute`, and put the doc in the folder.

A check list before this PR is merged:
* [ ] An issue template for MEP
* [ ] An issue template for normal issues (optional)
* [ ] Issue tags `MEP` and `F-request-for-MEP`.

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
+MMTk Enhancement Proposal (MEP)
+===============================
+
+An MMTk Enhancement Proposal (MEP) is a more formal variant of issue.  It has a special format, and
```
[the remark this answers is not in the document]
wks: I used your sentence as the definitive description.  But I think we should still include the details here because this paragraph is the summary of this document.  It helps the reader (especially those who have just heard of MEP) understand important details without going through the whole document.

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+
+An MEP is required when making a significant change to the design of the MMTk core.  It is
+applicable to any kind of significant changes, including but not limited to:
+
+-   Bug fixes, including performance bug fixes, that require change to a major part of the MMTk
+    core.
+-   Changes to the MMTk core to implement a feature demanded by bindings.
+-   Major refactoring to the MMTk core.
+
+One purpose of MEP is reducing the risks to the future development of MMTk.  Large-scale changes and
```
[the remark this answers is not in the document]
wks: I replaced this sentence with what Steve wrote in the "risks" section.  "The core purpose of the MEP process is to avoid the unintentional introduction of changes that bring long-term negative impacts to MMTk." Saying it is the "core purpose" sounds better than "one purpose is...", but it still doesn't rule out the possibility that there are other purposes.

I think we don't need to emphasize whether it is the only purpose or not.  If we find MEP may serve other purposes, we can update this document.

> I think it is a good idea to state what 'risks' means...

We now have a dedicated "risks" section to discuss the details of the risks.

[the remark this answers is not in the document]
wks: Personally, I think MEP is also useful if I want to make a major change to MMTk, but I don't know exactly how to implement it.  The MEP process should be able to help by allowing members to discuss the best implementation strategy together.  I'd really like the MEP process to help with the PR https://github.com/mmtk/mmtk-core/pull/1004 because it is important but I haven't figured out the satisfactory implementation, yet.  We can try that later and see if MEP can serve this purpose.

docs/team/mep.md
```diff
@@ -0,0 +1,265 @@
...
+public API changes usually indicate such risks, but these are only indicators, not criteria.  The
+assessment of risks is subjective, and we need to discuss in order to reach consensus.
```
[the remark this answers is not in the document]
wks: I have made the change and added the paragraph as you suggested.

docs/team/mep.md
wks: I agree.  I removed all places that mention JEP, keeping only one sentence that mentions that MEP is inspired by JEP.

docs/team/mep.md
wks: I removed this paragraph.

docs/team/mep.md
wks: I removed the sentences that mentiond the `{A,C,G}-*` tags, and only kept the `MEP` tag.

docs/team/mep.md
wks: I have removed that.

docs/team/mep.md
wks: I'll add it after this PR is merged, and the URL should be shown after that.

docs/team/mep.md
wks: This helps vim keep the same indentation and text width so that it can be automatically formatted (with the command `gqap` in vim).  Anyone who edit this document will have the same setting.

Generally speaking, letting text files wrap lines at a specific width makes it more friendly to readers who read the file locally with a text editor (Vim, VSCode, etc.).

docs/team/mep.md
wks: I changed the paragraph so that it no longer mentions JEP.

docs/team/mep.md
wks: We discussed this on Wednesday.  I think even though our reviewing process is public, not every user or contributor watch every issue or MEP.  I think we still need to announce such changes in a more dramatic way just in case some users may suddenly find things breaking in an unexpected way.

I changed the last sentence to "communicate with sponsors... if relevant."

qinsoon (approved):
LGTM

PR #1057, opened 2023-12-22
Update issue templates

wks:


wks:
We'll commit this together with the PR for the MEP doc.

Issue #1061, opened 2024-01-02
Remove or refactor `GcStatus`


wks:
The `GcStatus` type is defined as following:

```rust
pub enum GcStatus {
    NotInGC,
    GcPrepare,
    GcProper,
}
```

It is inherited from JikesRVM.  The status is set during different phases of a GC.

-   It is set to `GCPrepare`
    -   in `ScheduleCollection` in Rust MMTk, and
    -   in the `INITIATE` phase in JikesRVM.
-   It is set to `GCProper`
    -   when all mutator stacks have been scanned in Rust MMTk, and
    -   in the end of the `STACK_ROOTS`  and `ROOTS` phases in JikesRVM.  Also set to `GC_PROPER` in `processPhaseStack`
-   It is set to `NotInGC`
    -   right before calling `resume_mutators` in Rust MMTk, and
    -   in the `COMPLETE` phase in JikesRVM.  Also set to `NOT_IN_GC` in `processPhaseStack`

But its value is never read in the Rust MMTK.  Neither `MMTK::gc_in_progress()` nor `MMTk::gc_in_progress_proper()` is called anywhere in the Rust MMTk or any of its bindings.

In JikesRVM,

-   `gcInProgress()` is only used in assertions, especially for concurrent plans.
-   `gcInProgressProper()` is only called in `modifyCheck` which no longer exists in the Rust MMTk.

It looks like the GC status mainly helps in concurrent GC algorithms, but the lxr branch of Rust MMTk does not read the `gc_status` field, either.

So I think `GcStatus` can be safely removed from MMTK without any harm.

# Repurposing

~~When I was trying to remove the coordinator thread (See https://github.com/mmtk/mmtk-core/issues/1053), I realized that with the coordinator thread removed, we need a state machine to record the GC status, in a more fine-grained way than the current `GcStatus` and the current fields of `GCReqeuster`, and the states will affect the behavior of several things, including triggering GC, and the behavior of the last parked GC worker (to open new buckets or to wait for the GC to be triggered).  I think we can re-purpose the `GcStatus` type for this.~~

*(Update: https://github.com/mmtk/mmtk-core/pull/1067 which fixes #1053 now uses its own state for that purpose, and will not affect this issue.)*

wks:
[...]

Issue #1053, opened 2023-12-20
Remove the coordinator (controller) thread


wks:
It was created because OpenJDK requires the thread that stops the world to be the same thread that starts the world.  But now the OpenJDK binding uses a dedicated "companion" thread for that purpose.

The coordinator thread has brought much trouble to the synchronization with GC workers, and it has been changed many time for that.  With the concept of "coordinator work" removed in https://github.com/mmtk/mmtk-core/pull/794, the coordinator thread is now only used to respond to GC triggers.  We realized that we can remove the coordinator and the GC workers can coordinate themselves.

Removing the coordinator thread can also save one context switch between the last GC worker starting sleeping and the coordinator thread opening new buckets.

wks:
We used to believe that the coordinator (controller) thread was introduced to handle STW for OpenJDK (which requires the thread that starts and stops the world to be the same).  But from the source code of JikesRVM, it appears that the controller thread existed in JikesRVM, too.  The `ControllerCollectorContext` was the context of the controller thread, and what we currently have as `GCRequester` originated from that (I did the refactoring in https://github.com/mmtk/mmtk-core/pull/533).

I think we already agreed that we can remove the coordinator thread.  To implement this change, however, we need to acknowledge that we are deviating from JikesRVM.  We probably have a good reason to do this, that is, we now use work packets.  We still need to implement a GC-triggering algorithm that is equivalent to the old [ControllerCollectorContext](https://github.com/JikesRVM/JikesRVM/blob/master/MMTk/src/org/mmtk/plan/ControllerCollectorContext.java).  For now, it should be as simple as adding the `ScheduleCollection` work packet.  With concurrent GC added, it should handle the case where a mutator triggers a GC while GC is still running, in which case another GC should start when the current GC finishes.

wks:
The coordinator thread behaves like an implicit state machine: WaitForGC -> ScheduleGC -> OpenMoreBuckets -> OpenMoreBuckets -> ... -> GcFinished -> WaitForGC

After removing the coordinator thread, GC workers coordinate themselves.  Specifically when all GC workers parked, the last parked worker shall perform actions.

-   If MMTk has just been initialized, all GC workers should simply park waiting for the first GC.
-   If GC is requested (`poll` or `handle_user_collection_request`) and GC has not been scheduled, it should schedule the GC (by adding the `ScheduleCollection` work packet).  If it is already scheduled, however, it will be a no-op if another mutator also requests for GC.
[...]

PR #1062, opened 2024-01-05
Remove the coordinator (controller) thread

wks:
**DRAFT**: Need polishing.  Need performance evaluation.  Need to update bindings.

This PR removes the coordinator (a.k.a. controller) thread.  GC worker threads now coordinate themselves.

Major changes include:

-   `GCController` is removed.  All synchronization mechanisms for the communication between the controller and workers are removed, too.
-   The mechanism for triggering GC is redesigned.  Mutators now communicate with workers directly, and a variable `WorkerMonitorSync::should_schedule_gc` is added as the communication channel.  To trigger a GC, a mutator shall acquire the mutex `WorkerMonitor::sync`, set `sync.should_schedule_gc` to `true`, and notify one worker.  The last parked worker always checks this variable before blocking.
-   The last parked worker (instead of the coordinator) is now responsible for adding the `ScheduleCollection` work packet when a GC is requested.
-   The last parked worker (instead of the coordinator) is now responsible for opening work packets when open buckets are drained.
-   The work packet `EndOfGC` is removed.  The code is now executed by the last parked worker when GC finishes.
-   `GlobalState::gc_status` is now `Atomic`, and it is read by the last parked worker to decide whether GC is in progress.  Previously, it was not used in Rust MMTk, and was only used for assertion in the MMTk in JikesRVM
-   Added `GlobalState::gc_start_time` to record the start time of a GC.  It was a local variable in `GCController`.
-   Added method `WorkBucket::add_no_notify` to add work packets while holding the mutex `WorkerMonitor::sync`.
-   Updated comments and deleted outdated comments no longer relevant after this change.

Fixes: https://github.com/mmtk/mmtk-core/issues/1053
Closes: https://github.com/mmtk/mmtk-core/issues/1061


wks:
I ran `lusearch` from DaCapo Chopin on `bobcat.moma`, 2.5x min heap size w.r.t. OpenJDK's G1, 40 invocations, 5 iterations each, comparing master (build1) and this PR (build2).

The follow plot is normalised to build1:
image
plotty link: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2024-01-08-Mon-075256&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

The following plot is not normalised:
image
plotty link: https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2024-01-08-Mon-075256&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|41&Histogram%20(with%20CI)^build^mmtk_gc&

From the plots, it looks like this PR reduces the STW time if the absolute (not normalised) number of GC is large.  There is 1% to 2% STW time reduction for GenCopy, GenImmix, SemiSpace and StickyImmix.  For Immix, the STW time is increased by 0.3%, but the error bar is much larger than the difference.  From individual data, it looks like some outlier invocations triggered more GC than others.  The following plot shows the number of GCs executed in each invocation when running Immix, sorted and grouped by build.  It shows the distribution of the number of GCs is the same, but "build2" (yellow) contains some outliers on the right end of the curve.

image

We see this PR makes GC slightly faster, and mainly impacts the plans that trigger GC often (mainly generational plans).  I think the main reason is that this PR reduces one context switch every time all workers stopped (because there is no need to switch to the coordinator).  If a plan does more GC, but fewer work in each GC (mostly nursery GC), the synchronisation overhead will be significant.  But it is hard to explain why SemiSpace becomes faster this way because it is not generational.  Maybe SemiSpace simply triggered GC too many times given this heap size.

wks:
The same setting (lusearch, 2.5x min heap size, 40 invocations) ran on skunk.moma.  This Coffee Lake machine does not have small ("efficiency") cores, so it is expected to have smaller variance.

Normalised to build1:
image

https://squirrel.anu.edu.au/plotty/wks/noproject/#0|skunk-2024-01-10-Wed-093645&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

STW times are reduced for all plans, and Mutator (other) times are increased for all plans.

I realised that there is one problem with the measurement.  The definition of "STW time" changed in this PR.  On master, the STW time starts when the coordinator thread receives a message from the mutator (i.e. when the "requested flag" is set and the coordinator thread is notified by the CondVar), but in this PR, the STW time starts at the beginning of the `ScheduleCollection` work packet.  For this reason, a greater proportion of time is calculated as part of the mutator time.  This explains why the greater the speed-up of STW time is, the greater the slow-down of mutator time is.

The total time reflects the actual speed-up.  There is a 0.3% slow down for Immix, likely due to the slightly increased number of GC due to random effects.

The same data without normalisation:
image
https://squirrel.anu.edu.au/plotty/wks/noproject/#0|skunk-2024-01-10-Wed-093645&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|41&Histogram%20(with%20CI)^build^mmtk_gc&

At this heap size, GenCopy and GenImmix spent more time in STW than in mutator (other).  This explains why their normalised `time.other` values apparently increased so much, because we are dividing the same absolute difference with a smaller denominator.

wks:
I repeated the lusearch run on three machines:

Re-ran on skunk.moma:
image
https://squirrel.anu.edu.au/plotty/wks/noproject/#0|skunk-2024-01-11-Thu-100400&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

fisher.moma:
image
https://squirrel.anu.edu.au/plotty/wks/noproject/#0|fisher-2024-01-11-Thu-100421&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

mole.moma:
image
https://squirrel.anu.edu.au/plotty/wks/noproject/#0|mole-2024-01-11-Thu-100437&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&

On skunk.moma and fisher.moma, the mutator time (`time.other`) increased a lot for almost all plans.  But on mole.moma, it reduced.  This shows that the measured increase in mutator time is not random because it can be repeated.  I still can't find a good explanation, yet.

The GC time (`time.stw`) decreased for all plans on all machine, except Immix on fisher.

The total time (`time`) decreased for all plans on all machines, but not by much.

wks:
I added a `build3` which is https://github.com/mmtk/mmtk-core/pull/1067.  It further refactored over `build2` so that it is easy to express the idea of "letting all workers do one thing" (i.e. work towards one "goal") which can be either "GC" or   "stopping for fork" at this time.  I did the refactoring because I think the current code in this PR compilcates things a bit (although it already greatly simplified the synchronisation by removing coordinator) and made it hard to add forking support.

Result on Skunk.  All the same setting as before except adding `build3`. i.e.:

-   build1: master
-   build2: this PR
-   build3: https://github.com/mmtk/mmtk-core/pull/1067

image

`build2` and `build3` have the same performance characterestics.  `time.other` increased, `time.stw` decreased, `time` decreased.  And the geomean shows `build3` performs slightly better than `build2`, probably due to simplified synchronisation and one removed mutex.  But it is still only 0.5% improvement over `master` in total time.

wks:
I am closing this PR in favor for https://github.com/mmtk/mmtk-core/pull/1067 (Remove coordinator and support forking).  The reason is, after we remove the coordinator thread, we need a way for workers to coordinate themselves.  But the concrete way how they coordinate themselves is dictated by the requirement of supporting forking.  That is, we must implement a coordinating mechanism that (1) does not need a dedicated coordinator thread, and (2) is easy to bring down all GC workers when forking.  The code in this PR accomplished (1), but is obviously not suitable for (2), and https://github.com/mmtk/mmtk-core/pull/1067 has already done non-trivial refactoring over this PR.  Therefore, it is a waste of effort to merge this PR before applying https://github.com/mmtk/mmtk-core/pull/1067.

PR #1055, opened 2023-12-20
Bump version to v0.22

qinsoon:

PR #1065, opened 2024-01-11
Revert a mistake in v0.22 that prevents cargo publish

qinsoon:
https://github.com/mmtk/mmtk-core/pull/1055 removed the version for `mmtk-macros` in `Cargo.toml`. I thought it was fine, as it uses a local path as the dependency. But `cargo publish` actually requires all the dependencies to specify a version. So `cargo publish` failed for https://github.com/mmtk/mmtk-core/releases/tag/v0.22.0 in https://github.com/mmtk/mmtk-core/actions/runs/7284255107/job/19849376243. The action did not return a non-zero code so I was not aware of the issue. `v0.22` was not published to `crates.io`.

This PR adds back the version for `mmtk-macros`, adds a check for `cargo publish` dry run in the style check, and fixes the publish script so the script properly fails if all the attempts of `cargo publish` fail.

### DIFF
--- a/.github/workflows/cargo-publish.yml
+++ b/.github/workflows/cargo-publish.yml
@@ -37,7 +37,7 @@ jobs:
 
           for n in {1..5}; do
               echo "Attempt #"$n
-              false && { success=true; break; }
+              cargo publish && { success=true; break; }
               echo "Wait for Retry #"$n
               sleep 60
           done
--- a/.github/workflows/cargo-publish.yml
+++ b/.github/workflows/cargo-publish.yml
@@ -33,9 +33,16 @@ jobs:
       # The script will retry publish for 5 times with 60 seconds between the retries.
       - name: Public mmtk-core
         run: |
+          success=false
+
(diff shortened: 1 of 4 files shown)


.github/workflows/cargo-publish.yml
```diff
@@ -33,9 +33,16 @@ jobs:
...
         run: |
+          success=false
+
           for n in {1..5}; do
-            echo "Attempt #"$n
-            cargo publish && break
-            echo "Wait for Retry #"$n
-            sleep 60
+              echo "Attempt #"$n
+              false && { success=true; break; }
```
caizixian, line 40: `false` is probably from testing but it should be `cargo publish`

caizixian (changes requested):

.github/workflows/cargo-publish.yml
qinsoon: Right. Sorry for that.

caizixian (approved):

Issue #1070, opened 2024-01-15
Panic with 'VO bit already set' in GenImmix


qinsoon:
Build OpenJDK with `fastdebug` and `VO_BIT=1`.

Run dacapo 2006 with GenImmix.
```
MMTK_PLAN=GenImmix /home/runner/work/mmtk-openjdk/mmtk-openjdk/bundles/jdk/bin/java -XX:+UseThirdPartyHeap -server -XX:MetaspaceSize=100M -Xms20M -Xmx20M -jar /home/runner/work/mmtk-openjdk/mmtk-openjdk/dacapo/dacapo-2006-10-MR2.jar antlr
```

We will see.
```console
[2024-01-15T04:43:14Z INFO  mmtk::util::heap::gc_trigger] [POLL] nursery: Triggering collection (5127/5120 pages)
[2024-01-15T04:43:14Z INFO  mmtk::plan::generational::global] Nursery GC
[2024-01-15T04:43:15Z INFO  mmtk::scheduler::gc_work] End of GC (2692/5120 pages, took 7 ms)
thread '<unnamed>' panicked at '40000000: VO bit already set', /home/runner/.cargo/git/checkouts/mmtk-core-3306bdeb8eb4322b/79fb0bb/src/util/metadata/vo_bit/mod.rs:65:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
```

wks:
Did this happen in CI? We can set `RUST_BACKTRACE` to 1 when running on the CI so that we can always collect the stack trace.

qinsoon:
Yeah. In my PR https://github.com/mmtk/mmtk-openjdk/pull/263, I changed the heap size used in CI to a reasonable value (for https://github.com/mmtk/mmtk-core/issues/424), and started to see this error. I did not look into the problem though.

wks:
This is 100% reproducible on my machine.  I'll have a look at it.

wks:
The culprit is the check `current_chunk != self.common.start`:

```rust
impl<VM: VMBinding> CopySpace<VM> {
    // ... 
    #[cfg(feature = "vo_bit")]
    unsafe fn reset_vo_bit(&self) {
        let current_chunk = self.pr.get_current_chunk();
        if self.common.contiguous {
            // If we have allocated something into this space, we need to clear its VO bit.
            if current_chunk != self.common.start {  // ERROR!
[...]

PR #1071, opened 2024-01-15
Fix un-cleared VO bits for contiguous monotone PR

wks:
Fixed a bug that when using a generational plan, if mutators allocated less than one chunk of memory into the nursery between GCs, some VO bits will not be cleared.

This PR also ensures the field `MonotonePageResourceSync::current_chunk` grows monotonically during GC when the space is contiguous.

Fixes: https://github.com/mmtk/mmtk-core/issues/1070


qinsoon (approved):
LGTM

qinsoon:
The JikesRVM test failed, due to an issue for downloading classpath from its FTP server. It should be unrelated with the change. I will see if we can work around it. I also noticed a random failure for Ruby tests (see run attempt 3). It is probably unrelated as well.

PR #1075, opened 2024-01-23
Ask from binding if GC is disabled

udesou:
MMTk currently supports disabling GC (a.k.a. allowing the heap to grow above its limit) via two api functions: [`disable_collection`](https://github.com/mmtk/mmtk-core/blob/ef2bd6d043d8675badaa415db89be7b52439725f/src/memory_manager.rs#L539) and [`enable_collection`](https://github.com/mmtk/mmtk-core/blob/ef2bd6d043d8675badaa415db89be7b52439725f/src/memory_manager.rs#L519). 

This PR replaces this strategy since currently, MMTk relies on the binding ensuring [thread safety](https://github.com/mmtk/mmtk-core/blob/ef2bd6d043d8675badaa415db89be7b52439725f/src/memory_manager.rs#L515) when calling those functions. In this refactoring, MMTk asks if GC is disabled in `VM::VMCollection::is_collection_disabled()` function, and expects any thread safety logic to be implemented in the VM itself, and not necessarily depend on MMTk.

### DIFF
--- a/benches/alloc.rs
+++ b/benches/alloc.rs
@@ -2,11 +2,24 @@ use criterion::Criterion;
 
 use mmtk::memory_manager;
 use mmtk::util::test_util::fixtures::*;
+use mmtk::util::test_util::mock_method::*;
+use mmtk::util::test_util::mock_vm::{write_mockvm, MockVM};
 use mmtk::AllocationSemantics;
 
 pub fn bench(c: &mut Criterion) {
-    // Setting a larger heap so we won't trigger GC, but we should disable GC if we can
-    let mut fixture = MutatorFixture::create_with_heapsize(usize::MAX);
+    // Setting a larger heap (1TB) so we won't trigger GC, but we should disable GC if we can
+    let mut fixture = MutatorFixture::create_with_heapsize(1 << 40);
+    {
+        write_mockvm(|mock| {
+            *mock = {
+                MockVM {
+                    is_collection_enabled: MockMethod::new_fixed(Box::new(|_| false)),
+                    ..MockVM::default()
+                }
+            }
+        });
+    }
+
     c.bench_function("alloc", |b| {
         b.iter(|| {
             let _addr =
--- a/benches/alloc.rs
+++ b/benches/alloc.rs
@@ -5,9 +5,8 @@ use mmtk::util::test_util::fixtures::*;
 use mmtk::AllocationSemantics;
 
 pub fn bench(c: &mut Criterion) {
-    // Disable GC so we won't trigger GC
-    let mut fixture = MutatorFixture::create_with_heapsize(1 << 30);
-    memory_manager::disable_collection(fixture.mmtk());
+    // Setting a larger heap so we won't trigger GC, but we should disable GC if we can
+    let mut fixture = MutatorFixture::create_with_heapsize(1 << 60);
     c.bench_function("alloc", |b| {
         b.iter(|| {
             let _addr =
--- a/benches/alloc.rs
+++ b/benches/alloc.rs
@@ -6,7 +6,7 @@ use mmtk::AllocationSemantics;
 
 pub fn bench(c: &mut Criterion) {
     // Setting a larger heap so we won't trigger GC, but we should disable GC if we can
-    let mut fixture = MutatorFixture::create_with_heapsize(1 << 60);
+    let mut fixture = MutatorFixture::create_with_heapsize(1 << 50);
     c.bench_function("alloc", |b| {
         b.iter(|| {
             let _addr =
--- a/benches/alloc.rs
+++ b/benches/alloc.rs
@@ -7,8 +7,8 @@ use mmtk::util::test_util::mock_vm::{write_mockvm, MockVM};
 use mmtk::AllocationSemantics;
 
 pub fn bench(c: &mut Criterion) {
-    // Setting a larger heap (1TB) so we won't trigger GC, but we should disable GC if we can
-    let mut fixture = MutatorFixture::create_with_heapsize(1usize << 40);
+    // Setting a larger heap, although the GC should be disabled in the MockVM
+    let mut fixture = MutatorFixture::create_with_heapsize(1 << 30);
     {
         write_mockvm(|mock| {
             *mock = {
--- a/benches/alloc.rs
+++ b/benches/alloc.rs
@@ -6,7 +6,7 @@ use mmtk::AllocationSemantics;
 
 pub fn bench(c: &mut Criterion) {
     // Setting a larger heap so we won't trigger GC, but we should disable GC if we can
-    let mut fixture = MutatorFixture::create_with_heapsize(1 << 50);
+    let mut fixture = MutatorFixture::create_with_heapsize(usize::MAX);
     c.bench_function("alloc", |b| {
         b.iter(|| {
             let _addr =
--- a/benches/alloc.rs
+++ b/benches/alloc.rs
@@ -8,7 +8,7 @@ use mmtk::AllocationSemantics;
 
 pub fn bench(c: &mut Criterion) {
     // Setting a larger heap (1TB) so we won't trigger GC, but we should disable GC if we can
-    let mut fixture = MutatorFixture::create_with_heapsize(1 << 40);
+    let mut fixture = MutatorFixture::create_with_heapsize(1usize << 40);
     {
         write_mockvm(|mock| {
             *mock = {
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -510,42 +510,6 @@ pub fn initialize_collection<VM: VMBinding>(mmtk: &'static MMTK<VM>, tls: VMThre
     probe!(mmtk, collection_initialized);
 }
 
-/// Allow MMTk to trigger garbage collection when heap is full. This should only be used in pair with disable_collection().
-/// See the comments on disable_collection(). If disable_collection() is not used, there is no need to call this function at all.
-/// Note this call is not thread safe, only one VM thread should call this.
-///
-/// Arguments:
-/// * `mmtk`: A reference to an MMTk instance.
-pub fn enable_collection<VM: VMBinding>(mmtk: &'static MMTK<VM>) {
-    debug_assert!(
-        !mmtk.state.should_trigger_gc_when_heap_is_full(),
-        "enable_collection() is called when GC is already enabled."
-    );
-    mmtk.state
-        .trigger_gc_when_heap_is_full
-        .store(true, Ordering::SeqCst);
-}
-
-/// Disallow MMTk to trigger garbage collection. When collection is disabled, you can still allocate through MMTk. But MMTk will
-/// not trigger a GC even if the heap is full. In such a case, the allocation will exceed the MMTk's heap size (the soft heap limit).
-/// However, there is no guarantee that the physical allocation will succeed, and if it succeeds, there is no guarantee that further allocation
-/// will keep succeeding. So if a VM disables collection, it needs to allocate with careful consideration to make sure that the physical memory
-/// allows the amount of allocation. We highly recommend not using this method. However, we support this to accomodate some VMs that require this
-/// behavior. This call does not disable explicit GCs (through handle_user_collection_request()).
-/// Note this call is not thread safe, only one VM thread should call this.
-///
-/// Arguments:
-/// * `mmtk`: A reference to an MMTk instance.
-pub fn disable_collection<VM: VMBinding>(mmtk: &'static MMTK<VM>) {
-    debug_assert!(
-        mmtk.state.should_trigger_gc_when_heap_is_full(),
-        "disable_collection() is called when GC is not enabled."
-    );
-    mmtk.state
-        .trigger_gc_when_heap_is_full
-        .store(false, Ordering::SeqCst);
-}
-
 /// Process MMTk run-time options. Returns true if the option is processed successfully.
 ///
 /// Arguments:
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -454,7 +454,10 @@ pub fn gc_poll<VM: VMBinding>(mmtk: &MMTK<VM>, tls: VMMutatorThread) {
         "gc_poll() can only be called by a mutator thread."
     );
 
-    if mmtk.state.should_trigger_gc_when_heap_is_full() && mmtk.gc_trigger.poll(false, None) {
+    if !(VM::VMCollection::is_collection_disabled())
+        && mmtk.state.should_trigger_gc_when_heap_is_full()
+        && mmtk.gc_trigger.poll(false, None)
+    {
         debug!("Collection required");
         assert!(mmtk.state.is_initialized(), "GC is not allowed here: collection is not initialized (did you call initialize_collection()?).");
         VM::VMCollection::block_for_gc(tls);
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -35,9 +35,8 @@ use std::sync::atomic::Ordering;
 /// 2. Set command line options for MMTKBuilder by [`crate::memory_manager::process`] or [`crate::memory_manager::process_bulk`].
 /// 3. Initialize MMTk by calling this function, `mmtk_init()`, and pass the builder earlier. This call will return an MMTK instance.
 ///    Usually a binding store the MMTK instance statically as a singleton. We plan to allow multiple instances, but this is not yet fully
-///    supported. Currently we assume a binding will only need one MMTk instance.
-/// 4. Enable garbage collection in MMTk by [`crate::memory_manager::enable_collection`]. A binding should only call this once its
-///    thread system is ready. MMTk will not trigger garbage collection before this call.
+///    supported. Currently we assume a binding will only need one MMTk instance. Note that GC is enabled by default and the binding should
+///    implement `VMCollection::is_collection_disabled()` if it requires that the GC should be disabled at a particular time.
 ///
 /// Note that this method will attempt to initialize a logger. If the VM would like to use its own logger, it should initialize the logger before calling this method.
 /// Note that, to allow MMTk to do GC properly, `initialize_collection()` needs to be called after this call when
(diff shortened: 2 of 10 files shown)


wks:
If the VM needs to disable GC to implement critical sections where GC activities such as moving object may affect the correctness of the program (usuallly due to native programs accessing buffers in heap objects), the easiest way is simply telling the current thread not to yield.  Of course this approach only works for STW GC.  p.s. It is always better to pin objects for this purpose instead of relying on disabling GC.

The mechanism introduced in this PR (i.e. letting mmtk-core ask the binding whether it should trigger GC) is still not sound for implementing critical sections.  For example,

-   Mutator1 attempted to allocate a large object and triggered GC.  Mutator1 blocked itself for GC.
-   A GC thread responded and asked the VM binding if it was OK to do GC.
-   The VM binding answered yes.
-   The GC thread started GC, and was just about to stop all mutators.
-   Mutator2 entered a critical section, and native C code started reading from an array in a MMTk heap object.
-   Mutator2 executed a yieldpoint and yielded.
-   GC started tracing, and moved the heap object.

The problem is that one mutator is still active after another mutator triggered GC and a GC thread responded.

qinsoon:
I don't think this PR is trying to introduce something that can be used for implementing critical section. It is trying to replace the current `enable/disable_collection` API with a callback so that MMTk does not need to know about GC enable state and just asks the binding for it. arguably this will simply MMTk's implementation.

src/memory_manager.rs
```diff
@@ -454,7 +454,7 @@ pub fn gc_poll<VM: VMBinding>(mmtk: &MMTK<VM>, tls: VMMutatorThread) {
         "gc_poll() can only be called by a mutator thread."
     );
 
-    if mmtk.state.should_trigger_gc_when_heap_is_full() && mmtk.gc_trigger.poll(false, None) {
+    if !(VM::VMCollection::is_collection_disabled()) && mmtk.gc_trigger.poll(false, None) {
```
qinsoon, line 457: `should_trigger_gc_when_heap_is_full` may be unused for this change, and can be removed?

udesou:
> I don't think this PR is trying to introduce something that can be used for implementing critical section. It is trying to replace the current `enable/disable_collection` API with a callback so that MMTk does not need to know about GC enable state and just asks the binding for it. arguably this will simply MMTk's implementation.

@qinsoon is correct. 

> The mechanism introduced in this PR (i.e. letting mmtk-core ask the binding whether it should trigger GC) is still not sound for implementing critical sections. For example,
>
> Mutator1 attempted to allocate a large object and triggered GC. Mutator1 blocked itself for GC.
> A GC thread responded and asked the VM binding if it was OK to do GC.
> The VM binding answered yes.
> The GC thread started GC, and was just about to stop all mutators.
> Mutator2 entered a critical section, and native C code started reading from an array in a MMTk heap object.
> Mutator2 executed a yieldpoint and yielded.
> GC started tracing, and moved the heap object.

For critical sections, Julia has the notion of [GC state](https://github.com/mmtk/julia/blob/5c406d9bb20d76e2298a6101f171cfac491f651c/src/julia_threads.h#L217-L222). Essentially, if Mutator2 enters a critical section, Mutator1 should wait (when it's waiting for the world to stop) until Mutator2 leaves the critical section.

wks:
Currently, `enable_collection` is primarily used for postponing GC to the moment when GC is safe.  For example, the VM needs to initialize type info metadata (for Java, that include the `Klass` instances for the most primitive types such as `Object`; for Ruby, that include the global tables for strings, symbols, finalizers, etc.) before it can scan any object.  For this use, it is sufficient to let all operations that can possibly trigger GC happen after the `enable_collection`.  The current API says "`enable_collection` is not thread-safe".  It means the VM needs to use its own synchronization mechanisms to ensure no "potential GC-triggering operations" happen concurrently (neither happen before nor happen after) with `enable_collection`.  I think this semantics is fine because VM initialization is usually done in one thread (i.e. before other mutator threads are spawn).

If `disable_collection` cannot be used for critical sections, I don't know where it can be useful.  Maybe some languages allow the user to disable GC (e.g. PHP famously do it so that some programs run 100% faster).

As long as mmtk-core doesn't need to enforce any sychronization between mutators, this PR should be OK.  Specifically,

1.  It made clear in the doc of `is_collection_disabled` that it is called when allocating and when VM requests a GC.  That is, the VM developers know when and in which thread it is called.
2.  The default behavior is "always allowed" which is intuitive, but I guess most VMs will need to override it.

And the synchronization is left to the VM.  The VM can duplicate the current behavior by adding an atomic boolean variable in the binding.

udesou:
> Currently, `enable_collection` is primarily used for postponing GC to the moment when GC is safe. For example, the VM needs to initialize type info metadata (for Java, that include the `Klass` instances for the most primitive types such as `Object`; for Ruby, that include the global tables for strings, symbols, finalizers, etc.) before it can scan any object. For this use, it is sufficient to let all operations that can possibly trigger GC happen after the `enable_collection`. The current API says "`enable_collection` is not thread-safe". It means the VM needs to use its own synchronization mechanisms to ensure no "potential GC-triggering operations" happen concurrently (neither happen before nor happen after) with `enable_collection`. I think this semantics is fine because VM initialization is usually done in one thread (i.e. before other mutator threads are spawn).

I'm a bit confused. Are you referring to `initialize_collection`? Does Ruby actually use `enable_collection` and `disable_collection`? 

> If disable_collection cannot be used for critical sections, I don't know where it can be useful. Maybe some languages allow the user to disable GC (e.g. PHP famously do it so that some programs run 100% faster).

We can ask the Julia folks why they have it, but the semantics as far as I understand is that you may continue allocating (beyond any heap limit) and it will not trigger GC. Essentially disabling any GC pooling that may happen, including the case when the user asks for a GC.

> As long as mmtk-core doesn't need to enforce any sychronization between mutators, this PR should be OK. Specifically,
> 
> 1. It made clear in the doc of is_collection_disabled that it is called when allocating and when VM requests a GC. That is, the VM developers know when and in which thread it is called.
> 2. The default behavior is "always allowed" which is intuitive, but I guess most VMs will need to override it.

I'll try to improve the documentation to include (1). For (2), it means that bindings for VMs that do not have such "feature" don't need to change anything. Again, if we're talking about `enable_collection` and not `initialize_collection`, the latter should not be impacted by this change, I hope.

wks:
> I'm a bit confused. Are you referring to `initialize_collection`? Does Ruby actually use `enable_collection` and `disable_collection`?

Both.  Ruby currently uses both `initialize_collection`, `enable_collection` and `disable_collection` during VM start-up.  The current code is something like:

```c
        // The threading system is ready.  Initialize collection.
        mmtk_initialize_collection(th);
        // Bind the main thread.
        rb_mmtk_bind_mutator(th);
        // Temporarily disable GC.
        mmtk_disable_collection();

// After lots of initialization, in another function

        if (rb_mmtk_enabled_p()) {
            // Now that the VM says it's time to enable GC, we enable GC for MMTk, too.
            mmtk_enable_collection();
        }
```

Given that the only thing `initialize_collection` (currently) does is spawning GC threads, I think it is equivalent to postpone the `initialize_collection` to the place of `enable_collection`, and remove both the `disable_collection` and `enable_collection` above.  "Having GC threads but they cannot do GC" is equivalent to "not having GC threads at all".

qinsoon:
Is this PR ready for review? I would prefer @wks to review this. I will remove myself from the reviewer list.

src/vm/collection.rs
```diff
@@ -139,4 +139,21 @@ pub trait Collection<VM: VMBinding> {
...
+    /// not trigger a GC even if the heap is full. In such a case, the allocation will exceed MMTk's
+    /// heap size (the soft heap limit). However, there is no guarantee that the physical allocation
+    /// will succeed, and if it succeeds, there is no guarantee that further allocation will keep
+    /// succeeding. So if a VM disables collection, it needs to allocate with careful consideration
+    /// to make sure that the physical memory allows the amount of allocation. We highly recommend
+    /// to have GC always enabled (i.e. that this method always returns false). However, we support
+    /// this to accomodate some VMs that require this behavior. Note that this call also disables
+    /// explicit GCs (through handle_user_collection_request()). Note also that any synchronization
+    /// involving enabling and disabling collections by mutator threads should be implemented by the VM.
+    fn is_collection_disabled() -> bool {
```
wks, line 154: I feel `is_collection_enabled()` more intuitive because otherwise I need to deal with the case of GC "not disabled" which is a double negation.

p.s. Although Chinese love double negation (while English seem to hate double negation), it is still a bit mind-bending when I see double negation in the API.  I tried to implement this for mmtk-ruby, and I got it wrong on my first attempt.

benches/alloc.rs
```diff
@@ -5,9 +5,8 @@ use mmtk::util::test_util::fixtures::*;
 use mmtk::AllocationSemantics;
 
 pub fn bench(c: &mut Criterion) {
-    // Disable GC so we won't trigger GC
-    let mut fixture = MutatorFixture::create_with_heapsize(1 << 30);
-    memory_manager::disable_collection(fixture.mmtk());
+    // Setting a larger heap so we won't trigger GC, but we should disable GC if we can
+    let mut fixture = MutatorFixture::create_with_heapsize(usize::MAX);
```
wks, line 9: Given that our current GC trigger does align-up for the heap size, like this:

```rs
// GCTrigger::new()
                GCTriggerSelector::FixedHeapSize(size) => Box::new(FixedHeapSizeTrigger {
                    total_pages: conversions::bytes_to_pages_up(size),
                }),
```

`usize::MAX` will be aligned to 0.  So we can't use `usize::MAX` now.  Maybe we can choose a slightly smaller value, such as 1TB or 256TB

Because it is based on `MockVM`, we can mock `is_collection_disabled`, too.

src/vm/collection.rs
```diff
@@ -139,4 +139,21 @@ pub trait Collection<VM: VMBinding> {
...
+        // By default, MMTk assumes that collections are always enabled, and the binding should define
+        // this method if the VM supports disabling GC
```
wks, line 156: It is worth mentioning that some VMs cannot do GC until some metadata are set up.

```suggestion
        // this method if the VM supports disabling GC, or if the VM cannot safely trigger GC until certain
        // initialization is done, such as initializing class metadata for scanning objects.
```

wks:
Here is a PR for the Ruby binding: https://github.com/mmtk/mmtk-ruby/pull/52

qinsoon:
binding-refs
JULIA_BINDING_REPO=udesou/mmtk-julia
JULIA_BINDING_REF=feature/check-gc-disabled
RUBY_BINDING_REF=feature/check-gc-disabled

qinsoon:
The merge check somehow failed, which is strange. It seems observing extended tests from previous runs (not the most recent runs).
```console
[2024-02-01 01:57:33.902814] wait.DEBUG: Check "extended-tests-ruby (release)" has the following status "completed" and conclusion "failure"
[2024-02-01 01:57:33.902824] wait.DEBUG: Check (extended-tests-ruby (release)) failed, marking resolve and failure
[2024-02-01 01:57:33.902834] wait.DEBUG: Check "extended-tests-ruby (debug)" has the following status "completed" and conclusion "cancelled"
[2024-02-01 01:57:33.902843] wait.DEBUG: Check (extended-tests-ruby (debug)) failed, marking resolve and failure
[2024-02-01 01:57:33.902853] wait.DEBUG: Check "extended-tests-jikesrvm" has the following status "completed" and conclusion "failure"
[2024-02-01 01:57:33.902862] wait.DEBUG: Check (extended-tests-jikesrvm) failed, marking resolve and failure
[2024-02-01 01:57:33.902871] wait.DEBUG: Check "extended-tests-julia" has the following status "completed" and conclusion "failure"
[2024-02-01 01:57:33.902881] wait.DEBUG: Check (extended-tests-julia) failed, marking resolve and failure
```
Those tests passed in the most recent run.

udesou:
@wks I've just re-requested a review in case you have further comments.

qinsoon:
> Update: **They are not equivalent.** If `initialize_collection` is not called, but GC is enabled (by default), then `alloc` will trigger GC and wait for GC to finish, which will never happen because GC threads are not spawn yet.

Did you observe this? If so, it sounds like a bug. Without `initialize_collection`, MMTk does not set `intialized` to `true`, and should not trigger a GC at all.

src/vm/collection.rs
```diff
@@ -139,4 +139,22 @@ pub trait Collection<VM: VMBinding> {
...
     }
+
+    /// Callback function to ask the VM whether GC is enabled or disabled, allowing or disallowing MMTk
+    /// to trigger garbage collection. When collection is disabled, you can still allocate through MMTk,
+    /// but MMTk will not trigger a GC even if the heap is full. In such a case, the allocation will
+    /// exceed MMTk's heap size (the soft heap limit). However, there is no guarantee that the physical
+    /// allocation will succeed, and if it succeeds, there is no guarantee that further allocation will
+    /// keep succeeding. So if a VM disables collection, it needs to allocate with careful consideration
+    /// to make sure that the physical memory allows the amount of allocation. We highly recommend
+    /// to have GC always enabled (i.e. that this method always returns false). However, we support
```
wks, line 150:
```suggestion
    /// to have GC always enabled (i.e. that this method always returns true). However, we support
```

src/vm/collection.rs
```diff
@@ -139,4 +139,22 @@ pub trait Collection<VM: VMBinding> {
...
+    /// this to accomodate some VMs that require this behavior. Note that this call also disables
+    /// explicit GCs (through handle_user_collection_request()). Note also that any synchronization
```
wks, line 152: Be more specific about the behavior of `handle_user_collection_request()`.

```suggestion
    /// this to accomodate some VMs that require this behavior. Note that
    /// `handle_user_collection_request()` calls this function, too.  If this function returns
    /// false, `handle_user_collection_request()` will not trigger GC, either. Note also that any synchronization
```

src/vm/tests/mock_tests/mock_test_allocate_with_re_enable_collection.rs
```diff
@@ -4,8 +4,9 @@ use crate::util::test_util::mock_method::*;
 use crate::util::test_util::mock_vm::*;
 use crate::AllocationSemantics;
 
-// This test allocates after calling initialize_collection(). When we exceed the heap limit, MMTk will trigger a GC. And block_for_gc will be called.
-// We havent implemented block_for_gc so it will panic. This test is similar to allocate_with_initialize_collection, except that we once disabled GC in the test.
+// This test allocates after calling initialize_collection(). When we exceed the heap limit for the first time, MMTk will not trigger GC since GC has been disabled
+// However, the second 1MB allocation will trigger a GC since GC is enabled again. And block_for_gc will be called.
+// We havent implemented block_for_gc so it will panic. This test is similar to allocate_with_initialize_collection, except that GC is disabled once in the test.
```
wks, line 9: This should be a doc comment (`///`, three slashes instead of two).  And there are also some minor Markdown formatting that we can conveniently fix now.

```suggestion
/// This test allocates after calling `initialize_collection()`. When we exceed the heap limit for the first time, MMTk will not trigger GC since GC has been disabled
/// However, the second 1MB allocation will trigger a GC since GC is enabled again. And `block_for_gc` will be called.
/// We haven't implemented `block_for_gc` so it will panic. This test is similar to `allocate_with_initialize_collection`, except that GC is disabled once in the test.
```

src/vm/tests/mock_tests/mock_test_allocate_with_re_enable_collection.rs
```diff
@@ -31,29 +37,23 @@ pub fn allocate_with_re_enable_collection() {
...
         || {
-            // This is actually redundant, as we defined block_for_gc for this test.
-            // This just demostrates that we can check if the method is called.
+            // This ensures that block_for_gc is called for this test, and that the second allocation
+            // does not trigger GC since we expect is_collection_enabled to be called three times.
             read_mockvm(|mock| {
-                assert!(mock.block_for_gc.is_called());
+                assert!(
+                    mock.block_for_gc.is_called() && mock.is_collection_enabled.call_count() == 3
+                );
```
wks, line 56: I suggest splitting it into two assertions so that when one of them fails, we know exactly which one failed.

```suggestion
                assert!(mock.block_for_gc.is_called());
                assert!(mock.is_collection_enabled.call_count() == 3);
```

src/vm/tests/mock_tests/mock_test_allocate_with_re_enable_collection.rs
wks: About `block_for_gc` panicking:  I think we can re-attach another `MockMethod` after the first two `alloc` so that it actually expects `block_for_gc` will be called exactly once at the third `alloc`.

wks (changes requested):

wks:
> > Update: **They are not equivalent.** If `initialize_collection` is not called, but GC is enabled (by default), then `alloc` will trigger GC and wait for GC to finish, which will never happen because GC threads are not spawn yet.
> 
> Did you observe this? If so, it sounds like a bug. Without `initialize_collection`, MMTk does not set `intialized` to `true`, and should not trigger a GC at all.

No.  This is purely hypothetical.

I see that `Space::acquire` has this statement:

```rust
        // Is a GC allowed here? If we should poll but are not allowed to poll, we will panic.
        // initialize_collection() has to be called so we know GC is initialized.
        let allow_gc = should_poll && self.common().global_state.is_initialized();
```

And this:

```rust
        if should_poll && self.get_gc_trigger().poll(false, Some(self.as_space())) {
            debug!("Collection required");
            assert!(allow_gc, "GC is not allowed here: collection is not initialized (did you call initialize_collection()?).");
            // ...
            VM::VMCollection::block_for_gc(VMMutatorThread(tls)); // We have checked that this is mutator
            unsafe { Address::zero() }
```

So if `initial_collection()` is not called, but `poll` says it should do GC, then it will panic.

It looks like the actual contract of `initialize_collection` is that it not only spawns GC threads, but it is also a precondition for allocating memory.  (p.s. This is fine for forking because the forked process (both parent and child) should immediately call `initialize_collection` again after forking, before making any allocation.)

wks (approved):
LGTM

wks:
The "Merge Check" failed for the reason of "Status checks failed".  That's strange.

qinsoon:
> The "Merge Check" failed for the reason of "Status checks failed". That's strange.

I think the first attempt to run Julia tests failed, which caused the merge check fail. When we reran the failed tests (Julia tests), the merge check did not get a rerun and its status won't change.

Issue #1078, opened 2024-01-30
Sanity GC is broken


angussidney:
In the latest version of MMTk (mmtk-core commit 42754a5f23, mmtk-openjdk commit 54f6688fe4), the Sanity GC is broken.

```
MTK_PLAN=Immix ~/honours/openjdk-builds/20240130-mmtk-slowdebug-42754a5
f2-upstream/bin/java -XX:+UseThirdPartyHeap -Xms512M -Xmx512M -jar /usr/share/benchmarks/dacap
o/dacapo-23.11-chopin.jar lusearch                                                            
[2024-01-30T00:31:21Z INFO  mmtk::memory_manager] Initialized MMTk with Immix (FixedHeapSize(5
36870912))                                                                                    
Using scaled threading model. 24 processors detected, 24 threads used to drive the workload, i
n a possible range of [1,2048]                                                                
Version: lucene 9.7.0 (use -p to print nominal benchmark stats)                               
[2024-01-30T00:31:25Z INFO  mmtk::mmtk] User triggering collection                            
[2024-01-30T00:31:25Z INFO  mmtk::util::sanity::sanity_checker] Sanity GC prepare             
thread 'thread 'thread 'thread 'thread '<unnamed>thread 'thread 'thread '<unnamed><unnamed><un
named><unnamed>' panicked at 'thread '<unnamed><unnamed><unnamed><unnamed>thread '' panicked a
t 'thread '' panicked at 'thread '' panicked at 'thread 'internal error: entered unreachable c
ode: `MutatorConfig::prepare_func` must not be called for the current plan.' panicked at 'thre
ad 'internal error: entered unreachable code: `MutatorConfig::prepare_func` must not be called
 for the current plan.'  

[...]

fatal runtime error: failed to initiate panic, error 
5

Aborted (core dumped)
```

[...]

PR #1079, opened 2024-01-30
Fix issues in sanity GC

qinsoon:
This PR fixes two issues in sanity GC.
* It fixes https://github.com/mmtk/mmtk-core/issues/1078 by removing the code to prepare/release mutator/collector, which seems unnecessary.
* It replaces the use of `ScanObjects` for cached root nodes with `ProcessRootNode`. Otherwise, the assertion `assert!(!self.roots())` will fail in `ScanObjectsWork::do_work_common`.

### DIFF
--- a/.github/workflows/merge-check.yml
+++ b/.github/workflows/merge-check.yml
@@ -25,3 +25,9 @@ jobs:
           checkInterval: 600
         env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Check result
+        if: ${{ steps.waitforstatuschecks.outputs.status != 'success' }}
+        uses: actions/github-script@v3
+        with:
+          script: |
+            core.setFailed('Status checks failed')
--- a/docs/userguide/src/tutorial/code/mygc_semispace/gc_work.rs
+++ b/docs/userguide/src/tutorial/code/mygc_semispace/gc_work.rs
@@ -78,8 +78,8 @@ impl<VM: VMBinding> ProcessEdgesWork for MyGCProcessEdges<VM> {
         }
     }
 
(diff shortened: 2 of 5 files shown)


wks:
`ScanObjectsWork::do_work_common` used to contain code for adding root nodes into the sanity checker.  Since https://github.com/mmtk/mmtk-core/commit/61d20e2dcd5b4743ef04a8118eb807bcd6f6e2e2, the processing of root nodes (the objects pointed by root edges from outside the object graph, i.e. pointed by stack slots and global variables) is moved to `ProcessRootNode`, and `ScanObjects` never processes any "root nodes".  Therefore, it is right to replace the use of `ScanObjects` with `ProcessRootNode`.  (I wonder why https://github.com/mmtk/mmtk-core/commit/61d20e2dcd5b4743ef04a8118eb807bcd6f6e2e2 changed the code related to sanity GC but didn't change the actual `SanityChecker`.)

I also think it is safe to remove the `roots()` method from the `ScanObjectsWork` trait, and the `roots` field from both `ScanObjects` and `PlanScanObjects`.  The `roots()` method was only used by the code related to sanity checking.  The fact that it is asserted to be false in `do_work_common` means it is not used now.

qinsoon:
> I also think it is safe to remove the `roots()` method from the `ScanObjectsWork` trait, and the `roots` field from both `ScanObjects` and `PlanScanObjects`. The `roots()` method was only used by the code related to sanity checking. The fact that it is asserted to be false in `do_work_common` means it is not used now.

I removed `roots()` in the trait and the fields in its implementations.

wks (approved):
LGTM

wks:
There are some example code in the documentation that still uses the `roots` parameter.

qinsoon:
binding-refs
OPENJDK_BINDING_REF=add-test-for-sanity-gc

PR #1083, opened 2024-02-19
Allow binding to implement GC trigger

qinsoon:
This PR allows bindings to create a delegated GC trigger. 
* Make `GCTriggerPolicy` public. Add a new type `SpaceStats` which simply wraps `Space`.
* Add `Collection::create_gc_trigger`. This will be called when the `gc_trigger` option is set to `Delegated`.


src/util/heap/gc_trigger.rs
```diff
@@ -65,7 +67,10 @@ impl<VM: VMBinding> GCTrigger<VM> {
     /// * `space`: The space that triggered the poll. This could `None` if the poll is not triggered by a space.
     pub fn poll(&self, space_full: bool, space: Option<&dyn Space<VM>>) -> bool {
         let plan = unsafe { self.plan.assume_init() };
-        if self.policy.is_gc_required(space_full, space, plan) {
+        if self
+            .policy
+            .is_gc_required(space_full, space.map(|s| SpaceStats::new(s)), plan)
```
wenyuzhao, line 72: Unrelated to this PR: Why is `Plan` safe to be exposed to the binding, whereas `Space` is not? (See L126 below in this file)

wenyuzhao (approved):
LGTM

src/util/heap/gc_trigger.rs
wenyuzhao: Looks like it's still possible to access spaces from a plan reference? https://github.com/mmtk/mmtk-core/blob/7081e868c393ace57dc364a6739228703e4b5321/src/plan/global.rs#L356

qinsoon: > Unrelated to this PR: Why is `Plan` safe to be exposed to the binding, whereas `Space` is not? (See L126 below in this file)

`Space` is not public before this PR, and I don't think we have a good reason to expose it. So I decided to wrap `Space` into a public type `SpaceStats`.

For `Plan`, I don't think we want to expose `Plan` either (at least some methods in `Plan` should not be public). Maybe we should check the visibility of `Plan`'s methods, and have another PR to tidy it up.

qinsoon: > Looks like it's still possible to access spaces from a plan reference?
> 
> https://github.com/mmtk/mmtk-core/blob/7081e868c393ace57dc364a6739228703e4b5321/src/plan/global.rs#L142-L155
> 
> https://github.com/mmtk/mmtk-core/blob/7081e868c393ace57dc364a6739228703e4b5321/src/plan/global.rs#L326-L334

`BasePlan` is not public, though it appears in a public trait. It is the same case for `Space` in `Plan::collection_required` before this PR. I am not sure what is Rust's rule on this. There is a lint `private_in_public`, but it seems Rust does not warn for these cases.

Issue #1084, opened 2024-02-19
Dead Links in documentation


Pat-Lafon:
Hello!

I apologize for not using a template but this issue seems unrelated to the templates provided. For fun, I was attempting to follow the NoGC porting guide for a toy language however I had trouble following the guide because some of the links were dead. It seems some parts of the repository at least relating to the dummy vm have been moved.

Concretely, atleast two links in https://docs.mmtk.io/portingguide/howto/nogc.html pointing to the dummy vm like "https://github.com/mmtk/mmtk-core/tree/master/vmbindings/dummyvm".

It might be worth setting up a dead link checker as part of your github actions but that is up to the maintainers' interest.

PR #1085, opened 2024-02-19
Check and fix broken links in docs

qinsoon:
This PR adds a Github workflow to check if there is any broken links in the docs. It also fixes existing broken links found by the workflow. This PR closes https://github.com/mmtk/mmtk-core/issues/1084.


udesou (approved):
LGTM.

Issue #594, opened 2022-05-16
Generational plans are slower than their non-generational counterparts


k-sareen:
During some performance evaluation, I have noticed that our generational GC plans (`GenCopy` and `GenImmix`) are slower than their non-generational counterparts (`SemiSpace` and `Immix`). `Immix` has the best GC time as well as overall benchmark execution time. This is quite weird as generational garbage collection should improve the STW time instead of increase it.

### Results

**Table 1: STW time for different MMTk GC plans**
-- | GenCopy (core_3dbdd7ae-jdk_3cc0d71) | GenImmix (core_3dbdd7ae-jdk_3cc0d71) | Immix (core_3dbdd7ae-jdk_3cc0d71) | MarkCompact (core_3dbdd7ae-jdk_3cc0d71) | NoGC (core_3dbdd7ae-jdk_3cc0d71) | SemiSpace (core_3dbdd7ae-jdk_3cc0d71)
-- | -- | -- | -- | -- | -- | --
avrora | 4.535 -1.83%, +1.86% | 3.118 -8.83%, +8.86% | 1.000 -1.79%, +1.82% | 3.774 -2.12%, +2.15% | 0.000 | 4.288 -1.99%, +2.02%
cassandra | 2.698 -3.38%, +3.49% | 1.936 -3.40%, +3.51% | 1.000 -3.28%, +3.39% | 5.089 -2.38%, +2.49% | 0.000 | 2.934 -2.54%, +2.65%
eclipse | 7.899 -0.47%, +0.48% | 6.293 -3.46%, +3.47% | 1.000 -0.49%, +0.49% | 9.403 -0.58%, +0.58% | 0.000 | 8.122 -0.43%, +0.43%
fop | 2.710 -2.24%, +2.33% | 3.491 -9.58%, +9.67% | 1.000 -2.98%, +3.07% | 4.497 -2.09%, +2.19% | 0.000 | 2.540 -2.14%, +2.24%
h2o | 4.539 -1.34%, +1.36% | 2.381 -2.07%, +2.08% | 1.000 -1.23%, +1.24% | 7.424 -1.16%, +1.18% | 0.000 | 3.991 -1.44%, +1.45%
jython | * | 1.564 -0.76%, +0.76% | 1.000 -0.46%, +0.46% | 5.559 -0.34%, +0.34% | 0.000 | 2.921 -0.41%, +0.41%
[...]

PR #1087, opened 2024-02-22
Allow nursery size to be proportional to the heap size

qinsoon:
This PR introduces different kinds of nursery size options, and by default, we use a proportion of the heap size as the min nursery. This PR should generally improve the generational plans' performance by triggering a full heap GC more promptly. This PR mitigates the issue identified in https://github.com/mmtk/mmtk-core/issues/594, but does not fully fix the problem.


qinsoon:
```rust
/// Return upper bound of the nursery size (in number of bytes)
    pub fn get_max_nursery_bytes(&self) -> usize {
        self.nursery.max.unwrap_or_else(|| {
            if !vm_layout().force_use_contiguous_spaces {
                DEFAULT_MAX_NURSERY_32
            } else {
                DEFAULT_MAX_NURSERY
            }
        })
    }
```

This check is removed. I am not sure about the consequences. Maybe I should change the default value for max nursery, depending on the VM layout.

qinsoon:
I quickly tried `h2o`, for which generational Immix plans were reported slower than full heap Immix in https://github.com/mmtk/mmtk-core/issues/594#issuecomment-1913851147. By changing the min nursery size, generational immix can perform better than Immix. I will do more thorough performance evaluation before making this PR ready for review.

qinsoon:
The performance results for now (using 2.5x minheap from `temurin-21-G1-n-5`): 

GenImmix
===

Normalized to Immix/master
genix
[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-02-26-Mon-032759&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&bmtime&|10&stickyix^1^null|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.ix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

Normalized to GenImmix/master
genix-normalized-to-master
[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-02-26-Mon-032759&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&bmtime&|10&genix^1^1|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.genix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

StickyImmix
===

Normalized to Immix/master
stickyix
[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-02-26-Mon-032759&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&bmtime&|10&genix^1^null|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.ix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

Normalized to StickyImmix/master
stickyix-normalized-to-master
[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-02-26-Mon-032759&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&bmtime&|10&stickyix^1^1|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.stickyix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

Issues I will look into
===
* `jython` and `tomcat` shows slowdown for GenImmix
* `h2`, `jython` and `lusearch` show slowdown for StickyImmix

Issues I am not going to look into
===
* Gen immix and sticky immix is slower than immix (the two issues above may contribute to this issue. But if those are resolved and gen/sticky immix are still slower, I do not plan to look further into the issues.)
* Immix and sticky immix ran into OOM for `tomcat`. 
* Immix ran into OOM for `h2`.

k-sareen:
Hmm. `{Gen,Sticky}Immix` are still really bad in comparison to `Immix` for lusearch. That doesn't make any sense. Looking back at my old results, it seems like yeah, that was the case back then as well.

qinsoon:
I fixed a few mistakes based on the previous evaluation. The following is the new results. I ran with 2.5x of the G1 heap size (`temurin-21-G1-n-5`) with 5 iterations.

The results show that the PR is an improvement for both GenImmix and StickyImmix in most benchmarks, 8% improvement for GenImmix and 3% improvement for StickyImmix (using 0.25x heap size as the min nursery). However, both plans are still performing worse than Immix (2% and 1.5% slower, respectively). There are still other performance issues with our generational plans that should be looked into.

The results also show slowdown in `tomcat`. The G1 min heap for `tomcat` is 19M, and the min heap for MMTk is 42M (GenImmix) and 59M (Immix) in my measurement. So with 2.5x of the G1 min heap, we are almost running at the min heap for Immix plans. With a proportion of the heap size as the min nursery, we always trigger full heap GCs for `tomcat`, while with a fixed 2M min nursery (as in `master`), we may occasionally do nursery GCs. However, this is a separate issue about MMTk's min heap with OpenJDK.

GenImmix
===
min-nursery-genix

[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-03-26-Tue-000913&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&time^time.other^time.stw&|10&stickyix^1^null|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.genix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

StickyImmix
===
min-nursery-stickyix

[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-03-26-Tue-000913&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&time^time.other^time.stw&|10&genix^1^null|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.stickyix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

src/util/options.rs
```diff
@@ -389,105 +399,199 @@ impl FromStr for AffinityKind {
...
     /// for 64-bit systems.
-    Bounded,
+    Bounded {
+        /// The lower bound of the nursery size in bytes. Default to [`DEFAULT_MIN_NURSERY`].
+        min: usize,
+        /// The upper bound of the nursery size in bytes. Default to [`DEFAULT_MAX_NURSERY`].
+        max: usize,
+    },
+    /// A bounded nursery that is porportional to the current heap size. By default, a proportional bounded
+    // nursery has a lower bound of 20% of the heap size, and has an upper bound of 100% of the heap size.
```
k-sareen, line 417:
```suggestion
    // nursery has a lower bound of 25% of the heap size, and has an upper bound of 100% of the heap size.
```

src/util/options.rs
```diff
@@ -109,19 +107,22 @@ pub const DEFAULT_MIN_NURSERY: usize = 2 << LOG_BYTES_IN_MBYTE;
...
-/// The default nursery space size.
-#[cfg(target_pointer_width = "32")]
-pub const NURSERY_SIZE: usize = 32 << LOG_BYTES_IN_MBYTE;
 /// The default min nursery size. This does not affect the actual space we create as nursery. It is
 /// only used in the GC trigger check.
 #[cfg(target_pointer_width = "32")]
 pub const DEFAULT_MIN_NURSERY: usize = 2 << LOG_BYTES_IN_MBYTE;
-const DEFAULT_MAX_NURSERY_32: usize = 32 << LOG_BYTES_IN_MBYTE;
+/// The default max nursery size for 32 bits.
+pub const DEFAULT_MAX_NURSERY_32: usize = 32 << LOG_BYTES_IN_MBYTE;
```
k-sareen, line 115: These numbers were based off of JikesRVM and an older DaCapo benchmark suite. I'm not sure if they reflect good defaults for the larger workloads + compressed pointers would also use these defaults instead of the 64-bit one (which might explain the performance issues you saw).

src/util/options.rs
```diff
@@ -718,13 +822,14 @@ options! {
...
-    /// both the upper and lower bounds. The nursery size can be set like 'Fixed:8192', for example,
-    /// to have a Fixed nursery size of 8192 bytes
-    // FIXME: This is not a good way to have conflicting options -- we should refactor this
-    nursery:               NurserySize          [env_var: true, command_line: true]  [|v: &NurserySize| v.min > 0 && v.max.map(|max| max > 0 && max >= v.min).unwrap_or(true)]
-        = NurserySize { kind: NurseryKind::Bounded, min: DEFAULT_MIN_NURSERY, max: None },
+    /// The nursery size for generational plans. It can be one of Bounded, ProportionalBounded or Fixed.
+    /// The nursery size can be set like 'Fixed:8192', for example,
+    /// to have a Fixed nursery size of 8192 bytes, or 'ProportionalBounded:0.2,1.0' to have a nursery size
+    /// between 20% and 100% of the heap size. You can omit lower bound and upper bound to use the default
+    /// value for bounded nurseryby by using '_'. For example, 'ProportionalBounded:0.1,_' sets the min nursery
```
k-sareen, line 829:
```suggestion
    /// value for bounded nursery by by using '_'. For example, 'ProportionalBounded:0.1,_' sets the min nursery
```

src/util/heap/gc_trigger.rs
```diff
@@ -104,6 +109,62 @@ impl<VM: VMBinding> GCTrigger<VM> {
         let plan = unsafe { self.plan.assume_init() };
```
k-sareen, line 109: You can replace the `let plan = ...` with the new function you implemented.

src/util/options.rs
k-sareen: Ah wait -- this was a bit confusing: for `estimate_virtual_memory_in_bytes`, if `force_use_contiguous_space` is `false` (which it is for compressed pointers and 32-bit VMs), then 32 MB is used as the max nursery size when creating the space. But then in the above code in `GCTrigger` the `DEFAULT_MAX_NURSERY` is used which is much larger than 32 MB for 64-bit VMs with compressed pointers.

src/util/options.rs
```diff
@@ -389,105 +399,199 @@ impl FromStr for AffinityKind {
...
-            },
-        }
+    /// Estimate the virtual memory needed for the nursery space. This is used during space creation.
+    pub fn estimate_virtual_memory_in_bytes(&self) -> usize {
+        let virtual_memory_bytes = match *self {
+            NurserySize::Bounded { min: _, max } => max,
+            // Just use the default max nursery size -- the nursery won't get larger than that.
+            NurserySize::ProportionalBounded { min: _, max: _ } => {
+                if !crate::util::heap::vm_layout::vm_layout().force_use_contiguous_spaces {
+                    DEFAULT_MAX_NURSERY_32
```
wks, line 438: If the space is confined (not using contiguous spaces), the return value will be `DEFAULT_MAX_NURSERY_32` which is 32MB.  But the `Options::nursery` can be set to a larger proportion and can be larger than this.  As @k-sareen commented in https://github.com/mmtk/mmtk-core/pull/1087#discussion_r1548945572, `GCTrigger::get_max_nursery_bytes` may return a larger size.

wks (commented):
One major problem is the virtual memory size of the nursery for confined address space.  On 32-bit systems, or 64-bit systems with compressed oops, the nurser of `GenCopy` and `GenImmix`  occupies a contiguous range separated from discontiguous range shared by other spaces.  The size is determined by `estimate_virtual_memory_in_bytes`, and it returns 32MB on systems with confined address space.  This size may be smaller than the desired nursery size, regardless whether the nursery size is fixed, bounded or proportionally bounded.

I think the 32MB limitation is only relevant if the nursery is contiguous and the address space is confined.

-   If the nursery can be discontiguous (this is possible in theory, but I don't know why we didn't implement that.  But the conceptual nursery for StickyImmix is indeed discontiguous.), we can just let it share the discontiguous address range (managed by FreeList) with other spaces.
-   If the address space is not confined, the code beneath `VMRequest::fixed_extent` will ignore the size and use `1TB` instead (in `VMRequest::common64bit`).

I don't think we want to make the nursery discontiguous for now, because that may require changes to many different places, but not impossible.  And it may affect performance.

The easiest way to make the design consistent is making two explicit rules (at least temporarily) that

1.  The nursery of `GenCopy` and `GenImmix` is always contiguous, and
2.  If the address space is confined, the nursery of `GenCopy` and `GenImmix` can not be larger than 32MB, regardless whether the nursery size is fixed, bounded or proportional.

We can document the second rule on `Options::nursery`.  This may not be ideal, but we don't need to change the 32MB size limit for now.

Alternatively, instead of making 32MB a constant, we can let the VM configure it using `VMLayout`.

A slightly more elegant but more difficult way to fix it is making use of `VMRequest::fraction` which is currently unused in the Rust MMTk, and probably needs testing.

src/util/options.rs
```diff
@@ -389,105 +399,199 @@ impl FromStr for AffinityKind {
...
-        })?;
-        let value = ns[1]
-            .parse()
-            .map_err(|_| String::from("Failed to parse size"))?;
-        Ok(NurserySize::new(kind, Some(value)))
+    /// Return true if the values are valid.
+    fn validate(&self) -> bool {
+        match *self {
+            NurserySize::Bounded { min, max } => min < max,
+            NurserySize::ProportionalBounded { min, max } => min < max && max <= 1.0f64,
```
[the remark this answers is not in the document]
wks: Maybe it is useful to allow `min == max`, because we currently don't have a way to express that the nursery size shall be a particular proportion (not a range) of the heap size.
```suggestion
            NurserySize::ProportionalBounded { min, max } => 0.0f64 < min && min <= max && max <= 1.0f64
```

src/util/options.rs
qinsoon: Yeah. You are right. I have changed this. I added a check: if we do not have pressure about virtual memory, the virtual memory for nursery can be as large as the max nursery (even if the max nursery is 100% heap size). Otherwise, it just panics.

Let me know what you think. @wks @k-sareen 

I will rerun the benchmarks for this change.

k-sareen: Panicking is pretty drastic, but if we don't have a nice way of saying "the rest of the heap" then maybe for the time being it is fine. We should file an issue though. Also is there any way to test for a 32-bit architecture? The JikesRVM binding doesn't support generational collectors and the MockVM doesn't stress test the GC much.

qinsoon: I think it might be better to panic instead of running with an unexpected value silently -- we could change this if the panic turns out to be an issue. 

What do you think we should test? The OpenJDK uses compressed pointer which tests most of the implementation for 32bits layout. I also added two unit tests that check the virtual memory we set for nursery.

k-sareen: Yes I mainly wanted to test the virtual memory and ensuring that full-heap GCs happen more promptly. I guess OpenJDK with compressed pointers and the unit tests are good enough.

k-sareen (approved):
LGTM. But we should make an issue for resolving the virtual memory issue. Also did you get the performance results in the end?

qinsoon:
> LGTM. But we should make an issue for resolving the virtual memory issue. Also did you get the performance results in the end?

I am running the benchmarks. I will post the results once I got it.

qinsoon:
GenImmix
===
min-nursery-genix-fix
[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-04-09-Tue-050105&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&time^time.other^time.stw&|10&stickyix^1^null|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.genix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

StickyImmix
===
min-nursery-stickyix-fix
[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-04-09-Tue-050105&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&time^time.other^time.stw&|10&genix^1^null|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.stickyix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

The improvement for GenImmix is 9.5% and the improvement for Stickyimmix is 3.1% (compared to the 8% and 3% improvement in the previous evaluation). However, one issue is that `h2` panics in the run, as it sets the heap size as 1700M, and we cannot guarantee another 1700M virtual memory for the 100% max nursery.

If the nursery size is set explicitly, I think panicking may be a good idea when we cannot satisfy the virtual memory for the nursery. But when we use the default nursery size, we would expect it run, rather than panicking.

Performance-wise, I think this PR achieved what we want.

wks:
> ... But when we use the default nursery size, we would expect it run, rather than panicking.

Yes.  If the user doesn't specify a nursery size, we shouldn't blame the user for providing a bad value.  One way to solve this is to provide another option "force_nursery_size" which defaults to false.  When true, it will panic if the give nursery size cannot be satisfied; when false, MMTk core will adjust the size to the nearest possible value.

qinsoon:
> > ... But when we use the default nursery size, we would expect it run, rather than panicking.
> 
> Yes. If the user doesn't specify a nursery size, we shouldn't blame the user for providing a bad value. One way to solve this is to provide another option "force_nursery_size" which defaults to false. When true, it will panic if the give nursery size cannot be satisfied; when false, MMTk core will adjust the size to the nearest possible value.

I am thinking about adding `NurserySize::Adaptive` and use that as the default value. It tries to use [25%, 100%) heap size as nursery, and when it is not possible, it will be adaptive and use whatever that works (e.g. use a smaller extent for nursery virtual memory, or use discontiguous nursery).

k-sareen:
The performance looks good (better than master for most things) but still not much better than Immix which is odd. But this PR does not need to solve all the issues at once.

I agree with @wks and your `Adaptive` idea.

qinsoon:
We checked with Steve. The reason that nursery was using a contiguous space is that we can do range check for mature objects in barriers. However, this is not important for now. I changed the code to use discontiguous nursery. I am running another round of evaluation.

qinsoon:
The results with a discontiguous nursery.

GenImmix
===
min-nursery-genix-discontig
[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-04-12-Fri-060021&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&time^time.other^time.stw&|10&stickyix^1^null|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.genix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

StickyImmix
===
min-nursery-stickyix-discontig
[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-04-12-Fri-060021&benchmark^buildstring^genix^invocation^iteration^ix^stickyix&time^time.other^time.stw&|10&genix^1^null|10&iteration^1^4|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-master.ms.s.c2.tph.probes_cp.probes_rustmmtk.probes_native.stickyix.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&)

The results are pretty similar to the previous run (with contiguous nursery), but slower. The improvement is 8.7% for GenImmix and 2.4% for StickyImmix, compared with 9.5% and 3.1% in the previous run. There is a consistent 0.8%/0.7% difference.

The PR currently uses discontiguous nursery. But I can change it to contiguous nursery, and only use discontiguous nursery if we may run out of virtual address. Just let me know if that's desired.

qinsoon:
@wks @k-sareen Can you review this PR again if you think it is necessary? I would like to get issues resolved for this PR, and get it merged.

wks (approved):
LGTM

PR #1089, opened 2024-02-29
Adding feature to query the fragmentation of immixspace

udesou:
Currently we can use the `count_live_bytes_in_gc` feature to check the total fragmentation of the heap, by counting the bytes of all objects that are scanned and dividing that by the total number of pages used by MMTk. 
However, to provide an accurate idea about fragmentation in terms of how much of the heap could be (opportunistically) defragmented, this PR does the counting only for the immix space, enabling to print the fragmentation rate of immixspace as `live_bytes_in_immixspace / (live_pages_in_immixspace << log_bytes_on_page)`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,6 +152,9 @@ malloc_counted_size = []
 # Count the size of all live objects in GC
 count_live_bytes_in_gc = []
 
+# Count the size of live objects in immixspace (the ones that can actually be opportunistically moved)
+count_live_bytes_immixspace = []
+
 # Workaround a problem where bpftrace scripts (see tools/tracing/timeline/capture.bt) cannot
 # capture the type names of work packets.
 bpftrace_workaround = []
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -557,6 +557,19 @@ pub fn live_bytes_in_last_gc<VM: VMBinding>(mmtk: &MMTK<VM>) -> usize {
     mmtk.state.live_bytes_in_last_gc.load(Ordering::SeqCst)
 }
 
+/// Return the percentage of fragmentation of the immixspace (e.g. 42.98 percent). To do that we count the size of every live object
+/// in the immixspace. Since MMTk accounts for memory in pages, we return the ratio between this number and
+/// the number of used bytes (according to the used pages by the immixspace).
+/// The value returned by this method is only updated when we finish tracing in a GC. A recommended timing
+/// to call this method is at the end of a GC (e.g. when the runtime is about to resume threads).
+#[cfg(feature = "count_live_bytes_immixspace")]
+pub fn fragmentation_rate_in_immixspace<VM: VMBinding>(mmtk: &MMTK<VM>) -> f64 {
+    mmtk.state
+        .fragmentation_rate_in_immixspace
+        .load(Ordering::SeqCst) as f64
+        / 100.0
+}
+
 /// Return the starting address of the heap. *Note that currently MMTk uses
 /// a fixed address range as heap.*
 pub fn starting_heap_address() -> Address {
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -217,6 +217,13 @@ impl<VM: VMBinding> crate::policy::gc_work::PolicyTraceObject<VM> for ImmixSpace
             debug_assert!(self.in_space(object));
             self.mark_lines(object);
         }
+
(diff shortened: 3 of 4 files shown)


src/memory_manager.rs
```diff
@@ -557,6 +557,19 @@ pub fn live_bytes_in_last_gc<VM: VMBinding>(mmtk: &MMTK<VM>) -> usize {
...
 }
 
+/// Return the percentage of fragmentation of the immixspace (e.g. 42.98 percent). To do that we count the size of every live object
+/// in the immixspace. Since MMTk accounts for memory in pages, we return the ratio between this number and
+/// the number of used bytes (according to the used pages by the immixspace).
+/// The value returned by this method is only updated when we finish tracing in a GC. A recommended timing
+/// to call this method is at the end of a GC (e.g. when the runtime is about to resume threads).
+#[cfg(feature = "count_live_bytes_immixspace")]
+pub fn fragmentation_rate_in_immixspace<VM: VMBinding>(mmtk: &MMTK<VM>) -> f64 {
+    mmtk.state
```
qinsoon, line 567: You can do something like this:
```rust
#[cfg(feature = "count_live_bytes_immixspace")]
pub fn fragmentation_rate_in_immixspace<VM: VMBinding>(mmtk: &'static MMTK<VM>) -> usize {
    use crate::policy::immix::ImmixSpace;
    use crate::policy::space::Space;
    let mut rate = None;
    mmtk.get_plan().for_each_space(&mut |space: &dyn Space<VM>| {
        if let Some(immix) = space.downcast_ref::<ImmixSpace<VM>>() {
            assert!(rate.is_none(), "There are multiple Immix spaces in the plan.");
            // Get the stats here from ImmixSpace
            rate = Some(immix.fragmentation_rate);
        }
    });
    rate.expect("No Immix space in the plan.")
}
```

In this case, you can simply store those values in `ImmixSpace` rather than `GlobalState`.

src/policy/immix/immixspace.rs
```diff
@@ -467,6 +479,62 @@ impl<VM: VMBinding> ImmixSpace<VM> {
...
+                match block.get_state() {
+                    BlockState::Marked => {
+                        panic!("At this point the block should have been swept already");
+                    }
+                    BlockState::Unmarked => {
+                        // Block is unmarked and cannot be reused (has no holes)
+                        dist.live_lines += Block::LINES;
+                    }
+                    BlockState::Reusable { unavailable_lines } => {
+                        dist.live_lines += unavailable_lines as usize;
```
wenyuzhao, line 509: Is this number equivalent to linearly scan the line mark table and manually check the line mark bytes? Probably better to add an assertion to double check. Same as L506.

Cargo.toml
```diff
@@ -50,6 +50,7 @@ static_assertions = "1.1.0"
 strum = "0.25"
 strum_macros = "0.25"
 sysinfo = "0.29"
+chrono = "*"
```
k-sareen, line 53: I'm being pedantic, sorry, but the dependencies list is sorted in alphabetical order to make it easier to scan dependencies. Would you mind moving `chrono` above to make it alphabetical?

udesou: No worries. I'm probably going to cleanup the code before merging, and might end up removing `chromo` altogether. But in the final version, if I leave it, I'll move it up to keep the alphabetical order.

k-sareen: If you just need chrono for timestamps and don't actually care about the date/time then you can just use UNIX time (in milliseconds) as the prefix to the print statement

src/policy/immix/immixspace.rs
udesou: I've added an assertion in the latest commit that compares the number of live lines by just iterating the lines in the block and checking the mark table with the number from the block state. Is that what you meant?

src/policy/immix/immixspace.rs
```diff
@@ -489,6 +591,98 @@ impl<VM: VMBinding> ImmixSpace<VM> {
...
+                        live_lines_from_block_state += Block::LINES;
+                    }
+                    BlockState::Reusable { unavailable_lines } => {
+                        dist.live_lines += unavailable_lines as usize;
+                        live_lines_from_block_state += unavailable_lines as usize;
+                    }
+                    BlockState::Unallocated => {}
+                }
+
+                assert_eq!(live_lines_in_table, live_lines_from_block_state);
```
qinsoon, line 642: I observed this assertion failed with left = 25 and right = 128.

PR #1090, opened 2024-03-13
Create separate work packet types for object tracing from mod buf

qinsoon:
Currently all the tracing in a nursery GC is done in `GenNurseryProcessEdges`, including objects reachable from root, and objects reachable from mod buf. This PR uses seperate work packet types for objects that are from mod buf. This helps us understand the nursery GC time better.

### DIFF
--- a/src/plan/generational/barrier.rs
+++ b/src/plan/generational/barrier.rs
@@ -10,9 +10,7 @@ use crate::vm::edge_shape::MemorySlice;
 use crate::vm::VMBinding;
 use crate::MMTK;
 
-use super::gc_work::GenNurseryProcessEdges;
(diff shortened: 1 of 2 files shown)


wks:
Alternatively, we can add a field in `struct GenNurseryProcessEdges` to identify their sources.  For example,

```rust
pub struct GenNurseryProcessEdges<VM: VMBinding, P: GenerationalPlanExt<VM> + PlanTraceObject<VM>> {
    plan: &'static P,
    base: ProcessEdgesBase<VM>,
    source: PacketSource,
}

enum PacketSource {
    Root,
    ModBuf,
    RegionModBuf,
}
```

PR #1094, opened 2024-03-20
Use SFTDenseChunkMap on 64bits when vm_space is enabled

qinsoon:
This PR resolves the issues found in https://github.com/mmtk/mmtk-julia/pull/143.
```console
[2024-03-19T05:23:33Z INFO  mmtk::policy::vmspace] Set [78624DC00000, 786258000000) as VM region (heap available range [20000000000, 220000000000))
thread '<unnamed>' panicked at 'start = 78624DC00000 + bytes = 171966464 should be smaller than space_start 380000000000 + max extent 2199023255552, index 28, table size 31', /home/runner/.cargo/git/checkouts/mmtk-core-89cdc7bf360cce7f/46b0abe/src/policy/sft_map.rs:202:21
```

Basically we should not use `SFTSpaceMap` when we have off-heap memory. Julia tries to set VM space at an address range outside our heap range, and that causes issues when we compute index for `SFTSpaceMap`. Using `SFTDenseChunkMap` will solve the issue.

Changes:
* Use `SFTDenseChunkMap` if `vm_space` is enabled.
* Add a range check for `SFTSpaceMap::has_sft_entry()`. The address/index computation is only correct if the address is in the range.
* Add a simple test case for https://github.com/mmtk/mmtk-julia/pull/143.

### DIFF
--- a/src/policy/sft_map.rs
+++ b/src/policy/sft_map.rs
@@ -202,7 +202,15 @@ mod space_map {
                 // Make sure the range is in the space
                 let space_start = Self::index_to_space_start(index);
                 assert!(start >= space_start);
-                assert!(start + bytes <= space_start + vm_layout().max_space_extent(), "The range of {} + {} bytes does not fall into the space range {} and {}, and it is probably outside the address range we use.", start, bytes, space_start, space_start + vm_layout().max_space_extent());
+                assert!(
+                    start + bytes <= space_start + vm_layout().max_space_extent(),
+                    "The range of {} + {} bytes does not fall into the space range {} and {}, \
+                    and it is probably outside the address range we use.",
+                    start,
+                    bytes,
+                    space_start,
+                    space_start + vm_layout().max_space_extent()
+                );
             }
 
             self.sft.get_unchecked(index).store(space);
--- a/src/policy/sft_map.rs
+++ b/src/policy/sft_map.rs
@@ -67,17 +67,23 @@ pub trait SFTMap {
 
 pub(crate) fn create_sft_map() -> Box<dyn SFTMap> {
     cfg_if::cfg_if! {
-        if #[cfg(all(feature = "malloc_mark_sweep", target_pointer_width = "64"))] {
-            // 64-bit malloc mark sweep needs a chunk-based SFT map, but the sparse map is not suitable for 64bits.
-            Box::new(dense_chunk_map::SFTDenseChunkMap::new())
-        } else if #[cfg(target_pointer_width = "64")] {
+        if #[cfg(target_pointer_width = "64")] {
+            // For 64bits, we generally want to use the space map, which requires using contiguous space and no off-heap memory.
+            // If the requirements do not meet, we have to choose a different SFT map implementation.
             use crate::util::heap::layout::vm_layout::vm_layout;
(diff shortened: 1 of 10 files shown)


k-sareen:
I presume 32-bit compressed pointer VMs are unaffected? (Selfish question given ART is 32-bit and has a bootimage)

qinsoon:
> I presume 32-bit compressed pointer VMs are unaffected? (Selfish question given ART is 32-bit and has a bootimage)

Right, 32 bits are not affected. I will update the title to make it clearer.

src/policy/sft_map.rs
```diff
@@ -261,7 +266,7 @@ mod space_map {
 
             let assert_for_index = |i: usize| {
                 let (start, end) = SFTSpaceMap::index_to_space_range(i);
-                debug!("Space: Index#{} = [{}, {})", i, start, end);
+                println!("Space: Index#{} = [{}, {})", i, start, end);
```
wks, line 269: `println` will print the string to stdout, and will be executed regardless of debug level.  Did you set this to `println` while debugging but forgot to change it back to `debug`?

src/policy/sft_map.rs
```diff
@@ -186,21 +194,15 @@ mod space_map {
...
                 let space_start = Self::index_to_space_start(index);
-                // FIXME: Curerntly skip the check for the last space. The following works fine for MMTk internal spaces,
-                // but the VM space is an exception. Any address after the last space is considered as the last space,
-                // based on our indexing function. In that case, we cannot assume the end of the region is within the last space (with MAX_SPACE_EXTENT).
-                if index != table_size - 1 {
-                    assert!(start >= space_start);
-                    assert!(start + bytes <= space_start + vm_layout().max_space_extent());
-                }
+                assert!(start >= space_start);
+                assert!(start + bytes <= space_start + vm_layout().max_space_extent(), "The range of {} + {} bytes does not fall into the space range {} and {}, and it is probably outside the address range we use.", start, bytes, space_start, space_start + vm_layout().max_space_extent());
```
wks, line 205: I am surprised that `cargo fmt` simply ignores this line.  It's likely that the string literal is too long.  After breaking the string literal so that it can fit into 100-character line width, `cargo fmt` formats it properly.

And it is more concise and more accurate to use the half-open interval notation `[start_inclusive, end_exclusive)`

```suggestion
                assert!(
                    start + bytes <= space_start + vm_layout().max_space_extent(),
                    "The range [{}, {}) ({} bytes) does not fall into the space range [{}, {}), \
                    and it is probably outside the address range we use.",
                    start,
                    start + bytes,
                    bytes,
                    space_start,
                    space_start + vm_layout().max_space_extent()
                );
 ```

wks (changes requested):

src/policy/sft_map.rs
qinsoon: Sounds good.

src/policy/sft_map.rs
qinsoon: I deliberately left it as `println!()`. Rust logs won't get printed as we do not initialise the logger. This is in the tests, and normally the output will be captured anyway.

Issue #1092, opened 2024-03-18
Segfault in unit tests on 32bits


qinsoon:
I recently saw this error in https://github.com/mmtk/mmtk-core/actions/runs/8258885514/job/22591917311?pr=1090. It should not be introduced in that PR. This test failed in 32 bits with Rust 1.71.1.

```console
++ cargo test --features vm_space,ro_space,code_space,vo_bit,is_mmtk_object,object_pinning,immix_non_moving,sticky_immix_non_moving_nursery,immix_smaller_block,immix_zero_on_release,sanity,analysis,nogc_lock_free,nogc_no_zeroing,single_worker,extreme_assertions,nogc_multi_space,work_packet_stats,malloc_counted_size,count_live_bytes_in_gc,bpftrace_workaround,malloc_jemalloc
   Compiling mmtk v0.23.0 (/home/runner/work/mmtk-core/mmtk-core)
    Finished test [unoptimized + debuginfo] target(s) in 10.84s
     Running unittests src/lib.rs (target/debug/deps/mmtk-e1b2bcfa71dea606)

running 419 tests
test build_info::tests::test_features ... ok
test build_info::tests::test_pkg_version ... ok
test build_info::tests::test_git_version ... ok
test build_info::tests::test_full_build_version ... ok
test util::address::tests::align_down ... ok
test util::address::tests::align_up ... ok
test util::address::tests::bit_or ... ok
test util::address::tests::bit_and ... ok
test util::conversions::tests::test_bytes_to_formatted_string ... ok
test util::address::tests::is_aligned_to ... ok
test util::conversions::tests::test_chunk_align ... ok
test util::conversions::tests::test_page_align ... ok
test util::heap::layout::byte_map_mmapper::tests::address_to_mmap_chunks ... ok
error: test failed, to rerun pass `--lib`

Caused by:
  process didn't exit successfully: `/home/runner/work/mmtk-core/mmtk-core/target/debug/deps/mmtk-e1b2bcfa71dea606` (signal: 11, SIGSEGV: invalid memory reference)
```

It did not show in which test we got segfault.

It may be related with #1091. But there is no prove that they are related except that they were revealed in about the same time in 32bits Rust 1.71.1.

qinsoon:
[...]

Issue #1091, opened 2024-03-15
RawMemoryFreeList can't get more space from mmap


wks:
I observed the following error when the CI was running for https://github.com/mmtk/mmtk-core/pull/1067

The GitHub Action link: https://github.com/mmtk/mmtk-core/actions/runs/8293161626/job/22695757005?pr=1067

```
---- util::raw_memory_freelist::tests::alloc_exhaust stdout ----
thread 'util::raw_memory_freelist::tests::alloc_exhaust' panicked at 'Can't get more space with mmap()', src/util/raw_memory_freelist.rs:202:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: mmtk::util::raw_memory_freelist::RawMemoryFreeList::mmap
             at ./src/util/raw_memory_freelist.rs:202:9
   3: mmtk::util::raw_memory_freelist::RawMemoryFreeList::raise_high_water
             at ./src/util/raw_memory_freelist.rs:196:9
   4: mmtk::util::raw_memory_freelist::RawMemoryFreeList::grow_list_by_blocks
             at ./src/util/raw_memory_freelist.rs:144:13
   5: mmtk::util::raw_memory_freelist::RawMemoryFreeList::grow_freelist
             at ./src/util/raw_memory_freelist.rs:134:9
   6: mmtk::util::raw_memory_freelist::tests::new_raw_memory_freelist
             at ./src/util/raw_memory_freelist.rs:279:9
   7: mmtk::util::raw_memory_freelist::tests::alloc_exhaust
             at ./src/util/raw_memory_freelist.rs:474:40
   8: mmtk::util::raw_memory_freelist::tests::alloc_exhaust::{{closure}}
             at ./src/util/raw_memory_freelist.rs:473:24
   9: core::ops::function::FnOnce::call_once
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/ops/function.rs:250:5
  10: core::ops::function::FnOnce::call_once
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[...]

PR #1095, opened 2024-03-21
Resolve issues for unit tests on 32 bits

qinsoon:
This PR resolves a few issues we recently saw for running unit tests in our CI.
* Fix an issue in the tests for byte map mmapper. In the clean up step, we do munmap for a larger area than what we actually mmapped in the test. This caused an issue that we may munmap regions that are used by other code. https://github.com/mmtk/mmtk-core/issues/1092 is caused by this -- the munmap after the tests unmaps the memory used by Rust, and caused Rust to segfault after the tests.
* Change the test address for mmap tests to reduce the chance that we may fail in mmap in the tests.
* ~~Make raw_memory_freelist only compiled in 64bits. This mitigates the issue we saw in https://github.com/mmtk/mmtk-core/issues/1091 for raw_memory_freelist tests.~~

### DIFF
--- a/src/util/heap/layout/byte_map_mmapper.rs
+++ b/src/util/heap/layout/byte_map_mmapper.rs
@@ -229,16 +229,16 @@ mod tests {
     #[test]
     fn ensure_mapped_1page() {
         serial_test(|| {
+            let pages = 1;
+            let start_chunk = ByteMapMmapper::address_to_mmap_chunks_down(FIXED_ADDRESS);
+            let end_chunk =
+                ByteMapMmapper::address_to_mmap_chunks_up(FIXED_ADDRESS + pages_to_bytes(pages));
+            let test_memory_size = (end_chunk - start_chunk) * MMAP_CHUNK_BYTES;
             with_cleanup(
                 || {
                     let mmapper = ByteMapMmapper::new();
-                    let pages = 1;
                     mmapper.ensure_mapped(FIXED_ADDRESS, pages).unwrap();
 
-                    let start_chunk = ByteMapMmapper::address_to_mmap_chunks_down(FIXED_ADDRESS);
-                    let end_chunk = ByteMapMmapper::address_to_mmap_chunks_up(
-                        FIXED_ADDRESS + pages_to_bytes(pages),
-                    );
                     for chunk in start_chunk..end_chunk {
                         assert_eq!(
                             mmapper.mapped[chunk].load(Ordering::Relaxed),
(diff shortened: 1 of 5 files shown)


src/util/heap/layout/byte_map_mmapper.rs
```diff
@@ -312,6 +315,7 @@ mod tests {
     #[test]
     fn protect() {
         serial_test(|| {
+            let test_memory_size = MMAP_CHUNK_BYTES * 2;
```
wks, line 318: Inside the `with_cleanup( || { ... })` block, it computes the number of pages to mmap separately (the expression `pages_per_chunk * 2`).  We have to manually make sure this size is the same as `MMAP_CHUNK_BYTES * 2` which is the definition of `test_memory_size`.

It is better if we can hoist the computations before the invocation of `with_cleanup`, like how you changed other test cases, and compute the number of pages to mmap and mprotect from the `test_memory_size`.  That will make the two values dependent on each other, and make the test case more robust.  For example:

```rs
let test_memory_bytes = MMAP_CHUNK_BYTES * 2;
let protect_memory_bytes = test_memory_bytes / 2; // Now this depends on test_memory_bytes
let test_memory_pages = test_memory_bytes >> LOG_BYTES_IN_PAGE as usize; // This also depends on test_memory_bytes
let protect_memory_pages = protect_memory_bytes >> LOG_BYTES_IN_PAGE as usize; // This also depends on test_memory_bytes
with_cleanup(|| {
    // ...
    mmapper.ensure_mapped(FIXED_ADDRESS, test_memory_pages).unwrap();
    mmapper.protect(FIXED_ADDRESS, protect_memory_pages);
    // ...
}, || {
    memory::munmap(FIXED_ADDRESS, test_memory_pages).unwrap();
})
```

src/util/heap/layout/byte_map_mmapper.rs
```diff
@@ -344,6 +348,7 @@ mod tests {
     #[test]
     fn ensure_mapped_on_protected_chunks() {
         serial_test(|| {
+            let test_memory_size = MMAP_CHUNK_BYTES * 2;
```
wks, line 351: The same for this one.

src/util/heap/layout/byte_map_mmapper.rs
qinsoon: Done

wks (approved):
LGTM

Issue #1098, opened 2024-03-25
VM_LAYOUT: References to mut static


wks:
In the Rust 2024 Edition it will be a compilation error to create `&` and `&mut` of `static mut` variables.  See https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-reference.html

Although the 2024 edition not yet released, the current stable Rust (1.77.0) warns against this.  See: https://doc.rust-lang.org/1.77.0/rustc/lints/listing/warn-by-default.html#static-mut-refs

This is causing some CI tests to fail.  See https://github.com/mmtk/mmtk-core/actions/runs/8417778415/job/23046912436?pr=1067

Currently, only `crate::util::heap::layout::vm_layout::VM_LAYOUT` is affected.

(Note: There is another use case of `static mut` in mock_test_doc_avoid_resolving_allocator.rs, but it seems unnecessary.)

A simple workaround is possible.  Change `&VM_LAYOUT` to `unsafe { &*addr_of!(VM_LAYOUT) }`, and the warning goes away.

But it reveals a problem.  `VM_LAYOUT` is global.  The function `fn vm_layout() -> &'static VMLayout` is used in 90+ different places in mmtk-core, and they all treat it as a global constant struct.  But `VM_LAYOUT` is actually mutable (by `set_custom_vm_layout`).  More precisely, it is immutable after initialization for each `MMTK` instance, as is explained in the doc comment:

```rust
    /// ...
    /// This function must be called before MMTk::new()
    pub(crate) fn set_custom_vm_layout(constants: VMLayout) {
```

The proper fix is moving the `VMLayout` structure to a field of `MMTK`, and make all functions that depend on it a method of `MMTK`

For example, `fn memory_manager::starting_heap_address() -> Address` and `fn memory_manager::last_heap_address() -> Address` should be methods of `MMTK` because each `MMTK` instance should have its own heap range.

As another example, `VMLayout` is frequently used when creating spaces and page resources.  Spaces and page resources belong to each `MMTK` instance.  Understandably, if `VMLayout` is a per-MMTK constant, it should be a parameter to the constructor of `MMTK`, and therefore it can be passed to the constructors of spaces and page resources.  We discussed refactoring the creation of spaces and page resources in another issue: https://github.com/mmtk/mmtk-core/issues/932

PR #1099, opened 2024-03-25
Fix warnings with the stable Rust toolchain

wks:
As getting a ref `&` of `static mut` variables will become a compilation error in the Rust 2024 Edition, we use `unsafe { &*addr_of!(VM_LAYOUT) }` to work around it.  See: https://github.com/mmtk/mmtk-core/issues/1098

Changed the `static mut` in mock_test_doc_avoid_resolving_allocator.rs to an ordinary local variable.

Initialize the thread-local variable `WORKER_ORDINAL` with a `const` block.  It avoids lazy initialization.  We also bumped the `rust-version` field in Cargo.toml to 1.71.1 because the semantics of the `const` block in the `thread_local!` macro was undocumented in 1.70.0.


Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository = "https://github.com/mmtk/mmtk-core"
 readme = "README.md"
 categories = ["memory-management"]
 keywords = ["gc", "garbage", "collection", "garbage-collection", "allocation"]
-rust-version = "1.70.0"
+rust-version = "1.71.1"
```
qinsoon, line 13: This breaks our MSRV policy. Our pinned Rust version is `1.71.1` and the MSRV should be at max `1.70`. We can think about bumping our pinned version (to 1.77).

wks: I bumped `rust-toolchain` to 1.77.0.

qinsoon (approved):
LGTM

PR #1100, opened 2024-03-26
Use mark sweep space as the non moving space

qinsoon:


k-sareen:
I have a commit here https://github.com/k-sareen/mmtk-core/commit/d12fe11c886e8bdad988d61bc4e6450515f0db6e which does the same but allows MarkSweepSpace to either be malloc or native-based.

qinsoon:
binding-refs
JIKESRVM_BINDING_REF=use-ms-as-nonmoving

qinsoon:
See https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/.60ChunkMap.60.20with.20ImmixSpace.20and.20MarkSweepSpace/near/468596380. The chunk map is causing issues: Immix may iterate and free mark sweep chunks using chunk map.

PR #1105, opened 2024-04-09
Bump dependencies after version 0.24

wks:
This PR updates Cargo crates dependencies and GitHub Actions dependencies after v0.24 was released, with two exceptions:

-   We do not update `criterion` because a later version unintentionally requires MSRV 1.74.  The problem has been fixed in the upstream criterion repo, but has not been released, yet.
-   We replaced `actions-rs/toolchain` with manual invocations of `rustup` because `actions-rs/toolchain` is no longer maintained.  For tasks that don't require a specific toolchain, we rely on `rustup` and `cargo` to automatically install the toolchain specified by the file `rust-toolchain` upon first invocation.

Also disabled Clippy checks on Darwin because the `clippy` program itself randomly crashes on Darwin.

### DIFF
--- a/.github/scripts/ci-style.sh
+++ b/.github/scripts/ci-style.sh
@@ -2,6 +2,15 @@
 
 export RUSTFLAGS="-D warnings -A unknown-lints"
 
+# --- Check format ---
+cargo fmt -- --check
+cargo fmt --manifest-path=macros/Cargo.toml -- --check
+
+# All versions of Clippy randomly crash on Darwin.  We disable Clippy tests for Darwin for now.
+if [[ $(uname) == "Darwin" ]]; then
+    exit 0
+fi
+
 # Workaround the clippy issue on Rust 1.72: https://github.com/mmtk/mmtk-core/issues/929.
 # If we are not testing with Rust 1.72, or there is no problem running the following clippy checks, we can remove this export.
 CLIPPY_VERSION=$(cargo clippy --version)
--- a/.github/workflows/api-check.yml
+++ b/.github/workflows/api-check.yml
@@ -29,6 +29,8 @@ jobs:
       - name: Install nightly Rust toolchain
         run: rustup toolchain install nightly
 
+      # Show the Rust toolchain we are actually using
+      - run: rustup show
       - run: cargo --version
       - run: cargo +nightly --version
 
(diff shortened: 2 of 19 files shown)


qinsoon:
Some dependencies need Rust 1.74 or newer, which does not comply with our MSRV 1.71.1. I don't think it is worth to bump MSRV for this reason. Maybe revert the dependency?

wks:
> Some dependencies need Rust 1.74 or newer, which does not comply with our MSRV 1.71.1. I don't think it is worth to bump MSRV for this reason. Maybe revert the dependency?

The dependency is pulled in by criterion 0.5.1.  `clap` is a command line parsing library, and I wonder why criterion needs that.  I'll see if we can disable that dependency by disabling some of its Cargo features.  But our "N-1" MSRV policy does allow us to bump MSRV up to 1.76, given that the version in rust-toolchain is 1.77.0.

wks:
Criterion depends on `clap` version 4, but `clap` started to require Rust 1.74 after version 4.4.9.  The Criterion developers noticed this problem, too (because Criterion intends to support MSRV 1.70), and locked the version of `clap` in a recent commit: https://github.com/bheisler/criterion.rs/commit/9f1db4a8314c874d5b18826900b34fc6673fa11d which has not been released, yet.

I don't see what new feature of Rust 1.74 Clap is using, but according to Clap's doc, it only intends to supports N-2 MSRV.  So it may be bumping MSRV for no reason.  While that's compatible with our N-1 MSRV policy, I don't think it's necessary for us to bump our `rust-version` in Cargo.toml because (1) we don't directly depend on clap, and (2) the next release of criterion is likely to revert to MSRV 1.70, and (3) we don't use new features of criterion 0.5, either.  So I reverted the criterion dependency to 0.4 and kept our `rust-version` unchanged.

qinsoon (approved):
LGTM

wks:
I also removed dependencies on `actions-rs/toolchain`.

-   For `minimal-test-core/${target-triple}/${version}`, I directly use `rustup` to install the needed toolchain.
-   For other tests, such as `check-public-api-changes` and `msrv`, I don't invoke `rustup` explicitly, but upon the first invocation of `rustup show`, `cargo --version` or any other commands related to `cargo`, it will automatically install the toolchain specified in the file `mmtk-core/rust-toolchain`.  The output of `rustup show` and `cargo --version` show that those tests are using the desired toolchains.

There is one small problem.  `rustup` gave a warning when trying to install the `i686-*` toolchain.  See https://github.com/mmtk/mmtk-core/issues/1107 for more details.  That would require much changes to the CI scripts and I am not going to do that change in this PR.  This PR focuses on removing outdated dependencies.

Issue #1103, opened 2024-04-05
Block cell size is sometimes zero in native MarkSweep


wks:
This bug is observed in the CI for a PR, but is reproducible on the master branch.  https://github.com/mmtk/mmtk-core/actions/runs/8548408967/job/23453824271?pr=1067

The bug manifests as GC hanging during the first GC, with 100% CPU usage.

The bug is related to the native MarkSweep plan and eager sweeping.  The "eager_sweeping" feature must be enabled to trigger this bug.  The "vo_bit" feature implies "eager_sweeping".

To reproduce this bug, we need to compile mmtk-openjdk with VO-bit enabled (which will enable "eager_sweeping").

```
export DEBUG_LEVEL=fastdebug
make CONF=linux-x86_64-normal-server-$DEBUG_LEVEL THIRD_PARTY_HEAP=$PWD/../mmtk-openjdk/openjdk MMTK_VO_BIT=1
```

One benchmark that is know to reproduce this error is `antlr` in `dacapo-2006-10-MR2.jar`, with the following command line:

```
MMTK_THREADS=128 MMTK_PLAN=MarkSweep /path/to/openjdk/build/linux-x86_64-normal-server-fastdebug/jdk/bin/java -XX:+UseThirdPartyHeap -server -XX:MetaspaceSize=100M -Xms40M -Xmx40M -jar dacapo-2006-10-MR2.jar antlr
```

When the error occurs, it will hang during *the very first GC*.

Here is a backtrace captured using GDB:

```
#0  0x00007ffff5414956 in mmtk::util::metadata::side_metadata::helpers::meta_byte_mask (metadata_spec=0x7ffff5a6a508)
    at src/util/metadata/side_metadata/helpers.rs:135
#1  0x00007ffff5459b5b in mmtk::util::metadata::side_metadata::global::{impl#0}::store_atomic::{closure#0}<u8> ()
    at src/util/metadata/side_metadata/global.rs:592
#2  0x00007ffff5458c78 in mmtk::util::metadata::side_metadata::global::SideMetadataSpec::side_metadata_access<u8, (), mmtk::util::metadata::side_metadata::global::{impl#0}::store_atomic::{closure_env#0}<u8>, mmtk::util::metadata::side_metadata::global::{impl#0}::store_atomic::{closure_env#1}<u8>> (self=0x7ffff5a6a508, data_addr=..., input=..., access_func=..., verify_func=...)
    at src/util/metadata/side_metadata/global.rs:483
[...]

PR #1106, opened 2024-04-10
Lock mark sweep block list during release

qinsoon:
This PR resolves the issue in https://github.com/mmtk/mmtk-core/issues/824#issuecomment-1558661794. It closes https://github.com/mmtk/mmtk-core/issues/824.
* Lock block lists in `AbandonedBlockLists` when they are accessed. This resolves the data race between `AbandonedBlockLists` and `SweepChunk` -- they both access blocks.
* Add a feature `ms_block_list_sanity`. This embeds a sanity block list `Vec<Block>` in `BlockList`, and checks if the actual block list with side metadata and linked list matches the sanity list. `ms_block_list_sanity` is enabled for `extreme_assertions`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -125,6 +125,9 @@ immix_smaller_block = []
 # Zero the unmarked lines after a GC cycle in immix. This helps debug untraced objects.
 immix_zero_on_release = []
 
+# Run sanity checks for BlockList in mark sweep.
+ms_block_list_sanity = []
+
 # Run sanity GC
 sanity = []
 # Run analysis
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,7 +141,7 @@ nogc_no_zeroing = ["nogc_lock_free"]
 single_worker = []
 
 # To run expensive comprehensive runtime checks, such as checking duplicate edges
-extreme_assertions = []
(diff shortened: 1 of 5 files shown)


wks:
We can add the assertions I mentioned in https://github.com/mmtk/mmtk-core/issues/1103 (not the early `exit(0)` of course) into this PR so that we can check if that problem persists after this PR.

qinsoon:
> We can add the assertions I mentioned in #1103 (not the early `exit(0)` of course) into this PR so that we can check if that problem persists after this PR.

I tried. The issue can still be reproduced. This PR does not fix #1103.

wks (commented):
If this PR fixes the issue, I am OK with it because the main part is just adding a lock.

But I think there is something fundamentally wrong with our implementation.  We can discuss more about it on Zulip.  My hypothesis is that SweepChunk doesn't really need to sweep the blocks, but only needs to count marked blocks.  Once we remove that, there should be no data race anymore.  That'll fundamentally address the race this PR is addressing.

Using `try_lock` on the sanity check should make it fail faster if a race occurs.

qinsoon:
I am closing this PR. See discussion in https://github.com/mmtk/mmtk-core/issues/1103, and the new PR https://github.com/mmtk/mmtk-core/pull/1112.

Issue #1097, opened 2024-03-22
StickyImmix nursery GC moves objects pointed by pinning roots


wks:
In default setting, if a nursery GC is triggered in StickyImmix, and an object is pointed by a pinning root (transitive or not), it will still attempt to copy the object.  This contradicts with the contract of "pinning roots".

# Details:

The VM binding can deliver lists of pinning roots via `RootsWorkFactory::create_process_pinning_roots_work(nodes: Vec<ObjectReference>)`.  The object graph edges pointing from roots (stack slots or global variables) to heap objects are represented as `ObjectReference` of the heap objects.  That's sufficient because the objects won't be moved, therefore there is no need to update the stack slots or global variables.

```rust
impl<VM: VMBinding, E: ProcessEdgesWork<VM = VM>, I: ProcessEdgesWork<VM = VM>>
    RootsWorkFactory<EdgeOf<E>> for ProcessEdgesWorkRootsWorkFactory<VM, E, I>
{
    // ...
    fn create_process_pinning_roots_work(&mut self, nodes: Vec<ObjectReference>) {
        crate::memory_manager::add_work_packet(
            self.mmtk,
            WorkBucketStage::PinningRootsTrace,
            ProcessRootNode::<VM, I, E>::new(nodes, WorkBucketStage::Closure),
        );
    }
    // ...
```

Here, `E` and `I` are type parameters.  `E` is used for non-pinning edges, while `I` is used for pinning edges.  In this case, `ProcessRootNode::<VM, I, E>::new(nodes, ...` tells the newly created `ProcessRootNode` work packet that the edges represented by the `nodes` vector shall be traced by `I`, and their transitively reached edges shall be traced by `E`.

For StickyImmix, the `I` and `E` come from the `TPProcessEdges` and `ProcessEdgesWorkType` members of `StickyImmixNurseryGCWorkContext`:

```rust
pub struct StickyImmixNurseryGCWorkContext<VM: VMBinding>(std::marker::PhantomData<VM>);
impl<VM: VMBinding> crate::scheduler::GCWorkContext for StickyImmixNurseryGCWorkContext<VM> {
    type VM = VM;
    type PlanType = StickyImmix<VM>;
    type ProcessEdgesWorkType = GenNurseryProcessEdges<VM, Self::PlanType>;
    type TPProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType>;
}
```

[...]

PR #1108, opened 2024-04-10
Allow roots to be pinned for StickyImmix nursery collections

k-sareen:
This PR allows StickyImmix to properly deal with transitive pinning trace. Before this PR, StickyImmix may move objects in the transitive pinning trace, as it simply uses `GenNurseryProcessEdges` which allows moving. With this PR, for transitive pinning trace, StickyImmix uses `GenNurseryProcessEdges<..., TRACE_KIND_TRANSITIVE_PIN>`. This PR fixes https://github.com/mmtk/mmtk-core/issues/1097.

### DIFF
--- a/src/plan/sticky/immix/gc_work.rs
+++ b/src/plan/sticky/immix/gc_work.rs
@@ -12,7 +12,8 @@ impl<VM: VMBinding> crate::scheduler::GCWorkContext for StickyImmixNurseryGCWork
     type VM = VM;
     type PlanType = StickyImmix<VM>;
     type DefaultProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType, DEFAULT_TRACE>;
-    type PinningProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType, TRACE_KIND_TRANSITIVE_PIN>;
+    type PinningProcessEdges =
+        GenNurseryProcessEdges<VM, Self::PlanType, TRACE_KIND_TRANSITIVE_PIN>;
 }
 
 pub struct StickyImmixMatureGCWorkContext<VM: VMBinding, const KIND: TraceKind>(
--- a/src/plan/sticky/immix/gc_work.rs
+++ b/src/plan/sticky/immix/gc_work.rs
@@ -1,4 +1,5 @@
 use crate::policy::gc_work::TraceKind;
+use crate::policy::gc_work::DEFAULT_TRACE;
 use crate::policy::gc_work::TRACE_KIND_TRANSITIVE_PIN;
 use crate::scheduler::gc_work::PlanProcessEdges;
 use crate::{plan::generational::gc_work::GenNurseryProcessEdges, vm::VMBinding};
@@ -6,11 +7,12 @@ use crate::{plan::generational::gc_work::GenNurseryProcessEdges, vm::VMBinding};
 use super::global::StickyImmix;
 
 pub struct StickyImmixNurseryGCWorkContext<VM: VMBinding>(std::marker::PhantomData<VM>);
+
 impl<VM: VMBinding> crate::scheduler::GCWorkContext for StickyImmixNurseryGCWorkContext<VM> {
     type VM = VM;
     type PlanType = StickyImmix<VM>;
-    type DefaultProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType>;
-    type PinningProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType>;
+    type DefaultProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType, DEFAULT_TRACE>;
+    type PinningProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType, TRACE_KIND_TRANSITIVE_PIN>;
 }
 
 pub struct StickyImmixMatureGCWorkContext<VM: VMBinding, const KIND: TraceKind>(
--- a/src/plan/generational/barrier.rs
+++ b/src/plan/generational/barrier.rs
@@ -3,6 +3,7 @@
 use crate::plan::barriers::BarrierSemantics;
 use crate::plan::PlanTraceObject;
 use crate::plan::VectorQueue;
+use crate::policy::gc_work::DEFAULT_TRACE;
 use crate::scheduler::WorkBucketStage;
 use crate::util::constants::BYTES_IN_INT;
 use crate::util::*;
@@ -45,7 +46,7 @@ impl<VM: VMBinding, P: GenerationalPlanExt<VM> + PlanTraceObject<VM>>
         let buf = self.modbuf.take();
         if !buf.is_empty() {
             self.mmtk.scheduler.work_buckets[WorkBucketStage::Closure]
-                .add(ProcessModBuf::<GenNurseryProcessEdges<VM, P>>::new(buf));
+                .add(ProcessModBuf::<GenNurseryProcessEdges<VM, P, DEFAULT_TRACE>>::new(buf));
         }
     }
 
@@ -54,7 +55,7 @@ impl<VM: VMBinding, P: GenerationalPlanExt<VM> + PlanTraceObject<VM>>
         if !buf.is_empty() {
             debug_assert!(!buf.is_empty());
             self.mmtk.scheduler.work_buckets[WorkBucketStage::Closure].add(ProcessRegionModBuf::<
-                GenNurseryProcessEdges<VM, P>,
+                GenNurseryProcessEdges<VM, P, DEFAULT_TRACE>,
             >::new(buf));
         }
     }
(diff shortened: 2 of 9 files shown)


k-sareen:
Again reinforcing that transitively pinning is still unsupported for the StickyImmix nursery GC (though it should be possible to support it). The main question is how to treat objects in the modbuffer

src/plan/sticky/immix/gc_work.rs
```diff
@@ -1,16 +1,20 @@
...
+    std::marker::PhantomData<VM>
+);
+
+impl<VM: VMBinding, const KIND: TraceKind> crate::scheduler::GCWorkContext for StickyImmixNurseryGCWorkContext<VM, KIND> {
     type VM = VM;
     type PlanType = StickyImmix<VM>;
-    type ProcessEdgesWorkType = GenNurseryProcessEdges<VM, Self::PlanType>;
-    type TPProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType>;
+    type ProcessEdgesWorkType = GenNurseryProcessEdges<VM, Self::PlanType, KIND>;
+    type TPProcessEdges = UnsupportedProcessEdges<VM>;
```
k-sareen, line 17: Will changing this break the Julia binding? @qinsoon @udesou

qinsoon: I think this won't work if a binding uses transitive pinning. Can we use `TRACE_KIND_FAST` here? Will that work?

k-sareen: That's what I had before in my fork, but then Eduardo mentioned that transitively pinning has unclear semantics for the modbuffer which I did not consider since ART does not do transitively pinning roots. I decided to make it explicit that transitively pinning is still unsupported instead of having a broken implementation in this PR.

k-sareen: My question was that if this will completely break the Julia binding since it's changed to "unsupported" instead of "just working" tm since Julia disables copying for nursery GCs (i.e. sets `PREFER_COPY_ON_NURSERY_GC` to `false`).

wks, line 17: That breaks Ruby, too.  Basically if you set `TPProcessEdges` to `UnsupportedProcessEdges`, any unmovable roots (delivered via `ProcessEdgesWorkRootsWorkFactory::create_process_pinning_roots_work`) will create an work packet that uses `UnsupportedProcessEdges` to trace the roots, which crashes.

Suggested changes:

```suggestion
    type ProcessEdgesWorkType = GenNurseryProcessEdges<VM, Self::PlanType, DEFAULT_TRACE>;
    type TPProcessEdges = GenNurseryProcessEdges<VM, Self::PlanType, TRACE_KIND_TRANSITIVE_PIN>;
```

And you can remove the generic parameter `<KIND>` from `StickyImmixNurseryGCWorkContext`, too.

I think the type member `type TPProcessEdges` is misleading.  It's not really related to "transitive pinning" per se.  The semantics is:

-    If an edge is updateable (i.e. its target is allowed to move), the GC should create a `ProcessEdgesWorkType` to process that edge.
-    If an edge cannot be updated (i.e. its target must not move), the GC should create a `TPProcessEdges` to process that edge.

What really happens is that when creating root packets, there are three options:
| method | process edges from root to object using | process edges between objects using |
|---|---|---|
| `create_process_edge_roots_work` | `ProcessEdgesWorkType` | `ProcessEdgesWorkType` |
| `create_process_pinning_roots_work` | `TPProcessEdges` | `ProcessEdgesWorkType` |
| `create_process_tpinning_roots_work` | `TPProcessEdges` | `TPProcessEdges` |

The kind of root selects the strategy of choosing which packets to use.  Non-transitive pinning roots process roots using `TPProcessEdges` but after that everything else can be processed with `ProcessEdgesWorkType`.  Transitive pinning is implemented with forcing all edges reachable from those roots to be traced using `TPProcessEdges`.

So the difference is just whether the edge can be updated.  I think there is much room for refactoring.

wks:
> Again reinforcing that transitively pinning is still unsupported for the StickyImmix nursery GC (though it should be possible to support it). The main question is how to treat objects in the modbuffer.
> 
> @wks Can you try this PR out for Ruby to see if it allows objects to move during the nursery GC?

I confirmed that with this PR (and the suggested changes I mentioned in the last comment), Ruby is able to pin object during nursery GC, and the GC is also able to move objects during GC.  I am not sure whether it works perfectly because the Ruby binding itself has not been fully tested against moving GC, yet.

src/plan/sticky/immix/gc_work.rs
k-sareen: Ohkay -- I should probably have tested that one-line change with ART before making this PR. It probably worked in my fork because I implemented `TPProcessEdges` instead of using the "unsupported" one.

This is _very_ confusing. I re-read the `create_process_pinning_roots_work` function and it's difficult to reason where the type parameters `I` and `E` are actually coming from and what they will be when using pinning roots vs transitively pinning roots. Your explanation makes sense but this is nowhere in the doc comment for the `ProcessEdgesWorkRootsWorkFactory` type.

k-sareen:
Done @wks

k-sareen:
But now with the fix, transitively pinning is _actually_ unsupported (given the issue regarding modbuffer) but MMTk will allow a VM developer to proceed regardless. This can blowup in glorious ways.

wks:
> But now with the fix, transitively pinning is _actually_ unsupported (given the issue regarding modbuffer) but MMTk will allow a VM developer to proceed regardless. This can blowup in glorious ways.

I think the *fields* of the objects in the modbuf should be considered non-pinning roots.  If an object is added to the modbuf, it simply means the object has been modified before the current GC, and some of its children may point to young objects.  Being in the modbuf doesn't mean the field of the object cannot be updated.  But if the same young object is also reachable from pinning roots, the object will be pinned; if it's reachable from transitive pinning roots, it and its children will all be pinned.

I think the current implementation of `ProcessModBuf` is fine:

```rs
        if is_nursery_gc() {
            // Scan objects in the modbuf and forward pointers
            let modbuf = std::mem::take(&mut self.modbuf);
            GCWork::do_work(
                &mut ScanObjects::<E>::new(modbuf, false, WorkBucketStage::Closure), // Put into the Closure bucket
                worker,
                mmtk,
            )
        }
```

As long as the `ScanObjects` work packet is put into the `Closure` bucket, it will be done *after* all the transitive pinning roots and their children are traced, and also after all non-transitive pinning roots are traced.

k-sareen:
Hm. I see. It'd be better if Julia could actually test this PR out to ensure that nothing breaks since neither ART nor Ruby use the transitively pinning roots.

k-sareen:
@qinsoon @wks The JikesRVM test failure was due to a timeout. Have we seen something like this before? I am not sure why this PR would cause a timeout.

qinsoon:
> @qinsoon @wks The JikesRVM test failure was due to a timeout. Have we seen something like this before? I am not sure why this PR would cause a timeout.

I think we have seen timeouts, but not frequently.

k-sareen:
JikesRVM just seems like it's a flaky CI. This code doesn't even touch the GCs that JikesRVM supports. I don't know why we get the failing instruction not in RVM address error. Should I investigate further?

k-sareen:
Idk why the ready-to-merge check timed out. I just re-ran it.

wks:
https://github.com/mmtk/mmtk-core/actions/runs/8657791364/job/23740512154
It looks like it timed out while running GenImmix.  This is different from https://github.com/mmtk/mmtk-core/issues/1103

It may be an unknown bug.

k-sareen:
> Idk why the ready-to-merge check timed out. I just re-ran it.

@qinsoon I think the ready-to-merge check is broken when we have the binding tests. They're called "extended-tests-<binding name>" instead of "<binding name>-binding-test".

k-sareen:
> https://github.com/mmtk/mmtk-core/actions/runs/8657791364/job/23740512154 It looks like it timed out while running GenImmix. This is different from #1103 which is caused by a race condition in MarkSweep.
> 
> It may be an unknown bug.

Yikes. OpenJDK shouldn't time out since this should only change how the pinning/transitively pinning roots work. I'll investigate locally when I have the time.

wks:
> OR This could be the result of the new change where we removed the coordinator. ...

Yes.  This is possible.  By removing the coordinator, the mechanism for triggering GC has changed.  There could be some corner cases that are not handled properly, but has not been discovered, yet.

k-sareen:
@qinsoon I'm not sure if the ready to merge check was fixed properly. It seems like it's still waiting on something

qinsoon:
> @qinsoon I'm not sure if the ready to merge check was fixed properly. It seems like it's still waiting on something

Not sure what is happening there. The problem of wrong ignored action names was fixed.

qinsoon:
> @qinsoon I'm not sure if the ready to merge check was fixed properly. It seems like it's still waiting on something

Yeah. https://github.com/mmtk/mmtk-core/pull/1119 should fix it. The action name for OpenJDK was changed due to the reused workflow. So the merge check was actually waiting for the OpenJDk tests to finish (as it was not ignored).

qinsoon (approved):
LGTM

Issue #1107, opened 2024-04-10
Rustup warns against installing i686 toolchains on x86_64


wks:
`rustup` gave a warning when trying to install the `i686-*` toolchain:

```
error: DEPRECATED: future versions of rustup will require --force-non-host to install a non-host toolchain.
warning: toolchain '1.71.1-i686-unknown-linux-gnu' may not be able to run on this system.
warning: If you meant to build software to target that platform, perhaps try `rustup target add i686-unknown-linux-gnu` instead?
```

What's happening is that the host is actually an x86_64 (using the OS image `ubuntu-22.04` from GitHub), although it is possible to run 32-bit programs using lib32 libraries.  The ideal solution is

-   installing an x86_64-unknown-linux-gnu toolchain, and
-   adding an i686-unknonwn-linux-gnu target, and
-   invoking `cargo build` and other cargo commands with the `--target` command line option.
    -   Alternatively we can override the default build target in `Cargo.toml`.

PR #1109, opened 2024-04-11
Use host toolchain and explicit target in CI script

wks:
This fixes a problem of installing i686 toolchain on an x86_64 host when running "Minimal tests - mmtk-core" on GitHub CI. Instead, we use the x86_64 toolchain, but add the i686 target and force all cargo commands to use that target using the `CARGO_BUILD_TARGET` environment variable.

Fixes: https://github.com/mmtk/mmtk-core/issues/1107


qinsoon (approved):
LGTM

PR #1110, opened 2024-04-11
Don't use `PROT_EXEC` with `mmap` on macos

eileencodes:
On macos I was running `./miniruby --mmtk -e 'puts "Hello world!"'` and seeing a panic:

```
thread '<unnamed>' panicked at /src/util/raw_memory_freelist.rs:202:9:
Can't get more space with mmap()
Abort trap: 6
```

On macos, you can't combine `PROT_EXEC` with `PROT_WRITE` and `PROT_READ` in `mmap`, it will return a permission denied error. It's not clear from the `Can't get more space with mmap()` that's what was happening, but I could see it when inspecting the `ret` from `mmap_fixed` in `dzmmap_noreplace`.

I found a thread on apple confirming that `PROT_EXEC` doesn't make sense for macos in this context. https://forums.developer.apple.com/forums/thread/740017

Note: there are still failing tests for `mmap` on macos. The main change with this PR is that it no longer panics when runnning a simple hello world with Ruby on macos. Previously the build wasn't finishing and would throw the following error:

```
process didn't exit successfully: `/mmtk-core/target/debug/deps/mmtk-ef0f14e36a03fbba` (signal: 11, SIGSEGV: invalid memory reference)
```

Now the tests do finish, with 25 failures.

### DIFF
--- a/src/util/memory.rs
+++ b/src/util/memory.rs
@@ -40,7 +40,7 @@ pub fn set(start: Address, val: u8, len: usize) {
 /// may corrupt others' data.
 #[allow(clippy::let_and_return)] // Zeroing is not neceesary for some OS/s
 pub unsafe fn dzmmap(start: Address, size: usize, strategy: MmapStrategy) -> Result<()> {
-    let prot = PROT_READ | PROT_WRITE | PROT_EXEC;
+    let prot = MMAP_PROT;
     let flags = libc::MAP_ANON | libc::MAP_PRIVATE | libc::MAP_FIXED;
     let ret = mmap_fixed(start, size, prot, flags, strategy);
     // We do not need to explicitly zero for Linux (memory is guaranteed to be zeroed)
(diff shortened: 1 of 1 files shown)


src/util/memory.rs
```diff
@@ -58,6 +58,12 @@ const MMAP_FLAGS: libc::c_int = libc::MAP_ANON | libc::MAP_PRIVATE | libc::MAP_F
 // MAP_FIXED is used instead of MAP_FIXED_NOREPLACE (which is not available on macOS). We are at the risk of overwriting pre-existing mappings.
 const MMAP_FLAGS: libc::c_int = libc::MAP_ANON | libc::MAP_PRIVATE | libc::MAP_FIXED;
 
+#[cfg(target_os = "linux")]
+const MMAP_PROT: libc::c_int = PROT_READ | PROT_WRITE | PROT_EXEC;
+#[cfg(target_os = "macos")]
+// PROT_EXEC cannot be used with PROT_READ on Apple Silicon
+const MMAP_PROT: libc::c_int = PROT_READ | PROT_WRITE;
```
caizixian, line 65: Thanks for the PR. I think for now getting something to work is better than nothing. But we do need to consider how MMTk handles code allocation for JIT.

MMTk might need to do this on macOS https://developer.apple.com/documentation/apple-silicon/porting-just-in-time-compilers-to-apple-silicon#Enable-the-JIT-entitlements-for-the-Hardened-Runtime

k-sareen: Thank you for the PR! I wonder if we should do things differently for Intel Macs vs Apple Silicon? Though maybe it makes more sense to have the same mechanism regardless of the underlying architecture. @qinsoon Do you have any thoughts?

qinsoon: > I wonder if we should do things differently for Intel Macs vs Apple Silicon?

Maybe. But our support for intel macOS is minimal as well so this is not a big concern for now.

qinsoon (approved):
LGTM

Issue #688, opened 2022-10-26
Mimalloc/marksweep issues


qinsoon:
This tracks leftover issues with our mimalloc allocator and the mark sweep plan that uses it.

### Zeroing

Currently zeroing happens when we allocate in a cell, that is very fine grained zeroing. We should use a more coarse grained zeroing. For example, we can do zeroing when we sweep the whole block, and when we sweep contiguous cells.

### Min heap

The min heap of the current implementation (based on https://github.com/mmtk/mmtk-core/pull/643) is:

| benchmark | marksweep | marksweep (without abandoning blocks in GC) | non-moving immix with 64K block | semispace |
|------------|-------------|----------------------------|--|--|
| dacapo2006-antlr | 37 | 38 | 20 | 10 |
| dacapo2006-luindex | 57 | 65 | 24 | 12 |
| dacapo2006-lusearch | 92 | 186 | 25 | 15 |
| dacapo2006-pmd | 97 | 97 | 65 | 52 |
| dacapo2006-xalan | 66 | 102 | 69 | 67 |

Mimalloc uses thread-local free lists. So instead of having a global free list for each size class, mimalloc allocator has a thread-local free list for each size class. In the original implementation, an allocator always allocate from its own free list, not others. And we do the same in our GC'd version: GC identifies live objects, and let each allocator sweep their dead cells. In this case, if an allocator has recyclable blocks for certain size classes and it does not use it frequently, other allocators cannot reuse those blocks, but have to allocate new blocks from the global memory.

We solved part of the problem by abandoning blocks in GC. So for each GC, each allocator will abandon their local block lists, and return the blocks back to the global pool (in the same way as if those threads die). So for new allocation, allocators will first attempt to get blocks from the global pool. This helps the min heap (from Column2 to Column1). 

However, in comparison with non-moving Immix, we are still worse. We will need to investigate more.

qinsoon:
I measured the min heap for our malloc mark sweep with mimalloc library. I am not doing bulk xor for mark bits in mark sweep in the measurement, as it is known to have [issues with detecting used pages](https://github.com/mmtk/mmtk-core/blob/master/src/policy/mallocspace/global.rs#L518-L524). Instead, I iterate through each object and check their mark bit.

The results are as below.

| benchmark | mark sweep with library mimalloc counting 4K page | mark sweep with library mimalloc counting 64K page |
|------------|-------------|----------------------------|
| dacapo2006-antlr | 10 | 16 |
[...]

PR #1112, opened 2024-04-12
Remove data race for marksweep block lists

qinsoon:
This PR refactors mark sweep.
* It removes the data race for accessing block lists, as discussed in https://github.com/mmtk/mmtk-core/issues/1103#issuecomment-2050805848.
* It fixes an issue that block state did not get reset before GC and caused marked blocks to stay alive forever: https://github.com/mmtk/mmtk-core/issues/688#issuecomment-2058193888
* It fixes an issue that empty blocks did not get released for lazy sweeping.

### DIFF
--- a/src/policy/marksweepspace/native_ms/block_list.rs
+++ b/src/policy/marksweepspace/native_ms/block_list.rs
@@ -164,12 +164,19 @@ impl BlockList {
         BlockListIterator { cursor: self.first }
     }
 
-    /// Sweep all the blocks in the block list.
-    pub fn sweep_blocks<VM: VMBinding>(&self, space: &super::MarkSweepSpace<VM>) {
+    /// Release unmarked blocks, and sweep other blocks in the block list. Used by eager sweeping.
+    pub fn release_and_sweep_blocks<VM: VMBinding>(&self, space: &super::MarkSweepSpace<VM>) {
         for block in self.iter() {
             if !block.attempt_release(space) {
                 block.sweep::<VM>();
             }
+        }
+    }
+
+    /// Release unmarked blocks, and do not sweep any blocks. Used by lazy sweeping
+    pub fn release_blocks<VM: VMBinding>(&self, space: &super::MarkSweepSpace<VM>) {
+        for block in self.iter() {
+            block.attempt_release(space);
         }
     }
 }
--- a/src/policy/marksweepspace/native_ms/block_list.rs
+++ b/src/policy/marksweepspace/native_ms/block_list.rs
@@ -1,4 +1,4 @@
-use super::Block;
+use super::{Block, BlockState};
 use crate::util::alloc::allocator;
 use crate::util::linear_scan::Region;
 use crate::vm::VMBinding;
@@ -167,6 +167,8 @@ impl BlockList {
     /// Release unmarked blocks, and sweep other blocks in the block list. Used by eager sweeping.
     pub fn release_and_sweep_blocks<VM: VMBinding>(&self, space: &super::MarkSweepSpace<VM>) {
         for block in self.iter() {
+            // We should not have unallocated blocks in a block list
+            debug_assert_ne!(block.get_state(), BlockState::Unallocated);
             if !block.attempt_release(space) {
                 block.sweep::<VM>();
             }
@@ -176,6 +178,8 @@ impl BlockList {
     /// Release unmarked blocks, and do not sweep any blocks. Used by lazy sweeping
     pub fn release_blocks<VM: VMBinding>(&self, space: &super::MarkSweepSpace<VM>) {
         for block in self.iter() {
+            // We should not have unallocated blocks in a block list
+            debug_assert_ne!(block.get_state(), BlockState::Unallocated);
             block.attempt_release(space);
         }
     }
--- a/src/policy/marksweepspace/native_ms/global.rs
+++ b/src/policy/marksweepspace/native_ms/global.rs
@@ -42,6 +42,26 @@ pub enum BlockAcquireResult {
 }
 
 /// A mark sweep space.
+/// 
+/// The space and each free list allocator own some block lists.
+/// A block that is in use belongs to exactly one of the block lists. In this case,
+/// whoever owns a block list has exclusive access on the blocks in the list.
+/// There should be no data race to access blocks. A thread should NOT access a block list
+/// if it does not own the block list.
+/// 
+/// The table below roughly describes what we do in each phase.
+/// 
+/// | Phase          | Allocator local block lists                     | Global abandoned block lists                 | Chunk map |
+/// |----------------|-------------------------------------------------|----------------------------------------------|-----------|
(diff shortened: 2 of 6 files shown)


src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -340,42 +369,42 @@ impl<VM: VMBinding> MarkSweepSpace<VM> {
...
-        {
-            if !block.attempt_release(self.space) {
-                // Block is live. Increment the allocated block count.
-                allocated_blocks += 1;
-            }
-        }
-        // Set this chunk as free if there is not live blocks.
-        if allocated_blocks == 0 {
+            .count();
+        if occupied_blocks == 0 {
```
wks, line 398: One problem with setting chunk state as `Free` in `Prepare` is that some blocks may become unoccupied *after* sweeping, and the chunks containing them may become free then.  Those chunks will not be marked as `Free` in this GC.

One way to solve the problem is that we keep the `SweepChunk` work packet, but we only let it identify fully free chunks (no blocks are marked), and don't let it actually release any blocks.  That should avoid the race on blocks.

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -340,42 +369,42 @@ impl<VM: VMBinding> MarkSweepSpace<VM> {
...
+            // Set this chunk as free if there is no live blocks.
             self.space.chunk_map.set(self.chunk, ChunkState::Free)
+        } else {
+            // Otherwise this chunk is occupied, and we reset the mark bit if it is on the side.
+            if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::LOCAL_MARK_BIT_SPEC {
+                for chunk in self.space.chunk_map.all_chunks() {
+                    side.bzero_metadata(chunk.start(), Chunk::BYTES);
```
wks, line 405: I found a strange thing.  We clear the mark bits, but we never clear block states before GC.  That means once a block becomes `BlockState::Marked`, it remains `BlockState::Marked` forever.

wks (commented):
I added some comments.  One strange thing is that once a Block becomes `BlockState::Marked`, it remain marked forever.  But I think the block state of all marked blocks should be reset to unmarked so that we can find blocks that actually have marked objects.

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -294,6 +314,15 @@ impl<VM: VMBinding> MarkSweepSpace<VM> {
         }
     }
 
+    pub fn end_of_gc(&mut self) {
+        let mut from = self.abandoned_in_gc.lock().unwrap();
+        let mut to = self.abandoned.lock().unwrap();
```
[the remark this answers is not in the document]
qinsoon: Good point. I changed to `get_mut()`.

src/policy/marksweepspace/native_ms/global.rs
qinsoon: I don't think this is a problem. The only use of the chunk map is to iterate through the blocks in use and bulk zeroing the side metadata. Doing it in the beginning of the current GC (prepare) or doing it at the end of the previous GC (release) should not make a difference.

src/policy/marksweepspace/native_ms/global.rs
qinsoon: Good observation. If what you said is true, it basically means we never release any block if it is ever marked. This could explain https://github.com/mmtk/mmtk-core/issues/688. I added code in this work packet to reset the block mark, and I will look further into this.

qinsoon: I pushed two changes. https://github.com/mmtk/mmtk-core/pull/1112/commits/19e87aa72cf0e8a2ee2bcf2d26f09fa693961912 resets the block state in prepare. https://github.com/mmtk/mmtk-core/pull/1112/commits/a920da3892c7c9b81189aee7a3ae0a6b7b7a4b71 also releases unmarked block for lazy sweeping (otherwise dacapo tests still run OOM).

The min heap results after these fixes are here: https://github.com/mmtk/mmtk-core/issues/688#issuecomment-2058193888. They are improved a lot.

wks:
Style check failed due to trailing spaces.  If you use VSCode, there are extensions for highlighting trailing spaces and removing them.

src/util/alloc/free_list_allocator.rs
```diff
@@ -428,43 +425,45 @@ impl<VM: VMBinding> FreeListAllocator<VM> {
...
     #[cfg(not(feature = "eager_sweeping"))]
     fn reset(&mut self) {
         trace!("reset");
         // consumed and available are now unswept
         for bin in 0..MI_BIN_FULL {
             let unswept = self.unswept_blocks.get_mut(bin).unwrap();
-            unswept.lock();
+
+            // We should have no unswept blocks here. Blocks should be either in avialable, or consumed.
+            debug_assert!(unswept.is_empty());
```
wks, line 437: This should only be true if `ABANDON_BLOCKS_IN_RESET == true`.  Otherwise, after this GC, one mutator may still have unswept blocks when another mutator has exhausted all available blocks and triggers a GC.

wks: Another minor spelling mistake: "avialable" -> "available".

src/policy/marksweepspace/native_ms/block_list.rs
```diff
@@ -164,14 +164,21 @@ impl BlockList {
         BlockListIterator { cursor: self.first }
     }
 
-    /// Sweep all the blocks in the block list.
-    pub fn sweep_blocks<VM: VMBinding>(&self, space: &super::MarkSweepSpace<VM>) {
+    /// Release unmarked blocks, and sweep other blocks in the block list. Used by eager sweeping.
+    pub fn release_and_sweep_blocks<VM: VMBinding>(&self, space: &super::MarkSweepSpace<VM>) {
```
wks, line 168: In `release_and_sweep_block`, `sweep` is called if the block `attempt_release` returns `false`.  But if a block is unallocated, `attempt_release` will still return `false`, and this function will still call `sweep`.  But it doesn't make sense to sweep an unallocated block.

I think it doesn't even make sense to call `attempt_release` on an unallocated block in the first place.  If we guarantee that we never add any unallocated block into any `BlockList`, we can just let `attempt_release` panic when it sees a block is unallocated.

It may be necessary to be a bit sloppy about block states if the race between `SweepChunk` and `BlockList::release_and_sweep_blocks` still exists because the block state may suddenly change.  Now that we removed that race, we may simply assume the block cannot be unallocated before we call `attempt_release`.

wks (commented):
The field `BlockList::lock` and its methods `BlockList::lock()` and `BlockList::unlock()` are unused.  We can remove them now.

qinsoon:
> The field `BlockList::lock` and its methods `BlockList::lock()` and `BlockList::unlock()` are unused. We can remove them now.

I kept the methods just in case we may need them in the future.

wks (approved):
LGTM

PR #1073, opened 2024-01-19
Rename CI tests to minimal/extended tests

qinsoon:
This PR refactors the CI workflows for https://github.com/mmtk/mmtk-core/issues/935, and run some OpenJDK tests as minimal tests.

PR #1119, opened 2024-04-17
Fix the openjdk test name in merge check

qinsoon:
The extended tests for OpenJDK reuses a workflow, so its action name is different. We did not use the correct action name since https://github.com/mmtk/mmtk-core/pull/1073. It manifested as the merge check was waiting for the OpenJDK tests to finish and checked its result before it finished.


wks (approved):
LGTM

PR #1122, opened 2024-04-17
Use to_address for SFT access

wks:
Changed convenient methods in `ObjectReference` so that if they use SFT, we use `self.to_address()` to get the address instead of `self.0` which returns the raw address.


qinsoon (approved):
LGTM

Issue #1125, opened 2024-04-19
ReferenceProcessor does not trace transitively discovered SoftReference instances


wks:
After retaining and scanning `SoftReference`, it generates `ProcessEdgesWork` work packets to trace the children of retained `SoftReference` instances.  During this time, more `SoftReference` instances will be discovered, but they will not be traced.  That will cause those transitively discovered `SoftReference` to contain un-updated reference to the children in the from-space, or crash MarkCompact due to those object not having forwarding pointers.

This bug was originally discovered when running CI tests for the PR removing `ObjectReference::NULL` when running `fop` in DaCapo 2006 using MarkCompact.  See: https://github.com/mmtk/mmtk-openjdk/actions/runs/8750247972/job/24016142656?pr=265

# The program

```java
import java.lang.ref.*;

class StrongNode {
    public WeakNode next;
    public StrongNode(WeakNode next) {
        this.next = next;
    }
}

class WeakNode extends SoftReference<StrongNode> {
    public WeakNode(StrongNode next) {
        super(next);
    }
}

public class TwoLevelDeadWeak {
    public static WeakNode mkWeak(Object object) {
        StrongNode strong1 = new StrongNode(null);
        WeakNode weak1 = new WeakNode(strong1);
        StrongNode strong2 = new StrongNode(weak1);
        WeakNode weak2 = new WeakNode(strong2);
        return weak2;
    }

    public static void main(String[] args) {
        Object obj = new Object();

        System.out.println("Making objects");
        WeakNode rooted = mkWeak(obj);
        System.out.println("Triggering GC...");
        System.gc();

        System.out.format("Rooted: %s%n", rooted);
        StrongNode myStrong = rooted.get().next.get();
        System.out.format("myStrong: %s%n", myStrong);

        System.out.println("Done.");
    }
}
```

# How to crash SemiSpace?

Run it with `MMTK_NO_REFERENCE_TYPE=true`, preferrably with stack trace and logs enabled.

```
export RUST_LOG=info,mmtk::util::reference_processor=trace
export RUST_BACKTRACE=1
MMTK_THREADS=1 MMTK_NO_REFERENCE_TYPES=false MMTK_PLAN=SemiSpace /path/to/openjdk/build/linux-x86_64-normal-server-slowdebug/jdk/bin/java -XX:+UseThirdPartyHeap -cp /path/to/the/program TwoLevelDeadWeak
```

It will print:

```
...
[...]

PR #1126, opened 2024-04-20
Ensure reference objects are either traced or cleared

wks:
After retaining soft references, we postpone the scanning of soft references until the transitive closure from retained soft references is fully expanded.  While tracing and scanning new objects reached from the referents of soft references, new soft references may be discovered, and they will be added to the reference processor.  By postponing the scanning, we ensure the newly discovered soft references are scanned, too.

To ensure we do not accidentally enqueue more objects than we should,  we no longer call `trace_object` when *scanning* soft, weak and phantom references.  Instead, we call `ObjectReference::get_forwarded_object` to get the forwarded references and referents.  We still call `trace_object` when *retaining* soft references, and when forwarding references in MarkCompact during the second transitive closure.

Fixes: https://github.com/mmtk/mmtk-core/issues/1125

### DIFF
--- a/src/policy/markcompactspace.rs
+++ b/src/policy/markcompactspace.rs
@@ -251,7 +251,9 @@ impl<VM: VMBinding> MarkCompactSpace<VM> {
             queue.enqueue(object);
         }
 
-        Self::get_header_forwarding_pointer(object)
+        let result = Self::get_header_forwarding_pointer(object);
+        assert!(!result.is_null(), "Object {object} does not have a forwarding pointer");
+        result
     }
 
     pub fn test_and_mark(object: ObjectReference) -> bool {
--- a/src/util/reference_processor.rs
+++ b/src/util/reference_processor.rs
@@ -81,15 +81,12 @@ impl ReferenceProcessors {
 
     // Methods for scanning weak references. It needs to be called in a decreasing order of reference strengths, i.e. soft > weak > phantom
 
+    pub fn retain_soft_refs<E: ProcessEdgesWork>(&self, trace: &mut E, mmtk: &'static MMTK<E::VM>) {
+        self.soft.retain::<E>(trace, is_nursery_gc(mmtk.get_plan()));
+    }
+
     /// Scan soft references.
     pub fn scan_soft_refs<E: ProcessEdgesWork>(&self, trace: &mut E, mmtk: &'static MMTK<E::VM>) {
-        // For soft refs, it is up to the VM to decide when to reclaim this.
-        // If this is not an emergency collection, we have no heap stress. We simply retain soft refs.
-        if !mmtk.state.is_emergency_collection() {
-            // This step only retains the referents (keep the referents alive), it does not update its addresses.
-            // We will call soft.scan() again with retain=false to update its addresses based on liveness.
-            self.soft.retain::<E>(trace, is_nursery_gc(mmtk.get_plan()));
-        }
         // This will update the references (and the referents).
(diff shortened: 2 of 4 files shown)


wks:
I marked this PR as ready for review.  This fixes some obvious problems described in https://github.com/mmtk/mmtk-core/issues/1125, but there are still crashes when running mmtk-openjdk with the reference processor in mmtk-core.  For example, running lusearch with StickyImmix with DaCapo Chopin with reference processing enabled will still crash due to SIGSEGV, regardless whether this PR is applied to the master branch or not.

As we discussed yesterday, we should enable binding-side reference processing for the OpenJDK binding ASAP.

src/util/reference_processor.rs
```diff
@@ -236,6 +240,18 @@ impl ReferenceProcessor {
         e.trace_object(referent)
     }
 
+    /// This function is called when forwarding the references and referents (for MarkCompact). It
+    /// -   enqueues the referent if not yet reached, so that its children will be visited and
```
qinsoon, line 244: When you say 'enqueue' here, do you mean the tracing queue? I would suggest avoiding using the word 'enqueue' here, as this method is actually used in a method named 'enqueue' which enqueues dead weak references to the reference queue. Just say it 'traces' the referent.

src/util/finalizable_processor.rs
```diff
@@ -140,6 +141,15 @@ pub struct Finalization<E: ProcessEdgesWork>(PhantomData<E>);
 
 impl<E: ProcessEdgesWork> GCWork<E::VM> for Finalization<E> {
     fn do_work(&mut self, worker: &mut GCWorker<E::VM>, mmtk: &'static MMTK<E::VM>) {
+        // Rescan soft and weak references at the end of the transitive closure from resurrected
```
qinsoon, line 144: * Do we need this `ReScanReferences` if `no_reference_type` is set to `false`? 
* Sentinels are supposed to prevent the current bucket being drained and prevent the next bucket being opened. However, this work packet is in `FinalRefClosure` which happens after `SoftRefClosure`. So it means `SoftRefClosure` was already drained and the next bucket was already activated. What does it mean by adding another packet as the sentinel here? It sounds like this should just be a normal packet in `SoftRefClosure`.

wks: > Do we need this ReScanReferences if `no_reference_type` is set to false?

If `no_reference_type` is set to true, we can omit that because all references would be treated as strong references.  If false, we will need that.  I'll make the packet added conditionally according to the value of `no_reference_type`.

> ... However, this work packet is in FinalRefClosure which happens after SoftRefClosure. ...

That's a typo.  I meant to add the work packet to `FinalRefClosure`.  Soft and weak references should be re-scanned after all objects reachable from resurrected objects are traced, because some `SoftReference` and `WeakReference` objects will be resurrected, too.

wks:
The Ruby test failed due to a known problem https://github.com/mmtk/ruby/issues/67  I'll re-run the Ruby test.

qinsoon (approved):
LGTM

src/policy/markcompactspace.rs
```diff
@@ -251,7 +251,12 @@ impl<VM: VMBinding> MarkCompactSpace<VM> {
             queue.enqueue(object);
         }
 
-        Self::get_header_forwarding_pointer(object)
+        let result = Self::get_header_forwarding_pointer(object);
+        assert!(
```
qinsoon, line 255: Probably use `debug_assert!` instead. This is hot code for object tracing.

wks: Done.

PR #1127, opened 2024-04-23
Dump process map on `mmap` failure and fix typos

k-sareen:


src/util/memory.rs
```diff
@@ -147,6 +147,8 @@ pub fn handle_mmap_error<VM: VMBinding>(error: Error, tls: VMThread) -> ! {
     match error.kind() {
         // From Rust nightly 2021-05-12, we started to see Rust added this ErrorKind.
         ErrorKind::OutOfMemory => {
+            #[cfg(target_os = "linux")]
+            info!("{}", get_process_memory_maps());
```
wks, line 151: 1.  It is better if we make the call regardless of the current OS, but we may implement the function platform-specific.  At this moment, we can make the function do nothing (or return empty string or "(unavailable)") for systems other than Linux.
2.  We should just print using `eprint!` or `eprintln!` (or do the printing inside `get_process_memory_maps`, or introduce a `print_process_memory_maps`).  It will be very helpful to have the information when the release build fails.  And if it fails, it doesn't matter whether we print more things to the console or less.

k-sareen: Sure

src/util/memory.rs
```diff
@@ -147,6 +147,7 @@ pub fn handle_mmap_error<VM: VMBinding>(error: Error, tls: VMThread) -> ! {
     match error.kind() {
         // From Rust nightly 2021-05-12, we started to see Rust added this ErrorKind.
         ErrorKind::OutOfMemory => {
+            eprintln!("{}", get_process_memory_maps());
```
wks, line 150: Well, I just realised that the out-of-memory event is not always lethal because the VM binding can implement `VM::VMCollection::out_of_memory`.  If the VM can recover from OOM (like OpenJDK), it is probably unnecessary to print the memory maps.  But if it crashes (like when OpenJDK generates something like `hs_err_pidXXXXX.log`), the ability to print mmap would be helpful.  I think if there is a call to `VM::VMCollection::out_of_memory` following this, we should move it to the default implementation of `Collection::out_of_memory` (because it panics anyway).

src/util/memory.rs
```diff
@@ -159,14 +160,18 @@ pub fn handle_mmap_error<VM: VMBinding>(error: Error, tls: VMThread) -> ! {
             if let Some(os_errno) = error.raw_os_error() {
                 // If it is OOM, we invoke out_of_memory() through the VM interface.
                 if os_errno == libc::ENOMEM {
+                    eprintln!("{}", get_process_memory_maps());
```
wks, line 163: This is OOM, too.  We may rely on the default impl of `Collection::out_of_memory` to print the mmap.

src/util/memory.rs
```diff
@@ -159,14 +160,18 @@ pub fn handle_mmap_error<VM: VMBinding>(error: Error, tls: VMThread) -> ! {
...
                     // Signal `MmapOutOfMemory`. Expect the VM to abort immediately.
                     trace!("Signal MmapOutOfMemory!");
                     VM::VMCollection::out_of_memory(tls, AllocationError::MmapOutOfMemory);
                     unreachable!()
                 }
             }
         }
-        ErrorKind::AlreadyExists => panic!("Failed to mmap, the address is already mapped. Should MMTk quanrantine the address range first?"),
+        ErrorKind::AlreadyExists => {
+            eprintln!("{}", get_process_memory_maps());
```
wks, line 172: Well, this is an error condition, and usually indicates a bug elsewhere in MMTk-core.  It is legit to directly print here.

src/util/memory.rs
k-sareen: Ideally we want more information (what address was MMTk trying to `mmap`), but our current internal API for all this is quite wonky. And it seems like `handle_mmap_failure` is only used for metadata when `mmap` can fail in normal space allocation as well

wks (commented):
I just realised in the case of OOM, we should let `Collection::out_of_memory` choose whether to print the mmap or silently recover from the error.  But if other mmap errors occur, we can still just print the error.

src/util/memory.rs
k-sareen: No this is different. This is an `mmap` error, not a simple OOM error. The simple OOM error is `AllocationError::HeapOutOfMemory`

wks (approved):
LGTM

Issue #1129, opened 2024-04-28
Improper but "silently succeeding" use of post write barrier


wks:
This is a sub-problem of https://github.com/mmtk/mmtk-core/issues/1038 which we decided to postpone after the MEP for removing `ObjectReference::NULL` (https://github.com/mmtk/mmtk-core/issues/1043).

MMTk's write barrier functions take `target: ObjectReference` as parameter. For example,

```rs
pub fn object_reference_write_pre<VM: VMBinding>(
    mutator: &mut Mutator<VM>,
    src: ObjectReference,
    slot: VM::VMEdge,
    target: ObjectReference,
) {
    mutator
        .barrier()
        .object_reference_write_pre(src, slot, target);
}
```

With https://github.com/mmtk/mmtk-core/issues/1043 merged, this form no longer works if `target` is `NULL` or any other non-reference values.  But real-world programs may overwrite a field previously holding an object reference with NULL or a non-reference value.

However, all bindings are still working.  That's because they transmute 0 to `ObjectReference` in C/C++ and pass them to Rust.  Currently, the object-remembering write barrier simply ignores the `slot` and the `target` fields, so it is unaware that the `ObjectReference` variable is holding an illegal value.  But we have to acknowledge that it is a bug and should be fixed.

We can simply change `target` to `Option<ObjectReference>`, with `None` representing a deleted edge.

However, `object_reference_write` will no longer work if `target` is `None` because MMTk core cannot write `NULL` or non-reference values to the slot.

```
pub fn object_reference_write<VM: VMBinding>(
    mutator: &mut Mutator<VM>,
    src: ObjectReference,
    slot: VM::VMEdge, // Edge::store(&self, target: ObjectReference) does not accept null target, which is intentional.
    target: ObjectReference,
);
```

wks:
An alternative solution is that `object_reference_write_{pre,post}` still have `target: ObjectReference` parameter, but the VM binding shall not call `object_reference_write_pre` if the old value in the slot is not an object reference, and shall not call `object_reference_write_post` if the new value written to the slot is not an object reference.

This scheme should work for the generational barrier because the generational barrier is an insertion barrier.  It only cares about newly added edges from old to young objects.

It should work for the snapshot-at-the-beginning (SATB) barrier, too.  It only tries to remember the old target if it exists.

It should work for naive RC, too.  If there is no old target, there is no object to decrement its RC; if there is no new target, there is no object to increment its RC, either.

[...]

PR #1130, opened 2024-04-29
Fix write barrier parameter type

wks:
Change the target type of write barrier functions to `Option<ObjectReference>`.  This allows VM bindings to use the write barrier API when storing NULL references or non-references values to slots.

Fixes: https://github.com/mmtk/mmtk-core/issues/1129

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -232,7 +232,7 @@ pub fn post_alloc<VM: VMBinding>(
 ///
 /// # Deprecated
 ///
-/// This form of write barrier has multiple issues.
+/// This function needs to be redesigned.  Its current form has multiple issues.
 ///
 /// -   It is only able to write non-null object references into the slot.  But dynamic language
 ///     VMs may write non-reference values, such as tagged small integers, special values such as
@@ -243,9 +243,10 @@ pub fn post_alloc<VM: VMBinding>(
 ///     type information, the offset (if it is an interior pointer), etc.  A write barrier is
 ///     associated to an assignment operation, which usually updates such information instead.
 ///
-/// VM bindings should use `object_reference_write_pre` and `object_reference_write_post` instead.
-/// Those functions leaves the actual write of the slot to the VM binding.
-#[deprecated = "Use `object_reference_write_pre` and `object_reference_write_post` instead"]
+/// We will redesign a more general subsuming write barrier to address those problems and replace
+/// the current `object_reference_write`.  Before that happens, VM bindings should use
+/// `object_reference_write_pre` and `object_reference_write_post` instead.
+#[deprecated = "Use `object_reference_write_pre` and `object_reference_write_post` instead, until this function is redesigned"]
 pub fn object_reference_write<VM: VMBinding>(
     mutator: &mut Mutator<VM>,
     src: ObjectReference,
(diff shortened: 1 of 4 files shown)


wks:
binding-refs
OPENJDK_BINDING_REF=fix/write-barrier-option
JULIA_BINDING_REPO=wks/mmtk-julia
JULIA_BINDING_REF=fix/write-barrier-option

wks:
It is now ready for review.

Binding PRs:

-   https://github.com/mmtk/mmtk-openjdk/pull/273
-   https://github.com/mmtk/mmtk-julia/pull/148

Only the OpenJDK and the Julia bindings are affected.  The master branch of the Ruby binding does not yet contain StickyImmix support, but I tested it offline.  JikesRVM and V8 currently do not use write barriers when using Rust MMTk.

src/memory_manager.rs
```diff
@@ -229,6 +229,23 @@ pub fn post_alloc<VM: VMBinding>(
...
+///     `null`, `undefined`, `true`, `false`, etc. into a field that previous contains an object
+///     reference.
+/// -   It relies on `slot.store` to write `target` into the slot, but `slot.store` is designed for
+///     forwarding references when an object is moved by GC, and is supposed to preserve tagged
+///     type information, the offset (if it is an interior pointer), etc.  A write barrier is
+///     associated to an assignment operation, which usually updates such information instead.
+///
+/// VM bindings should use `object_reference_write_pre` and `object_reference_write_post` instead.
+/// Those functions leaves the actual write of the slot to the VM binding.
+#[deprecated = "Use `object_reference_write_pre` and `object_reference_write_post` instead"]
```
qinsoon, line 248: I don't recall we have any discussion about removing the subsuming barrier. If I missed any previous discussion, just let me know.

If I understand you right, there are a few issues with this method right now:
1. `target: ObjectReference` cannot represent all the cases. If MMTk does the write for the binding, `target` needs to be able to represent any value the runtime may hold. `Option<ObjectReference>` is not sufficient either. We may want something like `target: Value`.
2. `slot: VM::VMEdge` cannot handle writing values that are not references. Though `Edge` (we want to rename it to `Slot`) is introduced for updating slots for copying GC, I think it is a valid case of reusing this trait to update slots in write barriers. We could extend the trait to allow it to store any value.

Is there any fundamental reason that we should remove the subsuming barrier instead of trying to correct it?

The reason why `write_pre/post` work is that we currently do not use the `slot` argument, and we do not care the `target` value if it is not a reference (so `Option<ObjectReference>` is sufficient).

qinsoon (changes requested):

src/memory_manager.rs
wks: We don't *remove* the subsuming barrier, but we need to *redesign* it.  That's why I said "*This form of* write barrier has multiple issues", not subsuming barriers in general.  We will replace this `object_reference_write` with another function, maybe with the same name, but different parameters.

There are ways to allow the VM customize the write operation of the subsuming barrier.  One way is letting the caller givt it a callback that actually writes to the field so that it can write the field in whatever way it likes.  The subsuming barrier effectively executes the pre and post barrier between the callback.

wks: I changed the comment and emphasized that we would redesign that function and replace the current one.

qinsoon (approved):
LGTM

PR #1132, opened 2024-05-06
Fix compiler errors and warnings for Rust 1.78.0

wks:
`ScanObjectsWork::make_another` is unused.  It was used when `ScanObjectsWork` was used for scanning node roots.  Now that node roots are scanned by the dedicated `ProcessRootNode` work packet, we can remove it.

`WorkCounter::get_base_mut` is never used.  All derived counters use `merge_val` to update all fields at the same time.

We use `Box::as_ref()` to get the reference to its underlying element. It fixes a compilation error related to CommonFreeListPageResource.  But we should eventually remove CommonFreeListPageResource completely as it is a workaround for mimicking the legacy design from JikesRVM that allow VMMap to enumerate and patch existing FreeListPageResource instances by registering them in a global list, which is not idiomatic in Rust.  See https://github.com/mmtk/mmtk-core/pull/934 and https://github.com/mmtk/mmtk-core/pull/953.

### DIFF
--- a/src/policy/sft_map.rs
+++ b/src/policy/sft_map.rs
@@ -361,7 +361,7 @@ mod dense_chunk_map {
 
             let space_name = unsafe { &*space }.name().to_string();
             // We shouldn't have this space in our map yet. Otherwise, this method is called multiple times for the same space.
-            assert!(self.index_map.get(&space_name).is_none());
+            assert!(self.index_map.contains_key(&space_name));
             // Index for the space
             let index = self.sft.len();
             // Insert to hashmap and vec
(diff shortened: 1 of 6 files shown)


qinsoon (approved):
LGTM

src/policy/sft_map.rs
```diff
@@ -361,7 +361,7 @@ mod dense_chunk_map {
 
             let space_name = unsafe { &*space }.name().to_string();
             // We shouldn't have this space in our map yet. Otherwise, this method is called multiple times for the same space.
-            assert!(self.index_map.get(&space_name).is_none());
+            assert!(self.index_map.contains_key(&space_name));
```
qinsoon, line 364: This should not be `!...contains_key()`. See the failed tests.

PR #1133, opened 2024-05-06
API migration guide.

wks:
An *API Migration Guide* section is added to the porting guide.

A symbolic link is created as `mmtk-core/API-MIGRATION.md` which can be easily discovered from the root of the repository.

### DIFF
--- a/docs/userguide/src/SUMMARY.md
+++ b/docs/userguide/src/SUMMARY.md
@@ -36,6 +36,7 @@
     - [Performance Tuning](portingguide/perf_tuning/prefix.md)
         - [Link Time Optimization](portingguide/perf_tuning/lto.md)
         - [Optimizing Allocation](portingguide/perf_tuning/alloc.md)
+- [API Migration Guide](api_migration.md)
 
 -----------
 
--- a/API-MIGRATION.md
+++ b/API-MIGRATION.md
@@ -1,0 +1,1 @@
+docs/userguide/src/api_migration.md
--- a/docs/userguide/src/api_migration.md
+++ b/docs/userguide/src/api_migration.md
@@ -1,0 +1,137 @@
+<!--
+The canonical location of this document is `mmtk-core/doc/userguide/src/api_migration.md`.
+It will be part of the MMTk User Guide, available online at <https://docs.mmtk.io/index.html>.
+A symbolic link is available at `mmtk-core/API-MIGRATION.md` to make this document easier to find in
+the repository.
+-->
+
+# API Migration Guide
+
+This document lists changes to the MMTk-VM API that require changes to be made by the VM bindings.
+VM binding developers can use this document as a guide to update their code to maintain
+compatibility with the latest release of MMTk.
+
+<!--
+Developers of mmtk-core:
+
+Add an item when making API-breaking changes, but edit existing item if the same API is changed
+again before the upcoming release.  No need to add item if a change only adds new API functions, or
+if it is source-compatible with the previous version so that VM binding code does not need to be
+changed.
+
+Check the current version in `Cargo.toml` before adding items.
+New items should be added to the section for the *upcoming* release.
+
+Use URLs of the pull requests to link to the relevant revisions.  Do not use commit hashes because
+they will change after squash-merging.
+
+Maintain a line width of 100 characters so that developers who read this file in an IDE or text
+editor can still read comfortably.
+-->
+
+## 0.25.0
+
+
+### `ObjectReference` is no longer nullable
+
+-   mmtk-core PRs:
+    -   <https://github.com/mmtk/mmtk-core/pull/1064>
+    -   <https://github.com/mmtk/mmtk-core/pull/1130>
+-   Binding pull requests for reference
+    -   <https://github.com/mmtk/mmtk-openjdk/pull/265>
+    -   <https://github.com/mmtk/mmtk-openjdk/pull/273> (for write barriers)
+
+#### MMTk-side API changes
+
+`ObjectReference` is no longer nullable.  `ObjectReference::NULL` and `ObjectReference::is_null()`
+have been removed.
+
+The write barrier APIs changed the type of the `target` parameters from `ObjectReference` to
+`Option<ObjectReference>`.  It affects:
+
+-   `mmtk::memory_manager::object_reference_write_pre`
+-   `mmtk::memory_manager::object_reference_write_post`
+
+The subsuming barrier `object_reference_write` is labelled as `#[deprecated]` and needs to be
+redesigned.  Before a replacement is introduced, VM bindings should use `object_reference_write_pre`
+and `object_reference_write_post` instead.
+
+#### Changes required for callbacks implemented by VM bindings
+
+`Edge::load()` now returns `Option<ObjectReference>`.  If the slot is holding a `null` reference or
+any non-reference values (such as small integer), it can just return `None`, and MMTk core will skip
+that slot.
+
+Several methods in `ObjectModel` are affected by this change because they usually involve converting
+`Address` to `ObjectReference`.
+
+-   `ObjectModel::copy` usually calls `copy_context.alloc_copy` to allocate the to-space copy, and
+    convert the returned address into an `ObjectReference`.  Since `alloc_copy` never returns 0, you
+    may use `ObjectReference::from_raw_address_unchecked`.
+-   `get_reference_when_copied_to` never passes 0 to the address parameter, so you may use
+    `ObjectReference::from_raw_address_unchecked`, too.
+-   `address_to_ref` is the inverse operation of `ref_to_address` which always converts from a
+    non-nullable `ObjectReference`.  You may use `ObjectReference::from_raw_address_unchecked`, too.
+
+Some methods of `ReferenceGlue` are affected, too.  *(Note, however, that if your VM binding is
+still using `ReferenceGlue` and the reference processor and finalization processor in mmtk-core, it
+is strongly recommended to switch to the `Scanning::process_weak_refs` method and implement weak
+reference and finalization processing on the VM side.)*
+
+-   `ReferenceGlue::get_referent` now returns `None` to indicate that the referent has been cleared,
+    and `ReferenceGlue::is_referent_cleared` is removed.
+-   `ReferenceGlue::clear_referent` now needs to be explicitly implemented.  However, if you were
+    using `Edge::store` to store NULL reference to the slot, you can no longer do it because
+    `Edge::store` now takes a non-nullable `ObjectReference` parameter.  You have to implement your
+    own method for storing NULL to a slot.
+
+#### Changes that affect native API
+
+Some public API functions that used to return `Option<ObjectReference>` still return
+`Option<ObjectReference>`.  Examples include:
+
+-   `memory_manager::get_finalized_object`
+-   `ObjectReference::get_forwarded_object`
+
+However, if you want to expose this API function to native (C/C++) programs, you can no longer
+return `ObjectReference::NULL`.  You may use `mmtk::util::api_util::NullableObjectReference` in
+`extern "C"` functions to gracefully encode `None` as 0 and pass it to native programs.
+
+
+### Instance methods of `ObjectReference` changed
+
+-   mmtk-core PR: <https://github.com/mmtk/mmtk-core/pull/1122>
+-   Binding PR for reference: <https://github.com/mmtk/mmtk-openjdk/pull/272>
+
+Some instance methods of `ObjectReference` now require a generic argument `<VM: VMBinding>` because
+they involve dispatching through the SFT and use the VM-specific `to_address` method under the hood.
+These methods include:
+
+-   `is_reachable`
+-   `is_live`
+-   `is_movable`
+-   `get_forwarded_object`
+-   `is_in_any_space`
+-   `is_sane`
+
+`ObjectReference::value()` is removed.  The user should call `ObjectReference::to_raw_address()` and
+then `Address::as_usize()` to get the underlying address as `usize`.
+
+
+### The GC controller (a.k.a. coordinator) is removed
+
+-   mmtk-core PR: <https://github.com/mmtk/mmtk-core/pull/1067>
+-   Binding PR for reference: <https://github.com/mmtk/mmtk-openjdk/pull/268>
+
+The GC controller thread is removed from MMTk core.
+
+`Collection::spawn_gc_thread` will no longer ask the binding to spawn the controller thread.  The
+binding should delete the code path for creating the controller thread.
+
+`GCWorker::run` and its wrapper `memory_manager::start_worker` now take ownership of the
+`Box<GCWorker>` instance instead of having a `&mut GCWorker` parameter.
+
+
+<!--
+vim: tw=100
+-->


docs/userguide/src/migration/prefix.md
```diff
@@ -0,0 +1,135 @@
...
+they will change after squash-merging.
+
+Maintain a line width of 100 characters so that developers who read this file in an IDE or text
+editor can still read comfortably.
+-->
+
+## 0.25.0
+
+
+### `ObjectReference` is no longer nullable
```
qinsoon, line 33: We would expect most people to just bump the MMTk version and attempt to fix on their own. When they struggle to do that, the migration guide needs to help them locate the issues they are facing. We should expect people to glance through the list and find related information for them. Thus I suggest using a more concise and more organised format. 

I propose something like below:

```markdown
### `ObjectReference` is no longer nullable: https://github.com/mmtk/mmtk-core/pull/1064
* Affected: everyone
* Expected changes:
  * `ObjectReference` cannot refer to a null reference.
  * Use `Option<ObjectReference>` (Rust) or `mmtk::util::api_util::NullableObjectReference` (for native API).
* Example migration: https://github.com/mmtk/mmtk-openjdk/pull/265

### Write barriers use `Option<ObjectReference>` for `target`: https://github.com/mmtk/mmtk-core/pull/1130
* Affected: bindings using generational plans
* Expected changes: 
  * Bindings should properly pass `target: Option<ObjectRefernece>` for `object_reference_write_pre/post`.
  * `object_reference_write` is marked as deprecated and will be redesigned. Use `object_reference_write_pre/post` instead, for now.
* Example: https://github.com/mmtk/mmtk-openjdk/pull/273

### Instance methods of `ObjectReference` changed: https://github.com/mmtk/mmtk-core/pull/1122
* Affected: everyone
* Expected changes:
  * Some methods in `ObjectReference` expect a type parameter `<VM: VMBinding>`.
* Example: https://github.com/mmtk/mmtk-openjdk/pull/

### The GC controller (a.k.a. coordinator) is removed: https://github.com/mmtk/mmtk-core/pull/1067
* Affected: everyone
* Expected changes:
  * Code related with the GC controller is removed, and the bindings no longer need to spawn and run a GC controller.
* Example: https://github.com/mmtk/mmtk-openjdk/pull/268
```

qinsoon (commented):
I made some changes to the PR:
1. Add a CI check to make sure migration guide is updated if we have API changes.
2. Remove the symbolic link (symbolic links don't work for OS like windows). Instead, refer to the migration guide in README.
3. Move migration guide to a folder so we may extend it and split it based on versions.

I proposed using a more concise and more organised format for the migration guide.

wks:
> Add a CI check to make sure migration guide is updated if we have API changes.

This should give a warning instead of failing immediately.  Some API changes add new methods (such as adding new APIs), while others are source compatible (such as adding a type parameter `<T>` which can be automatically inferred).  Those changes don't need the binding to be changed.

docs/userguide/src/migration/prefix.md
wks: When the binding developers find an API changed, they want to know why this change is made, and more importantly, *how* they should change their existing usage.  The problem with this format is that it is overly concise.  For example, if we mention "`Edge::load()` now returns `Option<ObjectReference>`", the reader will think "What should it return after this change?" That's where the long explanation is needed.

But you are right.  Most people expect to quickly locate the related information.  I think we can add the "expected changes" list to the beginning of each item as a concise TL;DR-style hint, while the reader can continue reading if they find relevant.

qinsoon:
> > Add a CI check to make sure migration guide is updated if we have API changes.
> 
> This should give a warning instead of failing immediately. Some API changes add new methods (such as adding new APIs), while others are source compatible (such as adding a type parameter `<T>` which can be automatically inferred). Those changes don't need the binding to be changed.

Sounds good. I added the new check to the ignore list of `merge-check`.

docs/userguide/src/migration/prefix.md
qinsoon: I prefer a list of changes in the migration guide, such as https://getbootstrap.com/docs/5.0/migration/. Besides, the links to PRs should include all the information which people can continue reading to know more details. 

However, we can maintain a list with optional links to detailed description for a PR, like https://v3-migration.vuejs.org/breaking-changes/. e.g. 
```markdown
### `ObjectReference` is no longer nullable: https://github.com/mmtk/mmtk-core/pull/1064
* Affected: everyone
* Details: See here (this is optional)
* Expected changes:
  * `ObjectReference` cannot refer to a null reference.
  * Use `Option<ObjectReference>` (Rust) or `mmtk::util::api_util::NullableObjectReference` (for native API).
* Example: https://github.com/mmtk/mmtk-openjdk/pull/265
```

qinsoon: > For example, if we mention "`Edge::load()` now returns `Option<ObjectReference>`", the reader will think "What should it return after this change?" That's where the long explanation is needed.

We state that `ObjectReference` cannot be null any more, ad `Option<ObjectReference>` should be used. I think this should be enough for users to know how to deal with a return value with `Option<ObjectReference>`. Plus, we have updated comments for the function `Edge::load()` for details.

wks: I made another version in the bullet points style: https://github.com/wks/mmtk-core/blob/0db7a487a28d54a22949a81033878d862fa11081/docs/userguide/src/migration/prefix.md

It does look clearer at first glance, but I feel it is not right.  It looks good as a summary of API changes, but that's probably not what VM binding developers need.  When a binding developer pulls the `mmtk-core` repository, the binding will fail to compile, and the compiler and the IDE will highlight functions that have problems.  That will be the same set of functions we listed in the bullet points version.  I don't expect the VM binding developers to read through the bullet points in the Migration Guide because the compile errors will tell them what needs to be changed.  If the change is obvious, the binding developer will just make the change and fix the compilation error.

The problem is, what if the binding developer finds one of the functions no longer compiles, and but does not know how to update the old code to adapt to the change?  They will eventually find the answer by reading the API docs of affected functions.  But it is easier if they can go to the Migration Guide, and find detailed descriptions about the old behavior and the new behavior.  The Migration Guide should tell the reader what to do if the binding wants to do this or that.

I find [the migration guide of the sysinfo crate](https://github.com/GuillaumeGomez/sysinfo/blob/master/migration_guide.md) quite helpful, especially the following part:

> A new `System::refresh_memory_specifics` method and a new `MemoryRefreshKind` type were added, allowing you to control whether you want both RAM and SWAP memories to be updated or only one of the two. This change was needed because getting SWAP information on Windows is very slow.

MMTk used `RefreshKind::new().with_memory()` in `get_system_total_memory`.  After [bumping the version](https://github.com/mmtk/mmtk-core/issues/1105), it no longer compiles.  I figured out how to make the change by reading that paragraph which told me I should switch to `MemoryRefreshKind` and also decide if I want the RAM size or the swap size. Because MMTk wants the physical memory size, I changed it to `MemoryRefreshKind::new().with_ram().

qinsoon: > The problem is, what if the binding developer finds one of the functions no longer compiles, and but does not know how to update the old code to adapt to the change?

Exactly. I think most developers won't go straight to the migration guide, but instead they just pull the new version and try to fix compilation errors. They will turn to the migration guide when they find that they cannot fix some issues themselves. Hopefully the migration guide can help them locate where the compilation error is from, and provide them a link to an example -- that should be enough for most cases. So I suggest the migration guide to be concise and organized so that they can easily locate the issue they are searching for, and they can find the solution with an example, or with detailed migration description and the related PR.

wks: I made a third version https://github.com/wks/mmtk-core/pull/new/feature/api-change-log3

It still has the bullet points style, but is not as concise as the [second version](https://github.com/wks/mmtk-core/blob/0db7a487a28d54a22949a81033878d862fa11081/docs/userguide/src/migration/prefix.md), and contains more explanation.  For the convenience of navigation, it now organises the changes in a hierarchy of (1) type, trait or module, and then (2) method.  It is probably only needed for big changes like removing `ObjectReference`.

wks:
I changed the style so that the first two levels of the lists form an outline of the changes, and the third level of the list contains more details and explanations.

I also use JavaScript to let the reader collapse and expand the third level of the lists.  When collapsed, the list will look like a terse outline.  This functionality is only available in the rendered HTML version.

I also moved the template from the comments into a dedicated page.

qinsoon:
This is what the current form looks like for one API breaking PR, when we hide 'collapsible' sections.

api-migration

Users need to be able to locate the information they need by looking at the list. If we have 10 entries like above in the list, the users won't be able to easily find the entry that they are looking for. You can just try this yourself: copy and paste that one entry 10 times, and see if you can easily locate the title for each entry. I find it hard. If users cannot even find the titles easily, they will have a hard time to find the right entry for their needs.

I suggest something like this.

```markdown
### `ObjectReference` is no longer nullable

**TL;DR** `ObjectReference` can no longer represent a NULL reference.  Some methods of
`ObjectReferences` and write barrier functions are changed.  VM bindings need to re-implement
methods of the `Edge`, `ObjectModel` and `ReferenceGlue` traits.

COLLAPSIBLE START
All the detailed description goes here
COLLAPSIBLE END
```

We can use something like https://crates.io/crates/mdbook-admonish for [collapsible blocks](https://tommilligan.github.io/mdbook-admonish/#collapsible).

wks:
> Users need to be able to locate *the information they need* by looking at the list.

In my opinion, the information the VM binding developers need the most is the detailed guide to change a particular type/method when they don't know how to fix that compilation error.  So there needs to be an option to expand everything (or just the type and method names) and let the developer search for the method name they want to locate.

But it is nice to have an option to hide everything other than the TL;DR part just in case someone wants to skim through all changes and see if it is worth upgrading to a particular version.  I'll make the change.

wks:
> We can use something like https://crates.io/crates/mdbook-admonish for [collapsible blocks](https://tommilligan.github.io/mdbook-admonish/#collapsible).

Thank you.  That's something I was looking for.  I even tried to replicate what mdbook-admonish does.  But one problem with that is that Admonish is intended to draw attention.  But I want the folded details to be as quiet as possible so that the reader can focus on things that are *not* folded.  I think we can use Admonish for the TL;DR part, instead.

qinsoon:
> So there needs to be an option to expand everything (or just the type and method names) and let the developer search for the method name they want to locate.

In your implementation, you use the `<details>` tag so those collapsible sections are rendered in DOM. You can do in-page search from browsers to find text in collapsed sections without the need to expanding them first. I think that is nice.

qinsoon:
> I think we can use Admonish for the TL;DR part, instead.

If you mean use admonish to highlight TLDR, that's fine. But don't make TLDR collapsible.

wks:
Now we have the option to fold everything except the TL;DR.

docs/userguide/src/SUMMARY.md
```diff
@@ -36,6 +36,8 @@
     - [Performance Tuning](portingguide/perf_tuning/prefix.md)
         - [Link Time Optimization](portingguide/perf_tuning/lto.md)
         - [Optimizing Allocation](portingguide/perf_tuning/alloc.md)
+- [API Migration Guide](migration/prefix.md)
+    - [Template (for mmtk-core developers)](migration/template.md)
```
qinsoon, line 40: Do we even need to list the template in mdbook? For people who need to edit the migration guide, they will view the template as the raw markdown file rather than the rendered HTML version. For people who do not edit the migration guide, they do not need to see this template in the migration guide at all.

docs/userguide/assets/snippets/view-controls.html
```diff
@@ -0,0 +1,3 @@
+<button class="api-migration-details-show-tldr" type="button">Show TL;DR only</button>
```
qinsoon, line 1: Is it possible to also include the JavaScript snippet in this HTML file?
```javascript
<script type="text/javascript">
const isApiMigrationGuide = true;
</script>
```

qinsoon (commented):
Format-wise, I think it is good now. I have some questions about the content.

Also there is one question we need to answer: do we expect external contributors to write the migration guide when they introduce a API breaking change? We could make it optional, and require maintainers help write the migration guide instead.

docs/userguide/src/migration/prefix.md
```diff
@@ -0,0 +1,175 @@
...
+VM binding developers can use this document as a guide to update their code to maintain
+compatibility with the latest release of MMTk.
+
+
+## View control
+
+Collapse details to show only two levels of the list.  This allows you to browse through the
+changes quickly and selectively expand the details of your interested items.
+
+<button class="api-migration-details-collapse-all" type="button">Collapse all details</button>
```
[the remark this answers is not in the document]
wks: You may be looking at an outdated version.  The newest version has three buttons: (1) show only TL;DR, (2) show outline, (3) show all.

Each individual collapsible block can actually be individually expanded.  Currently I made it one-way expandable (i.e. once expanded, it cannot be collapsed again) so as not to let the "> show details..." label clutter the view.  But we can keep a "< hide details" label after each TL;DR when the user chooses "show only TL;DR".

docs/userguide/src/migration/template.md
```diff
@@ -0,0 +1,127 @@
...
+reading through the following list.
+
+Keep in mind that the reader can use the buttons above to hide everything but the TL;DR.  Make sure
+the TL;DR part covers all types/modules that are changed so that the readers know what changed by
+reading the TL;DR alone.
+```
+
+API changes:
+
+*   type `Foo`
```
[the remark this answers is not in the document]
wks: Optional.  The type name is usually unambiguous.  There are rare cases, though.  For example, we have two `Block` and two `BlockState` in mmtk-core.  In that case, we shall use fully-qualified names.

docs/userguide/src/migration/template.md
```diff
@@ -0,0 +1,127 @@
...
+    -   `defg()`
+        +   It now takes a new argument `foobarbaz` which requires blah blah...
+        +   And you need to blah blah blah blah blah blah blah blah blah blah blah blah blah blah
+            blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah
+            blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah...
+        +   *(Note 1: If the change is more complicated than "... is removed", feel free to put it
+            down one level.)*
+        +   *(Note 2: Since the third level is collapsible, feel free to add more details.)*
+*   module `aaa::bbb::ccc`
+    -   **Only affects users of feature "xxxx"** *(Note: When omitted, it affects everyone.)*
```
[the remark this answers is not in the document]
wks: The note can be applied at any level, not just modules.  I'll add it to other places to make this clear.

docs/userguide/src/migration/template.md
```diff
@@ -0,0 +1,127 @@
...
+    -   `method1()`
+        +   What happened to it...
+        +   Suggestions...
+    -   `method2()`
+        +   What happened to it...
+        +   Suggestions...
+
+VM bindings need to re-implement the following traits:
```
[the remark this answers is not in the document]
wks: What about "Things changed in mmtk-core"?

I think "VM->MMTk" and "MMTk->VM" is more accurate, but the reader needs to think what the arrows mean, and it is not immediately clear what the reader should do about them.  "Things changed in mmtk-core" tells them that those things changed on the mmtk-core side, but we don't know where you called them, so make changes if you need; while "VM bindings need to reimplement ..." clearly tells them they must rewrite those functions to conform to the new signatures.

docs/userguide/src/migration/prefix.md
```diff
@@ -0,0 +1,187 @@
...
+    -   `get_referent()`
+        +   The return type is changed to `Option<ObjectReference>`.
+        +   It now returns `None` if the referent is cleared.
+    -   `clear_referent()`
+        +   It now needs to be explicitly implemented because `ObjectReference::NULL` no longer
+            exists.
+        +   Note: The `Edge` trait does not have a method for storing NULL to a slot.  The VM
+            binding needs to implement its own method to store NULL to a slot.
+
+Miscellaneous changes:
```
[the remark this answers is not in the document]
wks: Because strictly speaking, the API is *not* changed w.r.t. `get_finalized_object` and `get_forwarded_object`.  They returned `Option<ObjectReference>` before, and they still return `Option<ObjectReference>`.  However, it is worth mentioning because the binding may expose them to C/C++, and that will fail to compile if the `extern "C"` functions return `ObjectReference` which was previously nullable.

docs/userguide/src/SUMMARY.md
wks: I prefer them rendered so that editors know what to expect when looking at the rendered version. But I agree that ordinary readers may not need to read it.  I'll see if we can make it a hidden section.

docs/userguide/assets/snippets/view-controls.html
wks: The `view-controls.html` is included too early in the process of loading page, therefore I only put HTML elements in it, but not the code.

It is recommended to load the JavaScript in the end of an HTML file (the HTML page rendered from prefix.md) because the browser has to interpret the whole `<script>` (which may generate additional HTML contents in the place of `<script>`) before continuing to load the page.  Running scripts too early will make the page laggy.

docs/userguide/src/migration/prefix.md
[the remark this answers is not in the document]
qinsoon: Yeah. That was an outdated comment. Sorry.

docs/userguide/src/migration/template.md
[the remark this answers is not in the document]
qinsoon: > What about "Things changed in mmtk-core"?
> 
> I think "VM->MMTk" and "MMTk->VM" is more accurate, but the reader needs to think what the arrows mean, and it is not immediately clear what the reader should do about them. 

I cannot find proper terms that are both ambiguous and straightforward to describe these two kinds of APIs. We may ask suggestions from others. One possibility is to put the 'VM binding' part under the 'API change' part as `trait VMBinding` -- maybe it is more proper.

> "Things changed in mmtk-core" tells them that those things changed on the mmtk-core side, but we don't know where you called them, so make changes if you need; while "VM bindings need to reimplement ..." clearly tells them they must rewrite those functions to conform to the new signatures.

In this case, you are talking about whether an API provided by MMTk is mandatory or optional for a binding. "VM->MMTK" API can be mandatory (e.g. `gc_init`), or optional (e.g. `gc_poll`). "MMTK->VM" API can be mandatory (e.g. `Collection::stop_all_mutators`), or optional (e.g. `Collection::post_forwarding`). Also the 'only affects users' note should cover this.

docs/userguide/src/migration/prefix.md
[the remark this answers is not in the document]
qinsoon: I see. You mean the API is compliant but the semantic is changed. Are there any changes other than semantic changes that should be put as 'miscellaneous changes'?

docs/userguide/assets/snippets/view-controls.html
qinsoon: The performance difference should be insigifnicant. But anyway, this is not a big deal. I was thinking about how we need to include when creating a new migration page. It is very possible that at some point, we will have multiple migration pages for different versions (e.g. pre-1.0, 2.0, etc). But that also happens rarely -- so it doesn't matter.

docs/userguide/src/migration/template.md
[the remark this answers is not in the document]
wks: >  In this case, you are talking about whether an API provided by MMTk is mandatory or optional for a binding.

No.  I am not.  It is a "VM calls this" (MMTk implements it) vs "VS implements this" (MMTk calls it) dichotomy.  The accurate term is still `VM -> MMTk` and `MMTk -> VM`, but I am describing them from the perspective of the VM.

I am not talking about whether an API is mandatory or optional.  For example, the VM must call `gc_init` somewhere, but MMTk core does not know where the VM calls it because the concrete place to initialize MMTk is VM-specific.  So we can't tell the VM binding to "change `MMTkHeap::initialize()` and call `gc_init` differently" because that's OpenJDK-specific.  I can only tell the VM binding "`gc_init` changed, and you call it differently wherever you call it".  On the contrary, `Scanning` is a trait defined by mmtk-core, so we can just tell the VM binding "Reimplement `Scanning`!"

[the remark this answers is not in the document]
qinsoon: We could use 'downcall API' vs 'upcall API'. We may ask Steve for some suggestions.

docs/userguide/src/SUMMARY.md
wks: Now I use `mdbook-hide` to hide the chapter, but only in the published version.  It is shown if we build the mdbook locally.

docs/userguide/src/migration/prefix.md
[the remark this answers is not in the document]
wks: The semantics is not changed, either.  `get_finalized_object` used to return `Option<ObjectReference>` which meant "there is no finalized object".  It still returns `Option<ObjectReference>` and still means the same.  The problem is how the VM binding exposes it to C++.  MMTk does not provide an official C/C++ API, so the exposed `extern "C"` wrapper is not controlled by MMTk-core.  I mentioned that because many bindings do that, i.e. they use `ObjectReference::NULL` to encode `None`, which is no longer possible.

Strictly speaking, the root cause of that is that `ObjectReference::NULL` became non-nullable.  The `extern "C"` wrappers of `get_finalized_object` and `get_forwarded_object` are just two possible use cases of `ObjectReference::NULL`.  But since we also introduced a `NullableObjectReference`, we mention that in "miscellaneous changes".

If there are other things that have non-obvious but better way to do, we can mention them in "miscellaneous changes", too.  And of course there are other use cases that I can't foresee now.

docs/userguide/book.toml
```diff
@@ -11,3 +11,12 @@ edition = "2021"
...
 edit-url-template = "https://github.com/mmtk/mmtk-core/edit/master/docs/userguide/{path}"
+additional-css = ["assets/css/api-migration-details.css", "assets/css/mdbook-admonish.css"]
+additional-js = ["assets/js/api-migration-details.js"]
+
+[preprocessor.admonish]
+command = "mdbook-admonish"
+assets_version = "3.0.2" # do not edit: managed by `mdbook-admonish install`
+
+[preprocessor.hide]
+hide = false # Change this to true when publishing.
```
qinsoon, line 22:
```suggestion
hide = false # This is overwritten to true by env var MDBOOK_preprocessor__hide__hide when we render the user guide for publishing.
```

wks:
I simply removed the sub-title "VM bindings should reimplement ..." and merged the items into "API changes".

qinsoon:
Broken link found:
```
### Errors in docs/userguide/src/portingguide/howto/nogc.md

* [404] [https://docs.mmtk.io/api/mmtk/vm/edge_shape/trait.MemorySlice.html](https://docs.mmtk.io/api/mmtk/vm/edge_shape/trait.MemorySlice.html) | Failed: Network error: Not Found
```

It should be probably from the last PR that renamed slot.

qinsoon:
> I simply removed the sub-title "VM bindings should reimplement ..." and merged the items into "API changes".

The template is not updated.

docs/userguide/src/migration/template.md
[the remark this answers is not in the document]
wks: I removed this sub-title.

docs/userguide/src/migration/prefix.md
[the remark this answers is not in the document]
wks: I changed it to "Not API change, but worth noting".  The purpose should be clear now.

wks:
I have just updated the template.

qinsoon (approved):
LGTM

PR #1138, opened 2024-05-22
Clear side forwarding bits properly

wks:
This PR fixes bugs related to the clearing of side forwarding bits.

The side forwarding bits of ImmixSpace are now cleared after moving GCs instead of before every major GC.  This eliminates redundant clearing operations during non-moving full-heap GCs (i.e. when not doing defrag), and also ensures the subsequent nursery GCs do not see stale forwarding bits when using StickyImmix in which case young objects are also allocated in the ImmixSpace.

Also fixed the code for clearing side forwarding bits in discontiguous CopySpace.


src/policy/copyspace.rs
```diff
@@ -168,32 +168,26 @@ impl<VM: VMBinding> CopySpace<VM> {
...
     }
 
     pub fn release(&self) {
-        unsafe {
+        for (start, size) in self.pr.iterate_allocated_regions() {
+            // Clear the forwarding bits if it is on the side.
+            if let MetadataSpec::OnSide(side_forwarding_status_table) =
+                *<VM::VMObjectModel as ObjectModel<VM>>::LOCAL_FORWARDING_BITS_SPEC
+            {
+                side_forwarding_status_table.bzero_metadata(start, size);
```
qinsoon, line 179: Is there any particular reason that this is moved to the release stage? It was in prepare, as we generally prefer to do things lazily. Probably it matters too much for this case.

src/policy/immix/immixspace.rs
```diff
@@ -911,6 +910,17 @@ impl<VM: VMBinding> GCWork<VM> for SweepChunk<VM> {
                     // Block is live. Increment the allocated block count.
                     allocated_blocks += 1;
                 }
+
+                // Clear side forwarding bits.
+                // In the beginning of the next GC, no side forwarding bits shall be set.
+                if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::LOCAL_FORWARDING_BITS_SPEC {
+                    if is_moving_gc {
+                        let objects_may_move = !is_defrag_gc || block.is_defrag_source();
```
qinsoon, line 918: Probably add some comments here to make it more clear: if this is a defrag GC, we only need to clear forwarding bits for defrag source blocks; if this is a moving GC but not a defrag GC, we have no other information, and we have to clear forwarding bits for all the blocks.

src/policy/immix/immixspace.rs
```diff
@@ -911,6 +910,17 @@ impl<VM: VMBinding> GCWork<VM> for SweepChunk<VM> {
                     // Block is live. Increment the allocated block count.
                     allocated_blocks += 1;
                 }
+
+                // Clear side forwarding bits.
+                // In the beginning of the next GC, no side forwarding bits shall be set.
```
qinsoon, line 915: Is it possible to assert this? We probably don't need the following line, and we can add some assertions there: https://github.com/mmtk/mmtk-core/blob/170630fc84bc82070ca6e06ea4fed76894aa703f/src/policy/immix/immixspace.rs#L631

qinsoon (commented):
Can you run benchmarks for Immix and SemiSpace? We should expect no perf difference for SemiSpace, and speedup/no diff for Immix, right?

src/policy/copyspace.rs
wks: Well, for CopySpace, it doesn't matter whether we do it in Prepare or Release.  But I find it easy to coalesce it with `bzero_vo_bits` because we need to iterate through "allocated regions" anyway.

For ImmixSpace, we do it in Release because the next GC may not be a copying GC (maybe a full-heap non-defrag GC, or a nursery GC but with non-moving nursery).

src/policy/immix/immixspace.rs
wks: It is possible to assert that there is no forwarding bits set in the beginning of a GC.  But that'll be extreme costly.  For the release build, it is unnecessary; for the debug build, such array/memory traversal operation is especially costly due to lack of optimisations including inlining.

The invocations of `clear_forwarding_bits` on line 622 and line 631 are unrelated to this.  It is called when one GC thread won the forwarding race, but later found that it cannot forward the object because the object is already marked (line 622) or the object is pinned or the defrag space exhausted (line 631).  It needs to revert it to the `FORWARDING_NOT_TRIGGERED_YET` state.  If the state were left in the `BEING_FORWARDED` state, other GC workers will block on this object forever, waiting for it to be forwarded, which will never happen; if we change it to `FORWARDED` state, other GC workers will attempt to load from its forwarding pointer which is never written.  So the state has to be `FORWARDING_NOT_TRIGGERED_YET`.  When another worker visits this object, it will won the forwarding race again, see it is already marked, and revert it to `FORWARDING_NOT_TRIGGERED_YET` again.

wks:
> Can you run benchmarks for Immix and SemiSpace? We should expect no perf difference for SemiSpace, and speedup/no diff for Immix, right?

This PR only changes the code that involves on-the-side forwarding bits.  If the forwarding bits are in the header, there should be no performance difference with the default configuration of OpenJDK (which uses in-header forwarding bits).

If the forwarding bits are on the side, this PR fixes correctness issues because both CopySpace and ImmixSpace (in StickyImmix) would crash.

I'll run benchmarks using the default settings (in-header forwarding bits) to verify that there is no performance difference.

src/policy/immix/immixspace.rs
wks: I ran some tests locally with StickyImmix and OpenJDK with side forwarding bits, and observed that in nursery GC right after a defrag GC, a worker may observe the forwarding bits being set, but the forwarding pointer is 0.  The problem disappears if I remove the optimisation about defrag sources (i.e. force clearing forwarding bits for all blocks).  This is strange.  It looks like some forwarding bits are set in non-defrag-source blocks.

qinsoon:
> I'll run benchmarks using the default settings (in-header forwarding bits) to verify that there is no performance difference.

If we can use side forwarding bits for OpenJDK, it is worth measuring the performance.

src/policy/immix/immixspace.rs
wks: It ended up that whether a block is a defrag source or not is determined by `DEFRAG_STATE_TABLE`, but `block.sweep()` overwrites `DEFRAG_STATE_TABLE` with the number of holes.  That caused my code to consider fewer blocks as defrag sources, and that's why the problem disappears if it clears forwarding bits regardless whether a block is a defrag source.  The fix is simply swapping the order of clearing forwarding bits and sweeping the block.

wks:
> > I'll run benchmarks using the default settings (in-header forwarding bits) to verify that there is no performance difference.
> 
> If we can use side forwarding bits for OpenJDK, it is worth measuring the performance.

I'll add a feature to OpenJDK to force the forwarding bits to be on the side.  And it shall be added to minimal-tests-openjdk.  Otherwise the CI has no way to even check if this PR is correct.

src/policy/immix/immixspace.rs
wks: There is another invocation of `clear_forwarding_bits::<VM>(object);` in `post_copy`.  I changed it to be executed unconditionally when I observed that the copy context may allocate the to-space copy into a free block with stale forwarding bits.  That was because we lazily clear the forwarding bits.  Now that we clear forwarding bits at the end of copy GCs, we are sure there must not be stale forwarding bits.  I will revert that commit and elide the `clear_forwarding_bits` if the forwarding bits are on the side.

wks: More precisely, it is re-applying https://github.com/mmtk/mmtk-core/commit/d341506b70a4465dbbc05e87d27d0b7bd23844bb which made that line conditional in the first place.

wks:
binding-refs
OPENJDK_BINDING_REF=fix/side-fwd-bits

wks:
There is a deadlock happening in the minimal binding test.  It is running `lusearch` in dacapo 2006.  The bug is reproducible locally, and it does not require the forwarding bits to be on the side (i.e. not related to this PR or the [OpenJDK PR](https://github.com/mmtk/mmtk-openjdk/pull/277)).  I have reported it in https://github.com/mmtk/mmtk-openjdk/issues/278.  Since the deadlock is so easy to trigger, I'll dig into that bug before continuing with this PR.

wks:
The benchmark result.  Lusearch from DaCapo Chopin, 3x min heap size, 40 invocations. 3 builds:

-   build1: master
-   build2: this PR and [the mmtk-openjdk PR](https://github.com/mmtk/mmtk-openjdk/pull/277)
-   build3: this PR and [the mmtk-openjdk PR](https://github.com/mmtk/mmtk-openjdk/pull/277), using side forwarding bits

Note: There is no master branch with side forwarding bits because that'll crash.

Plots (normalised to build1): [plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-05-30-Thu-092115&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&)

image

Plots (not normalised): [plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-05-30-Thu-092115&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|41&Histogram%20(with%20CI)^build^mmtk_gc&)

image

Note that the number of GC is stuck at 2047 for GenImmix.  The actual number of GCs should be much higher.

# build1 vs build2

The difference between build1 and build2 is the effect of this PR without enabling side forwarding bits.

Plots (normalised to build1): [plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-05-30-Thu-092115&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&build^2^build3|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&)

image

Mostly unchanged.  For some plans, such as GenCopy and SemiSpace, STW time becomes higher, but other time becomes lower.  It's likely some changes in the code influenced the way it measures STW time, counting some other time into STW time.  The total time is different with non-overlapping error bars, but is within 1%.

# build2 vs build3

The difference between build2 and build3 indicates the cost of enabling side forwarding bits.

Plots (normalised to build2): [plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-05-30-Thu-092115&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&build^2^build1|20&1^invocation|30&1&benchmark^mmtk_gc&build;build2|41&Histogram%20(with%20CI)^build^mmtk_gc&)

Obviously, the number of GC is increased.  That's because the side forwarding bits occupies part of the heap space.  As a result, the effective heap size for objects becomes smaller.  The STW time increased due to increased number of GCs and the cost of clearing side forwarding bits.  The plans involving CopySpace (SemiSpace, GenCopy and GenImmix) are impacted to a greater degree.  The `time.other` is mostly unchanged, with GenImmix as an exception.  But the total time is consistently increased.

image

qinsoon:
The slowdown of using side forwarding bits is significant (build2 vs build3). It would be good to measure the overhead of side forwarding bits on master -- if it cannot run with Immix or with discontiguous spaces, you can try find a configuration that runs (e.g. no compress pointer, semispace). 

Also please include all the benchmarks in the evaluation. It is okay to focus on one benchmark to address some issues, but we need to know the results on more benchmarks in order to know if the PR has performance regression.

wks:
> The slowdown of using side forwarding bits is significant (build2 vs build3).

Yes.  Side forwarding bits are not free.  It is significant for CopySpace because the entire CopySpace is copied away, and the clearing involves the entire space.  But it is much lower for ImmixSpace.  For Immix and StickyImmix, the total time overhead is within 1%, and the STW time overhead is about 2%-4%.

> It would be good to measure the overhead of side forwarding bits on master -- if it cannot run with Immix or with discontiguous spaces, you can try find a configuration that runs (e.g. no compress pointer, semispace).

I am doing a 4-way comparison, comparing master-header, master-side, PR-header and PR-side, using SemiSpace and GenCopy, without compressed oops.  Let's wait for the results.

wks:
I added build4: master with side forwarding bits enabled.  To avoid confusion, I'll use "master-header", "PR-header", "master-side", "PR-side" as the labels of the builds.  I re-ran lusearch, but only using SemiSpace and GenCopy which only involve CopySpace.

Plot (un-normalised): [plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-05-31-Fri-043010&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|70&build^sidefwd-4way|41&Histogram%20(with%20CI)^build^mmtk_gc&)

image

Plot (normalised to "master-header"): [plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-05-31-Fri-043010&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|70&build^sidefwd-4way|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&)

image

The increase in the number of GC and the times due to moving the forwarding bits on the side (i.e. the ratio of `side / header` for each number) is similar for both the master branch and this PR.   (p.s. I wonder why the number of GC didn't overflow this time.)

For SemiSpace, there is a 1.7% increase in STW time and 0.5% increase in total time, and it is consistent with the result of the last 3-way comparison.

wks:
I had a look at the timeline using eBPF.  THe following plots are from running lusearch, 64M heap size (the same as on lynx.moma), on my laptop.

This PR, header, no compressed oops, SemiSpace
image

This PR, side, no compressed oops, SemiSpace
image

There is one obvious problem with the current implementation.  The clearing of the side forwarding bits is not parallelised.  But for lusearch alone, that's not the bottleneck.  Any load-imbalance in the Prepare and Closure stage would have greater impact in STW time.

This PR, side, no compressed oops, GenCopy nursery collection
image

The impact should be greater for GenCopy because the impact of the Release work packet not parallelized is greater on nursery collection.  However, the benchmark results from lynx.moma showed that the impact of side forwarding bits on STW time is greater for SemiSpace.  I think that means the un-parallelised Release work packet is only a secondary factor.  The primary impact factor is the decreased effective heap size (due to side metadata taking space) as well as cache locality problems with side metadata.

wks:
I ran other benchmarks on on lynx.moma and bobcat.moma (identical hardware and identical binary). 3x min heap size w.r.t. G1 in OpenJDK. No compressed OOPs (even when running StickyImmix).

Note that all four builds use the mmtk-openjdk binding from https://github.com/mmtk/mmtk-openjdk/pull/277, but only master-header and PR-header have the `forwarding_bits_on_side` Cargo feature enabled in mmtk-openjdk.

SemiSpace. ([plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-05-31-Fri-165718&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|70&build^sidefwd-4way|30&1&benchmark&build;build1|41&Histogram%20(with%20CI)^build^benchmark&)) Note that biojava and fop failed to run at 3x min heap size with SemiSpace

image

StickyImmix. ([plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2024-05-31-Fri-165735&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|70&build^sidefwd-4way|30&1&benchmark&build;build1|41&Histogram%20(with%20CI)^build^benchmark&)

image

Comparing master-header and pr-header, most differences are within the error bars.  For SemiSpace, the stw.time for pr-header is noticeably higher (but still small) than master-header.  That could be a result of executing `for (start, size) in self.pr.iterate_allocated_regions() {` *unconditionally* in `Release`, even though it does a no-op within the for look if VO bit is not enabled, either.  I'll verify if that's the cause, but I am not sure if it is worth optimising.  For StickyImmix, the time.stw for pr-header is sometimes higher and sometimes lower, but within the error bars.  I think they are just noise because the clearing of side forwarding bits for ImmixSpace is always guarded by an `if` statement with a constant condition, and can be eliminated by the compiler.  The generated code should be identical if the forwarding bits are in the header.

Comparing XXX-header and XXX-side, we see the overhead of forcing forwarding bits to be on the side.  The overhead is significantly higher for SemiSpace (only CopySpace) than StickyImmix (only ImmixSpace).  One reason may be that ImmixSpace distributes the clearing operation to `SweepChunk`, and that can be paralellised.  Another reason may be that if StickyImmix executes full-heap GC often (which is the reality for tighter heap), it will not do copying GC unless it is a defrag collection and if the block is a defrag source.  That should greatly reduced the amount of clearing to do.  However, from the plot of "majorGC / GC" for StickyImmix, I don't see any connection between "majorGC / GC" and the increase in time.stw.

MajorGC per GC: ([plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2024-05-31-Fri-165735&benchmark^build^invocation^iteration^mmtk_gc&GC^majorGC&majorGC%20/%20GC|10&iteration^1^4|20&1^invocation|70&build^sidefwd-4way|41&Histogram%20(with%20CI)^build^benchmark&))

image

There are two interesting benchmarks: "avrora" and "jython".  Their side-metadata overhead is so big in SemiSpace, but "avrora" sped up when using side forwarding bits, the "jython" only had a very tiny overhead when using side forwarding bits.  The speed-up could be noise (as their error bars are large).

qinsoon (approved):
LGTM. The performance looks reasonable to me. If you would like to investigate further, that's fine. The PR in its current state is good to merge.

wks:
Running jython locally, with SemiSpace, 76M heap size (3x min heap size w.r.t. G1 in OpenJDK), and no compressed oops:

master-header (number of GC = 1697, time.stw = 6489.81 ms, time = 19773ms)
image

pr-side (number of GC > 2047 (counter overflowed), time.stw = 11954.42 ms, time = 24658ms)
image

It looks like the 76M heap size is still too small for jython when using SemiSpace.  Individual GCs don't look very different, and the Release work packet is trivial compared to the Closure stage.  But when using side metadata, it is simply doing more GCs.  Let's zoom in to the middle of the execution.

master-header
image

During some periods of execution (maybe when the program demands more memory), the program is spending almost 100% of the time doing GC, and the mutators hardly have any time to execute.  From time to time, the muators can get an extended period of execution time without triggering GC (probably when the program keeps fewer objects alive).  If the heap gets even a little bit smaller (or some portion of the heap is occupied by the side metadata), the thrashing will become significantly worse, amplifying the overhead of side forwarding bits.

Given that SemiSpace can only make use of 50% of the heap size, I doubled the heap size to 152M, and the problem became much milder.

master-header (GC = 141, time.stw = 590.08 ms, time = 13926 ms)
image

pr-side (GC = 149, time.stw = 735.66 ms, time = 14202 ms)
image

The STW time still increased significantly when using side metadata, but the total time is less affected.  At this heap size, it still occasionally have bursts of frequent GCs.  See the following plot:

master-header
image

Each individual GC still takes the similar time (about 3.5ms) as 76M, but the density is much lower.  I think it is the nature of the `jython` benchmark to have short bursts of allocations, but the amount of memory it keeps alive is not very high (that's why each GC can still finish in 3.5ms regardless of heap size).  Ideally, this kind of workload could benefit from MemBalancer.

FYI, the following timelines are recorded from StickyImmix with 76M heap size:

master-header (GC = 178, time.stw = 458.02, time = 14110)
image

pr-side (GC = 186, time.stw = 423.02, time = 13823) (Note: I recorded it on my laptop and only ran once.  The speed-up may just be noise, but the *shape* of the timeline is important.)
image

zoom-in of master-header
image

The allocation is still bursty, and there are some long GCs (which take up to 9.8ms).  But overall it only spent a small amount of time doing GC.

wks:
`avrora` is simpler.  GC happens at regular intervals.  The shape of the timeline of individual GCs doesn't change, but every GC becomes longer, and it runs more GCs.  It simply takes longer to do the Closure.  I think it is a result of cache locality.

master-header SemiSpace (GC = 137, time.stw = 283 ms, time = 6078 ms)
image

pr-side Semispace (GC = 164, time.stw = 387, time = 6105ms)
image

The 100th GC of master-header
image

The 100th GC of PR-side
image

wks:
I re-ran lusearch with SemiSpace, but this time with three builds

-   build1: master
-   build2: this PR
-   build3: this PR plus a conditional if statement so that the Release work packet won't iterate through the chunks if not using VO bits or side forwarding bits.  (https://github.com/wks/mmtk-core/commit/b6d24bf246e044eaffc489e423df2329ee836334)

All builds are using *header* forwarding bits.  I added Build3 to see if the unconditional iteration of chunks is the cause of the tiny overhead of STW time.

But experiments show that build3 has even greater overhead than build2. ([plotty](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|bobcat-2024-06-02-Sun-114142&benchmark^build^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^mmtk_gc&build;build1|41&Histogram%20(with%20CI)^build^mmtk_gc&))

image

I think the effect of the `if` statement is trivial, and maybe some inlining decisions made by the compiler can have more significant effects than that.

Issue #1140, opened 2024-05-24
`harness_begin` shouldn't trigger GC


wks:
Benchmarks use `harness_begin` to start collecting statistics.  It is a good practice for benchmarks to trigger a GC before starting collecting statistics because that will reduce the variance caused by the state of the heap (including the amount of free space available), which will in turn influence the behaviour of subsequent GCs (such as whether they will be nursery GCs or full-heap GCs).  Currently, `harness_begin` triggers a forced and exhaustive GC before starting collecting stats.

However, when using the benchmark (e.g. the DaCapo benchmark suite) on other GCs (such as G1 in OpenJDK), their GC and statistics framework may not have the same behaviour of triggering GC when starting collecting stats.  That makes the comparison unfair if the benchmark relies on `harness_begin` to trigger GC.

The desired behaviour is, `harness_begin` shall not trigger GC, but the benchmark shall trigger GC manually before calling `harness_begin`.

wks:
But I also remember that when running benchmarks, we also disable `System.gc()` and completely rely on `harness_begin` to trigger GC.  See [the `ignore_system_gc` option](https://github.com/mmtk/mmtk-core/blob/170630fc84bc82070ca6e06ea4fed76894aa703f/src/util/options.rs#L800) and [the `ignoreSystemGC` option in JikesRVM](https://github.com/JikesRVM/JikesRVM/blob/5072f19761115d987b6ee162f49a03522d36c697/MMTk/src/org/mmtk/utility/options/IgnoreSystemGC.java#L18)

PR #1141, opened 2024-05-24
Remove GC in harness_begin. full_heap_system_gc defaults to true

qinsoon:
This PR fixes https://github.com/mmtk/mmtk-core/issues/1140.

### DIFF
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -318,12 +318,10 @@ impl<VM: VMBinding> MMTK<VM> {
         self.scheduler.respawn_gc_threads_after_forking(tls);
     }
 
-    /// Generic hook to allow benchmarks to be harnessed. MMTk will trigger a GC
-    /// to clear any residual garbage and start collecting statistics for the benchmark.
+    /// Generic hook to allow benchmarks to be harnessed.
     /// This is usually called by the benchmark harness as its last step before the actual benchmark.
-    pub fn harness_begin(&self, tls: VMMutatorThread) {
+    pub fn harness_begin(&self, _tls: VMMutatorThread) {
         probe!(mmtk, harness_begin);
-        self.handle_user_collection_request(tls, true, true);
         self.inside_harness.store(true, Ordering::SeqCst);
         self.stats.start_all();
         self.scheduler.enable_stat();
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -710,8 +710,9 @@ pub fn add_phantom_candidate<VM: VMBinding>(mmtk: &MMTK<VM>, reff: ObjectReferen
     mmtk.reference_processors.add_phantom_candidate(reff);
 }
 
-/// Generic hook to allow benchmarks to be harnessed. We do a full heap
-/// GC, and then start recording statistics for MMTk.
+/// Generic hook to allow benchmarks to be harnessed. It is recommended that
+/// the application/benchmark should trigger a GC through [`crate::memory_manager::handle_user_collection_request`]
+/// before calling `harness_begin`.
 ///
 /// Arguments:
 /// * `mmtk`: A reference to an MMTk instance.
--- a/src/util/options.rs
+++ b/src/util/options.rs
@@ -807,7 +807,7 @@ options! {
     nursery:               NurserySize          [env_var: true, command_line: true]  [|v: &NurserySize| v.validate()]
         = NurserySize::ProportionalBounded { min: DEFAULT_PROPORTIONAL_MIN_NURSERY, max: DEFAULT_PROPORTIONAL_MAX_NURSERY },
     /// Should a major GC be performed when a system GC is required?
-    full_heap_system_gc:   bool                 [env_var: true, command_line: true]  [always_valid] = false,
+    full_heap_system_gc:   bool                 [env_var: true, command_line: true]  [always_valid] = true,
     /// Should finalization be disabled?
     no_finalizer:          bool                 [env_var: true, command_line: true]  [always_valid] = false,
     /// Should reference type processing be disabled?


qinsoon:
@wks Feel free to make code changes in the PR.

docs/userguide/src/migration/prefix.md
```diff
@@ -75,6 +75,23 @@ See also:
...
+Not API change, but worth noting:
+
+* module `mmtk::memory_manager`
+  * `harness_begin`
+    * The function used to trigger a forced full heap GC before starting collecting statistics.
+      Now it no longer triggers the GC.
+    * The user applications and benchmarks are responsible to trigger a GC before calling `harness_begin`.
+  * `handle_user_collection_request`
+    * The runtime option `full_heap_sytem_gc` defaults to `true` now (it used to be `false` by default).
+      The GC triggered by this function will be a full heap GC by default.
```
wks, line 94: You can add link to this PR in a "See also:" section.

docs/userguide/src/migration/prefix.md
```diff
@@ -75,6 +75,23 @@ See also:
 -   PR: <https://github.com/mmtk/mmtk-core/pull/1134>
 -   Example: <https://github.com/mmtk/mmtk-openjdk/pull/274>
 
+### Remove GC in `harness_begin`
```
wks, line 78: Because the versions are in the reversed order, we should list changes in the reversed order, too.

src/mmtk.rs
```diff
@@ -403,35 +401,22 @@ impl<VM: VMBinding> MMTK<VM> {
...
     /// * `tls`: The mutator thread that requests the GC
-    /// * `force`: The request cannot be ignored (except for NoGC)
-    /// * `exhaustive`: The requested GC should be exhaustive. This is also a hint.
-    pub fn handle_user_collection_request(
-        &self,
-        tls: VMMutatorThread,
-        force: bool,
-        exhaustive: bool,
-    ) {
+    pub fn handle_user_collection_request(&self, tls: VMMutatorThread) {
```
wks, line 408: Some VMs gives the user fine-grained control of whether to trigger a full-heap GC or a minor GC.  In CRuby, the [`GC::start`](https://docs.ruby-lang.org/en/3.3/GC.html#method-c-start) method uses the `full_mark` parameter for that.  The current master branch of `mmtk-ruby` ignores this because it still doesn't support StickyImmix, but will enable StickyImmix once we fix the side forwarding bits bug.  We should leave a public API for the VM to choose whether to perform an exhaustive GC.

Note: `memory_manager::handle_user_collection_request` is public, and `MMTK::handle_user_collection_request` is a public method, too.

The OpenJDK binding calls `memory_manager::handle_user_collection_request` which does not have the `force` and `exhaustive` parameters.  I suggest we leave `Option::full_heap_system_gc` and `MMTK::handle_user_collection_request` unchanged, but change `memory_manager::handle_user_collection_request` to pass `true` and `true` to `MMTK::handle_user_collection_request`.

wks (commented):
Considering that some VMs, such as CRuby, allows the user to choose between triggering a nursery or full-heap GC, we should retain a public API to allow triggering nursery and full-heap GC, too.

src/mmtk.rs
qinsoon: > change memory_manager::handle_user_collection_request to pass true and true to MMTK::handle_user_collection_request

In this case, MMTk basically ignores those two options: `ignore_system_gc` is ignored, as we always force GC, and `full_heap_system_gc` is also ignored as we always do full heap/exhaustive GC.

wks: No.  `MMTK::handle_user_collection_request` is public, too.  The Ruby VM, for example, can directly call `mmtk.handle_user_collection_request(tls, false, true)` to trigger full-heap GC, and call `mmtk.handle_user_collection_request(tls, false, false)` to trigger nursery GC.  OpenJDK can call `memory_manager::handle_user_collection_request(tls)` to use the default `force` and `exhaustive` arguments.

wks: Well, I think a more elegant solution is to change the argument of `MMTK::handle_user_collection_request` to take one single `UserCollectionArgs` value:

```rs
struct UserCollectionArgs {
    force: bool,
    exhaustive: bool,
}

impl default for UserCollectionArgs {
    fn default() -> Self {
        Self { force: true, exhaustive: true }
    }
}
```

This will allow the VM binding to always call `mmtk.handle_user_collection_request(tls, UserCollectionArgs::default())` if they want the default value, or call `mmtk.handle_user_collection_request(tls, UserCollectionArgs { exhaustive: true, ..Default::default() }` if they want to customise.  But that'll be an API-breaking change.

qinsoon:
As we discussed earlier, I remove the option `full_heap_system_gc`, and add an argument `exhaustive` for `handle_user_triggered_collection`.

@wks Can you take a look at this PR? If the changes look fine, I will fix the bindings for the breaking change.

src/mmtk.rs
```diff
@@ -318,12 +318,10 @@ impl<VM: VMBinding> MMTK<VM> {
         self.scheduler.respawn_gc_threads_after_forking(tls);
     }
 
-    /// Generic hook to allow benchmarks to be harnessed. MMTk will trigger a GC
-    /// to clear any residual garbage and start collecting statistics for the benchmark.
+    /// Generic hook to allow benchmarks to be harnessed.
     /// This is usually called by the benchmark harness as its last step before the actual benchmark.
-    pub fn harness_begin(&self, tls: VMMutatorThread) {
+    pub fn harness_begin(&self, _tls: VMMutatorThread) {
```
wks, line 323: Since we decided that this is going to be an API-breaking change, we can remove this unused `_tls` argument.

```suggestion
    pub fn harness_begin(&self) {
```

src/mmtk.rs
```diff
@@ -403,35 +401,26 @@ impl<VM: VMBinding> MMTK<VM> {
     }
 
     /// The application code has requested a collection. This is just a GC hint, and
-    /// we may ignore it.
+    /// we may ignore it (See `ignore_system_gc` and `full_heap_system_gc` in [`crate::util::options::Options`]).
```
wks, line 404: We removed `full_heap_system_gc`.

```suggestion
    /// we may ignore it (See `ignore_system_gc` in [`crate::util::options::Options`]).
```

wks (changes requested):

Issue #986, opened 2023-10-16
Add `visit_slice` to `EdgeVisitor`


wks:
It is stupid to enqueue the address of each and every element of an array into a `ProcessEdgesWork`.  We can enqueue slices directly.

For example, when scanning an `Object[]`, instead of enqueuing `&object[0]`, `&object[1]`, `&object[2]`, ..., `&object[object.length - 1]`, we can enqueue one single `OpenJDKEdgeRange { range: &object[0]..&object[object.length - 1] }`.  That saves both space and time.

But we need to take into account the effect of prefetching as well as eager edge loading (See: https://github.com/mmtk/mmtk-core/issues/584).  If we decide that it is better to load edges eagerly when scanning the object, we may want to enqueue the edge address together with the value it holds.  If that's the case, there will be no space advantage to enqueue slices because we need to enqueue values, too.  But there may be a temporal advantage.  If we know that consecutive edges in a ProcessEdgesWork work packet are loaded sequentially from an array, we may take advantage of it and do prefetching more efficiently.

wks:
https://github.com/mmtk/mmtk-core/issues/797 is related.  The region mod buf should have recorded slices instead of individual slots to avoid excessive space usage.

PR #1142, opened 2024-06-03
Add back `DummyVM` as a part of the porting guide. Minor changes to MMTk initialization in the porting guide.

qinsoon:
Following the discussion here: https://mmtk.zulipchat.com/#narrow/stream/315620-Porting/topic/Porting.20MMTK.20to.20Clasp.20Common.20Lisp/near/442123897. It is useful for the language implementers to have a Rust binding crate that implements all the boilerplate code and can compile to start with. This PR adds back `DummyVM` to the porting guide, and includes some minor changes to the porting guide.


docs/userguide/src/portingguide/howto/nogc.md
```diff
@@ -110,6 +112,29 @@ We recommend going through the [list of constants in the documentation](https://
...
+   the GC plan to use, heap sizes, etc. This is [a full list of runtime options](https://docs.mmtk.io/api/mmtk/util/options/struct.Options.html).
+   It is a common practice that the VM parses its command line arguments, then sets some GC-related options
+   to MMTk here. You can also set virtual memory layout for MMTk. Some runtimes may require special layouts, such as using compressed pointers
+   with a fixed heap range. However, both setting options and VM layouts are optional -- MMTk will use the default values if none is set.
+2. Create an `MMTK` instance via [`memory_manager::mmtk_init()`](https://docs.mmtk.io/api/mmtk/memory_manager/fn.mmtk_init.html). This
+   enables the binding to use most of the MMTk APIs in [`memory_manager`](https://docs.mmtk.io/api/mmtk/memory_manager/index.html), as most
+   APIs require a reference to `MMTK`.
+3. When the runtime is ready for GCs (including getting its thread system ready to spawn GC threads), it is expected to call [`memory_manager::initialize_collection`]
+   (https://docs.mmtk.io/api/mmtk/memory_manager/fn.initialize_collection.html). Once the function returns, MMTk may trigger a GC at any time.
```
k-sareen, line 131:
```suggestion
   (https://docs.mmtk.io/api/mmtk/memory_manager/fn.initialize_collection.html). Once the function returns, MMTk may trigger a GC at any appropriate time.
```

"appropriate" because we only really do it in the slow-path.

qinsoon:
> We shouldn't bury the `dummyvm` directory too deep into the docs directory. MMTk core developers will need to keep maintaining it as we change the API of mmtk-core. It is better to move it to a more accessible location. I suggest we place it in `mmtk-core/dummyvm`, directly. We may add a README.md in `mmtk-core/dummyvm` to state its purpose and link to the porting guide.

I put it as a part of the porting guide, as I feel it is only relevant for people who start a new binding. I don't think it is important and relevant enough to be at the top level of the project root directory. It was placed in `vmbindings/dummyvm`, because it was introduced when we had other bindings in the core repo such as `vmbindings/jikesrvm` and `vmbindings/openjdk`. 

For maintainability, I modified the CI scripts to build and style-check  `DummyVM` in the CI. We will know if it no longer compiles. Also I changed the `DummyVM` a bit to have more methods as `unimplemented!()`, as we no longer need it to run with the default implementation. We just need it to compile so people can have a reasonable start point for porting.

If you do think we should put it as a place that is easier for people to browse to from the root directory, I think we can put it in `mmtk-core/docs/dummyvm`.

wks:
> If you do think we should put it as a place that is easier for people to browse to from the root directory, I think we can put it in `mmtk-core/docs/dummyvm`.

Yes.  `mmtk-core/docs/dummyvm` is good enough, at least shorter than the previous `mmtk-core/vmbindings/dummyvm`.  We can add a sentence in its own `README.md` which says it is probably only relevant for people who start a new binding. 

And we can remove `mmtk-core/docs/header` given that we acknowledge that `dummyvm` is part of the documentation.  There is another `mmtk.h` in `docs/header`, and we can remove that in favor for the one in dummyvm.

docs/userguide/src/portingguide/code/dummyvm/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+lazy_static! {
+    pub static ref BUILDER: Mutex<MMTKBuilder> = Mutex::new(MMTKBuilder::new());
+    pub static ref SINGLETON: MMTK<DummyVM> = {
+        let builder = BUILDER.lock().unwrap();
+        debug_assert!(!MMTK_INITIALIZED.load(Ordering::SeqCst));
+        let ret = mmtk::memory_manager::mmtk_init(&builder);
+        MMTK_INITIALIZED.store(true, std::sync::atomic::Ordering::Relaxed);
+        *ret
+    };
+}
```
wks, line 51: This may not be the best way to initialize a global `MMTK` singleton.

1.  The `MMTKBuilder` only needs to exist during creating the `MMTK` instance.  It shouldn't be static, but it has to be due to the way `SINGLETON` is initialized.
2.  The `MMTK_INITIALIZED` variable plus the `SINGLETON` lazy static variable seem to reinvent the [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) type in the standard library (introduced in 1.70.0).

We used to do that because there were no better alternatives when we were using 1.61 or 1.66.  But I don't want to recommend this way of initialization to the readers as it is not very elegant, although our OpenJDK binding has been doing it this way without problem.

If we are not sure if other ways of initialization will work, I am OK with merging the current code in this PR.  We can start experimenting in the OpenJDK binding, and change the dummy VM in a subsequent PR.  That may take some time and some performance evaluation.

The Ruby binding currently uses `once_cell::sync::OnceCell<RubyBinding>` (and should probably switch to the stdlib, too), and the `RubyBinding` struct has a field `mmtk: &'static MMTK<Ruby>`.  It has been working well so far.

qinsoon: > If we are not sure if other ways of initialization will work, I am OK with merging the current code in this PR.

We can introduce a test in this crate to make sure that the initialisation works fine. 

> We can start experimenting in the OpenJDK binding, and change the dummy VM in a subsequent PR.

We don't have to make OpenJDk and other bindings to use the same initialization approach we will use in the dummyvm.

> That may take some time and some performance evaluation.

As long as we don't introduce overhead when we access `SINGLETON`, it should be fine. Initialization just happens once.

qinsoon: I am happy to make changes in this PR about the initialisation. If I understand you right, you suggested not to have the `MMTKBuilder` as a static, and try use `OnceLock`/`OnceCell` from the stdlib for `MMTK`.

The reason why `MMTKBuilder` is static is that we need to set options to `MMTKBuilder` from the VM when the VM parses its command line args. How do you deal with this?

wks: > The reason why `MMTKBuilder` is static is that we need to set options to `MMTKBuilder` from the VM when the VM parses its command line args. How do you deal with this?

The Ruby binding parses all command line args and store the results in variables in the C program.  Then it creates an `MMTKBuilder` instance, set the options, and instantiate `MMTK`.  See [these lines](https://github.com/mmtk/ruby/blob/310837dbbf6ca8abe662782bdef429369b43bb9f/mmtk_support.c#L283-L309).  The Rust part of the binding exposes several methods for setting options in the `MMTKBuilder` and C code calls them.  The `mmtk_init_binding` function gives the ownership of the builder back to the Rust code, and it instantiates `RubyBinding` and `MMTK`.

Alternatively, the VM binding creates an `MMTKBuilder` before parsing command line args.  During parsing, it sets options in the `MMTKBuilder`.  After parsing, it instantiates `MMTK`.

docs/dummyvm/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[profile.release]
+lto = true
+
+[dependencies]
+# We use a local path as the MMTk dependency here, as we want to test the code with the current version.
+# Generally for a binding, you would like to a specific version, or a git commit.
+# mmtk = "0.25.0"
+# mmtk = { git = "https://github.com/mmtk/mmtk-core.git", branch = "master" }
+mmtk = { path = "../../../../../../" }
```
wks, line 21: The path changed.

```suggestion
mmtk = { path = "../../" }
```

docs/userguide/src/portingguide/code/dummyvm/src/lib.rs
qinsoon: I made some changes. I expose a function to create an MMTk builder, process and mmtk_init require a pointer to the MMTk builder. I expose `InitializeOnce` from mmtk-core, and use it for singleton. I add a test to make sure that the init process works fine.

Can you review this PR again?

qinsoon:
The 3 optional checks failed, but they are expected:
* The broken link check failed as I added a link to the dummyvm directory in the github repo, which does not exist yet.
* The API check failed, as I exposed `InitializeOnce` from mmtk-core so it can be used for `SINGLETON`.
* The above API change does not require an entry in the migration check. The migration guide check failed, as we have an API change but no migration guide change.

docs/userguide/src/portingguide/code/dummyvm/src/lib.rs
wks: I made some changes upon that.  See https://github.com/mmtk/mmtk-core/commit/0436112477ac5281c974d173deb1656b18161cbd

I reordered the list of functions and placed `mmtk_create_builder` first, and then option-setting functions, and then `mmtk_init`.  That's probably the order the VM binding will call.

I replaced `InitializeOnce` with `OnceLock` for the `SINGLETON`.  `InitializeOnce` lacks a method that sets the value directly (`OnceLock::set`), but it's probably not worth adding it because `InitializeOnce` is intended for internal use.

I also created a wrapper function `fn mmtk() -> &'static MMTK<DummyVM>` because `OnceLock` does not implement `Deref`.  It calls `OnceLock::get()` which involves a check.  I think it's safer to recommend safe functions instead of faster unsafe ones.  If the binding developers find that it's a performance bottleneck, they can switch to the `once_cell` crate and use the unsafe method `once_cell::sync::OnceCell::get_unchecked` instead.

I changed the `extern "C"` wrapper functions for setting options a little bit.  I feel the name `mmtk_process` is not very straightforward (it's inherited from `OptionSet.process` from JikesRVM).  I categorized the method of setting options into "setting by value" and "setting by string", and each can be either done directly using Rust methods or done with `extern "C"` wrapper functions.  I added a test case which uses `extern "C"` wrappers for them.

docs/dummyvm/src/scanning.rs
```diff
@@ -0,0 +1,40 @@
...
+        unimplemented!()
+    }
+    fn scan_vm_specific_roots(_tls: VMWorkerThread, _factory: impl RootsWorkFactory<DummyVMSlot>) {
+        unimplemented!()
+    }
+    fn scan_object<EV: SlotVisitor<DummyVMSlot>>(
+        _tls: VMWorkerThread,
+        _object: ObjectReference,
+        _edge_visitor: &mut EV,
+    ) {
```
wks, line 28: We have renamed it to `SlotVisitor`.  The generic parameter and value parameter should be changed, too.

```suggestion
    fn scan_object<SV: SlotVisitor<DummyVMSlot>>(
        _tls: VMWorkerThread,
        _object: ObjectReference,
        _slot_visitor: &mut SV,
    ) {
```

docs/dummyvm/src/slots.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+impl Iterator for DummyVMMemorySliceIterator {
+    type Item = DummyVMSlot;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.cursor >= self.limit {
+            None
+        } else {
+            let edge = self.cursor;
```
wks, line 101:
```suggestion
            let slot = self.cursor;
```

docs/dummyvm/src/slots.rs
```diff
@@ -0,0 +1,233 @@
...
+            self.cursor = unsafe { self.cursor.add(1) };
+            Some(DummyVMSlot::Simple(SimpleSlot::from_address(
+                Address::from_ptr(edge),
```
wks, line 104:
```suggestion
                Address::from_ptr(slot),
```

docs/dummyvm/src/slots.rs
```diff
@@ -0,0 +1,233 @@
...
+            )))
+        }
+    }
+}
+
+/// Compressed OOP edge only makes sense on 64-bit architectures.
```
wks, line 110: There are other occurrences of "edge" in this file.  It's helpful to do a find/replace.

wks (changes requested):
There are still some occurrences of "edge" in "slots.rs".  They need to be renamed.

But in practice, no VM uses so many kinds of slots.  I introduced those slots into the old DummyVM in order to test if all those kinds of slots work, and now we have test cases in `mock_test_slots.rs`.  To make a template for new bindings, we may just set `DummyVM::SlotType` to `mmtk::vm::slot::SimpleSlot` and remove this `slots.rs`.  Alteranatively, we may implement a `DummyVMSlot` by copying from `SimpleSlot`.  That should be enough for documentation and demonstration.

docs/dummyvm/src/object_model.rs
```diff
@@ -0,0 +1,93 @@
...
+use mmtk::util::{Address, ObjectReference};
+use mmtk::vm::*;
+
+pub struct VMObjectModel {}
+
+// This is the offset from the allocation result to the object reference for the object.
+// Many methods like `address_to_ref` and `ref_to_address` use this constant.
+// For bindings that this offset is not a constant, you can implement the calculation in the methods, and
+// remove this constant.
+pub const OBJECT_REF_OFFSET: usize = 4;
```
wks, line 12: Like all the different kinds of slots, this non-zero and unaligned value was introduced for testing the VO bits.  Few real-world VMs use this value.  We may set it to zero, and comment that it can be set to a different value if the ObjectReference in the VM has an offset from the starting address.

docs/userguide/src/portingguide/code/dummyvm/src/lib.rs
qinsoon: Do we know if `OnceLock` would be no worse than lazy_static? We should get performance right here for the developers.

The reason that I used `InitializeOnce` is that we maintain `InitializeOnce`, and we know  it is zero-overhead abstraction. Getting this wrong could potentially make everything slow.

docs/userguide/src/portingguide/howto/nogc.md
```diff
@@ -110,6 +112,29 @@ We recommend going through the [list of constants in the documentation](https://
...
 Now that we have most of the boilerplate set up, the next step is to initialize MMTk so that we can start allocating objects.
 
+In short, MMTk uses the builder pattern. The binding needs to create an [`MMTKBuilder`](https://docs.mmtk.io/api/mmtk/struct.MMTKBuilder.html),
+create an [`MMTK`](https://docs.mmtk.io/api/mmtk/mmtk/struct.MMTK.html) instance from the builder, and then initialize MMTk's collection
+when the runtime system is ready for GCs.
+The following steps describes details. In an actual binding implementation, the binding may choose to combine several
+steps into one function call to make things simpler.
+
+1. Create an `MMTKBuilder` using [`MMTKBuilder::new()`](https://docs.mmtk.io/api/mmtk/struct.MMTKBuilder.html#method.new). You can set
+   runtime options via [`set_option()`](https://docs.mmtk.io/api/mmtk/struct.MMTKBuilder.html#method.set_option) for things like
```
wks, line 122: It's worth mentioning that it is possible to set options directly using statements like `builder.options.threads.set(4);`, and perhaps it is better to mention that before mentioning `set_option` which parses string.  Wenyu once tried to convert integer to strings just for passing the string to `set_option` because he didn't know it was possible to set options using Rust values directly.

docs/userguide/src/portingguide/code/dummyvm/src/lib.rs
wks: The source code of `OnceLock` shows that it is implemented just like `InitializeOnce`. (See https://doc.rust-lang.org/src/std/sync/once_lock.rs.html#77-99)

The author of the `once_cell` crate recommended against using `lazy_static!`.  (See https://docs.rs/once_cell/latest/once_cell/#faq)

We can use `cargo expand` to see the macro expansion of `lazy_static!`.  The following snippet is from `cargo expand` in `mmtk-openjdk`.

```rust
#[allow(missing_copy_implementations)]
#[allow(non_camel_case_types)]
#[allow(dead_code)]
pub struct SINGLETON_COMPRESSED {
    __private_field: (),
}
#[doc(hidden)]
pub static SINGLETON_COMPRESSED: SINGLETON_COMPRESSED = SINGLETON_COMPRESSED {
    __private_field: (),
};
impl ::lazy_static::__Deref for SINGLETON_COMPRESSED {
    type Target = MMTK<OpenJDK<true>>;
    fn deref(&self) -> &MMTK<OpenJDK<true>> {
        #[inline(always)]
        fn __static_ref_initialize() -> MMTK<OpenJDK<true>> {
            {
                if !use_compressed_oops() {
                    ::core::panicking::panic("assertion failed: use_compressed_oops()")
                }
                let mut builder = BUILDER.lock().unwrap();
                if !!MMTK_INITIALIZED.load(Ordering::Relaxed) {
                    ::core::panicking::panic(
                        "assertion failed: !MMTK_INITIALIZED.load(Ordering::Relaxed)",
                    )
                }
                set_compressed_pointer_vm_layout(&mut builder);
                let ret = mmtk::memory_manager::mmtk_init(&builder);
                MMTK_INITIALIZED.store(true, std::sync::atomic::Ordering::SeqCst);
                slots::initialize_compressed_oops_base_and_shift();
                *ret
            }
        }
        #[inline(always)]
        fn __stability() -> &'static MMTK<OpenJDK<true>> {
            static LAZY: ::lazy_static::lazy::Lazy<MMTK<OpenJDK<true>>> = ::lazy_static::lazy::Lazy::INIT;
            LAZY.get(__static_ref_initialize)
        }
        __stability()
    }
}
impl ::lazy_static::LazyStatic for SINGLETON_COMPRESSED {
    fn initialize(lazy: &Self) {
        let _ = &**lazy;
    }
}
```

A dereference `&SINGLETON` is expanded to `__stability()` which is `LAZY.get(__static_ref_initialize)` which is defined as:

```rust
pub struct Lazy<T: Sync>(Cell<MaybeUninit<T>>, Once);

impl<T: Sync> Lazy<T> {
    #[allow(deprecated)]
    pub const INIT: Self = Lazy(Cell::new(MaybeUninit::uninit()), ONCE_INIT);

    #[inline(always)]
    pub fn get<F>(&'static self, f: F) -> &T
    where
        F: FnOnce() -> T,
    {
        self.1.call_once(|| {
            self.0.set(MaybeUninit::new(f()));
        });

        // `self.0` is guaranteed to be initialized by this point
        // The `Once` will catch and propagate panics
        unsafe { &*(*self.0.as_ptr()).as_ptr() }
    }
}
```

This is very similar to `OnceLock::get_or_init` and our `InitializeOnce::initialize_once`, except it is using `Cell<MaybeUninit<T>>` instead of `UnsafeCell<MaybeUninit<T>>`.

In theory,

-   `InitializeOnce::get_ref` is the fastest because it involves no checking.  It is as fast as `once_cell::sync::OnceCell::get_unchecked`.
-   `std::sync::OnceLock::get` is as fast as `once_cell::sync::OnceCell::get`, but slower than the above.  They involve a check before getting the reference, but will not attempt to initialize it.
-   Dereferencing a `lazy_static!` should be as fast as `std::sync::OnceLock::get_or_init` and `once_cell:sync::OnceCell::get_or_init`.  They all check for completion before doing initialization.  But in order to call this function, the call site must involve the initializer function and a call to the underlying `once.call_once`, increasing the code complexity, which may affect the code cache friendliness of inlined code.

qinsoon: I quickly benchmarked the 4 options with criterion. The time showed `[lower_bound estimate upper_bound]`.
```console
once/once_lock/0        time:   [276.70 ps 277.88 ps 279.29 ps]
once/crate_once_cell/0  time:   [278.82 ps 282.20 ps 285.91 ps]
once/initialize_once/0  time:   [274.92 ps 275.69 ps 276.58 ps]
once/lazy_static/0      time:   [552.82 ps 555.12 ps 557.96 ps]
```

It is clear that we should stop using `lazy_static`. Other options are all good.

docs/dummyvm/src/slots.rs
qinsoon: Theses slots code are removed.

docs/dummyvm/src/object_model.rs
qinsoon: Done. The default value is 0 with comments to explain it.

docs/userguide/src/portingguide/howto/nogc.md
qinsoon: Done.

docs/userguide/src/portingguide/code/dummyvm/src/lib.rs
wks: > I quickly benchmarked the 4 options with criterion.

Which methods did you call, `get` or `get_unchecked`?  Is there any difference between them?

docs/dummyvm/src/object_model.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn get_type_descriptor(_reference: ObjectReference) -> &'static [i8] {
+        unimplemented!()
+    }
+
+    fn ref_to_object_start(object: ObjectReference) -> Address {
+        object.to_raw_address().sub(OBJECT_REF_OFFSET)
+    }
+
+    fn ref_to_header(object: ObjectReference) -> Address {
+        object.to_raw_address()
```
wks, line 79: We may introduce a constant `OBJECT_HEADER_OFFSET` (default to 0, too) so that the reader knows how it is usually computed.  Of course real-world VMs may simply return `object.to_raw_address()`.

```suggestion
        // Compute the address of the header from the raw address
        object.to_raw_address().sub(OBJECT_HEADER_OFFSET)
```

docs/dummyvm/src/object_model.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    fn ref_to_address(object: ObjectReference) -> Address {
+        // Just use object start.
+        Self::ref_to_object_start(object)
```
wks, line 84:
```suggestion
        // Just use the raw address if it is guaranteed to be inside the object.
        // But if the raw address may point to an address outside the object,
        // you'll need to subtract an offset from it so that the result is always inside the object.
        object.to_raw_address()
```

If `OBJECT_REF_OFFSET` is not zero, `ref_to_address` usually doesn't return the start address, but return `object`'s raw address, to reduce the amount of computing.  There are exceptions, however.  For example, JikesRVM may let ObjectRefrence point to the second field which is usually where the array payloads are.  That may need some special handling.

docs/dummyvm/src/object_model.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    fn address_to_ref(addr: Address) -> ObjectReference {
+        unsafe { ObjectReference::from_raw_address_unchecked(addr.add(OBJECT_REF_OFFSET)) }
```
wks, line 88:
```suggestion
        // This shall be the inverse operation of `ref_to_address`
        // You can assume `addr` is never zero because the zero address is never part of any object.
        unsafe { ObjectReference::from_raw_address_unchecked(addr) }
```

Changed to match the definition of `ref_to_address`.

wks (commented):
The methods `ref_to_object_start`, `ref_to_header`, `ref_to_address` and `address_to_ref` were previously defined for the convenience of some test cases, too.  We can change them to be more instructive.

docs/dummyvm/src/slots.rs
```diff
@@ -0,0 +1,73 @@
...
+    util::{Address, ObjectReference},
+    vm::slot::{MemorySlice, SimpleSlot},
+};
+
+// A binding may implement their own type of slots.
+// See https://github.com/mmtk/mmtk-core/blob/master/src/vm/tests/mock_tests/mock_test_slots.rs for different kinds of slots.
+pub type DummyVMSlot = SimpleSlot;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct DummyVMMemorySlice(*mut [ObjectReference]);
```
wks, line 12: MemorySlice is an advanced topic.  Currently it is only useful for the `memory_region_copy` barrier (and can potentially be used by `Scanning::scan_object`, too. See [this issue](https://github.com/mmtk/mmtk-core/issues/986)).  The VM binding developer will only need it when they start to support generational GC, which, in my opinion, is the fourth step in VM binding development (after (1) supporting NoGC, (2) supporting a non-copying GC such as MarkSweep, and (3) supporting a copying non-generational GC such as SemiSpace).  That may be one year or two into the VM binding development.  At that time, the VM binding developers should have enough knowledge to implement its own `MemorySlice`.  The Ruby binding has been developed for more than two years, and is just about to support StickyImmix.  But it is currently using a write barrier in the form of "this object has been modified", so it still doesn't use `MemorySlice`.

We should simply define `DummyVMMemorySlice` as `mmtk::vm::slot::UnimplementedMemorySlice` in order not to clutter the boilerplate.

But we should push an implementation like this into `mmtk::vm::slot`, probably named `SimpleMemorySlice`, to be paired with `SimpleSlot`, and then deprecate the `impl MemorySlice for Range<Address>` just like `impl Slot for Address`.

docs/dummyvm/src/object_model.rs
qinsoon: Done

docs/dummyvm/src/object_model.rs
qinsoon: No. It should be default to `ref_to_object_start`. `ref_to_object_start` is always correct as an implementation for `ref_to_address`. The binding may optimize more. I will make it clear in the comment.

docs/dummyvm/src/slots.rs
qinsoon: I removed the `slots` module.

docs/userguide/src/portingguide/code/dummyvm/src/lib.rs
qinsoon: I was using `get`.

docs/dummyvm/src/object_model.rs
qinsoon: I added some comments to make it clear that if `ref_to_address` is changed, the `address_to_ref` needs to be chagned as well.

docs/userguide/src/portingguide/howto/nogc.md
wks: Did you forget to commit the change? The document is not changed.

qinsoon: Yeah. Sorry. I had the changes on one machine, and switched to another machine to reproduce the test bug and forgot to push the changes on the first machine. I pushed the changes now.

wks (approved):
LGTM

Issue #1144, opened 2024-06-11
PrepareChunkMap (mark-sweep) unnecessarily clearing mark bits of all chunks


wks:
https://github.com/mmtk/mmtk-core/blob/401803ce44d851c19e4d04f152b494095f9b71e6/src/policy/marksweepspace/native_ms/global.rs#L435-L437

Each `PrepareChunkMap` work packet for `native_ms` is supposed to initialize the chunk-level metadata for a single chunk.  But the code shown above clears the side mark bits for all blocks.  This is totally unnecessary, and is taking up a considerable amount of time during a GC.

![image](https://github.com/mmtk/mmtk-core/assets/370317/c9641482-f83b-4f27-95b0-e7dc92d78a01)

The problem becomes more serious when I changed the native `MarkSweepSpace` to use `BlockPageResource` instead of the raw `FreeListPageResource`.  The reason may be that `BlockPageResource` caches blocks in the block queue, and releases chunks lazily, resulting in a higher number of allocated chunks during each GC.

(p.s. `ReleaseMutator` becomes faster because `BlockPageResource::release_block` is lock-free, while `FreeListPageResource::release_pages` needs a mutex to work, which is bad for parallel GC.  See: https://github.com/mmtk/mmtk-core/issues/1145)

![image](https://github.com/mmtk/mmtk-core/assets/370317/5e24d5a2-81f2-48db-a6a8-afda8df625c9)

It should only clear the side metadata for the single chunk that the work packet is responsible for.

qinsoon:
Yeah. That code is a clear bug. I guess I simply moved the mark bit clearing code from the plan's `prepare` to `PrepareChunkMap` without realizing that `PrepareChunkMap` is per chunk.

PR #1148, opened 2024-06-11
Fix mark bit clearing in PrepareChunkMap

wks:
Each PrepareChunkMap work packet (used by native MS) should only clear the side mark bits of the single chunk it is responsible for.

Fixes: https://github.com/mmtk/mmtk-core/issues/1144


qinsoon (approved):
LGTM

wks:
The broken link is a remnant from the previous PR, at which time we removed `docs/header/mmtk.h`.  I'll make another PR to fix the link.

PR #1149, opened 2024-06-11
Fix broken link to mmtk.h

wks:
`docs/header/mmtk.h` was removed in https://github.com/mmtk/mmtk-core/pull/1142 in favor for the header in the DummyVM, but the link was not updated.


qinsoon (approved):
LGTM

Issue #1145, opened 2024-06-11
MarkSweepSpace should use BlockPageResourcre


wks:
The native mark sweep space only ever allocates fixed-sized blocks from the page resource.  It is fully eligible for using `BlockPageResource` just like `ImmixSpace`.

The problem with using raw `FreeListPageResource` is that `FreeListPageResource::release_pages` needs to acquire a mutex (`FreeListPageResource::sync`).  The more GC workers there are, the worse the problem becomes.

When using a single GC worker thread:

![image](https://github.com/mmtk/mmtk-core/assets/370317/ed5eed6b-1dcd-4461-b0aa-58416ea81ed0)

When using two GC worker threads (running the same workload):

![image](https://github.com/mmtk/mmtk-core/assets/370317/8ea062ce-ba72-4621-9b67-79076895d7ce)

When using 16 GC worker threads:

![image](https://github.com/mmtk/mmtk-core/assets/370317/fd0d760f-6e73-4998-9d82-0e7cea209165)

And it is taking up a significant proportion of the GC time when running multiple workers.

![image](https://github.com/mmtk/mmtk-core/assets/370317/6ee38334-ebe3-43d3-bf82-164ad4450d36)

Issue #1146, opened 2024-06-11
Native MarkSweep: bad load balancing with single-threaded workloads


wks:
Currently, we parallelize the sweeping work by making one work packet for the global pool and one packet for each mutator.  It is OK for multi-threaded work loads, but when there is only one mutator, it hits a pathological case where the Release stage is dominated by a single long-running `ReleaseMutator` work packet.  Here is a timeline captured using eBPF when executing the Liquid benchmark using the Ruby binding (a single mutator, but multiple GC workers)

![image](https://github.com/mmtk/mmtk-core/assets/370317/b4a3baca-8af0-44ce-b787-15938b833fc9)

In comparison, here is the timeline for the lusearch benchmark in the DaCapo Chopin benchmark suite (with eager-sweeping force-enabled).  The parallel sweeping of mutators is better, but the `Release` work packet is not parallelized with `ReleaseMutator`

![image](https://github.com/mmtk/mmtk-core/assets/370317/cef45d0f-4b01-4870-9da7-d3fe32447764)

We should parallelize it by making work packets, each releasing a reasonable amount blocks.

qinsoon:
We may want to have a way to split `BlockLists` or each `BlockList` into several work packets. In that case, we can parallelize the global block lists and the thread local block lists for each mutator. The issues shown in the two graphs can be mitigated.

wks:
`ReleaseMutator` is not executed until `Release` finishes.  That's because the `Release` work packet spawns `ReleaseMutator` work packets after `Plan::release` returns.  It's a designed feature, but it means we shouldn't sweep the global pool in `Release`.

Since this problem affects all plans, I created a dedicated issue: https://github.com/mmtk/mmtk-core/issues/1147

PR #1150, opened 2024-06-11
Let MarkSweepSpace use BlockPageResource

wks:
Let `MarkSweepSpace` use the `BlockPageResource` instead of the raw `FreeListPageResource`.  This solves a problem where multiple GC workers contend for the mutex `FreeListPageResource::sync` when calling `FreeListPageResource::release_pages` concurrently during the `Release` stage.

We flush the `BlockPageResource` in `MarkSweepSpace::end_of_gc`.  To monitor the performance, we added a pair of USDT trace points before and after calling `Plan.end_of_gc`.  We also count the invocation of `end_of_gc` into the GC time.

We changed `BlockQueue` so that it uses `MaybeUninit` to hold uninitialized elements instead of relying on `B::from_aligned_address(Address::ZERO)` because the `Block` type used by `MarkSweepSpace` is based on `NonZeroUsize`, and cannot be zero.

Fixes: https://github.com/mmtk/mmtk-core/issues/1145

### DIFF
--- a/src/policy/marksweepspace/native_ms/block.rs
+++ b/src/policy/marksweepspace/native_ms/block.rs
@@ -4,6 +4,7 @@ use atomic::Ordering;
 
 use super::BlockList;
 use super::MarkSweepSpace;
+use crate::util::constants::LOG_BYTES_IN_PAGE;
 use crate::util::heap::chunk_map::*;
 use crate::util::linear_scan::Region;
 use crate::vm::ObjectModel;
@@ -48,6 +49,9 @@ impl Region for Block {
 }
 
 impl Block {
+    /// Log pages in block
+    pub const LOG_PAGES: usize = Self::LOG_BYTES - LOG_BYTES_IN_PAGE as usize;
+
     pub const METADATA_SPECS: [SideMetadataSpec; 7] = [
         Self::MARK_TABLE,
         Self::NEXT_BLOCK_TABLE,
--- a/src/policy/marksweepspace/native_ms/global.rs
+++ b/src/policy/marksweepspace/native_ms/global.rs
@@ -359,6 +359,11 @@ impl<VM: VMBinding> MarkSweepSpace<VM> {
 
         #[cfg(debug_assertions)]
         from.assert_empty();
+
+        // BlockPageResource uses worker-local block queues to eliminate contention when releasing
+        // blocks, similar to how the MarkSweepSpace caches blocks in `abandoned_in_gc` before
+        // returning to the global pool.  We flush the BlockPageResource, too.
+        self.pr.flush_all();
     }
 
     /// Release a block.
--- a/src/policy/marksweepspace/native_ms/global.rs
+++ b/src/policy/marksweepspace/native_ms/global.rs
@@ -7,7 +7,7 @@ use crate::{
     scheduler::{GCWorkScheduler, GCWorker},
     util::{
         copy::CopySemantics,
-        heap::{FreeListPageResource, PageResource},
+        heap::{BlockPageResource, PageResource},
         metadata::{self, side_metadata::SideMetadataSpec, MetadataSpec},
         ObjectReference,
     },
@@ -64,7 +64,7 @@ pub enum BlockAcquireResult {
 /// | GC - End of GC | -                                               | Merge the temp global lists                  | -         |
 pub struct MarkSweepSpace<VM: VMBinding> {
     pub common: CommonSpace<VM>,
-    pr: FreeListPageResource<VM>,
+    pr: BlockPageResource<VM, Block>,
     /// Allocation status for all chunks in MS space
     chunk_map: ChunkMap,
     /// Work packet scheduler
@@ -79,6 +79,8 @@ pub struct MarkSweepSpace<VM: VMBinding> {
     /// and will be moved to the abandoned lists above at the end of a GC.
     abandoned_in_gc: Mutex<AbandonedBlockLists>,
 }
+
+unsafe impl<VM: VMBinding> Sync for MarkSweepSpace<VM> {}
 
 pub struct AbandonedBlockLists {
     pub available: BlockLists,
@@ -278,9 +280,19 @@ impl<VM: VMBinding> MarkSweepSpace<VM> {
         let common = CommonSpace::new(args.into_policy_args(false, false, local_specs));
         MarkSweepSpace {
             pr: if is_discontiguous {
-                FreeListPageResource::new_discontiguous(vm_map)
+                BlockPageResource::new_discontiguous(
+                    Block::LOG_PAGES,
+                    vm_map,
+                    scheduler.num_workers(),
+                )
             } else {
-                FreeListPageResource::new_contiguous(common.start, common.extent, vm_map)
+                BlockPageResource::new_contiguous(
+                    Block::LOG_PAGES,
+                    common.start,
+                    common.extent,
+                    vm_map,
+                    scheduler.num_workers(),
+                )
             },
             common,
             chunk_map: ChunkMap::new(),
@@ -354,7 +366,7 @@ impl<VM: VMBinding> MarkSweepSpace<VM> {
         self.block_clear_metadata(block);
 
         block.deinit();
-        self.pr.release_pages(block.start());
+        self.pr.release_block(block);
     }
 
     pub fn block_clear_metadata(&self, block: Block) {
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -530,6 +530,12 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         // Tell GC trigger that GC ended - this happens before we resume mutators.
         mmtk.gc_trigger.policy.on_gc_end(mmtk);
 
+        // All other workers are parked, so it is safe to access the Plan instance mutably.
+        probe!(mmtk, plan_end_of_gc_begin);
+        let plan_mut: &mut dyn Plan<VM = VM> = unsafe { mmtk.get_plan_mut() };
+        plan_mut.end_of_gc(worker.tls);
+        probe!(mmtk, plan_end_of_gc_end);
+
         // Compute the elapsed time of the GC.
         let start_time = {
             let mut gc_start_time = worker.mmtk.state.gc_start_time.borrow_mut();
@@ -564,10 +570,6 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                 mmtk.get_plan().get_used_pages()
             );
         }
-
-        // All other workers are parked, so it is safe to access the Plan instance mutably.
-        let plan_mut: &mut dyn Plan<VM = VM> = unsafe { mmtk.get_plan_mut() };
-        plan_mut.end_of_gc(worker.tls);
 
         #[cfg(feature = "extreme_assertions")]
         if crate::util::slot_logger::should_check_duplicate_slots(mmtk.get_plan()) {
--- a/src/util/heap/blockpageresource.rs
+++ b/src/util/heap/blockpageresource.rs
@@ -8,10 +8,12 @@ use crate::util::heap::pageresource::CommonPageResource;
 use crate::util::heap::space_descriptor::SpaceDescriptor;
 use crate::util::linear_scan::Region;
 use crate::util::opaque_pointer::*;
+use crate::util::rust_util::zeroed_alloc::new_zeroed_vec;
 use crate::vm::*;
 use atomic::Ordering;
 use spin::RwLock;
 use std::cell::UnsafeCell;
+use std::mem::MaybeUninit;
 use std::sync::atomic::AtomicUsize;
 use std::sync::Mutex;
 
@@ -179,17 +181,29 @@ impl<VM: VMBinding, B: Region> BlockPageResource<VM, B> {
 
 /// A block list that supports fast lock-free push/pop operations
 struct BlockQueue<B: Region> {
+    /// The number of elements in the queue.
     cursor: AtomicUsize,
-    data: UnsafeCell<Vec<B>>,
+    /// The underlying data storage.
+    ///
+    /// -   `UnsafeCell<T>`: It may be accessed by multiple threads.
+    /// -   `Box<[T]>`: It holds an array allocated on the heap.  It cannot be resized, but can be
+    ///     replaced with another array as a whole.
+    /// -   `MaybeUninit<T>`: It may contain uninitialized elements.
+    ///
+    /// The implementaiton of `BlockQueue` must ensure there is no data race, and it never reads
+    /// uninitialized elements.
+    data: UnsafeCell<Box<[MaybeUninit<B>]>>,
 }
 
 impl<B: Region> BlockQueue<B> {
     /// Create an array
     fn new() -> Self {
-        let default_block = B::from_aligned_address(Address::ZERO);
+        let zeroed_vec = unsafe { new_zeroed_vec(Self::CAPACITY) };
+        let boxed_slice = zeroed_vec.into_boxed_slice();
+        let data = UnsafeCell::new(boxed_slice);
         Self {
             cursor: AtomicUsize::new(0),
-            data: UnsafeCell::new(vec![default_block; Self::CAPACITY]),
+            data,
         }
     }
 }
@@ -199,14 +213,14 @@ impl<B: Region> BlockQueue<B> {
 
     /// Get an entry
     fn get_entry(&self, i: usize) -> B {
-        unsafe { (*self.data.get())[i] }
+        unsafe { (*self.data.get())[i].assume_init() }
     }
 
     /// Set an entry.
     ///
     /// It's unsafe unless the array is accessed by only one thread (i.e. used as a thread-local array).
     unsafe fn set_entry(&self, i: usize, block: B) {
-        (*self.data.get())[i] = block
+        (*self.data.get())[i].write(block);
     }
 
     /// Non-atomically push an element.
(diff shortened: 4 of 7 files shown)


wks:
The `ImmixSpace` flushes the `BlockPageResource` when all the `SweepChunk` work packets finish.  This is hard to implement without significant changes because the `ReleaseMutator` work packet is created by the plan-agnostic `Release` work packet, without any context about the `MarkSweepSpace`.  It is hard to intercept the creation of `ReleaseMutator` work packet and inject an `Arc<FlushPageResource>` into each of them.  I just did what the `MarkSweepSpace` already does for flushing the `abandoned_in_gc` lists -- flushing it in `end_of_gc`.

The time taken to execute `end_of_gc` is trivial for SemiSpace and Immix (about 1us).  It is still trivial (but not that trivial) for MarkSweep (8us).  I'll take a look at whether it spent the time flushing the `abandoned_in_gc` or flushing the `BlockPageResource`.

wks:
The time spent flushing the `abandoned_in_gc` lists is slightly more than flushing the BlockPageResource.  I think it depends on the ratio of marked blocks and unmarked blocks.  In this particular GC (when running MarkSweep, lusearch, 72M heap size), it took 9.691us to flush them all.

image

When the heap is bigger, the time to flush will increase.  The following timeline is from the PMD benchmark (600M heap size).  The time to flush `abandoned_in_gc` is longer.  The time to flush the BlockPageResource remain constant.  I think that's because the block queue in the BlockPageResource flushes itself from time to time whenever it overflows.  That increased the chance of contention, but reduced the time to flush in the end of GC.

image

k-sareen:
> The `ImmixSpace` flushes the `BlockPageResource` when all the `SweepChunk` work packets finish. This is hard to implement without significant changes because the `ReleaseMutator` work packet is created by the plan-agnostic `Release` work packet, without any context about the `MarkSweepSpace`. It is hard to intercept the creation of `ReleaseMutator` work packet and inject an `Arc<FlushPageResource>` into each of them. I just did what the `MarkSweepSpace` already does for flushing the `abandoned_in_gc` lists -- flushing it in `end_of_gc`.
> 
> The time taken to execute `end_of_gc` is trivial for SemiSpace and Immix (about 1us). It is still trivial (but not that trivial) for MarkSweep (8us). I'll take a look at whether it spent the time flushing the `abandoned_in_gc` or flushing the `BlockPageResource`.

Why can't we use a sentinel to flush it at the end?

wks:
> Why can't we use a sentinel to flush it at the end?

There are two difficulties.

(1) To know if it is at the end, we need to know the number of units of work beforehand.  In `ImmixSpace`, the number of units is the number of `SweepChunk` work packets.  The number is known in `generate_sweep_tasks` (`tasks.len()`)

```rust
    fn generate_sweep_tasks(&self) -> Vec<Box<dyn GCWork<VM>>> {
        self.defrag.mark_histograms.lock().clear();
        // # Safety: ImmixSpace reference is always valid within this collection cycle.
        let space = unsafe { &*(self as *const Self) };
        let epilogue = Arc::new(FlushPageResource {
            space,
            counter: AtomicUsize::new(0),
        });
        let tasks = self.chunk_map.generate_tasks(|chunk| {
            Box::new(SweepChunk {
                space,
                chunk,
                epilogue: epilogue.clone(),
            })
        });
        epilogue.counter.store(tasks.len(), Ordering::SeqCst);
        tasks
    }
```

In `MarkSweepSpace`, the number of units of work is the number of `ReleaseMutator` work packets plus the `Release` work packet itself.  We could count the number of mutators, but as we discussed in https://github.com/mmtk/mmtk-core/issues/1146, for load balancing, it's better to break up each `ReleaseMutator` into fine-grained work packets, and the number of those work packets is not known until all `ReleaseMutator` packets have iterated through the block lists of all mutators.

We can't let some threads (such as workers executing `ReleaseMutator`) increment the tasks count (by generating fine-grained work packets) while other threads decrement the tasks count (by completing the fine-grained work packets).  In that case, the counter will oscillate between 0 and positive numbers.  As a result, the BlockPageResource may end up being flushed multiple times, and two threads may flush at the same time.  But `BockPageResource::flush_all()` is not thread-safe.

(2) In order to use the sentinel, we need to inject an `Arc<FlushPageResource>` into each unit of work.  In `ImmixSpace`, `generate_sweep_tasks` generates all `SweepChunk` work packets, and assigns the `epilogue` field of each of them.  But since `ReleaseMutator` is created in `Release::do_work` (which is plan-agnostic), we can't intercept it and add a field to the `ReleaseMutator` work packet (which is also plan-agnostic).  We could embed the `FlushPageResource` data structure in `MarkSweepSpace`, but that still has the first problem I mentioned above.

So the easiest way to do it is do it in `end_of_gc` like how `MarkSweep` currently flushes `abandoned_in_gc`.

Alternatively, we can make a "sentinel" that is executed when the `Release` bucket is drained.  That should be just as fast (or as slow) as doing it in `end_of_gc`.

wks:
Performance evaluation:

Three builds:

-   `master`: The master branch.
-   `masterfix`: The master branch, but count the time of `Plan.end_of_gc` into GC time, just like this PR.
-   `pr`: This PR.

I added the `masterfix` build because this PR changed the way we measure GC (STW) time.  The `masterfix` will be the representative of the old `MarkSweepSpace` that used the raw `FreeListPageResource`, and the difference between the `masterfix` and the `pr` build will be the effect of using `BlockPageResource` instead.

I ran `lusearch` from DaCapo Chopin on lynx.moma (only using the 8 big cores).  I used 4 heap sizes: 51MB, 64MB, 77MB and 92MB, which are 2.428x, 3.023x, 3.678x and 4.392x min heap size w.r.t. G1 in OpenJDK.  10 invocations for each heap size. (`running runbms -i 10 ... 8 3 4 5 6`)  I included both the MarkSweep and Immix plans.

Here are the line plots ([plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-12-Wed-111108&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|60&build&mmtk_gc|20&1^invocation|70&build-mmtk_gc^ms-block-pr|41&XY%20(with%20CI)^build-mmtk_gc^hfac&)).  Each line is a build-plan tuple.  The x axis is the heap factor (N/1000 min heap size).  The "geomean" near 0 should be "2428", and that's probably a bug in plotty.

image

We can see a tiny amount of difference between the builds for MarkSweep, but almost no difference for Immix.

The number of GC overflowed and stuck at 2048 at the smallest heap factor.

The `time.other` is much lower at the smallest heap factor for MarkSweep.  If the number is correct, it means mutators run faster when the heap is small. This is reproducible on my laptop, too.  I'll investigate if that's another performance anomaly of MarkSweep.

Here is a zoomed-in line plot for `time.stw`.

image

And the histogram of `time.stw` (only MarkSweep) normalized to `masterfix`. ([plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-12-Wed-111108&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&mmtk_gc^2^Immix|60&build&mmtk_gc|20&1^invocation|70&build-mmtk_gc^ms-block-pr|30&1&hfac&build-mmtk_gc;build3-MarkSweep|41&Histogram%20(with%20CI)^build-mmtk_gc^hfac&))

image

When the heap is small, the noise is large; but as the heap grows larger, this PR has smaller and smaller STW time.  That's because the larger the heap is, the more blocks there are to sweep, and the more blocks there are to release.  The lock contention problem with `FreeListPageResource::release_pages` becomes more serious, while advantage of the lock-free method `BlockPageResource::release_block` becomes greater.

FYI, the STW time when running Immix has no differences beyond noise between the three builds.  Here is a histogram.  ([plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-12-Wed-111108&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4&mmtk_gc^2^MarkSweep|60&build&mmtk_gc|20&1^invocation|70&build-mmtk_gc^ms-block-pr|30&1&hfac&build-mmtk_gc;build3-Immix|41&Histogram%20(with%20CI)^build-mmtk_gc^hfac&))

image

I'll try running the `pmd` benchmark which has a much larger min heap size than `lusearch`.

qinsoon:
The improvement looks marginal. I wonder if it is because you were using only 8 GC threads and there wasn't much contention in the release phase. Did you check the pathological case you found in https://github.com/mmtk/mmtk-core/issues/1145? Does this PR improve it? Nonetheless, I think switching to `BlockPageResource` is the right thing to do.

wks:
> The improvement looks marginal. I wonder if it is because you were using only 8 GC threads and there wasn't much contention in the release phase. Did you check the pathological case you found in #1145? Does this PR improve it? Nonetheless, I think switching to `BlockPageResource` is the right thing to do.

It was 16 GC threads.  I used `taskset -c 0-15` because each big core on `lynx.moma` has two SMT threads, appearing to the OS as two processors.

The pathological case in #1145 was using 100M heap size, so the difference is greater in that run.

master 72M (ran on my laptop):
image

master 100M (you can see the time to release mutators became much longer than 72M):
image

And yes.  This PR improves the speed of releasing mutators at 100M, too.
image

wks:
The `pmd` benchmark, with 7 different heap sizes ranging from 358M to 1134M, which are 1.892x to 6.000x min heap size w.r.t. G1in OpenJDK. It ran 20 invocations per heap size instead of 10.

Line plots: ([plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-12-Wed-162415&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|10&iteration^1^4|60&build&mmtk_gc|20&1^invocation|70&build-mmtk_gc^ms-block-pr|41&XY%20(with%20CI)^build-mmtk_gc^hfac&))

image

Histogram (MarkSweep-only), normalized to `masterfix`
image

The result isn't as exciting as I expected.  At 6x min heap size, this PR only has 6%-7% STW time reduction.  It may be because `pmd` doesn't have such a high heap turnaround ratio as `lusearch`.  In other words, it tends to retain more blocks rather than recycling them.

We observe a small increase in STW time from 1.8x to 3x min heap sizes, as a result of increased number of GC.  It may be a result of `BlockPageResource` not unmapping memory, making the heap tighter for other spaces (such as LOS), therefore triggering GC more often.  There is a comment in `BlockPageResource::flush_all`:

```rust
    pub fn flush_all(&self) {
        self.block_queue.flush_all()
        // TODO: For 32-bit space, we may want to free some contiguous chunks.
    }
```

qinsoon (approved):
The code looks good. We shall wait for Kunshan to run another round of evaluation for the PR to include all the benchmarks.

wks:
I ran many other benchmarks in DaCapo Chopin (excluding some that are known to produce inconsistent results as we increase the number of iterations, likely due to the lack of reference processing).  I ran with two builds (`masterfix` and `pr`), the MarkSweep, with three heap factors (3.023x, 3.678x and 4.392x).

There are too many plots, so I only post Plotty links.

-   [STW time, not normalized](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-13-Thu-115141&benchmark^build^hfac^invocation^iteration^mmtk_gc&bmtime^time^time.other^time.stw&|10&iteration^1^4|60&build&mmtk_gc|70&build-mmtk_gc^ms-block-pr|20&1^invocation|40&Histogram%20(with%20CI)^build-mmtk_gc^hfac&time.stw)
-   [STW time, normalized to `masterfix`](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-13-Thu-115141&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^bmtime^time^time.other^time.stw&|10&iteration^1^4|60&build&mmtk_gc|70&build-mmtk_gc^ms-block-pr|20&1^invocation|30&1&benchmark^hfac&build-mmtk_gc;build3-MarkSweep|40&Histogram%20(with%20CI)^build-mmtk_gc^hfac&time.stw)
-   [total time, not normalized](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-13-Thu-115141&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^bmtime^time^time.other^time.stw&|10&iteration^1^4|60&build&mmtk_gc|70&build-mmtk_gc^ms-block-pr|20&1^invocation|40&Histogram%20(with%20CI)^build-mmtk_gc^hfac&time)
-   [total time, normalized to `masterfix`](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-13-Thu-115141&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^bmtime^time^time.other^time.stw&|10&iteration^1^4|60&build&mmtk_gc|70&build-mmtk_gc^ms-block-pr|20&1^invocation|30&1&benchmark^hfac&build-mmtk_gc;build3-MarkSweep|40&Histogram%20(with%20CI)^build-mmtk_gc^hfac&time)

Most benchmarks have reduced STW time with this PR.

The benchmark that has the highest STW time reduction is graphchi, with more than 60% reduction at 4.392x min heap size (w.r.t. G1 in OpenJDK).  eBPF timeline shows graphchi spent a lot of time releasing mutators, likely because it simply has too many blocks to release in each GC, and the effect is amplified by the lock contention in `masterfix`.

masterfix:
image

pr:
image

The time to compute the transitive closure is also reduced, and it is likely to be a result of BlockPageResource reusing existing blocks, but I am not certain about it.

There are some benchmarks that have increased STW time instead.  They are cassandra, fop, jython and spring.  Their error bars are large, and I cannot reproduce their slowdown on my laptop.  The number of GC is small for both cassandra and fop (see [this plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-13-Thu-115141&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^bmtime^time^time.other^time.stw&|10&iteration^1^4|60&build&mmtk_gc|70&build-mmtk_gc^ms-block-pr|20&1^invocation|40&Histogram%20(with%20CI)^build-mmtk_gc^hfac&GC) for the number of GC), and the GCs concentrate near the end of the benchmark execution.  Such benchmarks are more susceptible to random effects.  For fop, some GCs may spend 50% of the time tracing descendants of finalizable objects.  So the variation of STW time is more related to when and how much finalization happened instead of the time spent executing ReleaseMutator. See the timeline below

image

The effect on total time is observable, but not that dramatic.

Overall, I think this PR is effective in reducing the time of ReleaseMutator, especially in the pathological cases where the heap is large and the number of blocks to be released is high.

qinsoon:
> There are too many plots, so I only post Plotty links.
> 
> * [STW time, not normalized](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Clynx-2024-06-13-Thu-115141&benchmark%5Ebuild%5Ehfac%5Einvocation%5Eiteration%5Emmtk_gc&bmtime%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4%7C60&build&mmtk_gc%7C70&build-mmtk_gc%5Ems-block-pr%7C20&1%5Einvocation%7C40&Histogram%20(with%20CI)%5Ebuild-mmtk_gc%5Ehfac&time.stw)
> * [STW time, normalized to `masterfix`](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Clynx-2024-06-13-Thu-115141&benchmark%5Ebuild%5Ehfac%5Einvocation%5Eiteration%5Emmtk_gc&GC%5Ebmtime%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4%7C60&build&mmtk_gc%7C70&build-mmtk_gc%5Ems-block-pr%7C20&1%5Einvocation%7C30&1&benchmark%5Ehfac&build-mmtk_gc;build3-MarkSweep%7C40&Histogram%20(with%20CI)%5Ebuild-mmtk_gc%5Ehfac&time.stw)
> * [total time, not normalized](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Clynx-2024-06-13-Thu-115141&benchmark%5Ebuild%5Ehfac%5Einvocation%5Eiteration%5Emmtk_gc&GC%5Ebmtime%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4%7C60&build&mmtk_gc%7C70&build-mmtk_gc%5Ems-block-pr%7C20&1%5Einvocation%7C40&Histogram%20(with%20CI)%5Ebuild-mmtk_gc%5Ehfac&time)
> * [total time, normalized to `masterfix`](https://squirrel.anu.edu.au/plotty/wks/noproject/#0%7Clynx-2024-06-13-Thu-115141&benchmark%5Ebuild%5Ehfac%5Einvocation%5Eiteration%5Emmtk_gc&GC%5Ebmtime%5Etime%5Etime.other%5Etime.stw&%7C10&iteration%5E1%5E4%7C60&build&mmtk_gc%7C70&build-mmtk_gc%5Ems-block-pr%7C20&1%5Einvocation%7C30&1&benchmark%5Ehfac&build-mmtk_gc;build3-MarkSweep%7C40&Histogram%20(with%20CI)%5Ebuild-mmtk_gc%5Ehfac&time)

The performance looks good. We see up to 10% improvement of STW time in geomean across all the measured benchmarks in 3x/3.6x/4.3x heap size, about 1% improvement in total time. You can merge the PR at any time.

PR #1151, opened 2024-06-17
Fix clippy warnings for Rust 1.79

wks:
This PR fixes new warnings generated by Clippy after upgrading Rust to 1.79.

-   Only define `ImmixSpaceArg::mixed_age` when the "vo_bit" feature is enabled.  This eliminates a dead code warning.
-   Use the associated constant `<integer_type>::MAX` (and `MIN`, too) instead of the deprecated `max_value()` method and the `::std::<integer_type>::MAX` constant.
-   Annotate the destination type of some `transmute` calls.
-   Use the safe `std::ptr::NonNull::from()` converter instead of `transmute`.
-   Initialize array elements of type `Atomic<MapState>` from constant.
-   Remove `MutatorContext::barrier_impl`.  It is not used by any bindings right now, and it depends on the layout of `&dyn Trait`.  Since `Barrier` has `Downcast` as a supertrait, if a binding needs a reference to a concrete `Barrier` implementation, it should just use downcast.

The warning against the `transmute` call in `mock_vm.rs` is suppressed because it is almost impossible to annotate.

Also fixes compilation problems:

-   Made benchmarks related to the "mock_test" feature conditionally compiled so that we don't get compilation error when running `cargo clippy --benches` or `cargo clippy --all-targets`.


wks:
Public API Check failed because I removed `MutatorContext::barrier_impl()`.  `DEFAULT_STRESS_FACTOR` is not changed, but `cargo-public-api` thought its type was `_`.

src/plan/mutator_context.rs
```diff
@@ -316,11 +309,6 @@ pub trait MutatorContext<VM: VMBinding>: Send + 'static {
     fn get_tls(&self) -> VMMutatorThread;
     /// Get active barrier trait object
     fn barrier(&mut self) -> &mut dyn Barrier<VM>;
-    /// Force cast the barrier trait object to a concrete implementation.
-    ///
-    /// # Safety
-    /// The safety of this function is ensured by a down-cast check.
-    unsafe fn barrier_impl<B: Barrier<VM>>(&mut self) -> &mut B;
```
qinsoon, line 323: I agree a binding can implement this method totally on the binding side if they need to. We can remove it.

The comments for write barriers methods mention using a specialized slow-path but they do not specifically mention this method. The comments seem to make sense even if we remove this method from our API. No change is required -- I just mention it in case you would like to improve the comments related to the function.
https://github.com/mmtk/mmtk-core/blob/47562b74c73ec74ab46e9b578fe3919346010b77/src/memory_manager.rs#L225

rust-toolchain
```diff
@@ -1 +1 @@
-1.77.0
+stable
```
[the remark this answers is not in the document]
wks: Oops. My bad.  I reverted this.

qinsoon (approved):
LGTM

src/plan/mutator_context.rs
wks: Yes.  The comment mentioned slow path, but it does not mention how to make the "specialized slow-path call".  Calling methods of the `Barrier` trait is only one step of it.  I think it is obvious to use downcast to eliminate a dynamic call for a trait that supports it.

PR #1154, opened 2024-06-25
Display number of slots in timeline visualization

wks:
In eBPF-based timeline visualization, we add the number of slots and whether the workpacket is from roots as arguments of the `ProcessEdgesWork` work packets.


wks:
FYI, the arguments (`is_roots` and `num_slots`) can be seen on the bottom panel of Perfetto UI when you click on a ProcessEdgesWork work packet.
image

qinsoon (approved):

PR #1155, opened 2024-06-25
Add a simple implementation for find_object_from_internal_pointer

qinsoon:
This PR adds a function to our API `find_object_from_internal_pointer`. It uses a simple implementation with `is_vo_bit_set_for_addr` which is not efficient. 

An efficient implementation should be able to bulk load side metadata, check the last bit set and deduce the data address from the bit. However, there are many corner cases for this that I haven't sorted out, and it would need more testing before we can make it ready. I think it might be a good idea to have the API first with a naive implementation which works. We can optimize it later.

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -658,10 +658,6 @@ pub fn find_object_from_internal_pointer<VM: VMBinding>(
     internal_ptr: Address,
     max_search_bytes: usize,
 ) -> Option<(Address, Address)> {
-    if !is_mapped_address(internal_ptr) {
-        return None;
-    }
-
     crate::util::metadata::vo_bit::search_vo_bit_before_addr::<VM>(internal_ptr, max_search_bytes)
 }
 
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -630,6 +630,36 @@ pub fn is_mmtk_object(addr: Address) -> bool {
     SFT_MAP.get_checked(addr).is_mmtk_object(addr)
 }
 
+/// Find if there is an object with VO bit set for the given address range.
+/// This is used instead of [`is_mmtk_object`] for conservative stack scanning if
+/// the binding may have internal pointers on the stack.
+///
+/// The function cannot directly return an object reference. Instead, it returns
+/// an address that is aligned to [`crate::util::is_mmtk_object::VO_BIT_REGION_SIZE`],
+/// and guarantees there is an object in `[return_address, return_address + VO_BIT_REGION_SIZE)`.
+/// The reason behind this is that we use the VO
+/// bit to represent a valid object for every 8 bytes. When we find a set VO bit,
+/// we only know that in the 8 bytes there is an object, and we cannot know where
+/// exactly the object is. The binding needs to use their knowledge about the alignment
+/// and offset for object reference to find out the object reference from the return value.
+/// See Case 2 in [`crate::memory_manager::is_mmtk_object`] for more explanation.
+///
+/// Argument:
+/// * `internal_ptr`: The internal pointer to start. We search backwards from this internal pointer to find the base reference.
+/// * `object_max_bytes`: The maximum of the possible object size. We will search for this many before deciding there is no object.
+///   The binding should keep this value low for better performance.
+#[cfg(feature = "is_mmtk_object")]
+pub fn find_object_from_internal_pointer<VM: VMBinding>(
+    internal_ptr: Address,
+    object_max_bytes: usize,
+) -> Option<Address> {
+    if !is_mapped_address(internal_ptr) {
+        return None;
+    }
+
+    crate::util::metadata::vo_bit::search_vo_bit_before_addr::<VM>(internal_ptr, object_max_bytes)
+}
+
 /// Return true if the `object` lies in a region of memory where
 /// -   only MMTk can allocate into, or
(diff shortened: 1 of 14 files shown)


src/util/metadata/vo_bit/mod.rs
```diff
@@ -98,45 +98,36 @@ pub fn is_vo_bit_set<VM: VMBinding>(object: ObjectReference) -> bool {
...
-    if !VO_BIT_SIDE_METADATA_SPEC.is_mapped(addr) {
-        return None;
-    }
-
-    if VO_BIT_SIDE_METADATA_SPEC.load_atomic::<u8>(addr, Ordering::SeqCst) == 1 {
-        Some(potential_object)
-    } else {
-        None
-    }
+pub fn is_vo_bit_set_for_addr<VM: VMBinding>(address: Address) -> bool {
```
qinsoon, line 101: This function used to return `ObjectReference`. However we cannot be sure that the `ObjectReference` is actually valid. I just change it to return boolean, and let the caller figure out the object reference.

src/util/metadata/vo_bit/mod.rs
```diff
@@ -98,45 +98,36 @@ pub fn is_vo_bit_set<VM: VMBinding>(object: ObjectReference) -> bool {
...
-pub unsafe fn is_vo_bit_set_unsafe<VM: VMBinding>(address: Address) -> Option<ObjectReference> {
-    let potential_object = ObjectReference::from_raw_address(address)?;
+pub unsafe fn is_vo_bit_set_unsafe<VM: VMBinding>(address: Address) -> bool {
+    _is_vo_bit_set::<false, VM>(address)
+}
 
-    let addr = potential_object.to_address::<VM>();
+fn _is_vo_bit_set<const ATOMIC: bool, VM: VMBinding>(address: Address) -> bool {
+    if let Some(potential_object) = ObjectReference::from_raw_address(address) {
+        let addr = potential_object.to_address::<VM>();
```
wks, line 117: I see this is a bit complicated.

Would it make things easier if we change the contract of `is_mmtk_object(addr)` so that `addr` must always be 8 byte aligned (or whatever the byte-per-region of `VO_BIT` is)?  In this case, we always check if a VO bit is set at address `addr`.  But when the VM binding does conservative stack scanning, it needs to decide which address to query according to its offset of `ref_to_address`.

I have already thought about `is_mmtk_object` being unsound in the sense that the argument `addr` is not always a valid object reference.  Maybe we just provide a safe variant of `is_vo_bit_set` that checks against the SFT and the memory mapping for the VM binding.  And we explicitly tell the VM that the VO bit is set at `objref.to_address()` rounded down to a multiple of 8 bytes, and let the VM binding figure out how to use it wisely.

qinsoon: > Would it make things easier if we change the contract of `is_mmtk_object(addr)` so that `addr` must always be 8 byte aligned (or whatever the byte-per-region of `VO_BIT` is)? In this case, we always check if a VO bit is set at address `addr`. But when the VM binding does conservative stack scanning, it needs to decide which address to query according to its offset of `ref_to_address`.

That will be annoying to the bindings, and expose unnecessary internal details.

I think our current code tries to be unnecessarily general. This results in more bugs, complex code and inefficiency. I would rather introduce some sane assumptions to make both our API and the internal implementation simpler.

wks: We already allows the VM binding to set the VO bit in the allocation fast path which inlines the `post_alloc` function.  So the VM binding already knows where VO bit should be set.  So changing the contract to explicitly mention VO bit does not add more burden to the VM binding.

qinsoon: > We already allows the VM binding to set the VO bit in the allocation fast path which inlines the `post_alloc` function. So the VM binding already knows where VO bit should be set. So changing the contract to explicitly mention VO bit does not add more burden to the VM binding.

What you said is an optimization. It just copies what we do in mmtk-core, so it naturally assumes our internal implementation. But what we discussed is the API -- it should not expose the internal implementation where possible.

src/memory_manager.rs
```diff
@@ -630,6 +630,37 @@ pub fn is_mmtk_object(addr: Address) -> bool {
...
+/// To minimize the cost, the user should also use a small `max_search_bytes`.
+///
+/// Argument:
+/// * `internal_ptr`: The internal pointer to start. We search backwards from this internal pointer to find the base reference.
+/// * `max_search_bytes`: The maximum number of bytes we may search for an object with VO bit set.
+#[cfg(feature = "is_mmtk_object")]
+pub fn find_object_from_internal_pointer<VM: VMBinding>(
+    internal_ptr: Address,
+    max_search_bytes: usize,
+) -> Option<(Address, Address)> {
```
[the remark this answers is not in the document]
k-sareen: `Option<Address>` is unfortunately larger than `usize`. Maybe we need to revisit `Address` and make it non-nullable as well to make `Option<Address>` easier to use as well as make it more idiomatic Rust

qinsoon:
Other than the inefficiency, one big issue for this PR is that the address we return may not be the base reference for the given address. It could be the previous object that ends before the address. We will have to check against the object size to know if the given address is within the object range. However, we cannot get the object size as we don't know the object reference.

I think we would want to enforce object reference alignment so we can know the object reference using VO bit. I feel that is essential.

k-sareen:
Superseded by #1165

Issue #804, opened 2023-05-05
Meta: Refactoring / fixing related to valid-object (VO) bit


wks:
A previous attempt to address many of the following problems was unsuccessful.  https://github.com/mmtk/mmtk-core/pull/680

We should do the following one by one in separate PRs.

-   [X] **Renaming alloc bit to VO bit.** https://github.com/mmtk/mmtk-core/pull/791
-   [ ] **Add an API for explicitly clearing the VO bit.** https://github.com/mmtk/mmtk-core/issues/648
-   [ ] **Remove the `is_mmtk_object` feature.**  It should be enabled as long as VO bit is available.
-   [ ] **Refactor `vo_bit.rs`**
    -   Clean up the API, and only use `ObjectReference` (not `Address`) to access the VO bit.  This emphasises that VO bits are set at object references' raw addresses.
    -   Use `fetch_and` and `fetch_or` for atomic setting/clearing of bits.
-   [ ] **MarkCompactSpace shall use local bits.** It should use space-local allocation bits to identify objects.
-   [ ] **MallocSpace shall use local alloc bits.** Its semantics is different from VO bit.  Even after the VO bit is cleared, malloc MarkSweep still needs the alloc bit to identify objects allocated by `malloc`, and call `free` on them.

PR #1157, opened 2024-06-26
Expose VO bit functions

udesou:
Currently Julia needs to bulk set the VO bit for boot image regions of code. 
While hacky, this solution seems to work and shouldn't be too problematic given we only do that for permalloc-ed objects. 
This PR makes the functions `bzero_vo_bit` and `bset_vo_bit` public via `util/metadata/side_metadata/global.rs` so that it can be called by the Julia binding.


src/util/metadata/side_metadata/global.rs
```diff
@@ -12,6 +12,18 @@ use std::fmt;
 use std::io::Result;
 use std::sync::atomic::{AtomicU8, Ordering};
 
+#[cfg(feature = "vo_bit")]
```
wks, line 15: It's just not right to put VO bit specific functions in `side_metadata::global`.  I suggest we make the `vo_bit` module public, and mark functions not intended to be public as `pub(crate)`.  It is actually part of the plan to make `vo_bit` public.  See: https://github.com/mmtk/mmtk-core/issues/804

qinsoon:
I am not sure if it is a good idea to bulk set vo bit. VO bit means a valid object. Bulk setting vo bits clearly violates the contract. Addresses with vo bits set may not be an object.

For example, when we have internal pointers and we try to find the base pointer, MMTk will give incorrect results, as the contract about VO bits is broken.

However, I dont know what we should do for Julia's boot image. Is it possible to find each object in the boot image? I looked at https://github.com/mmtk/julia/blob/d907a0630b4551cfbb321366da6e96242ada25d1/src/staticdata.c#L2844, and it doesn't look like an easy thing to do.

udesou:
> I am not sure if it is a good idea to bulk set vo bit. VO bit means a valid object. Bulk setting vo bits clearly violates the contract. Addresses with vo bits set may not be an object.
> 
> For example, when we have internal pointers and we try to find the base pointer, MMTk will give incorrect results, as the contract about VO bits is broken.
> 
> However, I dont know what we should do for Julia's boot image. Is it possible to find each object in the boot image? I looked at https://github.com/mmtk/julia/blob/d907a0630b4551cfbb321366da6e96242ada25d1/src/staticdata.c#L2844, and it doesn't look like an easy thing to do.

I guess the other option is to ignore immortal objects altogether. Would that make sense?

qinsoon:
This can be closed. https://github.com/mmtk/mmtk-core/pull/1248 exposed the VO bit under a feature, and Julia is using the feature.

Issue #1147, opened 2024-06-11
Do not do any significant amount of work in `Release`


wks:
In https://github.com/mmtk/mmtk-core/issues/1146, we observed that when using the native MarkSweep and the eager sweeping is enabled, no `ReleaseMutator` work packet is executed until `Release` finishes.  It can be seen from the eBPF timeline:

![image](https://github.com/mmtk/mmtk-core/assets/370317/b4a3baca-8af0-44ce-b787-15938b833fc9)

That is because the `Release` work packet calls `Plan::release` before spawning any `ReleaseMutator` work packet.  See:

https://github.com/mmtk/mmtk-core/blob/401803ce44d851c19e4d04f152b494095f9b71e6/src/scheduler/gc_work.rs#L126-L154

This is not a bug, because `Release` is designed to have exclusive access to the `Plan` when executed.  This means we shouldn't do any high-intensity work in `Plan::release` unless we don't need to release mutators.

**This affects all plans.**

`MarkSweep` does a non-trivial amount of work in `Plan::release` when doing eager sweeping.

Here is a zoomed-in timeline for Immix.  `ImmixSpace.release` generates `SweepChunk` work packets, so they start executing before `ReleaseMutator` work packets start.

![image](https://github.com/mmtk/mmtk-core/assets/370317/130bcda4-e6b0-40c4-a7a3-ed94d9df5312)

This is SemiSpace:

![image](https://github.com/mmtk/mmtk-core/assets/370317/32f9ac19-c325-438f-b15d-59575d0f93b7)

It looks OK.  But after I turn on the VO bits:

![image](https://github.com/mmtk/mmtk-core/assets/370317/028009b1-8f6b-4032-8371-d16bafa60491)

Clearly the `Release` work packet is spending a significant amount of time clearing the VO bit.

We should revisit all plans and all spaces, and see if they do any significant amount of work in `Plan.release` and `Space.release`.  If they do, they should be off-loaded to dedicated work packets so that they can be parallelized.

wks:
It raises another question:  If `Release` is supposed to have exclusive access to the Plan, and if it can spawn work packets (such as SweepChunk) that can start executing immediately, then it may not have exclusive access to the plan. because work packets such as `SweepChunk` may access the space which, according to Rust's concept of ownership, is part of the plan.

PR #1158, opened 2024-06-26
Add SweepChunk to native MarkSweepSpace

wks:
This PR re-introduce the `SweepChunk` work packet to the native `MarkSweepSpace` when using eager sweeping.  The main intention of this PS is fixing a pathological case where there is only one mutator and the single `ReleaseMutator` work packet cannot be parallelized.

The algorithm is unchanged for lazy sweeping.

When doing eager sweeping,

1.  We first release all unmarked blocks in `MarkSweepSpace::abandoned` and each mutator.
2.  And then we sweep blocks in parallel using `SweepChunk` work packets.
3.  We then go through all "consumed" blocks and move them into "available" lists if they have any free cells.

In step (1), we release blocks for each mutator in `ReleaseMutator`.  Releasing blocks is very fast, so parallelism is not a bottleneck.  During step (2), all blocks are either unallocated or marked, so we don't need to move any block out of linked lists, avoiding the data race we observed in https://github.com/mmtk/mmtk-core/issues/1103.  Step (3) is done by one thread, but it is fast enough not to be a performance bottleneck.

We also introduced a work packet `ReleaseMarkSweepSpace` which does what `MarkSweepSpace::release` did, but is executed concurrently with `ReleaseMutator` work packets.  In this way, we don't do too much work in the `Release` work packet, which is a problem we discussed in https://github.com/mmtk/mmtk-core/issues/1147.

We removed the constant `ABANDON_BLOCKS_IN_RESET` because it is obviously necessary to do so.  Otherwise one mutator will keep too many blocks locally, preventing other mutators to get available blocks to use.

We added an USDT tracepoint in `SweepChunk` in both `MarkSweepSpace` and `ImmixSpace` so that we can see the number of allocated blocks a `SweepChunk` work packet visited in the timeline generated by eBPF tracing tools.

We also changed `immixspace::SweepChunk` so that it now *asserts* that the chunk is allocated rather than skipping unallocated chunks.  `ChunkMap::generate_tasks` already filters out unallocated chunks.

Fixes: https://github.com/mmtk/mmtk-core/issues/1146

### DIFF
--- a/src/policy/marksweepspace/native_ms/global.rs
+++ b/src/policy/marksweepspace/native_ms/global.rs
@@ -529,26 +529,26 @@ struct SweepChunk<VM: VMBinding> {
 
 impl<VM: VMBinding> GCWork<VM> for SweepChunk<VM> {
     fn do_work(&mut self, _worker: &mut GCWorker<VM>, _mmtk: &'static MMTK<VM>) {
-        if self.space.chunk_map.get(self.chunk) == ChunkState::Allocated {
-            // number of allocated blocks.
-            let mut allocated_blocks = 0;
-            // Iterate over all allocated blocks in this chunk.
-            for block in self
-                .chunk
-                .iter_region::<Block>()
-                .filter(|block| block.get_state() != BlockState::Unallocated)
-            {
-                // We have released unmarked blocks in `ReleaseMarkSweepSpace` and `ReleaseMutator`.
-                // We shouldn't see any unmarked blocks now.
-                debug_assert_eq!(block.get_state(), BlockState::Marked);
-                block.sweep::<VM>();
-                allocated_blocks += 1;
-            }
-            probe!(mmtk, sweep_chunk, allocated_blocks);
-            // Set this chunk as free if there is not live blocks.
-            if allocated_blocks == 0 {
-                self.space.chunk_map.set(self.chunk, ChunkState::Free)
-            }
+        assert_eq!(self.space.chunk_map.get(self.chunk), ChunkState::Allocated);
+
+        // number of allocated blocks.
+        let mut allocated_blocks = 0;
+        // Iterate over all allocated blocks in this chunk.
+        for block in self
+            .chunk
+            .iter_region::<Block>()
+            .filter(|block| block.get_state() != BlockState::Unallocated)
+        {
+            // We have released unmarked blocks in `ReleaseMarkSweepSpace` and `ReleaseMutator`.
+            // We shouldn't see any unmarked blocks now.
+            debug_assert_eq!(block.get_state(), BlockState::Marked);
+            block.sweep::<VM>();
+            allocated_blocks += 1;
+        }
+        probe!(mmtk, sweep_chunk, allocated_blocks);
+        // Set this chunk as free if there is not live blocks.
+        if allocated_blocks == 0 {
+            self.space.chunk_map.set(self.chunk, ChunkState::Free)
         }
         self.epilogue.finish_one_work_packet();
     }
--- a/src/policy/marksweepspace/native_ms/global.rs
+++ b/src/policy/marksweepspace/native_ms/global.rs
@@ -101,25 +101,27 @@ impl AbandonedBlockLists {
         }
     }
 
-    fn release_blocks<VM: VMBinding>(&mut self, space: &MarkSweepSpace<VM>) {
-        for i in 0..MI_BIN_FULL {
-            // Release free blocks
-            self.available[i].release_blocks(space);
-            self.consumed[i].release_blocks(space);
-            self.unswept[i].release_blocks(space);
-        }
-    }
-
     fn sweep_later<VM: VMBinding>(&mut self, space: &MarkSweepSpace<VM>) {
         for i in 0..MI_BIN_FULL {
             // Release free blocks
             self.available[i].release_blocks(space);
             self.consumed[i].release_blocks(space);
-            self.unswept[i].release_blocks(space);
+            if cfg!(not(feature = "eager_sweeping")) {
+                self.unswept[i].release_blocks(space);
+            } else {
+                // If we do eager sweeping, we should have no unswept blocks.
+                debug_assert!(self.unswept[i].is_empty());
+            }
 
-            // Move remaining blocks to unswept
-            self.unswept[i].append(&mut self.available[i]);
-            self.unswept[i].append(&mut self.consumed[i]);
+            // For eager sweeping, that's it.  We just release unmarked blocks, and leave marked
+            // blocks to be swept later in the `SweepChunk` work packet.
+
+            // For lazy sweeping, we move blocks from available and consumed to unswept.  When an
+            // allocator tries to use them, they will sweep the block.
+            if cfg!(not(feature = "eager_sweeping")) {
+                self.unswept[i].append(&mut self.available[i]);
+                self.unswept[i].append(&mut self.consumed[i]);
+            }
         }
     }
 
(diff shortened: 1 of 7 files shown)


wks:
# DaCapo

I ran lusearch from DaCapo Chopin.  There are four builds: master and this PR, with lazy and eager sweeping for each of them.  Three heap factors (3.023x, 3.678x, 4.392x w.r.t. G1 in OpenJDK), 10 invocations each.  ([plotty link](https://squirrel.anu.edu.au/plotty/wks/noproject/#0|lynx-2024-06-26-Wed-135618&benchmark^build^hfac^invocation^iteration^mmtk_gc&GC^time^time.other^time.stw&|70&build^ms-sweep-chunk|10&iteration^1^4|20&1^invocation|41&Histogram%20(with%20CI)^build^hfac&))

image

The mutator time is not obviously changed.

The mutator time of master-eager is significantly lower at the lowest heap factor. That is a result of too frequent GCs (as seen in the number of GC).  Some mutator time is counted into STW time after one mutator triggers GC but before all mutators come to a stop, as seen in the timeline below.  This behavior is only reproducible if I run multiple iterations, and the timeline below is collected with the `-n5` cmdline option of DaCapo.  It may be a result of some existing bugs that caused MarkSweep not to release some memory.

image

As a result, the difference in STW time is mainly a result of increased number of GC and the memory leak bug. The STW time per GC is also unreasonably high for master-eager.

# eBPF

The following timeline is from master-eager, lusearch.  Although there are many mutators and `ReleaseMutator` can be parallelized, the `Release` work packet is busy releasing blocks in the abandoned list, blocking the `ReleaseMutator` work packets.

image

This is pr-eager running lusearch.  Using `SweepChunk` is *not* faster than `ReleaseMutator`.  It is not that load-balanced if the number of chunks is small and some chunks contain unallocated blocks.  And the sum of execution time of all `SweepChunk` packets is greater than that of all `ReleaseMutator` work packets.

image

This is master running Liquid using mmtk-ruby (must be eager).  Because there is only one mutator, the `ReleaseMutator` cannot be parallelized, and it spent most of the time doing sweeping.

image

This is this PR running Liquid.  Sweeping is still slow because of `naive_brute_force_sweep`, it is much more load-balanced.

image

# Conclusion

1.  There may be some bug in the master branch that caused MarkSweepSpace to not making some memory available for the mutators, causing the heap to become tighter as we increase the number of iterations.  This PR solves that problem.
2.  Although `SweepChunk` is not necessarily faster than doing sweeping in `ReleaseMutator` (depending on work load), it avoids the pathological case of single-threaded applications.

wks:
From the CI tests of the mmtk-ruby repo, this PR does *not* accelerate the test with MarkSweep.  See the execution time of the "Build and test (release, MarkSweep)" of the following two PRs:

-   Last PR that upgrades mmtk-ruby to the latest mmtk-core master: https://github.com/mmtk/mmtk-ruby/pull/78
-   This PR: https://github.com/mmtk/mmtk-ruby/pull/79

So the execution speed difference is more likely due to other factors, including the mutator speed (lack of allocation fast path) and GC speed (the time needed to run a whole GC), and the impact of the MemBalancer.

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -102,26 +106,31 @@ impl AbandonedBlockLists {
...
+            // allocator tries to use them, they will sweep the block.
+            if cfg!(not(feature = "eager_sweeping")) {
+                self.unswept[i].append(&mut self.available[i]);
+                self.unswept[i].append(&mut self.consumed[i]);
+            }
         }
     }
 
-    fn sweep<VM: VMBinding>(&mut self, space: &MarkSweepSpace<VM>) {
+    fn recategorize_blocks(&mut self) {
```
qinsoon, line 128: 'recycle' is more clear for what it does.
```suggestion
    fn recycle_blocks(&mut self) {
```

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -341,15 +351,17 @@ impl<VM: VMBinding> MarkSweepSpace<VM> {
...
-            let mut abandoned = self.abandoned.lock().unwrap();
-            abandoned.sweep(self);
-        } else {
-            // For lazy sweeping, we just move blocks from consumed to unswept. When an allocator tries
-            // to use them, they will sweep the block.
-            let mut abandoned = self.abandoned.lock().unwrap();
-            abandoned.sweep_later(self);
+            let num_mutators = VM::VMActivePlan::number_of_mutators();
+            // all ReleaseMutator work packets plus the ReleaseMarkSweepSpace packet
+            self.pending_release_packets
```
qinsoon, line 356: I suggest just using this mechanism for both eager and lazy sweeping. We always maintain the pending packets count. Just for eager sweeping, when all the packets are done, you generate sweep tasks. This could make things cleaner -- performance wise, the impact should be minimal. Otherwise, we see the conditionally compiled code for eager sweeping scattered everywhere to achieve this one thing.

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -419,6 +431,42 @@ impl<VM: VMBinding> MarkSweepSpace<VM> {
...
+        let old = self.pending_release_packets.fetch_sub(1, Ordering::SeqCst);
+        if old == 1 {
+            let work_packets = self.generate_sweep_tasks();
+            self.scheduler.work_buckets[WorkBucketStage::Release].bulk_add(work_packets);
+        }
+    }
+
+    fn generate_sweep_tasks(&self) -> Vec<Box<dyn GCWork<VM>>> {
+        let space = unsafe { &*(self as *const Self) };
+        let epilogue = Arc::new(RecategorizeBlocks {
```
qinsoon, line 445: I have a PR here that extracts the epilogue mechanism outside of the space code: https://github.com/mmtk/mmtk-core/pull/877. Do you want me to tidy it up and get it merged so you can base this PR on it?

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -78,6 +79,9 @@ pub struct MarkSweepSpace<VM: VMBinding> {
     /// these block lists in the space. These lists are only filled in the release phase,
     /// and will be moved to the abandoned lists above at the end of a GC.
     abandoned_in_gc: Mutex<AbandonedBlockLists>,
+    /// Count the number of pending release packets during the `Release` stage.
```
qinsoon, line 82: The comments seem too general. There might be other work packets that the plan is not aware of. The plan only cares about packets that mutate block lists, not all the packets in the release stage.

wks: Yes. That makes sense.  I'll explicitly mention `ReleaseMarkSweepSpace` and `ReleaseMutator` in the comment.

src/policy/marksweepspace/native_ms/global.rs
wks: I tried tidying it up locally, but I feel that #877 may not be general enough.  One example is the `pending_release_packets` field I put into `MarkSweepSpace`.  Due to our current mechanism of creating the `ReleaseMutator` work packets, we can't wrap `ReleaseMutator` into a `WorkPacketWithEpilogue`.  Therefore, I just use the space instance as the shared data.  Another thing I worry about is `WorkPacketEpilogue` wraps a `Box<dyn Fn>` inside it (it's better to use `FnOnce` instead).  We may want to mutate the epilogue in some ways after it is created, according to the work packets that will count down the counter.

I also think the "epilogue" may have some capability of collecting data in addition to counting.  For example, each unit of work could "abandon" some blocks to a pool which also plays the role of "epilogue".  Maybe the most general mechanism is a simple wrapper of `AtomicUsize` that provides a few intuitively named methods, and implement something like a countdown latch.

BTW, I have a utility class [`AfterAll`](https://github.com/mmtk/mmtk-ruby/blob/cf70302be2008626fb120e5353880e073b33ff09/mmtk/src/utils.rs#L60) in the Ruby binding.  It is more special purposed, but it's not that hard to implement or use.

src/policy/marksweepspace/native_ms/global.rs
wks: I made change as you suggested.

src/policy/marksweepspace/native_ms/global.rs
wks: I moved what's in `MarkSweepSpace::end_of_gc` into `recycle_blocks`.  That is now used by both lazy and eager sweeping.  When using lazy sweeping, we directly call `recycle_blocks` after all `ReleaseMutator` packets and the `ReleaseMarkSweepSpace` packet have finished.

qinsoon (approved):
LGTM

Issue #1044, opened 2023-12-08
Axioms that good definitions of ObjectReference must satisfy


wks:
*With issue https://github.com/mmtk/mmtk-core/issues/1170 addressed and PR https://github.com/mmtk/mmtk-core/pull/1195 merged, we now define `ObjectReference` as a non-zero word-aligned address within an object, and we agree with this definition at least for the current implementation of mmtk-core.  This issue summarizes discussions about the definition of `ObjectReference` before that change so that we can come back and find our previous discussions if we discuss this topic again.*

TL;DR: From time to time, we discuss the possibility to change the current definition of `ObjectReference`.  In theory, it should be opaque to mmtk-core.  But I have discovered that not all definitions are good.  This issue summarizes what a good definition should be, enumerate some popular definitions and discuss whether they are good.

Related links:

-   Issue:
    -   "ObjectReference should be opaque": https://github.com/mmtk/mmtk-core/issues/686
    -   Require ObjectReference to be inside an object: https://github.com/mmtk/mmtk-core/issues/1170
        -   PR: https://github.com/mmtk/mmtk-core/pull/1195
-   Zulip:
    -   Our last conversation.  https://mmtk.zulipchat.com/#narrow/stream/262679-General/topic/Terminology.3A.20canonical.20object.20reference.20address.3F
    -   General: https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/object.20reference
    -   Why handles won't work: https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/Don't.20create.20ObjectReferences.20during.20GC
    -   About internal references: https://mmtk.zulipchat.com/#narrow/stream/262673-mmtk-core/topic/Internal.20reference.20and.20handles

# Not all definitions of ObjectReference are good

There is an argument that `ObjectReference` may be opaque to mmtk-core.  However, some definitions won't work at all, and others won't work efficiently.  I am trying to list statements that needs to be true for all good definitions of ObjectReference.  Any definition that satisfies all of those statements should work.

## It must be able to instantiate ObjectReference from Address

Copying GCs will copy objects, and create a new `ObjectReference` for the to-space copy.

Linear scanning will identify objects at addresses (using the global VO bit or a local bitmap), and generate `ObjectReference` for those addresses.

Conservative stack scanning scans the stack for addresses with VO bit set, then we convert the Address to ObjectReference verbatim.

[...]

PR #1159, opened 2024-06-27
Require object reference to be aligned

qinsoon:
This PR adds some assumptions about object reference, and the conversion between object reference and in-object address. To be specific, this PR requires the following:
* Object reference needs to be word aligned
* There must be an constant offset between object reference and in-object address.

With these requirements, we can do the following computation to find object reference for internal pointers:
1. Use `ref_to_address` with the internal pointer to get the in-object address for the internal pointer. 
1. Find the last VO bit that is set.
2. Calculate the data address for the bit. This address is aligned to the VO bit region.
3. Use `address_to_ref` with the data address above. The return value is an object reference, but may not be a valid one.
4. Use the word alignment to find the actual object reference
5. Get the object size from the object reference, and make sure that the internal pointer is actually pointing into the object

Step 1 is necessary if the internal pointer is before the in-object address. If we concern about the invalid object reference in `ref_to_address` and `address_to_ref`, we could consider removing the two functions, and use a constant offset instead.

### DIFF
--- a/src/util/address.rs
+++ b/src/util/address.rs
@@ -212,6 +212,11 @@ impl Address {
         Address(self.0 - size)
     }
 
+    /// Apply an signed offset to the address.
+    pub const fn offset(self, offset: isize) -> Address {
+        Address((self.0 as isize).wrapping_add(offset) as usize)
+    }
+
     /// Bitwise 'and' with a mask.
     pub const fn and(self, mask: usize) -> usize {
         self.0 & mask
@@ -502,7 +507,7 @@ impl ObjectReference {
     ///
     /// MMTk should not make any assumption on the actual location of the address with the object reference.
     /// MMTk should not assume the address returned by this method is in our allocation. For the purposes of
-    /// setting object metadata, MMTk should use [`crate::vm::ObjectModel::ref_to_address()`] or [`crate::vm::ObjectModel::ref_to_header()`].
+    /// setting object metadata, MMTk should use [`crate::util::ObjectReference::to_address`] or [`crate::util::ObjectReference::to_header`].
     pub fn to_raw_address(self) -> Address {
         Address(self.0.get())
     }
@@ -512,8 +517,8 @@ impl ObjectReference {
     ///
     /// If `addr` is 0, the result is `None`.
     ///
-    /// MMTk should not assume an arbitrary address can be turned into an object reference. MMTk can use [`crate::vm::ObjectModel::address_to_ref()`]
-    /// to turn addresses that are from [`crate::vm::ObjectModel::ref_to_address()`] back to object.
+    /// MMTk should not assume an arbitrary address can be turned into an object reference. MMTk can use [`crate::util::ObjectReference::from_address`]
+    /// to turn addresses that are from [`crate::util::ObjectReference::to_address`] back to object.
     pub fn from_raw_address(addr: Address) -> Option<ObjectReference> {
         debug_assert!(
             addr.is_aligned_to(Self::ALIGNMENT),
@@ -540,12 +545,11 @@ impl ObjectReference {
     }
 
     /// Get the in-heap address from an object reference. This method is used by MMTk to get an in-heap address
-    /// for an object reference. This method is syntactic sugar for [`crate::vm::ObjectModel::ref_to_address`]. See the
-    /// comments on [`crate::vm::ObjectModel::ref_to_address`].
+    /// for an object reference.
     pub fn to_address<VM: VMBinding>(self) -> Address {
         use crate::vm::ObjectModel;
-        let to_address = VM::VMObjectModel::ref_to_address(self);
-        debug_assert!(!VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS || to_address == self.to_raw_address(), "The binding claims unified object reference address, but for object reference {}, ref_to_address() returns {}", self, to_address);
+        let to_address = Address(self.0.get()).offset(VM::VMObjectModel::IN_OBJECT_ADDRESS_OFFSET);
+        debug_assert!(!VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS || to_address == self.to_raw_address(), "The binding claims unified object reference address, but for object reference {}, in-object addr is {}", self, to_address);
         to_address
     }
 
@@ -563,17 +567,19 @@ impl ObjectReference {
     pub fn to_object_start<VM: VMBinding>(self) -> Address {
         use crate::vm::ObjectModel;
         let object_start = VM::VMObjectModel::ref_to_object_start(self);
-        debug_assert!(!VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS || object_start == self.to_raw_address(), "The binding claims unified object reference address, but for object reference {}, ref_to_address() returns {}", self, object_start);
+        debug_assert!(!VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS || object_start == self.to_raw_address(), "The binding claims unified object reference address, but for object reference {}, ref_to_object_start() returns {}", self, object_start);
         object_start
     }
 
-    /// Get the object reference from an address that is returned from [`crate::util::address::ObjectReference::to_address`]
-    /// or [`crate::vm::ObjectModel::ref_to_address`]. This method is syntactic sugar for [`crate::vm::ObjectModel::address_to_ref`].
-    /// See the comments on [`crate::vm::ObjectModel::address_to_ref`].
+    /// Get the object reference from an address that is returned from [`crate::util::address::ObjectReference::to_address`].
     pub fn from_address<VM: VMBinding>(addr: Address) -> ObjectReference {
         use crate::vm::ObjectModel;
-        let obj = VM::VMObjectModel::address_to_ref(addr);
-        debug_assert!(!VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS || addr == obj.to_raw_address(), "The binding claims unified object reference address, but for address {}, address_to_ref() returns {}", addr, obj);
+        let obj = unsafe {
+            ObjectReference::from_raw_address_unchecked(
+                addr.offset(-VM::VMObjectModel::IN_OBJECT_ADDRESS_OFFSET),
+            )
+        };
+        debug_assert!(!VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS || addr == obj.to_raw_address(), "The binding claims unified object reference address, but for address {}, the object reference is {}", addr, obj);
         debug_assert!(
             obj.to_raw_address().is_aligned_to(Self::ALIGNMENT),
             "ObjectReference is required to be word aligned"
--- a/src/util/address.rs
+++ b/src/util/address.rs
@@ -214,7 +214,7 @@ impl Address {
 
     /// Apply an signed offset to the address.
     pub const fn offset(self, offset: isize) -> Address {
-        Address((self.0 as isize).wrapping_add(offset) as usize)
+        Address(self.0.wrapping_add_signed(offset))
     }
 
     /// Bitwise 'and' with a mask.
--- a/src/util/address.rs
+++ b/src/util/address.rs
@@ -486,12 +486,18 @@ use crate::vm::VMBinding;
 /// `usize`.  For the convenience of passing `Option<ObjectReference>` to and from native (C/C++)
 /// programs, mmtk-core provides [`crate::util::api_util::NullableObjectReference`].
 ///
+/// Note that [`ObjectReference`] has to be word aligned.
+///
 /// [NPO]: https://doc.rust-lang.org/std/option/index.html#representation
 #[repr(transparent)]
 #[derive(Copy, Clone, Eq, Hash, PartialOrd, Ord, PartialEq, NoUninit)]
 pub struct ObjectReference(NonZeroUsize);
 
 impl ObjectReference {
+    /// The required minimal alignment for object reference. If the object reference's raw address is not aligned to this value,
+    /// you will see an assertion failure in the debug build.
+    pub const ALIGNMENT: usize = crate::util::constants::BYTES_IN_ADDRESS;
+
     /// Cast the object reference to its raw address. This method is mostly for the convinience of a binding.
     ///
     /// MMTk should not make any assumption on the actual location of the address with the object reference.
@@ -510,7 +516,7 @@ impl ObjectReference {
     /// to turn addresses that are from [`crate::vm::ObjectModel::ref_to_address()`] back to object.
     pub fn from_raw_address(addr: Address) -> Option<ObjectReference> {
         debug_assert!(
-            addr.is_aligned_to(crate::util::constants::BYTES_IN_ADDRESS),
+            addr.is_aligned_to(Self::ALIGNMENT),
             "ObjectReference is required to be word aligned"
         );
         NonZeroUsize::new(addr.0).map(ObjectReference)
@@ -527,7 +533,7 @@ impl ObjectReference {
     pub unsafe fn from_raw_address_unchecked(addr: Address) -> ObjectReference {
         debug_assert!(!addr.is_zero());
         debug_assert!(
-            addr.is_aligned_to(crate::util::constants::BYTES_IN_ADDRESS),
+            addr.is_aligned_to(Self::ALIGNMENT),
             "ObjectReference is required to be word aligned"
         );
         ObjectReference(NonZeroUsize::new_unchecked(addr.0))
@@ -569,8 +575,7 @@ impl ObjectReference {
         let obj = VM::VMObjectModel::address_to_ref(addr);
         debug_assert!(!VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS || addr == obj.to_raw_address(), "The binding claims unified object reference address, but for address {}, address_to_ref() returns {}", addr, obj);
         debug_assert!(
-            obj.to_raw_address()
-                .is_aligned_to(crate::util::constants::BYTES_IN_ADDRESS),
+            obj.to_raw_address().is_aligned_to(Self::ALIGNMENT),
             "ObjectReference is required to be word aligned"
         );
         obj
--- a/src/util/address.rs
+++ b/src/util/address.rs
@@ -500,7 +500,7 @@ pub struct ObjectReference(NonZeroUsize);
 
 impl ObjectReference {
     /// The required minimal alignment for object reference. If the object reference's raw address is not aligned to this value,
-    /// you will see an assertion failure in the debug build.
+    /// you will see an assertion failure in the debug build when constructing an object reference instance.
     pub const ALIGNMENT: usize = crate::util::constants::BYTES_IN_ADDRESS;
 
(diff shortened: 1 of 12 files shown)


qinsoon:
The following is some discussion with Steve on this topic.

('in-object' address/mapping or 'in-allocation' address/mapping refers to the address returned by `ref_to_address`)

> Steve Blackburn: I think we need to allow stronger assumptions about the alignment of the in-allocation address or the object reference (I think either will do). We cannot make assumptions about the start because it may be explicitly mis-aligned w.r.t. the object reference. Object references (meaning the value of the reference, not the location of the reference) usually have strong alignment assumptions in runtimes. The binding should declare what that alignment is. I'm not sure whether we should be using 'in-allocation' or 'object-reference'. The advantage of in-allocation is that it makes things simpler since the address (by definition) points into the object. But there's no natural reason for it to be aligned, so it might be a problem to demand an alignment. On the other hand object reference will in general require some arithmetic before its useful, but it generally will be naturally aligned since it is normal for them to have word alignment.

> Steve Blackburn: If the minimum allocation alignment is 1 word or greater (which I expect it is), then it should not be hard to get the binding to make a VO -> in-object mapping, which would allow the mark bit to be set. For example, if the alignment was 1 word and the in-object mapping had a constant relationship to the object reference, then the in-object address would either be word aligned, or would have a known constant distance to word alignment. If so it should not be hard to make a binding function that sets a mark bit based on a VO location. e.g.: interior pointer -> VO bit -> in-object address -> mark bit.

> Steve Blackburn: I assume that what makes most sense is to ask all language bindings to specify a minimum alignment on the object reference, and to require that the relationship between object reference and 'in-object pointer' be a constant. Thus the minimum alignment of the in-object pointer will be known (modulo the constant offset).

> Steve Blackburn: We should probably have an MMTk-level discussion about formalizing this. I think it is unavoidable that we formalize this (I don't think we can continue with this issue being ill-defined).

There was a little bit more discussion about allocation alignment and payload.

> Steve Blackburn: Each binding must specify an a minium alignment
> Steve Blackburn: My understanding is that that is with respect to the "object", i.e. the payload.
> Steve Blackburn: Payload means the non-header part of the object, ie its fields. So the start of the payload is the location of the first field.
> Steve Blackburn: NOT the start of the allocation
> Steve Blackburn: NOT the object reference
> Yi Lin: (Quote) If the minimum allocation alignment is 1 word or greater (which I expect it is)
> Yi Lin: We dont have this assumption yet. We allow alloctiaon alignment to be smaller than 1 word at the moment.
> Steve Blackburn: Why?
> Steve Blackburn: This seems very strange
> Steve Blackburn: You just need to fix that, which should be straightforward

wks:
Some high-level comments:

**Requiring `ObjectReference` to be an address**.  This PR adds a requirement that `ObjectReference` must be an address (and it must be aligned, and a constant offset from some address inside the object).  While I personally strongly support this change, we did have some previous discussion, and some of us argued the opposite.  The opposite idea is that "we should give the VM binding more freedom of what `ObjectReference` is, including whether it may include tag bits (which makes it unaligned)."  This is just a reminder.  Related link: https://github.com/mmtk/mmtk-core/issues/1044

**`ObjectReference` no longer needs to be what a reference field holds.**  If we require `ObjectReference` to be aligned, and a VM stores unaligned addresses in reference fields (due to offsets or tag bits), then such VM will require a conversion in `Slot::load()` to give MMTk-core an aligned address, and another conversion in `Slot::store()` to restore the unaligned-ness (by adding back the tag bits or offset) before writing back to the field.  This is not hard to do, and some VMs are already doing this.  But that leads to another question:

**If `ObjectReference` is not what's in the field, and we do almost everything with `objref.to_address()`, why don't we define `ObjectReference` to be equal to the "in-object address"?**  In other word, should we require that `ObjectReference` must always be inside the allocation?

If we do this, we can unify `to_raw_address` and `to_address`.  MMTk core almost never works with raw addresses.  The only not-for-testing use case of raw address in the current mmtk-core is the forwarding pointer of moved object.

If we do this, we can further simplify operations on VO bits such that "VO bits are always set at the address of ObjectReference".  That should greatly simplify things.

One drawback I can think of is debugging.  Previously, `ObjectReference` could be the same as the contents in the field for VMs like JikesRVM, but it will no longer be.  OpenJDK and Ruby are not impacted.  V8 reference fields always has tag bits, so the value should never have been the same as `ObjectReference`.

qinsoon:
> **Requiring `ObjectReference` to be an address**. This PR adds a requirement that `ObjectReference` must be an address (and it must be aligned, and a constant offset from some address inside the object). While I personally strongly support this change, we did have some previous discussion, and some of us argued the opposite. The opposite idea is that "we should give the VM binding more freedom of what `ObjectReference` is, including whether it may include tag bits (which makes it unaligned)." This is just a reminder. Related link: #1044

This is not introduced by this PR. We already stated this clearly in [`ObjectReference`](https://docs.mmtk.io/api/mmtk/util/address/struct.ObjectReference.html):
`MMTk expects each object reference to have a pointer to the object (an address) in each object reference, and that address should be used for this ObjectReference type.`

> **`ObjectReference` no longer needs to be what a reference field holds.** If we require `ObjectReference` to be aligned, and a VM stores unaligned addresses in reference fields (due to offsets or tag bits), then such VM will require a conversion in `Slot::load()` to give MMTk-core an aligned address, and another conversion in `Slot::store()` to restore the unaligned-ness (by adding back the tag bits or offset) before writing back to the field. This is not hard to do, and some VMs are already doing this. But that leads to another question:

This is also not introduced by this PR. We have agreed that we should use `Slot` to deal with the tag.

> **If `ObjectReference` is not what's in the field, and we do almost everything with `objref.to_address()`, why don't we define `ObjectReference` to be equal to the "in-object address"?** In other word, should we require that `ObjectReference` must always be inside the allocation?
> 
> If we do this, we can unify `to_raw_address` and `to_address`. MMTk core almost never works with raw addresses. The only not-for-testing use case of raw address in the current mmtk-core is the forwarding pointer of moved object.
> 
> If we do this, we can further simplify operations on VO bits such that "VO bits are always set at the address of ObjectReference". That should greatly simplify things.
> 
> One drawback I can think of is debugging. Previously, `ObjectReference` could be the same as the contents in the field for VMs like JikesRVM, but it will no longer be. OpenJDK and Ruby are not impacted. V8 reference fields always has tag bits, so the value should never have been the same as `ObjectReference`.

The idea of object reference and in-object address is not introduced in this PR, though this PR does add more documentation to clarify it. 

Currently it is up to the binding. It is not as simple as what you propose, but it is still clear and works for different bindings.
* If their object reference matches our object reference definition, and points to an address that is inside the object, they can use the same address for both object reference and the in-object address. Actually most of our bindings do that. 
* If their object reference matches our definition, and may not point to an addresss inside the object, they will have to use a different in-object address. JikesRVM is the example. 

`to_raw_address` is mostly for bindings to safely convert between object reference and address. MMTk core probably only uses it in some corner cases (like check VO bit, or find base references).

wks:
Some detailed comments.  Let's assume we do what you described in this PR (and forget about my aggressive suggestion of making raw address == in-allocation address).

# Alignment

> Object reference needs to be word aligned

Should we require `objref.to_address()` to always return an aligned address, too?  It should be a reasonable requirement, and it should simplify things (e.g. if we use that to set metadata, such as VO bit, we always set metadata at aligned addresses).  This PR currently only requires it to be a "constant offset" from the object reference, but the "constant offset" is not said to be a multiple of word size. 

> ... This address is aligned to the VO bit region. 4. Use address_to_ref with the data address above.  The return value is an object reference, but may not be a valid one.

This means `address_to_ref(addr)` where `addr` is word-aligned may return an `ObjectReference` whose raw value is *not* word-aligned.  This is a bit strange as it seems to imply the "constant offset" may not be a multiple of word size, and we need to "5. Use the word alignment to find the actual object reference".

# Using `address_to_ref` on internal pointer

> Use `ref_to_address` with the internal pointer to get the in-object address for the internal pointer.

We define an "in-object address *of an ObjectReference*" because an ObjectReference is not necessarily inside an object.  But an internal pointer is always pointing at a field and, by definition, is always inside an object.  I think it is pointless to call `ref_to_adderss` on an internal pointer.  But the last paragraph described its purpose:

> Step 1 is necessary if the internal pointer is before the in-object address

It's not sufficient.  If the object layout is `| GCHeader | MiscHeader | JavaHeader | length | elem0 | elem1 | ... |`, each header is 8 byte, and `ObjectReference` is at `start + 32` (after `length`), and its in-object address is at `start + 24`, this means `ref_to_address(objref) == objref - 8`, and `address_to_ref(addr) = addr + 8`.  If we have an internal reference pointing at `start + 0`, then `address_to_ref(start + 0)` will be `start + 8`, which is still before the in-object address which is `start + 24`.

And if the internal pointer points to the very last word of an object, and the offset from in-object address to ObjectReference is positive, then the result will be outside the object.

I think we may need VM-specific knowledge to find ObjectReference from an internal pointer.  If the in-object address is the same as the object start, we only need to search towards lower addresses.  If, in a VM, an internal pointer always points to the payload, and never points to the header (which is *not* always true for VMs mainly implemented in C/C++ because the runtime may be fiddling with the header while triggering GC), and the in-object address is after the header like the example above, we will only need to search towards lower addresses, too.  In general, we may need to search in both directions, but with different limits, and be careful not to go beyond page/block/chunk boundaries because metadata may not be mapped, and some spaces (such as MarkSweepSpace and ImmixSpace) never allocate objects across block boundaries.

But if we unify `to_address` and `to_raw_address`, this part can be simplified, too.

qinsoon:
For alignment, we can assume that most language use aligned object reference. So it is natural for us to require aligned object reference, rather than aligned in-object address.

For the corner case that the internal pointer is before the in-object address, you are right. We would have to search forwards if that happens. We should let bindings to tell us, and possibly we can leave it as unimplemented for a while.

wks:
> For alignment, we can assume that most language use aligned object reference. So it is natural for us to require aligned object reference, rather than aligned in-object address.

I mean, **I suggest adding a requirement that "both ObjectReference and in-object address needs to be word-aligned"**.  (That implies their difference must be a multiple of word size, too.)  I think all VMs can satisfy this, and this will simplify the VO bit searching algorithm.  Concretely, in step 4, if VO bit is set at an (aligned) address, then `address_to_ref(addr)` will be guaranteed to be aligned, too, and that must be a valid ObjectReference.

qinsoon:
> > For alignment, we can assume that most language use aligned object reference. So it is natural for us to require aligned object reference, rather than aligned in-object address.
> 
> I mean, **I suggest adding a requirement that "both ObjectReference and in-object address needs to be word-aligned"**. (That implies their difference must be a multiple of word size, too.) I think all VMs can satisfy this, and this will simplify the VO bit searching algorithm. Concretely, in step 4, if VO bit is set at an (aligned) address, then `address_to_ref(addr)` will be guaranteed to be aligned, too, and that must be a valid ObjectReference.

Calling `address_to_ref()` does not add much complexity -- it is just applying an offset. If your concern is the invalid object references that is passed or returned from `ref_to_address` and `address_to_ref`, we can just remove those methods and ask the binding for a constant offset instead.

wks:
> > I mean, **I suggest adding a requirement that "both ObjectReference and in-object address needs to be word-aligned"**. (That implies their difference must be a multiple of word size, too.) I think all VMs can satisfy this, and this will simplify the VO bit searching algorithm. Concretely, in step 4, if VO bit is set at an (aligned) address, then `address_to_ref(addr)` will be guaranteed to be aligned, too, and that must be a valid ObjectReference.
> 
> Calling `address_to_ref()` does not add much complexity -- it is just applying an offset. If your concern is the invalid object references that is passed or returned from `ref_to_address` and `address_to_ref`, we can just remove those methods and ask the binding for a constant offset instead.

I am not talking about the complexity of calling `address_to_ref()` itself.  I was talking about eliminating step 5 by requiring the result of `address_to_ref` to be aligned.  Actually this PR requires any `ObjectReference` to be aligned.

I don't understand the part "The return value is an object reference, but may not be a valid one."  If the in-object address `addr` is aligned, and the result of `address_to_ref(addr)` is an `ObjectReference` which must also be aligned, why the resulting `ObjectReference` can still be invalid?  What does "Use the word alignment to find the actual object reference" mean?

qinsoon:
> I don't understand the part "The return value is an object reference, but may not be a valid one." If the in-object address `addr` is aligned, and the result of `address_to_ref(addr)` is an `ObjectReference` which must also be aligned, why the resulting `ObjectReference` can still be invalid? What does "Use the word alignment to find the actual object reference" mean?

The in-object address `addr` may not be aligned.

wks:
> > I don't understand the part "The return value is an object reference, but may not be a valid one." If the in-object address `addr` is aligned, and the result of `address_to_ref(addr)` is an `ObjectReference` which must also be aligned, why the resulting `ObjectReference` can still be invalid? What does "Use the word alignment to find the actual object reference" mean?
> 
> The in-object address `addr` may not be aligned.

This will be a problem.  If, in a VM, the in-object address is not aligned, and the offset between ObjectReference and in-object address is a constant, then the offset must not be a multiple of word size.  This means if `addr` *is* aligned in the first place (as returned from step 3, "This address is aligned to the VO bit region."), the result of `address_to_obj(addr)` will *not* be aligned.  This will trigger an assertion error immediately because it attempts to construct an `ObjectReference` which is required to be aligned.

qinsoon:
> > > I don't understand the part "The return value is an object reference, but may not be a valid one." If the in-object address `addr` is aligned, and the result of `address_to_ref(addr)` is an `ObjectReference` which must also be aligned, why the resulting `ObjectReference` can still be invalid? What does "Use the word alignment to find the actual object reference" mean?
> > 
> > 
> > The in-object address `addr` may not be aligned.
> 
> This will be a problem. If, in a VM, the in-object address is not aligned, and the offset between ObjectReference and in-object address is a constant, then the offset must not be a multiple of word size. This means if `addr` _is_ aligned in the first place (as returned from step 3, "This address is aligned to the VO bit region."), the result of `address_to_obj(addr)` will _not_ be aligned. This will trigger an assertion error immediately because it attempts to construct an `ObjectReference` which is required to be aligned.

I see. We will have to require both object reference and in-object address to be aligned then.

qinsoon:
We may want to replace `ref_to_address` and `address_to_ref` with a constant. This can avoid using and producing 'invalid object reference' for those two methods when we do computation for VO bit.

qinsoon:
@wks Can you review this PR again? I am working on the binding PRs. But you can review this PR first -- the binding PRs will be straightforward.

qinsoon:
binding-refs
OPENJDK_BINDING_REF=update-mmtk-core-pr-1159
JIKESRVM_BINDING_REF=update-mmtk-core-pr-1159
V8_BINDING_REF=update-mmtk-core-pr-1159
JULIA_BINDING_REF=update-mmtk-core-pr-1159
RUBY_BINDING_REF=update-mmtk-core-pr-1159

qinsoon:
I have made changes based on the suggestions.

src/vm/object_model.rs
```diff
@@ -402,8 +432,12 @@ pub trait ObjectModel<VM: VMBinding> {
     /// mature space for generational plans.
     const VM_WORST_CASE_COPY_EXPANSION: f64 = 1.5;
 
-    /// If this is true, the binding guarantees that an object reference's raw address is always equal to the return value of the `ref_to_address` method
-    /// and the return value of the `ref_to_object_start` method. This is a very strong guarantee, but it is also helpful for MMTk to
+    /// If this is true, the binding guarantees that the object reference's raw address,
+    /// the in-object address, and the object start is always the same address. To be precise,
```
wks, line 436:
```suggestion
    /// the in-object address, and the object start are always the same address. To be precise,
```

src/util/address.rs
```diff
@@ -212,6 +212,11 @@ impl Address {
         Address(self.0 - size)
     }
 
+    /// Apply an signed offset to the address.
+    pub const fn offset(self, offset: isize) -> Address {
+        Address((self.0 as isize).wrapping_add(offset) as usize)
```
wks, line 217: We may use [wrapping_add_signed](https://doc.rust-lang.org/std/primitive.usize.html#method.wrapping_add_signed) introduced since Rust 1.66.0

```suggestion
        Address(self.0.wrapping_add_signed(offset))
```

Actually, when we compute an address at an offset from a given one, we are not expecting overflow.  It's better if it has undefined behavior in the release build and panics in the debug build.  Unfortunately, the standard library doesn't have such a method.  If we are pedantic, we can use [overflowing_add_signed](https://doc.rust-lang.org/std/primitive.usize.html#method.overflowing_add_signed) (since 1.66.0) instead and assert the boolean return value is false.

```suggestion
        let (result, overflow) = self.0.overflowing_add_signed(offset);
        debug_assert!(!overflow, "Overflowed when adding offset {offset} to {self}");
        Address(result)
```

src/util/address.rs
```diff
@@ -549,17 +567,23 @@ impl ObjectReference {
...
-    /// or [`crate::vm::ObjectModel::ref_to_address`]. This method is syntactic sugar for [`crate::vm::ObjectModel::address_to_ref`].
-    /// See the comments on [`crate::vm::ObjectModel::address_to_ref`].
+    /// Get the object reference from an address that is returned from [`crate::util::address::ObjectReference::to_address`].
     pub fn from_address<VM: VMBinding>(addr: Address) -> ObjectReference {
         use crate::vm::ObjectModel;
-        let obj = VM::VMObjectModel::address_to_ref(addr);
-        debug_assert!(!VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS || addr == obj.to_raw_address(), "The binding claims unified object reference address, but for address {}, address_to_ref() returns {}", addr, obj);
+        let obj = unsafe {
+            ObjectReference::from_raw_address_unchecked(
+                addr.offset(-VM::VMObjectModel::IN_OBJECT_ADDRESS_OFFSET),
```
wks, line 579: If possible, we can do a static assertion that `VM::VMObjectModel::IN_OBJECT_ADDRESS_OFFSET` is zero if `VM::VMObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS` is true.

wks (changes requested):

src/util/address.rs
qinsoon: Thanks. I will use `wrapping_add_signed`.

src/util/address.rs
qinsoon: I thought about it. `UNIFIED_OBJECT_REFERENCE_ADDRESS` requires the object start, the object reference and the in-object address are the same. The assertions **exactly match** the description.

We can assert `IN_OBJECT_ADDRESS_OFFSET == 0`, and it **implies** the object reference and the in-object address are the same.

I think the former is more straightforward.

wks (approved):
LGTM

qinsoon:
> The comments on the function `memory_manager::is_mmtk_object` should be updated, too. Because `ObjectReference` must be aligned, some of the description, such as `ObjectReference` in the range of `low <= objref < high`, is no longer relevant.
> 
> I also mentioned the possibility of moving `is_mmtk_object` into the module `mmtk::util::is_mmtk_object` and changing its return type in [#1165 (comment)](https://github.com/mmtk/mmtk-core/pull/1165#discussion_r1671727457). Since this PR also changes bindings, it is a good time to make the change.

I updated the doc for `is_mmtk_object`, but I didn't change the code or its return type. I will change the return type in the other PR https://github.com/mmtk/mmtk-core/pull/1165/files.

PR #1161, opened 2024-07-01
Bump version to v0.26

qinsoon:
This PR bumps the version to v0.26.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,29 @@
+0.26.0 (2024-07-01)
+===
+
+## What's Changed
+
+### Policy
+* Clear side forwarding bits properly by @wks in https://github.com/mmtk/mmtk-core/pull/1138
+* Fix mark bit clearing in PrepareChunkMap by @wks in https://github.com/mmtk/mmtk-core/pull/1148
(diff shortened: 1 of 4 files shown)


wks:
If we want to change the MSRV policy, we should discuss in a group, with @steveblackburn and @caizixian involved.  From our previous discussion, we acknowledge that MMTk is a research project and may be aggressive with the technology we use. Our lab machine images are updated automatically, so our burden of maintenance is not high.  We should also communicate with our collaborators and see if there are requirements for a less aggressive MSRV policy.

qinsoon:
I removed the changes about the MSRV policy in this PR. But I still suggest we adopt a less aggressive policy.

We currently use N-1. But in practice, we are not aggressively moving the MSRV, as we know sticking to N-1 all the time would bring troubles, and that is not desirable. However, sticking to N-1 is a valid option according to our policy. Using N-4 or something similar would set a more clear expectation for our users and contributors.

wks (approved):
LGTM

qinsoon:
binding-refs
OPENJDK_BINDING_REF=v0.26
JIKESRVM_BINDING_REF=v0.26

PR #1165, opened 2024-07-10
Internal pointer support

qinsoon:
This PR adds internal pointer support. It supersedes https://github.com/mmtk/mmtk-core/pull/1155 which provides a simple but inefficient implementation for internal pointers. This PR is based on https://github.com/mmtk/mmtk-core/pull/1159 which adds requirements for object reference alignment.

This PR
* adds `memory_manager::find_object_from_internal_pointer`
  * The call is dispatched using SFT to each space.
  * Large object space only checks the first word in VO bit for every page.
  * Mark sweep and immix space only searches for the max object size for those spaces.
* Allow iterating side metadata bits.
* Allow loading raw byte/word in side metadata.


benches/mock_bench/internal_pointer.rs
```diff
@@ -0,0 +1,97 @@
...
+    const NORMAL_OBJECT_SIZE: usize = 16 * 1024;
+    write_mockvm(|mock| {
+        *mock = MockVM {
+            get_object_size: MockMethod::new_fixed(Box::new(|_| NORMAL_OBJECT_SIZE)),
+            is_collection_enabled: MockMethod::new_fixed(Box::new(|_| false)),
+            ..MockVM::default()
+        }
+    });
+
+    c.bench_function("internal pointer - normal objects", |_b| {
```
qinsoon, line 24: This is some data for this micro benchmark.

The naive implementation (`find_prev_non_zero_value_simple`, similar to https://github.com/mmtk/mmtk-core/pull/1155): check every aligned data address until we find the object.
```console
internal pointer - normal objects
                        time:   [12.910 µs 12.911 µs 12.912 µs]

internal pointer - large objects
                        time:   [53.164 ns 53.486 ns 53.939 ns]
```

Check every _byte_ in side metadata until we find VO bit, and compute it back to the VO address
```console
internal pointer - normal objects
                        time:   [1.5500 µs 1.5504 µs 1.5509 µs]
                        change: [-88.001% -87.997% -87.994%] (p = 0.00 < 0.05)
                        Performance has improved.

internal pointer - large objects
                        time:   [58.115 ns 58.137 ns 58.161 ns]
                        change: [+8.0877% +8.6753% +9.1317%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Check every _word_ in side metadata until we find VO bit, and compute it back to the VO address
```console
internal pointer - normal objects
                        time:   [311.28 ns 312.38 ns 313.51 ns]
                        change: [-79.856% -79.804% -79.741%] (p = 0.00 < 0.05)
                        Performance has improved.

internal pointer - large objects
                        time:   [57.582 ns 57.607 ns 57.635 ns]
                        change: [-0.9006% -0.8311% -0.7566%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

src/memory_manager.rs
```diff
@@ -630,6 +630,52 @@ pub fn is_mmtk_object(addr: Address) -> bool {
...
+/// an address range and guarantees the object ference is in the range.
+/// The reason behind this is that we use the VO
+/// bit to represent a valid object for every 8 bytes ([`crate::util::is_mmtk_object::VO_BIT_REGION_SIZE`]).
+/// We use VO bits to find the object for an internal pointer.
+/// When we find a set VO bit, we only know that in the 8 bytes there is an object, and we cannot know where
+/// exactly the object is. The binding needs to use their knowledge about the alignment
+/// and offset for object references to find out the object reference from the return value.
+/// See Case 2 in [`crate::memory_manager::is_mmtk_object`] for more explanation.
+///
+/// Note that, we only consider pointers that points to an address that is equal or after the in-object addresss
```
wks, line 647:
```suggestion
/// Note that we only consider pointers that point to addresses that are equal to or greater than the in-object address
```

src/memory_manager.rs
```diff
@@ -630,6 +630,52 @@ pub fn is_mmtk_object(addr: Address) -> bool {
...
+/// out obvious non-pointers (e.g. alignment check, bound check, etc) before calling this function to avoid unnecessary
+/// cost. This method is not cheap.
+///
+/// To minimize the cost, the user should also use a small `max_search_bytes`.
+///
+/// Argument:
+/// * `internal_ptr`: The internal pointer to start. We search backwards from this internal pointer to find the base reference.
+/// * `max_search_bytes`: The maximum number of bytes we may search for an object with VO bit set.
+#[cfg(feature = "is_mmtk_object")]
+pub fn find_object_from_internal_pointer<VM: VMBinding>(
```
wks, line 664: I suggest moving this method to the module `mmtk::util::is_mmtk_object`.  As a Rust package, it is better to use modules to categorize functions instead of putting all static methods in one magical class like JikesRVM.

We should move `memory_manager::is_mmtk_object` into the `mmtk:util::is_mmtk_object` module, too, and leave a stub here for existing bindings that depend on `memory_manager::is_mmtk_object`.  The doc comments on `memory_manager::is_mmtk_object` needs to be updated, too, because we now require `ObjectReference` to be word-aligned, and we can remove some of the confusing descriptions there.  `memory_manager::is_mmtk_object` should return `Option<ObjectReference>`, too, for the sake of symmetry, and also because it is safe to convert an `Address` to `ObjectReference` once we know VO bit is set for its corresponding in-object address.  We may update `is_mmtk_object` in https://github.com/mmtk/mmtk-core/pull/1159 because that PR changes the requirement of `ObjectReference`.

wks: Sounds good.

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -25,6 +25,32 @@ pub(super) fn address_to_contiguous_meta_address(
     }
 }
 
+/// Performs reverse address translation from congitugous metadata bits to data addresses.
```
k-sareen, line 28:
```suggestion
/// Performs reverse address translation from contiguous metadata bits to data addresses.
```

src/util/metadata/vo_bit/mod.rs
```diff
@@ -160,3 +158,82 @@ pub fn bcopy_vo_bit_from_mark_bit<VM: VMBinding>(start: Address, size: usize) {
...
+        None
+    }
+}
+
+/// Turning a potential object reference into its in-object address (the ref_to_address address) where the metadata is set for.
+fn get_in_object_address_for_potential_object<VM: VMBinding>(potential_obj: Address) -> Address {
+    // This is hacky: the address could be unaligned, and we will see an assertion failure if we
+    // create the 'object reference' in the normal way.
+    let obj_ref: ObjectReference = unsafe { std::mem::transmute(potential_obj) };
+    obj_ref.to_address::<VM>()
```
wks, line 199: We may use the constant `IN_OBJECT_ADDRESS_OFFSET` here.

src/util/metadata/vo_bit/mod.rs
```diff
@@ -160,3 +158,82 @@ pub fn bcopy_vo_bit_from_mark_bit<VM: VMBinding>(start: Address, size: usize) {
...
+}
+
+/// Get the object reference from an aligned address where VO bit is set.
+fn get_object_ref_for_vo_addr<VM: VMBinding>(vo_addr: Address) -> ObjectReference {
+    // This is also hacky: address_to_ref may create an unaligned object reference.
+    let addr = <VM::VMObjectModel as ObjectModel<VM>>::address_to_ref(vo_addr).to_raw_address();
```
wks, line 205: Use `IN_OBJECT_ADDRESS_OFFSET` so that it won't panic.

src/policy/largeobjectspace.rs
```diff
@@ -86,6 +86,45 @@ impl<VM: VMBinding> SFT for LargeObjectSpace<VM> {
...
+        ptr: Address,
+        max_search_bytes: usize,
+    ) -> Option<ObjectReference> {
+        use crate::util::metadata::vo_bit;
+        // For large object space, it is a bit special. We only need to check VO bit for each page.
+        let mut cur_page = ptr.align_down(BYTES_IN_PAGE);
+        let low_page = ptr
+            .saturating_sub(max_search_bytes)
+            .align_down(BYTES_IN_PAGE);
+        while cur_page > low_page {
```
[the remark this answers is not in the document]
qinsoon: Right. `cur_page` could be the same as `low_page`, and we will need to check `cur_page`. I will fix this.

src/policy/largeobjectspace.rs
```diff
@@ -86,6 +86,45 @@ impl<VM: VMBinding> SFT for LargeObjectSpace<VM> {
...
+            // If the page start is not mapped, there can't be an object in it.
+            if !cur_page.is_mapped() {
+                return None;
+            }
+            if vo_bit::get_raw_vo_bit_word(cur_page) != 0 {
+                // Find the exact address that has vo bit set
+                for offset in 0..vo_bit::VO_BIT_WORD_TO_REGION {
```
wks, line 108: `VO_BIT_WORD_TO_REGION` is currently 512.  But what if the "in-object address" is 700 or 1000 bytes into the start of an object?  It is possible if the VM uses the Sable VM layout.

src/util/metadata/vo_bit/mod.rs
qinsoon: Yes. I will make changes once the other PR is merged.

src/policy/largeobjectspace.rs
qinsoon: It is very unlikely we will ever see an extreme case like that. However, I can try put in some assertions to make sure if that ever happens, we fail explicitly.

src/util/metadata/side_metadata/global.rs
```diff
@@ -946,6 +1075,194 @@ impl SideMetadataSpec {
...
+        let start_meta_shift = meta_byte_lshift(self, start_addr);
+        let end_meta_addr = address_to_contiguous_meta_address(self, end_addr);
+        let end_meta_shift = meta_byte_lshift(self, end_addr);
+
+        // The result will be set by one of the following closures.
+        // Use Cell so it doesn't need to be mutably borrowed by the two closures which Rust will complain.
+        let res = std::cell::Cell::new(None);
+
+        // Check and find the last bit that is set. We try load words where possible, and fall back to load bytes.
+        let check_bytes_backwards = |start: Address, end: Address| -> bool {
```
wks, line 1170: This is quite a big lambda.  I think it should be a very common operation to search for the last non-zero byte in an array, and I believe there are existing third-party crates that can do this, and probably with SIMD support.

If we want to implement it by ourselves, we may extract this into some utility modules, and implement this algorithm over anything that supports "loading at word or byte granularity".  That should include the side metadata, a `Vec<u8>`, or even a `String`, so that it will be easy to write unit tests for this algorithm without depending on the side metadata framework.

By the way, if we want to be aggressive and fully commit into search or updating side metadata bitmaps word by word, we can replace `check_bytes_backwards` and `check_bits_backwards` with `check_words_backwards` and `check_bits_in_word_backwards`.  But that itself may deserve a separate PR.

src/util/metadata/side_metadata/global.rs
```diff
@@ -946,6 +1075,194 @@ impl SideMetadataSpec {
...
+                    {
+                        BYTES_IN_ADDRESS
+                    } else {
+                        1
+                    };
+                    // Move to the address so we can load from it
+                    cur -= step;
+
+                    // We have to always check if the metadata address is mapped or not.
+                    if !cur.is_mapped() {
```
wks, line 1187: Rather than calling `is_mapped` at every `step`, we may set a sentinel (cur aligned down to page size) so that we don't need to call `is_mapped` until `cur` reaches into another page.

src/util/metadata/side_metadata/global.rs
```diff
@@ -946,6 +1075,194 @@ impl SideMetadataSpec {
...
+                        // Return true for early abort. If we see an unmapped metadata address, there is no point to keep checking.
+                        return true;
+                    }
+
+                    if step == BYTES_IN_ADDRESS {
```
wks, line 1192: This should be just like bits and bytes.  We may start iterating byte by byte, then word by word (or using SIMD if possible), and then byte by byte again.  It should be easier if we can encapsulate the algorithm for "find the last non-zero byte" into a module.

src/util/metadata/side_metadata/global.rs
```diff
@@ -946,6 +1075,194 @@ impl SideMetadataSpec {
...
+                            let byte_offset = (value.trailing_zeros() / 8) as usize;
+                            let byte_addr = cur + byte_offset;
+                            let byte_value = (value >> (byte_offset * 8)) & 0xFF;
+                            if byte_value != 0 {
+                                let bit = byte_value.trailing_zeros();
+                                debug_assert!(bit < 8);
+                                res.set(Some(contiguous_meta_address_to_address(
+                                    self, byte_addr, bit as u8,
+                                )));
+                                return true;
```
wks, line 1206: This looks identical to line 1213 to 1218.  This means if we can extract the current lambda into a function that returns the last non-zero byte, the code paths can be reused.

src/util/options.rs
```diff
@@ -25,7 +25,7 @@ pub enum NurseryZeroingOptions {
 }
 
 /// Select a GC plan for MMTk.
-#[derive(Copy, Clone, EnumString, Debug)]
+#[derive(Copy, Clone, EnumString, Debug, PartialEq)]
```
wks, line 28: It's OK for PlanSelector to implement `Eq`, too. `PartialEq` is designed for things like `f32` and `f64` where not all values are comparable (such as NaN).  `PlanSelector` is fully comparable.

src/util/metadata/side_metadata/global.rs
qinsoon: I thought about it. It requires changing `iterate_meta_bits`, and all the methods that use them. But most methods don't need to 'check words'. We could also add a similar function that allows 'check words', but it will duplicate code or have a very similar pattern as `iterate_meta_bits`. Neither sounds ideal. So I think it is probably a better idea to implement 'checking word' with 'check bytes'. 

I could extract some code from the closure and make it reusable.

src/util/metadata/side_metadata/global.rs
qinsoon: Sounds good. We do mmap at chunk granularity. We can use that to only check every chunk.

benches/mock_bench/internal_pointer.rs
qinsoon: After applying the suggestion in https://github.com/mmtk/mmtk-core/pull/1165#discussion_r1671931368, we only check if the metadata address is mapped for every chunk. The performance is further improved.
```console
internal pointer - normal objects
                        time:   [139.48 ns 139.65 ns 139.82 ns]
                        change: [-55.524% -55.388% -55.250%] (p = 0.00 < 0.05)
                        Performance has improved.

internal pointer - large objects
                        time:   [49.300 ns 49.361 ns 49.462 ns]
                        change: [-14.459% -14.367% -14.263%] (p = 0.00 < 0.05)
                        Performance has improved.
```

src/memory_manager.rs
qinsoon: I changed the return type for `is_mmtk_object`, and updated the document. But I did not move the functions to the `vo_bit` module. The current control flow is `memory_manager` -> SFT -> policy -> `vo_bit`. If we move the functions to `vo_bit`, the control flow would be `memory_manager` -> `vo_bit` -> SFT -> policy -> `vo_bit`, which seems confusing.

src/policy/largeobjectspace.rs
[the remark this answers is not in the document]
qinsoon: I added an assertion when we set VO bit for LOS. If the VO bit is not in the first 512 bytes, we will panic.

src/policy/largeobjectspace.rs
qinsoon: I added an assertion when we set VO bit for LOS. If the VO bit is not in the first 512 bytes, we will panic.

src/policy/marksweepspace/native_ms/global.rs
```diff
@@ -188,6 +188,17 @@ impl<VM: VMBinding> SFT for MarkSweepSpace<VM> {
...
     }
 
+    #[cfg(feature = "is_mmtk_object")]
+    fn find_object_from_internal_pointer(
+        &self,
+        ptr: Address,
+        max_search_bytes: usize,
+    ) -> Option<ObjectReference> {
+        // We don't need to search more than the max object size in the mark sweep space.
+        let search_bytes = usize::min(MAX_OBJECT_SIZE, max_search_bytes);
```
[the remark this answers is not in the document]
qinsoon: I tried this, and I saw segfaults happening in some tests. The reason is that for an address in the mark sweep space's virutal address range, we cannot guarantee that the address is actually in a block that is initialized. If the block is not initialized, loading cell size will segfault.

src/util/metadata/side_metadata/global.rs
qinsoon: I extract those code to be separate functions in `side_metadata::helpers`. They are not general to be used for things like `Vec<u8>`, as we need to check if the metadata address is mapped or not which only applies to our side metadata.

src/util/metadata/side_metadata/global.rs
qinsoon: Done. This makes the benchmark 50% faster. See https://github.com/mmtk/mmtk-core/pull/1165#discussion_r1674800993.

src/util/metadata/side_metadata/global.rs
qinsoon: See https://github.com/mmtk/mmtk-core/pull/1165#discussion_r1671930003.

src/util/options.rs
qinsoon: Yes. I added `Eq`.

qinsoon:
binding-refs
RUBY_BINDING_REF=update-mmtk-core-pr-1165

src/memory_manager.rs
wks: There is a module named `mmtk::util::is_mmtk_object`.  It currently only holds the constant `VO_BIT_REGION_SIZE` which the VM binding needs to know the granularity of the VO bits.  We can move the implementation of the `is_mmtk_object` function there.  But I acknowledge that the module name may not be the best (`mmtk::util::is_mmtk_object::is_mmtk_object` looks confusing, too).  If we don't have a better plan for reorganizing the functions, we can leave the `is_mmtk_object` function in the `memory_manager` module for now.

src/util/metadata/side_metadata/global.rs
wks: > ... They are not general to be used for things like `Vec<u8>`, as we need to check if the metadata address is mapped or not which only applies to our side metadata.

This is easy to solve.  We require the bit holder to implement a trait that must provide a `is_mapped` function, like this:

```rust
trait HoledByteArray {
    fn load_byte(&self, address: Address) -> u8;
    fn load_word(&self, address: Address) -> usize;
    fn is_mapped(&self, address: Address) -> bool;
}
```

And we can implement it for `Vec<u8>`, using byte index as `address`, and make `is_mapped` always return true.

```rust
impl HoledByteArray for Vec<u8> {
    fn load_byte(&self, address: Address) -> u8 { self[address.as_usize()] }
    fn load_word(&self, address: Address) -> usize { /* load 8 consecutive elements */ }
    fn is_mapped(&self, address: Address) -> bool { true }
```

And we can implement it for metadata, too.  We just load from raw address and actually check the mapping.

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -134,3 +160,193 @@ pub(super) fn meta_byte_mask(metadata_spec: &SideMetadataSpec) -> u8 {
...
+
+        if step == BYTES_IN_ADDRESS {
+            // Load and check a usize word
+            let value = unsafe { cur.load::<usize>() };
+            if value != 0 {
+                // Find the exact non-zero byte within the usize using bitwise operations
+                let byte_offset = (value.trailing_zeros() / 8) as usize;
+                let byte_addr = cur + byte_offset;
+                let byte_value: u8 = ((value >> (byte_offset * 8)) & 0xFF) as u8;
+                if let Some(bit) = find_last_non_zero_bit_in_u8(byte_value) {
```
wks, line 213: Actually if `value != 0`, the result must be `Some(bit)`.  We can just `unwrap()` the return value.

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -134,3 +160,193 @@ pub(super) fn meta_byte_mask(metadata_spec: &SideMetadataSpec) -> u8 {
...
+) -> FindMetaBitResult {
+    if !addr.is_mapped() {
+        return FindMetaBitResult::UnmappedMetadata;
+    }
+    let byte = unsafe { addr.load::<u8>() };
+    for cur_bit in (start_bit..end_bit).rev() {
+        if (byte & (1 << cur_bit)) != 0 {
+            return FindMetaBitResult::Found { addr, bit: cur_bit };
+        }
+    }
```
wks, line 245: The for loop probably won't hurt performance because it's just iterating bits, but we can use `trailing_zeros()` and check if it is within `start_bit..end_bit`, or mask the bits and call `find_last_non_zero_bit_in_u8`

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -25,6 +25,32 @@ pub(super) fn address_to_contiguous_meta_address(
...
+    let relative_meta_addr = metadata_addr - metadata_spec.get_absolute_offset();
+
+    let data_addr_intermediate = if rshift >= 0 {
+        relative_meta_addr << rshift
+    } else {
+        relative_meta_addr >> (-rshift)
+    };
+
+    let data_addr = (data_addr_intermediate << metadata_spec.log_bytes_in_region)
+        + ((shift as usize) << (metadata_spec.log_bytes_in_region - metadata_spec.log_num_of_bits));
```
wks, line 49: It is unlikely, but in theory `metadata_spec.log_bytes_in_region - metadata_spec.log_num_of_bits` may underflow.

If `log_num_of_bits >= 3` (i.e. `rshift <= 0`), then `shift` must be zero.  We can split this statement into the `if` and `else` branch above.  When `rshift >= 0`, we only use `metadata_spec.log_bytes_in_region`; when `rshift < 0`, we take `metadata_spec.log_num_of_bits` into consideration.

src/memory_manager.rs
qinsoon: I moved the actual implementation of the two methods to `is_mmtk_object`. But those are not public, and the public methods are still in `memory_manager`.

src/util/metadata/side_metadata/global.rs
qinsoon: Sounds good. But maybe we don't need to add it now: it adds complexity and there is no clear sign that we will use it any time soon. When anyone needs to reuse the code, they can do the refactoring and make the method more general.

src/memory_manager.rs
```diff
@@ -588,42 +588,69 @@ pub fn is_live_object<VM: VMBinding>(object: ObjectReference) -> bool {
...
 /// 1.  Return true if `ObjectReference::from_raw_address(addr)` is a valid object reference to an
 ///     object in any space in MMTk.
-/// 2.  Also return true if there exists an `objref: ObjectReference` such that
-///     -   `objref` is a valid object reference to an object in any space in MMTk, and
-///     -   `lo <= objref.to_address() < hi`, where
-///         -   `lo = addr.align_down(VO_BIT_REGION_SIZE)` and
-///         -   `hi = lo + VO_BIT_REGION_SIZE` and
-///         -   `VO_BIT_REGION_SIZE` is [`crate::util::is_mmtk_object::VO_BIT_REGION_SIZE`].
-///             It is the byte granularity of the valid object (VO) bit.
+/// 2.  Also return true if there exists an `objref: ObjectReference` for the address.
```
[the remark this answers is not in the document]
qinsoon: Done.

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -25,6 +25,32 @@ pub(super) fn address_to_contiguous_meta_address(
...
+pub(super) fn contiguous_meta_address_to_address(
+    metadata_spec: &SideMetadataSpec,
+    metadata_addr: Address,
+    shift: u8,
+) -> Address {
+    let rshift = (LOG_BITS_IN_BYTE as i32) - metadata_spec.log_num_of_bits as i32;
+    let relative_meta_addr = metadata_addr - metadata_spec.get_absolute_offset();
+
+    let data_addr_intermediate = if rshift >= 0 {
+        relative_meta_addr << rshift
```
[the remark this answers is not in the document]
qinsoon: I am not sure what is expected for the review. I renamed `rshift` to `shift`.

src/util/metadata/side_metadata/helpers.rs
qinsoon: Done. I added a few more tests to cover this case.

src/util/metadata/side_metadata/helpers.rs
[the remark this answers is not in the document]
wks: I don't know what's the best variable name for it, either, but "shift" looks good enough.

wks (approved):
LGTM

PR #1166, opened 2024-07-10
Clear pin bits

udesou:
If an object dies, we need to make sure his pin bit gets cleared. This PR enables clearing the pin bits when sweeping - clearing the whole block when `super::BLOCK_ONLY` is true, or at line granularity when a line is not marked. 
I've only added it to the immix policy, since for everything else pinning is either unsupported or it's a NOP.

### DIFF
--- a/src/policy/immix/block.rs
+++ b/src/policy/immix/block.rs
@@ -8,6 +8,8 @@ use crate::util::linear_scan::{Region, RegionIterator};
 use crate::util::metadata::side_metadata::{MetadataByteArrayRef, SideMetadataSpec};
 #[cfg(feature = "vo_bit")]
 use crate::util::metadata::vo_bit;
+#[cfg(feature = "object_pinning")]
+use crate::util::metadata::MetadataSpec;
 use crate::util::Address;
 use crate::vm::*;
 use std::sync::atomic::Ordering;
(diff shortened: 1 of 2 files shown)


src/policy/immix/block.rs
```diff
@@ -201,6 +201,9 @@ impl Block {
                     #[cfg(feature = "vo_bit")]
                     vo_bit::helper::on_region_swept::<VM, _>(self, false);
 
+                    #[cfg(feature = "object_pinning")]
```
wks, line 204: This should be under the condition that the pinning bit metadata is on the side.  If the pinning bit is in the header, we don't need to clear them because the object will be overwritten anyway.

src/policy/immix/block.rs
```diff
@@ -233,6 +236,10 @@ impl Block {
                     #[cfg(feature = "immix_zero_on_release")]
                     crate::util::memory::zero(line.start(), Line::BYTES);
 
+                    // We need to clear the pin bit as this line can be reused
+                    #[cfg(feature = "object_pinning")]
+                    crate::util::metadata::pin_bit::bzero_pin_bit::<VM>(line.start(), Line::BYTES);
```
wks, line 241: Ditto.

wks (commented):
We should only do the bulk clearing if the pinning bit is in side metadata.

~~BTW, if it is convenient, you can add the code for clearing pinning bits for CopySpace.~~  Sorry. We can't pin things in the CopySpace.

src/policy/immix/block.rs
udesou: I'm doing that check to only bulk clear if it's on the side in the `bzero_pin_bit` function (see [here](https://github.com/udesou/mmtk-core/blob/98f5a67ff7017649ba55302a0ac8a0a6c7615802/src/util/metadata/pin_bit.rs#L55-L63)), is your suggestion to instead "inline" that check when sweeping the block?

qinsoon: > I'm doing that check to only bulk clear if it's on the side in the `bzero_pin_bit` function (see [here](https://github.com/udesou/mmtk-core/blob/98f5a67ff7017649ba55302a0ac8a0a6c7615802/src/util/metadata/pin_bit.rs#L55-L63)), is your suggestion to instead "inline" that check when sweeping the block?

I think that is what Kunshan meant. If the pin bits is not on the side, there is no need to call bulk zero at all, as we cannot and we dont need to bulk zero the bits.

qinsoon (approved):
LGTM

PR #1168, opened 2024-07-17
Set the unlog bits for objects in the bootimage

k-sareen:


src/policy/vmspace.rs
```diff
@@ -224,6 +224,14 @@ impl<VM: VMBinding> VMSpace<VM> {
             start: start.align_down(BYTES_IN_PAGE),
             end: end.align_up(BYTES_IN_PAGE),
         });
+
+        if self.common.needs_log_bit {
+            if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC {
+                side.bset_metadata(start, size);
+            } else {
+                unimplemented!("We cannot bulk set unlogged bit.")
```
wks, line 232: If a VM uses in-header unlog bits, the VM can still set the unlog bit to 1 in the boot image and make it behave normally.  It is not necessary to panic.

src/policy/vmspace.rs
```diff
@@ -224,6 +224,14 @@ impl<VM: VMBinding> VMSpace<VM> {
             start: start.align_down(BYTES_IN_PAGE),
             end: end.align_up(BYTES_IN_PAGE),
         });
+
+        if self.common.needs_log_bit {
```
wks, line 228: Another concern is that this approach assumes the VM binding cannot set the unlog bits properly for reasons such as not being able to enumerate objects in the boot image.  This shouldn't be the default.  Instead, we can make it an option to bulk set side unlog bits.  It should be guarded by a Cargo feature so that the user knows it is a workaround.  (Alternatively, it can also be guarded by a constant in `ObjectModel`.  The bulk-setting operation can also be an API for the binding to invoke explicitly.)

src/policy/vmspace.rs
k-sareen: Ah that's fair

k-sareen: Done

src/policy/vmspace.rs
k-sareen: I've made it a feature

wks (commented):
We should add comments so that

1.  The VM binding developers know when they should use the feature introduced in this PR, and
2.  mmtk-core developers know what the code is for when they read the code.  When we see the code in the future, we may not immediately remember our discussion today.  Other developers not involved in today's discussion may not know its purpose because the code is intended for some very specific users.

Cargo.toml
```diff
@@ -88,6 +88,9 @@ mock_test = []
 # If a binding would need to trace/scan objects that is allocated and managed by the VM, `ActivePlan::vm_trace_object()` is an alternative.
 vm_space = []
 
+# Bulk set unlog bits for all objects inside the VM space
```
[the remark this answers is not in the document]
k-sareen: Done

src/policy/vmspace.rs
```diff
@@ -224,6 +224,13 @@ impl<VM: VMBinding> VMSpace<VM> {
             start: start.align_down(BYTES_IN_PAGE),
             end: end.align_up(BYTES_IN_PAGE),
         });
+
+        #[cfg(feature = "set_unlog_bits_vm_space")]
+        if self.common.needs_log_bit {
+            if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC {
+                side.bset_metadata(start, size);
+            }
+        }
```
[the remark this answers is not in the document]
k-sareen: Done

PR #1169, opened 2024-07-17
Ensure log bits are correctly maintained

k-sareen:
This fixes a bug wherein we were erroneously resetting the unlog bits
for objects in the modbuf for full-heap GCs. If we reset the unlog bits
in a full-heap GC, we may end up setting the unlog bit for an object
that actually died which may cause issues for future allocations.

In major GCs, we now bulk clear unlog bits for mark-sweep space and
immix space during the prepare stage and bulk clear them for copyspace
during the release stage. We clear log bits for large objects during the
sweep. We also set the unlog bits for mature objects when tracing. This
ensures that there are no stale unlog bits and that we don't accidentally
log nursery objects.

This PR also adds debug assertions checking that any object that has
been added to the modbuf is considered mature by MMTk.


src/plan/generational/gc_work.rs
```diff
@@ -108,29 +108,53 @@ impl<E: ProcessEdgesWork> ProcessModBuf<E> {
...
             // Scan objects in the modbuf and forward pointers
             let modbuf = std::mem::take(&mut self.modbuf);
             GCWork::do_work(
                 &mut ScanObjects::<E>::new(modbuf, false, WorkBucketStage::Closure),
                 worker,
                 mmtk,
             )
+        } else {
+            // Flip the per-object unlogged bits to "unlogged" state for objects inside the bootimage
+            #[cfg(feature = "vm_space")]
```
k-sareen, line 142: I didn't know what was better: to go through the list or bulk set the unlog bits in `vm_space.release()`. I thought this was better since it is more general than bulk setting the unlog bits (in case the VM binding is not using side-metadata for the unlog bits).

src/plan/barriers.rs
```diff
@@ -231,6 +231,7 @@ impl<S: BarrierSemantics> Barrier<S::VM> for ObjectBarrier<S> {
         target: Option<ObjectReference>,
     ) {
         if self.log_object(src) {
+            debug_assert!(src.is_live::<S::VM>(), "{} was logged but is not live", src);
```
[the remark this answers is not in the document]
k-sareen: Except the unlog bit should only be set for mature objects, so they _have_ to be considered alive. Or at the very least have to be marked. I previously had `is_marked` there but objects in the bootimage may not actually have mark bits set so it was triggering the assertion incorrectly.

src/plan/generational/gc_work.rs
qinsoon: I don't get why the VM space needs to be specially dealt here. Why do we need to set the unlog bits in mature GCs for the VM space? What about other spaces like immortal or LOS, how is the VM space different from those spaces?

k-sareen: Hm. So what we want to do is only reset the unlog bit for objects that are considered live. All ImmixSpace and CopySpace mature objects will set the unlog bit when you copy them over because of the `CopySemantics::Mature`. But objects in the VM space (or even immortal and LOS like you mentioned) don't perform copies and so we need to reset their bits _if_ those objects are live. VM space and immortal space objects are always live so we need to unconditionally reset them, but for LOS we need to reset only when the object is live. So really we're doing this reset at the wrong place.

qinsoon: > But objects in the VM space (or even immortal and LOS like you mentioned) don't perform copies and so we need to reset their bits iff those objects are live. 

Why not do that in the `trace_object` method for the policy where we know the object is reached? For ImmixSpace and CopySpace, the log bit is set during copying in `trace_object`.

k-sareen: Right that's what I suggested for LOS objects. But it's not clear to me if ART, for example, will actually end up tracing through all objects in the bootimage or immortal space for a full heap GC given their optimization(s). Maybe that's an ART-specific issue though.

k-sareen: Another alternative could be to create a new work packet to be executed after the transitive closure (or maybe release?) for major GCs. In that work packet we now have all the correct information about if a current object address is marked or not, and if it is, then we can reset the unlog bit for that object.

qinsoon: Another option is to unconditionally set the log bit in `ProcessModBuf` (no matter nursery or mature GCs), and bulk clear log bits when we release memory for dead objects.

In the current code base, for Immix space, we will set log bits during tracing if `unlog_object_when_traced` is true, and will clear log bits in prepare if `reset_log_bit_in_major_gc` is true. The former is needed, as we 'promote' nursery objects in immix space when we trace it (and we do not have to copy it to 'promote'). But the latter may not be necessary -- we can bulk clear log bits for the dead objects after a GC. For other spaces, we can always bulk clear log bits after a GC for reclaimed memory.

k-sareen: Maybe this warrants a group discussion. We can discuss this in the next group meeting. I'll do a hack for ART in the meantime since this current PR is probably still incorrect because of LOS objects' unlog bits not being reset.

k-sareen: What about cases where we recycle mature LOS objects? I think that is broken as well. If the unlog bit was set for it in a previous GC and it has died since then, if we allocate a new LOS object at the _same_ address, we have a young object with the unlog bit set which is also wrong. We need to clear the unlog bit for dead LOS objects in the sweep [here](https://github.com/mmtk/mmtk-core/blob/eb919f23e96feae6c8c00460e87d5ad4009083a4/src/policy/largeobjectspace.rs#L230-L235)

k-sareen:
Just fixed it like we discussed in a previous meeting

src/policy/vmspace.rs
```diff
@@ -275,6 +275,17 @@ impl<VM: VMBinding> VMSpace<VM> {
         );
         debug_assert!(self.in_space(object));
         if self.mark_state.test_and_mark::<VM>(object) {
+            // Flip the per-object unlogged bits to "unlogged" state for objects inside the
+            // bootimage
+            #[cfg(feature = "set_unlog_bits_vm_space")]
```
k-sareen, line 280: Perhaps we can remove the conditional compile feature here since it is likely objects in the VM space will want to have the unlog bit set

src/plan/barriers.rs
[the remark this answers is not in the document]
k-sareen: On second thought I actually agree with you and will remove the debug assert from the write barrier code. I think the debug assert inside the ProcessModBuf work packet is correct though

k-sareen:
@wks Can you re-review this?

wks:
> If we reset the unlog bits
> in a full-heap GC, we may end up setting the unlog bit for an object
> that actually died which may cause issues for future allocations.

I assume the "issues for future allocations" means newly allocated (young) objects may have unlog bit set, and may be logged when written into, and we will retain more objects than necessary.

But I don't think it is *sufficient* to ensure this by skipping setting unlog bits in mature GCs.  There may be dead mature objects that are not written to (i.e. unlogged).  When those objects die, their unlog bits will remain set.  If future allocations reuse the memory space of those objects, some young objects will also have unlog bits set.

If you want to ensure nursery objects never get logged, we may need to (bulk) clear unlog bits for reclaimed regions, like how we clear VO bits.  I think this only affects StickyImmix because we never set the unlog bits for objects in the GenCopy or GenImmix nursery in the first place.

k-sareen:
Hence why we do a bulk clear?

k-sareen:
Oh. How odd. This PR doesn't do the bulk clear. I must have messed up when cherry-picking from my branch

wks:
Instead of bulk clearing, it is also possible to copy the mark bits over to the unlog bits (and remove `ImmixSpaceArgs::unlog_object_when_traced` and `ImmixSpaceArgs::reset_log_bit_in_major_gc`).  I think that's a bit like VO bits where there are two different strategies, and we found that it is faster to just copy over mark bits.  We may refactor the code to unify the handling of unlog bits and VO bits.

p.s. I think the so-called "unlog bit" should really be named "is candidate for logging".  Obviously nursery objects are not such candidates.

k-sareen:
What that suggests is we need to be more systematic with how we deal with stale metadata values

k-sareen:
@wks What needs to be done for this PR to get it to merge? I think it's very important to merge bug fixes ASAP since it leads to duplication of debugging efforts otherwise. This is a very very subtle bug.

qinsoon:
If I understand correct, the issue is that for dead objects, we do not clear their unlog bits, and new objects allocated will have random unlog bits. The obvious fix would be to just clear the unlog bits before new allocation.

The PR tries to maintain the unlog bits before the objects become dead: if an object dies, its unlog bit is 0, and if an object lives, its unlog bit is 1. This doesn't sound like a straight solution to the problem. What is the reason for using this solution rather than just clearing the unlog bits?

k-sareen:
> The PR tries to maintain the unlog bits before the objects become dead: if an object dies, its unlog bit is 0, and if an object lives, its unlog bit is 1. This doesn't sound like a straight solution to the problem. What is the reason for using this solution rather than just clearing the unlog bits?

Fast-path allocation

k-sareen:
Really what we want is a meta-module that takes care of all relevant metadata when an object dies/a space is released. What we have currently is a hack since each metadata (like VO-bit or side forwarding bits) has been special cased. But I feel like that is out of scope of this PR.

src/plan/generational/gc_work.rs
```diff
@@ -108,22 +108,24 @@ impl<E: ProcessEdgesWork> ProcessModBuf<E> {
...
+        // Process and scan modbuf only if the current GC is a nursery GC
         if mmtk
             .get_plan()
             .generational()
             .unwrap()
             .is_current_gc_nursery()
         {
+            // Flip the per-object unlogged bits to "unlogged" state.
+            for obj in &self.modbuf {
+                debug_assert!((*obj).is_live(), "{} was logged but is not live", *obj);
```
wks, line 120: This assertion fails for GenCopy in the first nursery GC after the first full-heap GC.  It is because it is in the from space of the semispace mature space, and `is_live()` returns false for all objects in the from space.

wks:
> @wks What needs to be done for this PR to get it to merge? I think it's very important to merge bug fixes ASAP since it leads to duplication of debugging efforts otherwise. This is a very very subtle bug.

You still need to implement bulk clearing of unlog bits for ImmixSpace.  It is sufficient to do the clearing in `Block::sweep` in the two places where `vo_bit::helper::on_region_swept` is called.  You only need to do this for StickyImmix because we only allocate nursery objects in ImmixSpace in StickyImmix.  Unlike VO bits, stale unlog bits are benign in places where nursery objects cannot be allocated.  That includes non-empty lines (or blocks if `BLOCK_ONLY`).  By doing this, we maintain an invariant that newly allocated young objects must not have unlog bits set.

And I suggest adding a debug assertion that young objects cannot have unlog bits set.  We can do this while tracing or while allocating.  Possible places include `CommonGenPlan::trace_object_nursery` and `StickyImmix::trace_object_nursery`.

> EDIT: I can remove the `object.is_live()` debug assertions if you want, although I still think they're right.

Probably remove it for now because it fails in GenCopy.  We can introduce a similar assertion later.

I think the actual assertion you want to make is that the object is a mature object.  For StickyImmix, we use the mark bit to distinguish between old and young objects.  So `is_live` happens to work.  But for GenCopy, it is different.

k-sareen:
> I think the actual assertion you want to make is that the object is a mature object. For StickyImmix, we use the mark bit to distinguish between old and young objects. So is_live happens to work. But for GenCopy, it is different.

Yes exactly. My best approximation was `is_live`. Maybe negation of `is_object_in_nursery` is better?

wks:
> @wks Re: Immix log bits being cleared. We have a `ImmixSpaceArg` called `reset_log_bit_in_major_gc` which is only `true` for `StickyImmix`. I think it should be removed in favour of always resetting log bits right

It depends on what you mean by "always".  When using StickyImmix, the ideal outcome is:

-   With respect to the *time* of clearing unlog bits, we will clear unlog bits in both major GCs and minor GCs when using StickyImmix;
-   With respect to the *memory region* to clear unlog bits, we will no longer clear the unlog bits of the entire ImmixSpace in the beginning of a (full-heap) GC, but we will clear the unlog bits of completely free lines/blocks after a GC.

Currently `reset_log_bit_in_major_gc` governs the activity of clearing unlog bits for the entire ImmixSpace in the beginning of a full-heap GC.  It "always happens" before a major GC.  And this is no longer useful.  We should clear unlog bits

-   in every GC, regardless whether it is major or minor GC, and
-   in both GenImmix and StickyImmix.  We don't need it in Immix because it's not generational.

So if "always" means "in every GC and in both GenImmxi and StickyImmix", then yes.  We can remove `ImmixSpaceArgs::reset_log_bit_in_major_gc`.  But we still need a field in `ImmixSpace` or `ImmixSpaceArgs` to tell whether the `ImmixSpace` in the current plan uses unlog bits at all.  If not (i.e. the non-generational `Immix` plan), we don't clear unlog bits.

k-sareen:
> But we still need a field in ImmixSpace or ImmixSpaceArgs to tell whether the ImmixSpace in the current plan uses unlog bits at all. If not (i.e. the non-generational Immix plan), we don't clear unlog bits.

That field already exists in the [plan constraints](https://github.com/mmtk/mmtk-core/blob/c0f97884234b51b13c8ba5d1563a8e0f592d11c2/src/plan/plan_constraints.rs#L23). So yes, I was going to check that before I clear the bits

k-sareen:
@wks Can you review this again. I keep forgetting to get this merged. It's a very very subtle bug.

qinsoon:
We probably need to discuss this in the next meeting. I understand the problem is that for newly allocated objects, there might be arbitrary unlog bit values (rather than 0). But I don't get why you use the solution in the PR, instead of just clearing the bits when objects die, or when the memory is allocated. Is this solution supposed to be an optimization over the simple bulk clearing?

> If I understand correct, the issue is that for dead objects, we do not clear their unlog bits, and new objects allocated will have random unlog bits. The obvious fix would be to just clear the unlog bits before new allocation.
> 
> The PR tries to maintain the unlog bits before the objects become dead: if an object dies, its unlog bit is 0, and if an object lives, its unlog bit is 1. This doesn't sound like a straight solution to the problem. What is the reason for using this solution rather than just clearing the unlog bits?

This question still needs to be answered.

> > The PR tries to maintain the unlog bits before the objects become dead: if an object dies, its unlog bit is 0, and if an object lives, its unlog bit is 1. This doesn't sound like a straight solution to the problem. What is the reason for using this solution rather than just clearing the unlog bits?
> 
> Fast-path allocation.
> 
> EDIT: It seems like I haven't cherry-picked a further change from here: [k-sareen@8f6fcd8](https://github.com/k-sareen/mmtk-core/commit/8f6fcd8e8de8ac332577f4cc44af1c9cb8b0b065)
> 
> The above commit bulk clears the unlog bits for a global GC and then sets them again while tracing.

I don't get it. No matter the log bits is in header or on side, we can bulk clearing it before allocation.

src/plan/generational/global.rs
```diff
@@ -216,6 +216,10 @@ impl<VM: VMBinding> CommonGenPlan<VM> {
 
         // Evacuate nursery objects
         if self.nursery.in_space(object) {
+            debug_assert!(VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.is_unlogged::<VM>(
```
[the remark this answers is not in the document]
k-sareen: Oh sorry. Yes. It should be `!`. Copy-pasted it incorrectly. Thanks

wks (changes requested):
There is still no bzero operations at line level in nursery GCs for Sticky Immix.  StickyImmix can allocate young objects in the free lines in partially-occupied blocks.  If some lines used to hold mature objects and those objects die, their unlog bits will linger, and young objects will "inherit" their unlog bits.  See:  https://github.com/mmtk/mmtk-core/pull/1169#issuecomment-2507420394

Like Yi said in https://github.com/mmtk/mmtk-core/pull/1169#issuecomment-2506907659, we need to ensure that when a mature object dies, its unlog bit is set to 0.

And what I suggested is slightly looser: if an Immix line used to hold mature objects, and all objects in that line die, we clear all the unlog bits of that line;  If an Immix line still holds multiple mature objects and some of they die, we don't care because young objects cannot be allocated in to such lines anyway.

wks:
It looks good, but there is still a compilation error when the "object_pinning" feature is enabled.  And the branch needs to be merged or rebased with the master.

wks:
> I think the assertion is possibly failing due to a race condition between `ProcessModBuf` and the transitive closure. The `ProcessModBuf` work packet is executed in the [`Closure` bucket](https://github.com/mmtk/mmtk-core/blob/4e06434f965306507005bd67ae5852ba9640d974/src/plan/generational/barrier.rs#L48-L49).
> 
> I think this can be fixed by making the `ProcessModBuf` work packet execute in the `Unconstrained` bucket and then schedule the `ScanObjects` work packet generated by `ProcessModBuf` to the `Closure` bucket. Else, I can just remove the assertion. Thoughts @wks?

The mod buf can sometimes be very large, and there will be many `ProcessModBuf` work packets.  Forcing them to be processed before proceeding to Closure is a bit too constrainted and may affect performance.  I think we should remove the assertion.  Actually, the unlog bit of a mature object is 0 at the beginning of a GC if and only if it is in the modbuf, and it is 0 during Closure if it is still yet to be processed in the `ProcessModBuf` work packet.  So in `trace_object_nursery` it is undecided whether the unlog bit of a mature object should be 0 or 1.

wks:
By the way, if we want another level of safety, we can add all visited mature objects into a temporary buffer for the current GC, and use Sanity GC to assert that all mature objects have unlog bit = 1 at the end of a GC.

k-sareen:
I was going to say maybe why the assertion is failing is because the `is_object_in_nursery` function does not take the large object space into account (which might still be an issue), but the objects it's failing for are clearly not page-aligned. So is this actually a case where the assertion is being triggered correctly?

k-sareen:
@wks Yeah, either the OpenJDK binding is broken (somehow) or we still have stale unlog bits that were not cleared properly (or were accidentally set): 
```
thread '<unnamed>' panicked at /home/runner/work/mmtk-core/mmtk-core/mmtk-openjdk/repos/mmtk-core/src/plan/generational/gc_work.rs:116:17:
0x40829498 was logged but is not mature. Dumping process memory maps:
40000000-40400000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:nursery]
40400000-40800000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:los]
40800000-40c00000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:nursery]
40c00000-41000000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:los]
41000000-41800000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:nursery]
41800000-41c00000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:copyspace0]
```

wks:
One more bug:  In GenImmix, `ImmixSpaceArgs::unlog_object_when_traced` is set to `false`.  This means it will not set the unlog bits when tracing objects in the ImmixSpace.  However, this PR skips modbuf processing in major GCs.  This means if a live mature object is logged, its unlog bit will remain to be 0 after the major GC.  This object will not be logged again in the mutator time after the GC when written to, and this will cause its young children to be erroneously considered as garbage.

```diff
diff --git a/src/plan/generational/immix/global.rs b/src/plan/generational/immix/global.rs
index ba2850b3a4..207fde4f2c 100644
--- a/src/plan/generational/immix/global.rs
+++ b/src/plan/generational/immix/global.rs
@@ -252,7 +252,7 @@ impl<VM: VMBinding> GenImmix<VM> {
                 reset_log_bit_in_major_gc: false,
                 // We don't need to unlog objects at tracing. Instead, we unlog objects at copying.
                 // Any object is moved into the mature space, or is copied inside the mature space. We will unlog it.
-                unlog_object_when_traced: false,
+                unlog_object_when_traced: true,
                 // In GenImmix, young objects are not allocated in ImmixSpace directly.
                 #[cfg(feature = "vo_bit")]
                 mixed_age: false,
```

Setting `unlog_object_when_traced: true` is sufficient to fix this problem.  We don't need to set `reset_log_bit_in_major_gc` because we never allocate young objects into `ImmixSpace` in GenImmix.

It should be unnecessary for StickyImmix, too, because this PR now clear unlog bits when freeing lines.  I think I made a mistake to think there are stale unmark bits in StickyImmix, but since `reset_log_bit_in_major_gc == true`, there can't be stale unlog bits after major GC.  And there can't be any stale log bits after minor GCs, either, because no lines that contain mature objects are freed in minor GCs.

k-sareen:
> This shows we really need a centralized framework to clear the metadata bits.

Yes I totally agree. This is so incredibly confusing that I myself don't like this PR. But we need to have a fix for this bug. The centralized metadata framework is definitely the ideal.

wks:
I am considering a, possibly unlikely, scenario.  What if a whole chunk of ImmixSpace is reused for another space, say CopySpace?  If we allow stale bits to exist, the bits may end up being carried to a different space.  Well, this is impossible for Map64 because the *possible* address ranges of different spaces never overlap, but is should be possible for Map32, although very unlikely.

wks:
> I am considering a, possibly unlikely, scenario. What if a whole chunk of ImmixSpace is reused for another space, say CopySpace? If we allow stale bits to exist, the bits may end up being carried to a different space. Well, this is impossible for Map64 because the _possible_ address ranges of different spaces never overlap, but is should be possible for Map32, although very unlikely. This not only affects the unlog bits, but also any side metadata that allow stale values to exist.

Currently ImmixSpace uses `BlockPageResource`, and it returns pages to the `BlockPool`, and never return any pages (or chunks) back to the `VMMap`.  It means we are *temporarily* immune to this problem.  But we must reconsider this issue in the future.

k-sareen:
Well this is something the metadata framework should handle, right. Clear all metadata for a new chunk when it is (re)allocated.

wks:
Well, I think what happened to GenImmix is happening again for GenCopy.  After a major GC, we neither restore the log bits of objects in the modbuf nor set the log bits while tracing.  We do set the log bit in `post_copy`, but that only works for newly promoted objects, not surviving mature objects.

wks:
> I am considering a, possibly unlikely, scenario. What if a whole chunk of ImmixSpace is reused for another space, say CopySpace? If we allow stale bits to exist, the bits may end up being carried to a different space. Well, this is impossible for Map64 because the _possible_ address ranges of different spaces never overlap, but is should be possible for Map32, although very unlikely. This not only affects the unlog bits, but also any side metadata that allow stale values to exist.

I think this is actually happening for GenCopy.  Because the nursery, the from-space and the to-space are all CopySpace, they do return chunks back to VMMap, and some stale unlog bits are carried over to the nursery.

wks:
The following patch is sufficient to fix CopySpace.  Just leave no stale log bits behind when releasing.  (p.s. I wonder why this has been working for years.)

```diff
diff --git a/src/policy/copyspace.rs b/src/policy/copyspace.rs
index ddf8e5bb04..bbc66ccd4d 100644
--- a/src/policy/copyspace.rs
+++ b/src/policy/copyspace.rs
@@ -196,6 +196,12 @@ impl<VM: VMBinding> CopySpace<VM> {
                 side_forwarding_status_table.bzero_metadata(start, size);
             }
 
+            if self.common.needs_log_bit {
+                if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC {
+                    side.bzero_metadata(start, size);
+                }
+            }
+
             // Clear VO bits because all objects in the space are dead.
             #[cfg(feature = "vo_bit")]
             crate::util::metadata::vo_bit::bzero_vo_bit(start, size);
```

wks:
This patch can assert there is no stale log bits when a bump allocator acquires a page.  This is too aggressive, and we probably don't want this to be merged.

```diff
diff --git a/src/util/alloc/bumpallocator.rs b/src/util/alloc/bumpallocator.rs
index a436b2e07a..73301bab33 100644
--- a/src/util/alloc/bumpallocator.rs
+++ b/src/util/alloc/bumpallocator.rs
@@ -1,13 +1,13 @@
 use std::sync::Arc;
 
-use crate::util::Address;
+use crate::util::{Address, ObjectReference};
 
 use crate::util::alloc::Allocator;
 
 use crate::policy::space::Space;
 use crate::util::conversions::bytes_to_pages_up;
 use crate::util::opaque_pointer::*;
-use crate::vm::VMBinding;
+use crate::vm::{ObjectModel, VMBinding};
 
 /// Size of a bump allocator block. Currently it is set to 32 KB.
 const BLOCK_SIZE: usize = 8 << crate::util::constants::LOG_BYTES_IN_PAGE;
@@ -209,6 +209,16 @@ impl<VM: VMBinding> BumpAllocator<VM> {
                 block_size,
                 acquired_start
             );
+
+            for addr in (acquired_start.as_usize()..acquired_start.as_usize() + block_size).step_by(8) {
+                let objref = unsafe {
+                    ObjectReference::from_raw_address_unchecked(Address::from_usize(addr))
+                };
+                let unlogged = <VM as VMBinding>::VMObjectModel::GLOBAL_LOG_BIT_SPEC
+                    .is_unlogged::<VM>(objref, std::sync::atomic::Ordering::Relaxed);
+                assert!(!unlogged);
+            }
+
             if !stress_test {
                 self.set_limit(acquired_start, acquired_start + block_size);
                 self.alloc(size, align, offset)
```

k-sareen:
> The following patch is sufficient to fix CopySpace. Just leave no stale log bits behind when releasing. (p.s. I wonder why this has been working for years.)
> 
> ```diff
> diff --git a/src/policy/copyspace.rs b/src/policy/copyspace.rs
> index ddf8e5bb04..bbc66ccd4d 100644
> --- a/src/policy/copyspace.rs
> +++ b/src/policy/copyspace.rs
> @@ -196,6 +196,12 @@ impl<VM: VMBinding> CopySpace<VM> {
>                  side_forwarding_status_table.bzero_metadata(start, size);
>              }
>  
> +            if self.common.needs_log_bit {
> +                if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC {
> +                    side.bzero_metadata(start, size);
> +                }
> +            }
> +
>              // Clear VO bits because all objects in the space are dead.
>              #[cfg(feature = "vo_bit")]
>              crate::util::metadata::vo_bit::bzero_vo_bit(start, size);
> ```

Thanks. And yes, I wonder the same... This is a very non-trivial issue. We really should prioritize a better internal API for this.

k-sareen:
@wks The assertion failure comes from `SemiSpace` and it's an assertion for the VO-bit... Sigh.

wks:
> @wks The assertion failure comes from `SemiSpace` and it's an assertion for the VO-bit... Sigh.

The bug is in `MonotonicPageResource::iterate_allocated_regions`.  Particularly, this line:

```rust
                    let size = if self.pr.cursor().chunk_index() == start.chunk_index() {
                        self.pr.cursor() - start
```

The intention is that if the cursor of the PR is within the same chunk, it will only scan up to that cursor.

However, there is a corner case.  `self.pr.cursor()` may actually be "the end of the previous chunk" which has the same address as "the start of the current chunk".  It will happen if

1.  The Map32 gives chunk 2 to copyspace0.  The chunk range is 0x41400000-0x41800000.
2.  The MonotonicPageResource of copyspace0 allocated from 0x41400000 to 0x41800000, and used up chunk2.
3.  The Map32 gives chunk 1 to copyspace0.  The chunk range is 0x41000000-0x41400000
4.  The MonotonicPageResource of copyspace0 allocated from 0x41000000 to 0x41400000, and  used up chunk1, too.  The cursor stops at 0x41400000, the end of the 0x41000000-0x41400000 range.
5.  GC starts, and evacuated copyspace0.  The GC then starts to free all chunks the page resource obtained from Map32.
6.  When visiting the 0x41400000-0x41800000 range, it thinks the cursor 0x41400000 is within that range.  (Actually it is not. The cursor is at the end of the previous chunk.)
7.  The GC only clears the metadata from 0x41400000 to the cursor 0x41400000, which is exactly 0 bytes.
8.  The stale VO bits (and other bits, too) in 0x41400000-0x41800000 remained to the next GC, and caused assertion error.

So we should fix the iterator of the MonotonicPageResource.

k-sareen:
Wow. Thank you for catching that. Is the solution just to do `(self.pr.cursor() + 1).chunk_index()` or what? Seems like a tricky edge case.

wks:
I made a PR for this. https://github.com/mmtk/mmtk-core/pull/1299

I just used the simplest solution.  If it is intended for optimizing for the case that "the cursor is within that region", we just test if the cursor is within the region by `start < cursor && cursor < end`.

src/plan/generational/immix/global.rs
```diff
@@ -249,10 +249,8 @@ impl<VM: VMBinding> GenImmix<VM> {
         let immix_space = ImmixSpace::new(
             plan_args.get_space_args("immix_mature", true, false, VMRequest::discontiguous()),
             ImmixSpaceArgs {
-                reset_log_bit_in_major_gc: false,
-                // We don't need to unlog objects at tracing. Instead, we unlog objects at copying.
-                // Any object is moved into the mature space, or is copied inside the mature space. We will unlog it.
-                unlog_object_when_traced: false,
+                // We need to unlog objects at tracing time since we currently clear all log bits during a major GC
```
wks, line 252: Actually we don't clear all log bits during a major GC.  In the current code we set unlog bits during tracing, and clear unlog bits only for freed lines.

wks (commented):
Just a minor comment problem. It looks good otherwise.

src/plan/generational/immix/global.rs
k-sareen: Eh? This is not true. In the `ImmixSpace::prepare` we do a bulk clear of all the log bits. We do not clear the log bits in the `Block::sweep`.

wks: Oops.  My bad.  I forgot you reverted that change.

wks (approved):
LGTM

src/plan/generational/immix/global.rs
k-sareen: I mean technically, clearing bits lines may be more efficient. But this can be fixed in the internal metadata API refactor.

k-sareen:
@wks Do you know what commit details are used for the rebase commit? The details from the first one? Or does it use the description of the PR? I just wanted to clean up some stuff in the description before I merged it.

wks:
> @wks Do you know what commit details are used for the rebase commit? The details from the first one? Or does it use the description of the PR? I just wanted to clean up some stuff in the description before I merged it.

When we merge this PR, it will uses the title of this PR as the first line, followed by the description of this PR (i.e. the content of https://github.com/mmtk/mmtk-core/pull/1169#issue-2413530772).  You can edit them anytime before merging.

k-sareen:
Thank you. I've edited the description and title of the PR. If you think it's sufficient, please feel free to merge it. Thank you again for all your debugging help.

Issue #1170, opened 2024-07-19
Require ObjectReference to be inside an object


wks:
# Status quo

Currently, MMTk defines several addresses of an object.

| Name              | Definition                                    | Must be inside object |
|-------------------|-----------------------------------------------|-----------------------|
| starting address  | the return value of `memory_manager::alloc`   | Yes                   |
| ObjectReference   | an address that refers to an object           | No                    |
| in-object address | at a constant offset from ObjectReference, used to access SFT, side metadata, etc. | Yes |
| header address    | address used to access in-object metadata     | Yes                   |

The definition of `ObjectReference` is VM-specific.  We currently allow `ObjectReference` to be outside an object because some VMs do so.  For example, in JikesRVM, an `ObjectReference` is defined as the address to the array payload of an object if the object is an array.  That saves one offset computation for array element access, but when accessing scalar object fields or object headers, the VM will have to use negative offset from the `ObjectReference`.  When we port MMTk from JikesRVM to Rust, we inherited this type.  `ObjectReference` is now the standard way for mmtk-core to refer to an object.  We still allow `ObjectReference` to be outside an object so that when loading from a field in JikesRVM, we directly use the word stored in the field as `ObjectReference`.

However, because we only map side metadata memory for pages within spaces, addresses outside any space (or unmapped pages) may not have mapped metadata.   Similar is true for SFT entries which are allocated by chunk.  If we attempt to access metadata or SFT using an address outside the object, it will be a segmentation fault.  To solve this problem, we require the VM binding to implement `ObjectReference::ref_to_address` which computes the "in-object address" of an object which must be inside the object.  (https://github.com/mmtk/mmtk-core/pull/699)

Meanwhile, VMs that use conservative stack scanning needs to read a word from the stack, compute the "in-object address" from it, and see if the VO bit is set at the "in-object address".  Because we don't know if a word on the stack is an actual `ObjectReference` or not, the offset from the `ObjectReference` to the "in-object address" must be a constant (i.e. can be computed without reading any data from the object body).  (Also in https://github.com/mmtk/mmtk-core/pull/699)

[...]

PR #1174, opened 2024-07-24
Heap traversal

wks:
This PR adds a heap traversal API `MMTK::enumerate_objects` which enumerates all objects in the MMTk heap at the time of calling.

We added `SideMetadataSpec::scan_non_zero_values` to support enumerating objects by scanning the VO bit metadata.  It can also be used for scanning other side metadata when needed.  Note, however, that it is inefficient (but should work correctly) if the metadata is not contiguous or the metadata has more than one bit per region.  If there is any need for scanning such metadata, we need further refactoring.


wks:
This is my first attempt to create a heap traversal API.  The implementation is probably right, but we may need to refactor our current mechanism for visiting metadata in bulk (including reading, zeroing, setting, copying, as well as finding the last non-zero value, and finding all non-zero values in a range for heap traversal).  The main problem is that `iterate_meta_bits` has two call-back closures, and they can't share mutable borrows well.  We have already seen its consequence because `find_prev_non_zero_value_fast` needs to use a `Cell` to store the return value.  Heap traversal will need to use a `Cell` or `RefCell` for similar reason, and that won't perform well.

And things can be further simplified if we unify `ObjectReference` and the "in-object address" (see https://github.com/mmtk/mmtk-core/issues/1170).  Then the linear scanner will no longer depend on the `<VM: VMBinding>` type (for using `ObjectReference::from_address<VM>`).

wks:
> ... The main problem is that `iterate_meta_bits` has two call-back closures, and they can't share mutable borrows well. We have already seen its consequence because `find_prev_non_zero_value_fast` needs to use a `Cell` to store the return value. Heap traversal will need to use a `Cell` or `RefCell` for similar reason, and that won't perform well.

I reimplemented it upon https://github.com/mmtk/mmtk-core/pull/1181 and now we no longer use `RefCell`.

Benchmark (`taskset -c 0 cargo bench --features test_private -- bscan`) shows that it can scan an Immix block with 200 objects in 268 ns in the steady state.  In reality, it may take much longer due to cache misses.

GitHub CI reported an added API function `SideMetadataSpec::scan_non_zero_values` which is intended.  It did not report the added public method `MMTK::enumerate_objects`, probably because it is guarded by the feature "vo_bit".

src/util/object_enum.rs
```diff
@@ -0,0 +1,107 @@
...
+            let object = ObjectReference::from_address::<VM>(address);
+            (self.object_callback)(object);
+        })
+    }
+}
+
+/// Allow querying if a block may have objects. `MarkSweepSpace` and `ImmixSpace` use different
+/// `Block` types, and they have different block states. This trait lets both `Block` types provide
+/// the same `may_have_objects` method.
+pub(crate) trait BlockMayHaveObjects: Region {
```
qinsoon, line 70: Why not add the method to `Region`?

src/mmtk.rs
```diff
@@ -467,4 +469,46 @@ impl<VM: VMBinding> MMTK<VM> {
...
+    /// type information may not have been written.
+    ///
+    /// It has *undefined behavior* if GC happens while this function is being executed.  The VM
+    /// binding must ensure GC does not start while executing this function.  One way to prevent GC
+    /// from starting is not letting the current thread yield for GC.
+    ///
+    /// Some programming languages may provide an API that allows the user to allocate objects and
+    /// trigger GC while enumerating objects.  The VM binding may use the callback of this function
+    /// to save all visited objects in an array and let the user visit the array after this function
+    /// returned.
```
qinsoon, line 500: I don't see how it would be helpful to save the visited objects and visit them later. If the GC is happening, the visited objects may be reclaimed during this call, and the saved object references may become invalid.

**As this function uses VO bit (which is changed during GC and allocation), and the function does not block GC and does not block allocation, it should be undefined behavior if GC or allocation happens at the same time.** We cannot guarantee anything if GC/allocation is happenning. We may call the closure for an object that seems valid, but when the closure visits the object, the object may be reclaimed. We may visit objects that are copied (but their VO bits are not yet cleared). We may miss objects that are just allocated.

qinsoon (commented):
We have an old `ObjectIterator`, which is very similar to the simple version of `scan_non_zero_values` (expect that it skips the object size when it finds an object). After introducing `trait ObjectEnumerator`, it becomes confusing that we have `ObjectIterator` and `ObjectEnuemrator` and they have nothing to do with each other.

It would be good if we can simply remove `ObjectIterator` and replace its uses with the newly introduced enumerator. If this cannot be done in this PR, we should at least document the code to state what we plan to do with the old `ObjectIterator`.

src/util/object_enum.rs
wks: We only need it for the two `Block` types, one for `MarkSweepSpace` and the other for `ImmixSpace`.  Adding the method to `Region` will require all types that implement `Region` to implement that method, which is not good because

1.  It is unnecessary because we currently don't do the same testing on other levels, such as `Line`.  For `Chunk`, all spaces use the same `Chunk` type, and it is unnecessary to make it implement a generic trait.
2.  `Region` is a kind of address range (power of 2 sized and aligned), and all its methods are about address computing and comparing.  But the method `may_have_objects` is about querying its metadata.
3.  Not all `Region` types have enough metadata to identify if it may contain any object.  `Line`, for example, only has the line marking bits after tracing, and is unhelpful for traversing the heap during mutator time.

src/util/metadata/side_metadata/global.rs
```diff
@@ -1115,6 +1115,107 @@ impl SideMetadataSpec {
...
         res.map(|addr| addr.align_down(1 << self.log_bytes_in_region))
     }
+
+    /// Search for data addresses that have non zero values in the side metadata.  This method is
+    /// primarily used for heap traversal by scanning the VO bits.
+    ///
+    /// This function searches the side metadata for the data address range from `data_start_addr`
+    /// (inclusive) to `data_end_addr` (exclusive).  The data address range must be fully mapped.
+    ///
+    /// `visit_data` is called for each data address that has non-zero side metadata.
```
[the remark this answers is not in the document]
wks: OK. That means I still need to use the `IN_OBJECT_ADDRESS_OFFSET` constant and alignment requirement to find the right `ObjectReference`.  I'll fix it.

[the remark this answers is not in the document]
qinsoon: There are some methods in `vo_bit` that could be helpful to you. Also in the side metadata implementation, you don't need to know about `ObjectReference`. You can do that in the `object_enum` module.

src/mmtk.rs
wks: > I don't see how it would be helpful to save the visited objects and visit them later. If the GC is happening, the visited objects may be reclaimed during this call, and the saved object references may become invalid.

The VM binding needs to treat the saved list of object references as roots.  In the way, if GC happens, those objects will be guaranteed to live through the next GC.  The list may contain the last reference to some objects because the object may be unlinked from their original parents after the last GC.  That's expected.  Such heap traversal APIs are allowed to pick up objects that have become unreachable, finalized, etc., and the VM binding needs to be aware of that.  But if such unreachable objects points to other objects, their outgoing edges will remain valid because (1) if GC has not happened, their edges remain valid, and (2) if GC happens, those objects are rooted and will be traced and forwarded.

If the VM can update the roots (in the same way updating JNI handles or other native roots), the saved list can be updated if a copying GC happens.  Alternatively, if the VM only intends to support plans that support pinning, it can treat the list as pinning roots, and those objects will not be moved while iterating through the list.

I have tested it on Ruby by manually triggering a GC before saving the list (as a root) and iterating through the list.  It works.

qinsoon: What you said does not contradict with what I said. I suggested that 'it should be undefined behavior if GC or allocation happens at the same time (during the enumerating function call)'. After the enumerating function, the binding can do whatever they need to make an 'illusion' that the user is enumerating objects during a GC. But this function itself is not allowed during GC.
```suggestion
    /// trigger GC while enumerating objects.  The VM binding may use the callback of this function
    /// to save all visited objects in an array outside GCs and let the user visit the array after this function
    /// returned. The binding needs to carefully keep the object references in the array alive and updated properly if
    /// a GC happens.
```

wks:
> We have an old `ObjectIterator`, which is very similar to the simple version of `scan_non_zero_values` (expect that it skips the object size when it finds an object). After introducing `trait ObjectEnumerator`, it becomes confusing that we have `ObjectIterator` and `ObjectEnuemrator` and they have nothing to do with each other.
> 
> It would be good if we can simply remove `ObjectIterator` and replace its uses with the newly introduced enumerator. If this cannot be done in this PR, we should at least document the code to state what we plan to do with the old `ObjectIterator`.

I think `linear_scan::ObjectIterator` works similarly as the algorithm in this PR.  `ObjectIterator` outputs `ObjectReference`, too, and it also scans from low to high addresses.  A minor difference is that `ObjectIterator` attempts to jump a longer distance by calling `ObjectModel::get_current_size`.  I am not sure whether `ObjectModel::get_current_size` is faster than blindly scanning the VO bits at word granularity.  We need benchmarks for that.

Since removing `ObjectIterator` will involve refactoring MarkCompact, too, I'll do it in a separate PR.

src/mmtk.rs
wks: I agree that it should have undefined behavior if GC happens.

I thought allocation should be benign because allocation strictly adds VO bits but does not remove VO bits.  But after thinking about it, I find that it is easier to just specify that it has undefined behavior if allocation happens concurrently, too.  Reasons include

-   We currently accesses the VO bits at different granularities, including bit, byte, word, etc.  The interaction of such memory accesses in a multi-threaded environment is not well-defined in the C++11 or Rust memory model.  So it is hard to reason what the reader will see (or whether it has undefined behavior) if the read and the write happens concurrently.
-   In a multi-threaded program, different threads may not agree upon a global total order in which different VO bits are set.  (Thread 1 may see bit X is set before bit Y, but Thread2 may see bit Y is set before bit X.)  So "all bits that have been set when `enumerate_objects` is called" is not precisely defined unless we make all threads stop, or at least synchornize with the current thread.

It is OK for Ruby because we currently only supports a single Ractor (in which only one thread is active at any time).  For OpenJDK, it will require a `VMOperation` to stop the world before enumerating objects.

src/policy/space.rs
```diff
@@ -348,6 +349,28 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
...
+    ///
+    /// **Skipping empty ranges**: When enumerating address ranges, spaces can skip ranges (blocks,
+    /// chunks, etc.) that are guarenteed not to contain objects.
+    ///
+    /// **Dynamic dispatch**: Because `Space` is a trait object type and `enumerator` is a `dyn`
+    /// reference, invoking methods of `enumerator` involves a dynamic dispatching.  But the
+    /// overhead is OK if we call it a block at a time because scanning the VO bits will dominate
+    /// the execution time.  For LOS, it will be cheaper to enumerate individual objects than
+    /// scanning VO bits because it is sparse.
+    fn enumerate_objects_coarse(&self, enumerator: &mut dyn ObjectEnumerator);
```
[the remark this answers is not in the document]
wks: I renamed it back to `enumerate_objects`.

src/util/metadata/side_metadata/global.rs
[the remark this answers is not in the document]
wks: I now use `vo_bit::get_object_ref_for_vo_addr` to get object reference from raw Address.

src/mmtk.rs
wks: I updated the comment.  I made it clear that it is a data race if either allocation or GC happens concurrently with `enumerate_objects`.  I also added some comments on keeping the saved object references in roots if the VM binding implements an API that allows allocation while enumerating objects.

qinsoon (approved):
LGTM

PR #1178, opened 2024-07-26
Fix style check for Rust 1.80

qinsoon:


.github/workflows/api-check.yml
```diff
@@ -35,9 +35,9 @@ jobs:
       - run: cargo +nightly --version
 
       - name: Install cargo-public-api
-        run: cargo install cargo-public-api
+        run: cargo +nightly install cargo-public-api
```
wks, line 38: Why do we need to use the nightly toolchain to run `cargo-public-api`?

qinsoon: There is no guarantee that `cargo-public-api` will work with our pinned Rust version (and in this case, it only works with Rust 1.78+ due to some of its dependencies). We need to use a relatively new toolchain.

We don't have to use nightly. However, `cargo-public-api` requires a nightly toolchain to be installed. It is natural to just use nightly.

wks (approved):
LGTM

PR #1180, opened 2024-07-29
Refactor bulk metadata visiting

wks:
**DRAFT**:

-   There is still performance disadvantage over the old approach.  Find the cause.
-   (Re-)implement bulk update, VO bit searching (internal pointer) and VO bit scanning (traversal).

This PR refactors the mechanism for visiting (reading and/or updating) side metadata in bulk.  Comparing with the existing approach, the new approach

-   uses only one callback for both bit ranges an byte ranges (TODO: Try not to use callback.  Use iterator instead.)
-   allows arbitrary granularity instead of bytes-only
-   (TODO) allows breaking discontiguous metadata for a contiguous data address range into contiguous metadata sub-ranges. (Do we need it? We only need it for bulk visiting metadata for data ranges larger than 4K.)

Using one callback allows an arbitrary `FnMut` callback (instead two `Fn`s) to be used to visit things (such as valid object references via VO bits) in a metadata range.  This also removes the use of `Cell` when finding the last VO bit from an internal pointer.

Now the side metadata can be visited at any granularity, such as `usize`.  We no longer have to do it in two steps, from bits to bytes (`iterate_meta_bits`) and from bytes to words (`find_last_non_zero_bit_in_metadata_bytes`).  Now anything smaller than a grain (a `usize` for example) are treated as a "sub-grain range" and visited bit by bit (e.g. for each bit in a `usize`...).

(TODO) Make a generic mechanism to break a contiguous data address range into a single or multiple contiguous metadata ranges, so that concrete meta bit traversal mechanism can be expressed in several levels of nested `for` loops (or nested callbacks).

### DIFF
--- a/benches/bulk_meta/bzero.rs
+++ b/benches/bulk_meta/bzero.rs
@@ -21,6 +21,13 @@ pub fn bench(c: &mut Criterion) {
         })
     });
 
+    c.bench_function("bzero_bset_modern_callback", |b| {
+        b.iter(|| {
+            SideMetadataSpec::set_meta_bits_callback(start, 0, end, 0);
+            SideMetadataSpec::zero_meta_bits_callback(start, 0, end, 0);
+        })
+    });
+
     c.bench_function("bzero_bset_classic", |b| {
         b.iter(|| {
             SideMetadataSpec::set_meta_bits_classic(start, 0, end, 0);
--- a/benches/bulk_meta/bzero.rs
+++ b/benches/bulk_meta/bzero.rs
@@ -1,0 +1,30 @@
+use criterion::Criterion;
+use mmtk::util::{metadata::side_metadata::SideMetadataSpec, Address};
+
+pub fn bench(c: &mut Criterion) {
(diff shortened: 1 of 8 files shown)


k-sareen:
It might interest you to see how Android implements this: https://cs.android.com/android/platform/superproject/main/+/main:art/runtime/gc/accounting/space_bitmap-inl.h;l=103-229?q=visitMarkedRange

They have two functions `VisitMarkedRange` for arbitrary begin and end and `Walk` for going through the entire bitmap (you can optimize it to just load words, skipping all the extra complexity of the left- vs right-edge).

qinsoon:
If you would like to push further with this PR, you would need to show that the refactoring will result in improvement in favorable cases and will not slow down in other cases. Otherwise it is not worth to introduce the complexity.

wks:
Since this is too complicated and doesn't perform well, I am closing this PR in favor for https://github.com/mmtk/mmtk-core/pull/1181  I'll continue to do refactoring when needed in the future, bit by bit.

PR #1182, opened 2024-08-01
Extra assertions for mutators and epilogues

wks:
This commit adds assertions about the number of mutators and the list of mutators returned from the VM binding.

Also asserts that epilogues are not silently dropped before they are executed.  It could happen if a buggy VM binding returns a number of mutators that does not match the actual list of mutators.

This should detect bugs like https://github.com/mmtk/mmtk-ruby/issues/84

### DIFF
--- a/src/util/epilogue.rs
+++ b/src/util/epilogue.rs
@@ -1,0 +1,13 @@
+//! Utilities for implementing epilogues.
+//!
+//! Epilogues are operations that are done when several other operations are done.
+
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+/// A debugging method for detecting the case when the epilogue is never executed.
+pub fn debug_assert_counter_zero(counter: &AtomicUsize, what: &'static str) {
+    let value = counter.load(Ordering::SeqCst);
+    if value != 0 {
+        panic!("{what} is still {value}.");
+    }
+}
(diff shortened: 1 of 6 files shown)


wks:
binding-refs
RUBY_BINDING_REF=fix/mutator-count

src/util/epilogue.rs
```diff
@@ -0,0 +1,13 @@
...
+//!
+//! Epilogues are operations that are done when several other operations are done.
+
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+/// A debugging method for detecting the case when the epilogue is never executed.
+pub fn debug_assert_counter_zero(counter: &AtomicUsize, what: &'static str) {
+    let value = counter.load(Ordering::SeqCst);
+    if value != 0 {
+        panic!("{what} is still {value}.");
```
qinsoon, line 11: Instead of "FlushPageResource::counter is still 1", you can have a clearer error message like "The epilogue FlushPageResource is not executed. Something is wrong". Anyway this is a minor issue. I am okay if the PR is merged without any change.

qinsoon (approved):
LGTM

src/util/epilogue.rs
wks: I thought about that.  I used to include a "when" parameter to this function, too.  Then the more accurate message will be "`space.flush_page_resource()` is not executed when `FlushPageResource` is dropped", or "`self.generate_sweep_tasks()` or `self.recycle_blocks()` is not executed before end of GC".  That may become too detailed for the panic message.  I think the counter name is descriptive enough for a panic message, and the line number in the panic message can pinpoint the line of code where the error occurred.  That should be enough for developers to understand what was not executed.

PR #1184, opened 2024-08-07
Parallelize and workaround CI style checks

wks:
We now do style checks in parallel to `cargo test` so that style problems will fail fast.

Black-listed clippy 0.1.77 on x86_64-apple-darwin because it is known to crash randomly.

Also added missing style checks for non-mock benchmarks.


wks:
`cargo clippy` needs to cross-compile mimalloc if the host is different from the target.  That needs special setup for cross compilation, which is more complicated than running natively.  I will simply blacklist cargo clippy 1.77.0 on darwin.

qinsoon (approved):

PR #1188, opened 2024-08-26
Fix find_last_non_zero_bit, and align metadata address before converting to data address.

qinsoon:
This PR fixes some bugs in finding base references for internal references. Prominently two bugs:
1. The way we find the last non zero bit was wrong. We used to to use `trailing_zeroes` and compare the result with the given range of the starting and ending bits. This was simply wrong. Now we do a mask first to extract the value between the starting and the ending bits, and then use `leading_zeroes`. Performance-wise, the new code has similar performnace.
2. When we convert a metadata bit (address + bit offset) back to the data address, the metadata bit needs to be the start of the metadata value. If it is the middle of the metadata value, then the computed data address will be incorrect. This PR adds `align_metadata_address`.

This PR skips some tests for LOS in plans that do not use LOS.

### DIFF
--- a/src/util/metadata/side_metadata/helpers.rs
+++ b/src/util/metadata/side_metadata/helpers.rs
@@ -6,6 +6,7 @@ use crate::util::heap::layout::vm_layout::VMLayout;
 use crate::util::memory::MmapStrategy;
 #[cfg(target_pointer_width = "32")]
 use crate::util::metadata::side_metadata::address_to_chunked_meta_address;
+use crate::util::metadata::MetadataValue;
 use crate::util::Address;
 use crate::MMAPPER;
 use std::io::Result;
@@ -28,6 +29,7 @@ pub(super) fn address_to_contiguous_meta_address(
 }
 
 /// Performs reverse address translation from contiguous metadata bits to data addresses.
+/// The input address and bit shift should be aligned.
 ///
 /// Arguments:
 /// * `metadata_spec`: The side metadata spec. It should be contiguous side metadata.
@@ -38,6 +40,10 @@ pub(super) fn contiguous_meta_address_to_address(
     metadata_addr: Address,
     bit: u8,
 ) -> Address {
+    debug_assert_eq!(
+        align_metadata_address(metadata_spec, metadata_addr, bit),
+        (metadata_addr, bit)
+    );
     let shift = (LOG_BITS_IN_BYTE as i32) - metadata_spec.log_num_of_bits as i32;
     let relative_meta_addr = metadata_addr - metadata_spec.get_absolute_offset();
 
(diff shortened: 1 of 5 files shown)


src/util/metadata/side_metadata/helpers.rs
```diff
@@ -252,22 +281,41 @@ pub fn find_last_non_zero_bit_in_metadata_bits(
...
-        None
-    }
+macro_rules! impl_find_last_non_zero_bit {
+    ($int_ty: ty, $value: expr, $start: expr, $end: expr) => {{
+        let mask = match (1 as $int_ty).checked_shl(($end - $start) as u32) {
+            Some(shl) => (shl - 1) << $start,
+            None => <$int_ty>::MAX << $start,
+        };
+        let leading_zeroes = ($value & mask).leading_zeros();
+        if leading_zeroes >= <$int_ty>::BITS {
```
wks, line 297: We can just compare `$value & mask` with zero.  If not zero, there must be at least one non-zero bit.

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -197,15 +228,21 @@ pub fn find_last_non_zero_bit_in_metadata_bytes(
     let mut mapped_chunk = Address::MAX;
     while cur > meta_start {
         // If we can check the whole word, set step to word size. Otherwise, the step is 1 (byte) and we check byte.
-        let step = if cur.is_aligned_to(BYTES_IN_ADDRESS)
-            && cur.align_down(BYTES_IN_ADDRESS) >= meta_start
-        {
+        let step = if cur.is_aligned_to(BYTES_IN_ADDRESS) && cur - BYTES_IN_ADDRESS >= meta_start {
```
qinsoon, line 231: This is the other fix that is needed. The old code may check an address that is below `meta_start`.

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -252,22 +281,41 @@ pub fn find_last_non_zero_bit_in_metadata_bits(
...
+    // and does not worth the efforts.
+    impl_find_last_non_zero_bit!(u8, value, start, end)
+}
+
+fn find_last_non_zero_bit_usize(value: usize, start: u8, end: u8) -> Option<u8> {
+    // TODO: Ideally we implement the function as a generic function.
+    // However Rust does not have a generic trait for integers yet. We can use
+    // third party traits or our own trait (such as MetadataValue), but it is cumbersome,
+    // and does not worth the efforts.
+    impl_find_last_non_zero_bit!(usize, value, start, end)
```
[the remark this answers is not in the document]
qinsoon: Done. I used your implemenation. Thanks.

wks (approved):
LGTM

PR #1189, opened 2024-08-28
Skip iterating through unallocated memory for `MonotonePageResource`

k-sareen:


k-sareen:
It's wholly unnecessary to return the entire chunk when we know where the cursor is.

qinsoon (approved):
LGTM

PR #1191, opened 2024-08-29
Panic message and outdated comments

wks:
This commit contains some minor fixes.

-   Added the actual address to the panic message when the assertions
    about ObjectReference alignment fails.
-   Removed an outdated comment about the use of Cell.


qinsoon (approved):
LGTM

PR #1192, opened 2024-09-02
Filter the result for find_prev_non_zero_value_fast to make sure it is in the search range

qinsoon:


wks (approved):
LGTM.

This solves the immediate problem.  In the future, we may refactor the API so that the user always ask for a lower bound that is aligned to the region size of VO bits so that we won't need this filtering.

PR #1194, opened 2024-09-02
Assert OBJECT_REF_OFFSET_LOWER_BOUND invariant

wks:
Do a `debug_assert` on the invariant of `ObjectModel::OBJECT_REF_OFFSET_LOWER_BOUND` when calling `address.to_object_start::<VM>()`.  VM bindings may violate this invariant when changing their object models.


qinsoon (approved):
LGTM

PR #1195, opened 2024-09-05
Require ObjectReference to point inside object

wks:
Require the raw address of `ObjectReference` to be within the address range of the object it refers to.  The raw address is now used directly for side metadata access and SFT dispatching.  This makes "in-object address" unnecessary, and we removed the concept of "in-object address" and related constants and methods.

Methods which use the "in-object address" for SFT dispatching or side-metadata access used to have a `<VM: VMBinding>` type parameter.  This PR removes that type parameter.

Because `ObjectReference` is now both within an object an word-aligned, the algorithm for searching for VO bits from internal pointers is slightly simplified.  The method `is_mmtk_object` now has undefined behavior for arguments that are zero or misaligned because they are obviously illegal addresses for `ObjectReference`, and the user should have filtered them out in the first place.

Fixes: https://github.com/mmtk/mmtk-core/issues/1170

### DIFF
--- a/docs/userguide/src/portingguide/howto/nogc.md
+++ b/docs/userguide/src/portingguide/howto/nogc.md
@@ -98,9 +98,12 @@ We recommend going through the [list of metadata specifications](https://docs.mm
 A key principle in MMTk is the distinction between [`ObjectReference`](https://docs.mmtk.io/api/mmtk/util/address/struct.ObjectReference.html) and [`Address`](https://docs.mmtk.io/api/mmtk/util/address/struct.Address.html). The idea is that very few operations are allowed on an `ObjectReference`. For example, MMTk does not allow address arithmetic on `ObjectReference`s. This allows us to preserve memory-safety, only performing unsafe operations when required, and gives us a cleaner and more flexible abstraction to work with as it can allow object handles or offsets etc. `Address`, on the other hand, represents an arbitrary machine address. You might be interested in reading the [*Demystifying Magic: High-level Low-level Programming*][FBC09] paper which describes the above in more detail.
 
 In MMTk, `ObjectReference` is a special address that represents an object.  It is required to be
-with the address range of the object it refers to, and must be word-aligned.  This address is used
+within the address range of the object it refers to, and must be word-aligned.  This address is used
 by MMTk to access side metadata, and find the space or regions (chunk, block, line, etc.) that
-contains the object.
+contains the object.  It must also be efficient to locate the object header (where in-header MMTk
+metadata are held) and the object's VM-specific metadata, such as type information, from a given
+`ObjectReference`.  MMTk will need to access those information, either directly or indirectly via
+traits implemented by the binding, during tracing, which is performance-critical.
 
 The address used as `ObjectReference` is nominated by the VM binding when an object is allocated (or
 moved by a moving GC, which we can ignore for now when supporting NoGC).  VMs usually have their own
--- a/docs/userguide/src/portingguide/howto/nogc.md
+++ b/docs/userguide/src/portingguide/howto/nogc.md
@@ -95,13 +95,36 @@ We recommend going through the [list of metadata specifications](https://docs.mm
 
 #### `ObjectReference` vs `Address`
 
-A key principle in MMTk is the distinction between [`ObjectReference`](https://docs.mmtk.io/api/mmtk/util/address/struct.ObjectReference.html) and [`Address`](https://docs.mmtk.io/api/mmtk/util/address/struct.Address.html). The idea is that very few operations are allowed on an `ObjectReference`. For example, MMTk does not allow address arithmetic on `ObjectReference`s. This allows us to preserve memory-safety, only performing unsafe operations when required, and gives us a cleaner and more flexible abstraction to work with as it can allow object handles or offsets etc. `Address`, on the other hand, represents an arbitrary machine address. You might be interested in reading the *Demystifying Magic: High-level Low-level Programming* paper[^3] which describes the above in more detail.
+A key principle in MMTk is the distinction between [`ObjectReference`](https://docs.mmtk.io/api/mmtk/util/address/struct.ObjectReference.html) and [`Address`](https://docs.mmtk.io/api/mmtk/util/address/struct.Address.html). The idea is that very few operations are allowed on an `ObjectReference`. For example, MMTk does not allow address arithmetic on `ObjectReference`s. This allows us to preserve memory-safety, only performing unsafe operations when required, and gives us a cleaner and more flexible abstraction to work with as it can allow object handles or offsets etc. `Address`, on the other hand, represents an arbitrary machine address. You might be interested in reading the [*Demystifying Magic: High-level Low-level Programming*][FBC09] paper which describes the above in more detail.
 
-In MMTk, `ObjectReference` is a special address that represents an object. A binding may use tagged references, compressed pointers, etc.
-They need to deal with the encoding and the decoding in their [`Slot`](https://docs.mmtk.io/api/mmtk/vm/slot/trait.Slot.html) implementation,
-and always present plain `ObjectReference`s to MMTk. See [this test](https://github.com/mmtk/mmtk-core/blob/master/src/vm/tests/mock_tests/mock_test_slots.rs) for some `Slot` implementation examples.
+In MMTk, `ObjectReference` is a special address that represents an object.  It is required to be
+with the address range of the object it refers to, and must be word-aligned.  This address is used
+by MMTk to access side metadata, and find the space or regions (chunk, block, line, etc.) that
+contains the object.
 
-[^3]: https://users.cecs.anu.edu.au/~steveb/pubs/papers/vmmagic-vee-2009.pdf
+The address used as `ObjectReference` is nominated by the VM binding when an object is allocated (or
+moved by a moving GC, which we can ignore for now when supporting NoGC).  VMs usually have their own
+concepts of "object reference" which refer to objects.  Some of them, including OpenJDK and CRuby,
+uses addresses to the object (the starting address or at an offset within the object) to refer to an
+object.  Such VMs can directly use their "object reference" for the address of MMTk's
+`ObjectReference`.
+
+Some VMs, such as JikesRVM, refers to an object by an address at a constant offset after the header,
+and can be outside the object.  This does not satisfy the requirement of MMTk's `ObjectReference`,
+and the VM binding needs to make a clear distinction between the VM-level object reference and
+MMTk's `ObjectReference` type.  A detailed example for supporting such a VM can be found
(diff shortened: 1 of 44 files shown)


qinsoon:
This section in the porting guide should be updated: https://docs.mmtk.io/portingguide/howto/nogc.html#objectreference-vs-address. It would be better to add an example to reference the pattern in JikesRVM to deal with different language/VM-level object reference and MMTK-level object reference.

wks:
binding-refs
OPENJDK_BINDING_REF=features/objref-in-object2
JIKESRVM_BINDING_REPO=wks/mmtk-jikesrvm
JIKESRVM_BINDING_REF=features/objref-in-object2
V8_BINDING_REPO=wks/mmtk-v8
V8_BINDING_REF=features/objref-in-object2
RUBY_BINDING_REF=features/objref-in-object2
JULIA_BINDING_REPO=wks/mmtk-julia
JULIA_BINDING_REF=features/objref-in-object2

wks:
> This section in the porting guide should be updated: https://docs.mmtk.io/portingguide/howto/nogc.html#objectreference-vs-address. It would be better to add an example to reference the pattern in JikesRVM to deal with different language/VM-level object reference and MMTK-level object reference.

I added more details to that section.  I also changed the "Allocation" section below to emphasize that the VM binding shall nominate an address used for `ObjectReference`.

wks:
Related PRs:

- https://github.com/mmtk/mmtk-openjdk/pull/286
- https://github.com/mmtk/mmtk-jikesrvm/pull/180
- https://github.com/mmtk/mmtk-v8/pull/88
- https://github.com/mmtk/mmtk-ruby/pull/100
- https://github.com/mmtk/mmtk-julia/pull/173

qinsoon:
With the change of this PR, `ObjectReference` could be different from the VM-level object reference anyway, and the object start address always satisfies the definition of `ObjectReference`. Why don't we just use the object start address as `ObjectReference`? In that case, we don't need the term 'the object start address', the bindings don't need to pick a MMTk `ObjectReference`, and `alloc()` simply returns an `ObjectReference`. It further simplifies our API. 

It may be more costly for VMs like JikesRVM to convert between VM-level object reference and `ObjectReference`/'object start address'. But it would be interested to see if the cost is significant or not.

Anyway, this is just some thoughts. I will review the PRs tomorrow.

wks:
> ... Why don't we just use the object start address as `ObjectReference`? ...
> 
> It may be more costly for VMs like JikesRVM to convert between VM-level object reference and `ObjectReference`/'object start address'. But it would be interested to see if the cost is significant or not.
> 
> ...

It may be easy to compute the starting address from the ObjectReference (which *usually* points at the header, or at a constant offset from the header), but the other way (i.e. computing the ObjectReference, the header, or whatever needed to access in-object metadata) may not be easy, if possible at all.

One example is JikesRVM.  Given the staring address, MMTk may call `Scanning::scan_object(starting_address, field_visitor)`.  Then JikesRVM needs to find the TIB, and therefore needs to decide whether the JavaHeader is at `starting_address` (when not having a hash in the front) or `starting_address + 4` (with a 4-byte hash in the front).  This is difficult, but possible.  JikesRVM can reserve one bit in the JavaHeader, ensure that the bit is always 0 in the JavaHeader, but the same bit is always 1 in the hash.  The GC can tell whether the first word is the JavaHeader or the hash by looking at that bit.  That's complicated anyway, and may be invasive for other VMs.

Another example is SableVM.  Its header is in the middle of the object.  Reference and non-reference fields span towards both directions.  In this case, it is impossible to find the header when given the starting address because the fields can simply contain arbitrary data that may look like the header.

So it has to be decided by the VM binding for now, to find a middle ground between a simple API and make it efficient to find object metadata.

Speaking of metadata, I think `objref.to_header()` may or may not be necessary, given that the offsets of each metadata spec can be arbitrary.  It's convenient for JikesRVM because all metadata offsets were relative to the JikesRVM-level `ObjectReference`.  For other VMs, it is usually a constant offset from `ObjectReference` for efficiency reasons.  But let's discuss about that in the future.

docs/userguide/src/portingguide/howto/nogc.md
```diff
@@ -95,13 +95,36 @@ We recommend going through the [list of metadata specifications](https://docs.mm
...
-
-[^3]: https://users.cecs.anu.edu.au/~steveb/pubs/papers/vmmagic-vee-2009.pdf
+A key principle in MMTk is the distinction between [`ObjectReference`](https://docs.mmtk.io/api/mmtk/util/address/struct.ObjectReference.html) and [`Address`](https://docs.mmtk.io/api/mmtk/util/address/struct.Address.html). The idea is that very few operations are allowed on an `ObjectReference`. For example, MMTk does not allow address arithmetic on `ObjectReference`s. This allows us to preserve memory-safety, only performing unsafe operations when required, and gives us a cleaner and more flexible abstraction to work with as it can allow object handles or offsets etc. `Address`, on the other hand, represents an arbitrary machine address. You might be interested in reading the [*Demystifying Magic: High-level Low-level Programming*][FBC09] paper which describes the above in more detail.
+
+In MMTk, `ObjectReference` is a special address that represents an object.  It is required to be
+with the address range of the object it refers to, and must be word-aligned.  This address is used
```
qinsoon, line 101:
```suggestion
within the address range of the object it refers to, and must be word-aligned.  This address is used
```

qinsoon: We also require the VM to be able to easily find object metadata/header/type info from `ObjectReference`.

src/util/address.rs
```diff
@@ -467,32 +467,80 @@ mod tests {
...
+///
+/// In MMTk, `ObjectReference` holds a non-zero address.  It either points into one of MMTk's spaces
+/// or not.
+///
+/// -   When the address is in one of MMTk's spaces, it refers to an object allocated in that space.
+///     In this case, the address must satisfy the following requirements.
+///     -   The address must be within the address range of the object it refers to.  But it is not
+///         necessarily the starting address.
+///     -   The address must be word-aligned.
+/// -   When the address is outside any MMTk space, its semantics is VM-defined.
```
qinsoon, line 483: I don't think we have well-defined semantics for `ObjectReference` that points outside MMTk spaces. For example, if we access side metadata, we may see unmapped side metadata and we may panic.

src/util/address.rs
```diff
@@ -467,32 +467,80 @@ mod tests {
...
-/// ObjectReference represents address for an object. Compared with Address,
-/// operations allowed on ObjectReference are very limited. No address arithmetics
-/// are allowed for ObjectReference. The idea is from the paper
-/// High-level Low-level Programming (VEE09) and JikesRVM.
+/// `ObjectReference` represents address for an object. Compared with `Address`, operations allowed
+/// on `ObjectReference` are very limited. No address arithmetics are allowed for `ObjectReference`.
+/// The idea is from the paper [Demystifying Magic: High-level Low-level Programming (VEE09)][FBC09]
+/// and [JikesRVM].
+///
+/// In MMTk, `ObjectReference` holds a non-zero address.  It either points into one of MMTk's spaces
```
qinsoon, line 475: This is a good place to specify the requirements of `ObjectReference`:
1. uniquely references an object for MMTk
2. within allocated memory (VM space is an exception which needs to be discussed separately)
3. easy to get object info from `ObjectReference`

qinsoon: Probably don't say an `ObjectReference` 'points into one of MMTk's spaces'. When we have contiguous spaces, the entire address range is considered as the MMTk space, but not any address that is in the range can be an `ObjectReference`. An `ObjectReference` has to point inside the allocated memory.

src/vm/object_model.rs
```diff
@@ -432,10 +425,9 @@ pub trait ObjectModel<VM: VMBinding> {
     /// mature space for generational plans.
     const VM_WORST_CASE_COPY_EXPANSION: f64 = 1.5;
 
-    /// If this is true, the binding guarantees that the object reference's raw address,
-    /// the in-object address, and the object start are always the same address. To be precise,
-    /// 1. an object reference's raw address is always equal to the return value of the `ref_to_object_start` method,
-    /// 2. `IN_OBJECT_ADDRESS_OFFSET` is 0.
+    /// If this is true, the binding guarantees that the object reference's raw addressand the
```
qinsoon, line 428:
```suggestion
    /// If this is true, the binding guarantees that the object reference's raw address and the
```

qinsoon (commented):
The code is good. The document needs a bit improvement to make it clear about our reuirement for `ObjectReference`.

src/util/address.rs
wks: I did say it is "VM-defined".  But I can emphasize that MMTk won't access side metadata from an `ObjectReference` without knowing it is a valid reference to an MMTk object (or at least in one of MMTk's spaces), and that VM bindings shouldn't access side metadata using `ObjectReference`s that don't point to a valid MMTk object.

qinsoon: > (I can emphasize that) MMTk won't access side metadata from an `ObjectReference` without knowing it is a valid reference to an MMTk object (or at least in one of MMTk's spaces)

That's what I was concerned about. I don't think we can guarantee that with the current code base.

This is not the focus of this PR. So I think it is fine to say `ObjectReference` is only for our allocated memory (with VM space being an exception). We don't need to mention `ObjectReference` to non-MMTk memory. A binding may use `ObjectReference` to point to non-MMTk memory -- it is a hack, it may work or  it may not, we don't promise anything.

wks: Maybe I can move "ObjectReference to non-MMTk memory" down to the "Notes" section below, and briefly mention `ActivePlan::vm_trace_object` and the fact that this feature is experimental and may not work.  I just feel that the doc comment of `ObjectReference` should not contradict with that of `ActivePlan::vm_trace_object`.

qinsoon: Yeah. This sounds good.

docs/userguide/src/portingguide/howto/nogc.md
wks: I mentioned it must be efficient to access metadata and/or type info from `ObjectReference` in both the porting guide and the doc comment of `ObjectReference`.

src/util/address.rs
wks: Done.

qinsoon (approved):
LGTM

src/util/address.rs
wks: I added uniqueness to the doc, and mentioned that it must be easy to get object info from `ObjectReference`.  I also added a section for `VMSpace`.

I now say "`ObjectReference` that refers to an MMTk object" instead of "`ObjectReference" that points into one of MMTk's spaces".

src/vm/object_model.rs
wks: Done.

PR #1199, opened 2024-09-11
Update ci-perf-kit to 0.7.6

qinsoon:
Add a note about kernel update on 20240903 for CI machines.


qinsoon:
The public API check is failing. It seems that rust nightly updated its json output (`rustc 1.83.0-nightly (0ee7cb5e3 2024-09-10)`), and `cargo-public-api` isn't updated to that version yet.

wks (approved):
LGTM.

PR #1202, opened 2024-09-17
Add a feature immix_stress_copying

qinsoon:
This PR adds a feature `immix_stress_copying`. The bindings can turn on stress copying for Immix without modifying the source code. We plan to test with stress copying for Julia in the CI.


k-sareen (approved):
LGTM. Thanks!

PR #1205, opened 2024-09-20
Return if a GC ran or not for `handle_user_collection_request`

### LINKED ISSUE #1204: Return whether a GC was ran or not for `handle_user_collection_request`
This is useful if the binding wants to know if a GC was actually ran or not. For example, they may have set some state in preparation of a GC, but if a GC didn't run then they would want to reset the state.

k-sareen:
Closes #1204

### DIFF
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -401,7 +401,11 @@ impl<VM: VMBinding> MMTK<VM> {
     }
 
     /// The application code has requested a collection. This is just a GC hint, and
-    /// we may ignore it. Returns whether a GC was ran or not.
+    /// we may ignore it.
+    ///
+    /// Returns whether a GC was ran or not. If MMTk triggers a GC, this method will block the
+    /// calling thread and return true when the GC finishes. Otherwise, this method returns
+    /// false immediately.
     ///
     /// # Arguments
     /// * `tls`: The mutator thread that requests the GC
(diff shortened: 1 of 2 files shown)


src/mmtk.rs
```diff
@@ -432,7 +432,10 @@ impl<VM: VMBinding> MMTK<VM> {
                 .store(true, Ordering::Relaxed);
             self.gc_requester.request();
             VM::VMCollection::block_for_gc(tls);
+            return true;
```
qinsoon, line 435: The user only gets the return value after the GC is finished. Is that still useful?

k-sareen: I mean that's how it should be right? I'm not sure I understand your question.

src/mmtk.rs
```diff
@@ -401,7 +401,7 @@ impl<VM: VMBinding> MMTK<VM> {
     }
 
     /// The application code has requested a collection. This is just a GC hint, and
-    /// we may ignore it.
+    /// we may ignore it. Returns whether a GC was ran or not.
```
qinsoon, line 404:
```suggestion
    /// we may ignore it. If MMTk triggers a GC, this method will block the current thread
    /// and return true when the GC finishes. Otherwise, this method returns false immediately.
```

src/mmtk.rs
qinsoon: OK. Maybe document this more cleraly. See my other comment.

src/mmtk.rs
k-sareen: Done.

src/mmtk.rs
k-sareen: Done.

qinsoon (approved):
LGTM

PR #1208, opened 2024-09-24
Update ci-perf-kit to plot epoch

qinsoon:
Use https://github.com/mmtk/ci-perf-kit/pull/46 to plot performance data.


wks (approved):
LGTM

PR #1214, opened 2024-10-11
Fix a race between forwarding bits and VO bits.

wks:
The current code sets the forwarding bits before setting the VO bit when copying an object.  If another GC worker is attempting to forward the same object, it may observe the forwarding bits being `FORWARDED` but the VO bit is not set.  This violates the semantics of VO bits because VO bits should be set for both from-space and to-space copies.  This will affect VM bindings that assert slots always refer to a valid object when scanning objects and may update the same slot multiple times for some reasons.

This revision provides a mechanism to ensure that all necessary metadata are set before setting forwarding bits to `FORWARDED`.  Currently it affects the VO bits and the mark bits (which are used to update the VO bits in Immix-based plans).  It may be used for other metadata introduced in the future.

### DIFF
--- a/src/util/object_forwarding.rs
+++ b/src/util/object_forwarding.rs
@@ -74,12 +74,30 @@ pub fn spin_and_get_forwarded_object<VM: VMBinding>(
     }
 }
 
+/// Copy an object and set the forwarding state.
+///
+/// The caller can use `on_after_forwarding` to set extra metadata (including VO bits, mark bits,
+/// etc.) after the object is copied, but before the forwarding state is changed to `FORWARDED`. The
+/// atomic memory operation that sets the forwarding bits to `FORWARDED` has the `SeqCst` order.  It
+/// will guarantee that if another GC worker thread that attempts to forward the same object sees
+/// the forwarding bits being `FORWARDED`, it is guaranteed to see those extra metadata set.
+///
+/// Arguments:
+///
+/// *   `object`: The object to copy.
+/// *   `semantics`: The copy semantics.
+/// *   `copy_context`: A reference ot the `CopyContext` instance of the current GC worker.
+/// *   `on_after_forwarding`: A callback function that is called after `object` is copied, but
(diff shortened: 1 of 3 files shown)


qinsoon (approved):
LGTM

qinsoon:
```
error: cannot install package `cargo-msrv 0.16.2`, it requires rustc 1.78 or newer, while the currently active rustc version is 1.77.0
`cargo-msrv 0.1[6](https://github.com/mmtk/mmtk-core/actions/runs/11285882886/job/31389361498?pr=1214#step:5:7).0` supports rustc 1.76
```

Maybe we should pin to a specific `cargo-msrv` version before we bump our Rust toolchain version.

tianleq (approved):
LGTM

k-sareen (approved):
LGTM. Thanks!

src/util/object_forwarding.rs
```diff
@@ -74,12 +74,30 @@ pub fn spin_and_get_forwarded_object<VM: VMBinding>(
     }
 }
 
+/// Copy an object and set the forwarding state.
+///
+/// The caller can use `on_after_forwarding` to set extra metadata (including VO bits, mark bits,
+/// etc.) after the object is copied, but before the forwarding state is changed to `FORWARDED`. The
+/// atomic memory operation that sets the forwarding bits to `FORWARDED` has the `SeqCst` order.  It
```
k-sareen, line 81: Actually minor point. This is an implementation detail, so maybe we shouldn't document it in the API doc comment

qinsoon: I think it's fine. This function should be private to the crate, not public.

wks: That's OK.  The entire `util::object_forwarding` module is private, so the doc comments are only relevant to the developers of the mmtk-core.  We may update the doc comments if we change it to acquire-release later.

k-sareen: Sure 👍

PR #1216, opened 2024-10-16
Update CI macos image

qinsoon:
macos-12 will no longer be supported: https://github.com/actions/runner-images/issues/10721


wks (approved):
LGTM

PR #1218, opened 2024-10-29
Add alloc_with_options

qinsoon:
This PR adds an alternative allocation functions `alloc_with_options`. The attached `AllocationOptions` change the allocation behavior, such as avoiding GCs and allowing overcommit.

Our current `alloc` functions assume they are GC safe points and will trigger GCs internally. A runtime may have different assumptions. We see GHC has `allocateMightFail` (https://gitlab.haskell.org/ghc/ghc/-/blob/90746a591919fc51a0ec9dec58d8f1c8397040e3/rts/sm/Storage.c?page=2#L1089). Also Julia assumes perm alloc will not trigger a GC (https://github.com/mmtk/mmtk-julia/issues/172).
Having a variant of `alloc` that is not GC safepoints could be generally useful.


src/memory_manager.rs
```diff
@@ -159,24 +171,48 @@ pub fn alloc<VM: VMBinding>(
...
+    #[cfg(debug_assertions)]
+    crate::util::alloc::allocator::asset_allocation_args::<VM>(size, align, offset);
 
     mutator.alloc(size, align, offset, semantics)
 }
 
-/// Invoke the allocation slow path. This is only intended for use when a binding implements the fastpath on
+/// Allocate memory for an object. This function is NOT a GC safepoint. If MMTk fails
+/// to allocate memory, it will not attempt a GC, nor call [`crate::vm::Collection::out_of_memory`],
+/// MMTk instead return a zero address.
```
wks, line 182:
```suggestion
/// Allocate memory for an object. This function is guaranteed not to trigger GC and not call `Collection::block_for_gc` or [`crate::vm::Collection::out_of_memory`].  Instead it will return zero if MMTk fails to allocate memory.
```

src/memory_manager.rs
```diff
@@ -197,6 +233,28 @@ pub fn alloc_slow<VM: VMBinding>(
     mutator.alloc_slow(size, align, offset, semantics)
 }
 
+/// Invoke the allocation slow path. This function is NOT a GC safepoint. If MMTk fails
+/// to allocate memory, it will not attempt a GC, nor call [`crate::vm::Collection::out_of_memory`],
+/// MMTk instead return a zero address. See [`alloc_no_gc`] for more details.
+///
+/// Notes on [`alloc_slow_no_gc`] also apply to this function.
```
wks, line 240: I rephrased it a bit and refererence `alloc_no_gc`.

Instead of saying "notes on xxx also apply", I references `alloc_slow` and briefly quotes the related note.

```suggestion
/// Invoke the allocation slow path of [`alloc_no_gc`].
///
/// Like [`alloc_no_gc`], this function is guaranteed not to trigger GC and not to call
/// `Collection::block_for_gc` or [`crate::vm::Collection::out_of_memory`].  It returns zero on
/// allocation failure.
///
/// Like [`alloc_slow`], this function is also only intended for use when a binding implements the
/// fastpath on the binding side.
```

wks (commented):
I changed parts of the doc comments of `alloc`, `alloc_no_gc`, `alloc_slow` and `alloc_slow_no_gc`.  Particularly, I rewrote the comment of `alloc` to make it more ordered.

qinsoon:
I changed the documentation based on the suggestions.

src/memory_manager.rs
```diff
@@ -140,11 +140,34 @@ pub fn flush_mutator<VM: VMBinding>(mutator: &mut Mutator<VM>) {
...
+/// [`alloc_no_gc`] if it is undesirable to trigger GC at this allocation site.
+///
+/// If MMTk has attempted at least one GC, and still cannot free up enough memory, it will call
+/// [`crate::vm::Collection::out_of_memory`] to inform the binding. The VM binding
+/// can implement that method to handle the out-of-memory event in a VM-specific way, including but
+/// not limited to throwing exceptions or errors. If [`crate::vm::Collection::out_of_memory`] returns
+/// normally without panicking or throwing exceptions, this function will return zero.
+///
+/// This function in most cases returns a valid memory address.
+/// This function may return a zero address iif 1. MMTk attempts at least one GC,
```
k-sareen, line 164:
```suggestion
/// This function may return a zero address if 1. MMTk attempts at least one GC,
```

src/util/alloc/allocator.rs
```diff
@@ -130,8 +131,31 @@ pub fn get_maximum_aligned_size_inner<VM: VMBinding>(
     }
 }
 
+#[cfg(debug_assertions)]
+pub(crate) fn asset_allocation_args<VM: VMBinding>(size: usize, align: usize, offset: usize) {
```
k-sareen, line 135:
```suggestion
pub(crate) fn assert_allocation_args<VM: VMBinding>(size: usize, align: usize, offset: usize) {
```

src/plan/mutator_context.rs
```diff
@@ -278,7 +314,24 @@ pub trait MutatorContext<VM: VMBinding>: Send + 'static {
         offset: usize,
         allocator: AllocationSemantics,
     ) -> Address;
-    /// The slow path allocation. This is only useful when the binding
+    /// Allocate memory for an object. This is NOT a GC safepoint. If the allocation fails,
```
k-sareen, line 317: We don't want to use the safepoint terminology here like we discussed in the meeting, right? Same for the functions below.

src/memory_manager.rs
```diff
@@ -159,24 +182,46 @@ pub fn alloc<VM: VMBinding>(
...
+/// when the binding does not want GCs to happen at certain allocation sites (for example, places
+/// where stack maps cannot be generated), and is able to handle allocation failure if that happens.
+///
+/// Arguments:
+/// * `mutator`: The mutator to perform this allocation request.
+/// * `size`: The number of bytes required for the object.
+/// * `align`: Required alignment for the object.
+/// * `offset`: Offset associated with the alignment.
+/// * `semantics`: The allocation semantic required for the allocation.
+pub fn alloc_no_gc<VM: VMBinding>(
```
k-sareen, line 207: What are the semantics of `alloc_no_gc` with `oom_on_acquire`, i.e. what happens when you try to allocate a massive (let's say 25 TB) object? Is it possible to add this as a test?

src/util/alloc/allocator.rs
```diff
@@ -130,8 +131,31 @@ pub fn get_maximum_aligned_size_inner<VM: VMBinding>(
...
+    // Assert alignment
+    debug_assert!(align >= VM::MIN_ALIGNMENT);
+    debug_assert!(align <= VM::MAX_ALIGNMENT);
+    // Assert offset
+    debug_assert!(VM::USE_ALLOCATION_OFFSET || offset == 0);
+}
+
 /// The context an allocator needs to access in order to perform allocation.
 pub struct AllocatorContext<VM: VMBinding> {
+    /// Whether we should trigger a GC when the current alloccation fails.
```
k-sareen, line 153: Since it's actually the negation (i.e. `no_gc_on_fail = true` => no GC triggered), I think maybe something like this is more appropriate:
```suggestion
    /// Whether we should avoid triggering a GC when the current allocation fails.
```

src/memory_manager.rs
qinsoon: Actually this paragraph should be removed.

src/memory_manager.rs
qinsoon: Good point. I made some changes so we won't call `out_of_memory` when the `will_oom_on_acquire` check fails. I also added a test for it.

src/policy/space.rs
```diff
@@ -67,14 +67,16 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
     /// avoid arithmatic overflow. If we have to do computation in the allocation fastpath and
     /// overflow happens there, there is nothing we can do about it.
     /// Return a boolean to indicate if we will be out of memory, determined by the check.
-    fn will_oom_on_acquire(&self, tls: VMThread, size: usize) -> bool {
+    fn will_oom_on_acquire(&self, tls: VMThread, size: usize, allow_oom_call: bool) -> bool {
```
wks, line 70: I suggest splitting `will_oom_on_acquire` into two functions.

-   `will_oom_on_acquire`: This returns `bool`.  If it would OOM, it returns `true`, but doesn't actually call `out_of_memory`.
-   `handle_obvious_oom_request`: Call `will_oom_on_acquire` and, if it returns true, call `out_of_memory`.

One reason is that the name "will_..." suggests it only performs a check, and will not have side effects.  Another reason for this change is that there is a call site in `Space::acquire`:

```rust
        debug_assert!(
            !self.will_oom_on_acquire(tls, pages << LOG_BYTES_IN_PAGE, !no_gc_on_fail),
            "The requested pages is larger than the max heap size. Is will_go_oom_on_acquire used before acquring memory?"
        );
```

This assertion will never fail, but will panic in `Collection::out_of_memory`.  After splitting, `will_oom_on_acquire` will not call `out_of_memory`, and the assertion may fail if it returns false.

There were two more call sites of `will_oom_on_acquire`, in `bumpallocator.rs` and `large_object_allocator.rs`, respectively.  After the split, those two call sites shall call `handle_obvious_oom_request` instead.

I drafted a commit so you can cherry-pick: https://github.com/mmtk/mmtk-core/commit/b6da92ef89389ee0ecbf88ea7810c31e2ec3f2cd

qinsoon: Sounds good. I cherry-picked your changes.

wks (approved):
LGTM

qinsoon:
I’m awaiting feedback from @JunmingZhao42 and Ben before merging this PR, as they requested this feature.

src/policy/space.rs
```diff
@@ -104,17 +117,25 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
...
-            pr.clear_request(pages_reserved);
+            // Inform GC trigger about the pending allocation.
             self.get_gc_trigger()
                 .policy
                 .on_pending_allocation(pages_reserved);
-
             VM::VMCollection::block_for_gc(VMMutatorThread(tls)); // We have checked that this is mutator
-            unsafe { Address::zero() }
+
+            // Return zero -- the caller will handle re-allocation
```
bgamari, line 137:
```suggestion
            // Return zero -- the caller will handle re-attempting allocation
```

wks:
I think this PR still looks good now.

PR #1221, opened 2024-10-31
Change the default testing branch for Julia tests

qinsoon:
We recently re-organised branches in `mmtk-julia` and `julia`. Namely the previous `master` was renamed to `dev`, and we will use `master` for the version that works with Julia upstream. This PR extracts the default testing repos and branches, and changes the default testing branch for Julia.


wks (approved):
LGTM

Issue #744, opened 2023-01-30
Make env_logger optional


wks:
Currently, mmtk-core includes a `log` backend, that is, `env_logger`.  It also attempts to configure the `env_logger` in `memory_manager::mmtk_init`.  This approach, although convenient, may have a problem when the VM wants to include its own logging settings for various reasons.  For example,

1.  The VM binding also uses logs, and wants to log messages before `mmtk_init`.
2.  The VM has some parts written in Rust, too, and they also need logging. (Ruby's YJIT is written in Rust.)

We can have `env_logger` guarded behind a Cargo feature, or the other way around -- adding a Cargo feature to *not* include `env_logger` as part of our dependencies.

qinsoon:
Sounds good. This could also be a runtime option.

Currently we create `env_logger` in `mmtk_init`, and we allow that to fail. That means the binding can initialize a Rust logger before calling `mmtk_init`

PR #1226, opened 2024-11-05
Make env_logger an optional dependency

wks:
Now the built-in `env_logger` is guarded behind a Cargo feature "builtin_env_logger".  It is a default feature, but can be disabled in Cargo.toml by setting `dependencies.mmtk.default-features = false`. In this way, VM bindings that want to implement its own logger can remove the `env_logger` crate from its dependencies.

Fixes: https://github.com/mmtk/mmtk-core/issues/744

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ crossbeam = "0.8.1"
 delegate = "0.12.0"
 downcast-rs = "1.1.1"
 enum-map = "2.4.2"
-env_logger = "0.11.3"
+env_logger = { version = "0.11.3", optional = true }
 is-terminal = "0.4.7"
 itertools = "0.12.0"
 jemalloc-sys = { version = "0.5.3", features = ["disable_initial_exec_tls"], optional = true }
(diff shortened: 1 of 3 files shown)


wks:
JikesRVM failed once with "Failing instruction starting at f7cbd615 wasn't in RVM address space", a known bug.

qinsoon (approved):
LGTM

k-sareen:
I just have a quick question about this. Let's say in the future we add a new feature to the list of default features, is there a way to selectively disable `builtin_env_logger`? or will the binding implementer have to disable the entire `default` feature?

wks:
> I just have a quick question about this. Let's say in the future we add a new feature to the list of default features, is there a way to selectively disable `builtin_env_logger`? or will the binding implementer have to disable the entire `default` feature?

Unfortunately there is no way to selectively disable some default features for now.  There is an issue (https://github.com/rust-lang/cargo/issues/3126) for Cargo, and it is still open now.  But the user can always copy the list of default features and explicitly enable some of them.

k-sareen:
In that case, maybe it is better to make it a negation? As in `no_builtin_env_logger`. Then we can remove it from the default list and since the feature is disabled by default, we get the built-in `env_logger`.

PR #1228, opened 2024-11-06
Fix nightly build and add `inline` attributes to `{un,}likely`

k-sareen:


k-sareen:
I re-added the `inline` attributes to `{un,}likely` to make them consistent with where we [copied them from](https://github.com/rust-lang/hashbrown/blob/a41bd76de0a53838725b997c6085e024c47a0455/src/raw/mod.rs#L48-L70).

wks (approved):
LGTM

PR #1230, opened 2024-11-06
Fix auto merge branches

qinsoon:
This PR changes the auto merge workflow. For each binding, the workflow now allows inputs for base repo and base ref. This change is mostly for the Julia binding which uses `dev` instead of `master` as the default branch. https://github.com/mmtk/mmtk-core/pull/1221 only changed for the correctness testing, this PR made corresponding changes for auto merge.


wks (approved):
LGTM

PR #1236, opened 2024-11-15
Annotate mmap ranges using PR_SET_VMA

wks:
We demand that every invocation of `mmap` within mmtk-core to be accompanied with an "annotation" for the purpose of the mmap.  On Linux, we will use `PR_SET_VMA_ANON_NAME` to set the attribute after `mmap` so that it can be seen in `/proc/pid/maps`.  This will greatly improve the debugging experience.

### DIFF
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -135,6 +135,9 @@ unsafe impl<VM: VMBinding> Send for MMTK<VM> {}
 impl<VM: VMBinding> MMTK<VM> {
     /// Create an MMTK instance. This is not public. Bindings should use [`MMTKBuilder::build`].
     pub(crate) fn new(options: Arc<Options>) -> Self {
+        // Set the mmap annotation option now.  SFT may map memory.
+        crate::util::memory::MMAP_ANNO.store(*options.mmap_anno, Ordering::SeqCst);
+
         // Initialize SFT first in case we need to use this in the constructor.
         // The first call will initialize SFT map. Other calls will be blocked until SFT map is initialized.
         crate::policy::sft_map::SFTRefStorage::pre_use_check();
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -135,7 +135,7 @@ unsafe impl<VM: VMBinding> Send for MMTK<VM> {}
 impl<VM: VMBinding> MMTK<VM> {
     /// Create an MMTK instance. This is not public. Bindings should use [`MMTKBuilder::build`].
     pub(crate) fn new(options: Arc<Options>) -> Self {
-        // Set the mmap annotation option now.  SFT may map memory.
+        // Set the static variable for mmap annotation.
         crate::util::memory::MMAP_ANNO.store(*options.mmap_anno, Ordering::SeqCst);
 
         // Initialize SFT first in case we need to use this in the constructor.
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -136,7 +136,7 @@ impl<VM: VMBinding> MMTK<VM> {
     /// Create an MMTK instance. This is not public. Bindings should use [`MMTKBuilder::build`].
     pub(crate) fn new(options: Arc<Options>) -> Self {
         // Set the static variable for mmap annotation.
-        crate::util::memory::MMAP_ANNO.store(*options.mmap_anno, Ordering::SeqCst);
+        crate::util::memory::MMAP_ANNOTATION.store(*options.mmap_annotation, Ordering::SeqCst);
 
         // Initialize SFT first in case we need to use this in the constructor.
         // The first call will initialize SFT map. Other calls will be blocked until SFT map is initialized.
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -135,11 +135,6 @@ unsafe impl<VM: VMBinding> Send for MMTK<VM> {}
 impl<VM: VMBinding> MMTK<VM> {
     /// Create an MMTK instance. This is not public. Bindings should use [`MMTKBuilder::build`].
     pub(crate) fn new(options: Arc<Options>) -> Self {
-        // Set the static variable for mmap annotation.
-        // `Ordering::Relaxed`: This happens when MMTk is initialized, therefore happens before all
-        // invocations of `mmap`.
-        crate::util::memory::MMAP_ANNOTATION.store(*options.mmap_annotation, Ordering::Relaxed);
-
         // Initialize SFT first in case we need to use this in the constructor.
         // The first call will initialize SFT map. Other calls will be blocked until SFT map is initialized.
         crate::policy::sft_map::SFTRefStorage::pre_use_check();
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -136,7 +136,9 @@ impl<VM: VMBinding> MMTK<VM> {
     /// Create an MMTK instance. This is not public. Bindings should use [`MMTKBuilder::build`].
     pub(crate) fn new(options: Arc<Options>) -> Self {
         // Set the static variable for mmap annotation.
-        crate::util::memory::MMAP_ANNOTATION.store(*options.mmap_annotation, Ordering::SeqCst);
+        // `Ordering::Relaxed`: This happens when MMTk is initialized, therefore happens before all
+        // invocations of `mmap`.
+        crate::util::memory::MMAP_ANNOTATION.store(*options.mmap_annotation, Ordering::Relaxed);
 
         // Initialize SFT first in case we need to use this in the constructor.
         // The first call will initialize SFT map. Other calls will be blocked until SFT map is initialized.
--- a/src/policy/lockfreeimmortalspace.rs
+++ b/src/policy/lockfreeimmortalspace.rs
@@ -13,7 +13,7 @@ use crate::util::heap::gc_trigger::GCTrigger;
 use crate::util::heap::layout::vm_layout::vm_layout;
 use crate::util::heap::PageResource;
 use crate::util::heap::VMRequest;
-use crate::util::memory::MmapAnno;
+use crate::util::memory::MmapAnnotation;
 use crate::util::memory::MmapStrategy;
 use crate::util::metadata::side_metadata::SideMetadataContext;
 use crate::util::metadata::side_metadata::SideMetadataSanity;
@@ -246,7 +246,7 @@ impl<VM: VMBinding> LockFreeImmortalSpace<VM> {
             start,
             aligned_total_bytes,
             strategy,
-            &MmapAnno::Space {
+            &MmapAnnotation::Space {
                 name: space.get_name(),
             },
         )
--- a/src/policy/lockfreeimmortalspace.rs
+++ b/src/policy/lockfreeimmortalspace.rs
@@ -13,6 +13,7 @@ use crate::util::heap::gc_trigger::GCTrigger;
 use crate::util::heap::layout::vm_layout::vm_layout;
 use crate::util::heap::PageResource;
 use crate::util::heap::VMRequest;
+use crate::util::memory::MmapAnno;
 use crate::util::memory::MmapStrategy;
 use crate::util::metadata::side_metadata::SideMetadataContext;
 use crate::util::metadata::side_metadata::SideMetadataSanity;
@@ -241,10 +242,18 @@ impl<VM: VMBinding> LockFreeImmortalSpace<VM> {
             *args.options.transparent_hugepages,
             crate::util::memory::MmapProtection::ReadWrite,
         );
-        crate::util::memory::dzmmap_noreplace(start, aligned_total_bytes, strategy).unwrap();
+        crate::util::memory::dzmmap_noreplace(
+            start,
+            aligned_total_bytes,
+            strategy,
+            &MmapAnno::Space {
+                name: "LockFreeImmortalSpace",
+            },
+        )
+        .unwrap();
         if space
             .metadata
-            .try_map_metadata_space(start, aligned_total_bytes)
+            .try_map_metadata_space(start, aligned_total_bytes, "mmtk:LockFreeImmortalSpace")
             .is_err()
         {
             // TODO(Javad): handle meta space allocation failure
--- a/src/policy/lockfreeimmortalspace.rs
+++ b/src/policy/lockfreeimmortalspace.rs
@@ -247,18 +247,17 @@ impl<VM: VMBinding> LockFreeImmortalSpace<VM> {
             aligned_total_bytes,
             strategy,
             &MmapAnno::Space {
-                name: "LockFreeImmortalSpace",
+                name: space.get_name(),
             },
         )
         .unwrap();
-        if space
+        space
             .metadata
-            .try_map_metadata_space(start, aligned_total_bytes, "mmtk:LockFreeImmortalSpace")
-            .is_err()
-        {
-            // TODO(Javad): handle meta space allocation failure
-            panic!("failed to mmap meta memory");
-        }
+            .try_map_metadata_space(start, aligned_total_bytes, space.get_name())
+            .unwrap_or_else(|e| {
+                // TODO(Javad): handle meta space allocation failure
+                panic!("failed to mmap meta memory: {e}")
+            });
 
         space
     }
--- a/src/policy/marksweepspace/malloc_ms/metadata.rs
+++ b/src/policy/marksweepspace/malloc_ms/metadata.rs
@@ -100,7 +100,7 @@ fn map_active_chunk_metadata(chunk_start: Address) {
     );
 
     assert!(
-        CHUNK_METADATA.try_map_metadata_space(start, size).is_ok(),
+        CHUNK_METADATA.try_map_metadata_space(start, size, "mmtk:map_active_chunk_metadata:assert").is_ok(),
         "failed to mmap meta memory"
     );
 }
@@ -131,7 +131,7 @@ pub(super) fn map_meta_space(metadata: &SideMetadataContext, addr: Address, size
         // Note that this might fail. For example, we have marked a chunk as active but later we freed all
         // the objects in it, and unset its chunk bit. However, we do not free its metadata. So for the chunk,
         // its chunk bit is mapped, but not marked, and all its local metadata is also mapped.
-        let mmap_metadata_result = metadata.try_map_metadata_space(start, BYTES_IN_CHUNK);
+        let mmap_metadata_result = metadata.try_map_metadata_space(start, BYTES_IN_CHUNK, "mmtk:malloc_ms:map_meta_space");
         debug_assert!(
             mmap_metadata_result.is_ok(),
             "mmap sidemetadata failed for chunk_start ({})",
--- a/src/policy/marksweepspace/malloc_ms/metadata.rs
+++ b/src/policy/marksweepspace/malloc_ms/metadata.rs
@@ -76,7 +76,7 @@ fn is_meta_space_mapped_for_address(address: Address) -> bool {
 }
 
 /// Eagerly map the active chunk metadata surrounding `chunk_start`
-fn map_active_chunk_metadata(chunk_start: Address) {
+fn map_active_chunk_metadata(chunk_start: Address, space_name: &str) {
     debug_assert!(chunk_start.is_aligned_to(BYTES_IN_CHUNK));
     // We eagerly map 16Gb worth of space for the chunk mark bytes on 64-bits
     // We require saturating subtractions in order to not overflow the chunk_start by
@@ -99,16 +99,20 @@ fn map_active_chunk_metadata(chunk_start: Address) {
         chunk_start + (size / 2)
     );
 
-    assert!(
-        CHUNK_METADATA.try_map_metadata_space(start, size, "mmtk:map_active_chunk_metadata:assert").is_ok(),
-        "failed to mmap meta memory"
-    );
+    CHUNK_METADATA
+        .try_map_metadata_space(start, size, space_name)
+        .unwrap_or_else(|e| panic!("failed to mmap meta memory: {e}"));
 }
 
 /// We map the active chunk metadata (if not previously mapped), as well as the VO bit metadata
 /// and active page metadata here. Note that if [addr, addr + size) crosses multiple chunks, we
 /// will map for each chunk.
-pub(super) fn map_meta_space(metadata: &SideMetadataContext, addr: Address, size: usize) {
+pub(super) fn map_meta_space(
+    metadata: &SideMetadataContext,
+    addr: Address,
+    size: usize,
+    space_name: &str,
+) {
     // In order to prevent race conditions, we synchronize on the lock first and then
     // check if we need to map the active chunk metadata for `chunk_start`
     let _lock = CHUNK_MAP_LOCK.lock().unwrap();
@@ -118,7 +122,7 @@ pub(super) fn map_meta_space(metadata: &SideMetadataContext, addr: Address, size
         // Check if the chunk bit metadata is mapped. If it is not mapped, map it.
         // Note that the chunk bit metadata is global. It may have been mapped because other policy mapped it.
         if !is_chunk_mapped(start) {
-            map_active_chunk_metadata(start);
+            map_active_chunk_metadata(start, space_name);
         }
 
         // If we have set the chunk bit, return. This is needed just in case another thread has done this before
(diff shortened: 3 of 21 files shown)


k-sareen:
Why not just us the space/metadata spec name instead? Also there's a 80-byte limit for the string (including the NUL byte).

> However, for Map32, spaces may interleave at chunk granularity in the shared discontiguous memory range.

Why is this a problem? If a chunk is recycled, then we should just rename it.

wks:
> Why not just use the space/metadata spec name instead?

- For space, I just use its name.
- For side metadata, I use both space name and metadata spec name to distinguish between metadata for different spaces.
- And there are `mmap` invocations for other purposes, too, such as SFTMap.

> Also there's a 80-byte limit for the string (including the NUL byte).

Good point.  Although I don't think any space name + metadata name would exceed this limit, it's safe to perform the check when formatting.

> Why is this a problem? If a chunk is recycled, then we should just rename it.

The number of mmap entries is limited at the OS level.  So whenever possible, we should merge adjacent entries.  If no chunk in the discontiguous space has name (annotated with `PR_SET_VMA_ANON_NAME`), the OS may merge adjacent entries; but if adjacent entries have different names, they can't be merged.

k-sareen:
> And there are mmap invocations for other purposes, too, such as SFTMap.

Yes that is a good point. Though in my branch I just use an `Option<&str>` because it's not always clear what name to give to an `mmap` region such as the side metadata implementation for 32-bits and for `panic_if_unmapped`.

> The number of mmap entries is limited at the OS level. So whenever possible, we should merge adjacent entries. If no chunk in the discontiguous space has name (annotated with PR_SET_VMA_ANON_NAME), the OS may merge adjacent entries; but if adjacent entries have different names, they can't be merged.

I'm not sure this is a real concern? I don't imagine we'll ever be hitting the limit even for 32-bit machines. Even Android always has the annotations enabled.

wks:
> > The number of mmap entries is limited at the OS level. So whenever possible, we should merge adjacent entries. If no chunk in the discontiguous space has name (annotated with PR_SET_VMA_ANON_NAME), the OS may merge adjacent entries; but if adjacent entries have different names, they can't be merged.
> 
> I'm not sure this is a real concern? I don't imagine we'll ever be hitting the limit even for 32-bit machines. Even Android always has the annotations enabled. I strongly suggest having these on by default. If we ever hit the limit then we can disable it.

Well, perhaps it's not a problem.  I tried a few benchmarks.  We do mmap at chunk granularity.  This means we don't call `mmap` very often.  It is proportional to the heap size, but even memory-hungry benchmarks like h2 run just fine, with only a few hundred invocations of `mmap` in the entire execution.  And there are less than 20 entries in `/proc/pid/mmap`  The following is captured while running the h2 benchmark on OpenJDK with 1GB heap size.  They are the only mmaps created by mmtk-core (besides those created by `malloc` and `pthread_create`).

```
40000000-40400000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:nursery]
40400000-40800000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:los]
40800000-41800000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:nursery]
41800000-41c00000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:los]
41c00000-5f400000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:nursery]
5f400000-6f400000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:immix_mature]
6f400000-8f000000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:nursery]
8f000000-90800000 rw-p 00000000 00:00 0                                  [anon:mmtk:space:immix_mature]
100000000-1001e0000 rw-p 00000000 00:00 0 
1001e0000-140000000 ---p 00000000 00:00 0 
e0005000000-e0005800000 rw-p 00000000 00:00 0                            [anon:mmtk:sidemeta:nursery:VMGlobalLogBitSpec]
e0005800000-e0005c00000 rw-p 00000000 00:00 0                            [anon:mmtk:sidemeta:immix_mature:VMGlobalLogBitSpec]
e0005c00000-e0006400000 rw-p 00000000 00:00 0                            [anon:mmtk:sidemeta:nursery:VMGlobalLogBitSpec]
e0006400000-e0006800000 rw-p 00000000 00:00 0                            [anon:mmtk:sidemeta:immix_mature:VMGlobalLogBitSpec]
4e0800400000-4e0800c00000 rw-p 00000000 00:00 0                          [anon:mmtk:sidemeta:immix_mature:IX_LINE_MARK]
4e8800000000-4e8800400000 rw-p 00000000 00:00 0                          [anon:mmtk:sidemeta:immix_mature:IX_BLOCK_DEFRAG]
4e8900000000-4e8900400000 rw-p 00000000 00:00 0                          [anon:mmtk:sidemeta:immix_mature:IX_BLOCK_MARK]
4e8a00000000-4e8a00400000 rw-p 00000000 00:00 0                          [anon:mmtk:sidemeta:immix_mature:CHUNK_MARK]
4eaa82000000-4eaa82400000 rw-p 00000000 00:00 0                          [anon:mmtk:sidemeta:los:VMLocalLOSMarkNurserySpec]
4eac83000000-4eac84800000 rw-p 00000000 00:00 0                          [anon:mmtk:sidemeta:immix_mature:VMLocalMarkBitSpec]
```

So I think it's OK to leave it enabled by default.  The user can still set `MMTK_MMAP_ANNOTATION=false` to disable it if it ever becomes problematic.

k-sareen:
I think you need to update some of the tests to use the new annotation API as well.

Re: mmap chunks. Like I mentioned in a meeting, we can improve this further by allocating spaces from different ends of the address space

k-sareen:
Err.. Why is `sysinfo` broken 🤔

wks:
> Err.. Why is `sysinfo` broken 🤔

There is an upstream bug report: https://github.com/GuillaumeGomez/sysinfo/issues/1392

k-sareen:
Should we pin version of sysinfo/libc until this is resolved then?

src/mmtk.rs
```diff
@@ -135,6 +135,9 @@ unsafe impl<VM: VMBinding> Send for MMTK<VM> {}
 impl<VM: VMBinding> MMTK<VM> {
     /// Create an MMTK instance. This is not public. Bindings should use [`MMTKBuilder::build`].
     pub(crate) fn new(options: Arc<Options>) -> Self {
+        // Set the static variable for mmap annotation.
+        crate::util::memory::MMAP_ANNOTATION.store(*options.mmap_annotation, Ordering::SeqCst);
```
k-sareen, line 139: You don't need to be `SeqCst` here since there are no other threads right now (theoretically).

src/policy/space.rs
```diff
@@ -144,12 +144,19 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
...
+                                res.pages,
+                                self.common().mmap_strategy(),
+                                &memory::MmapAnnotation::Space {
+                                    name: self.get_name(),
+                                },
                             )
+                            .and(self.common().metadata.try_map_metadata_space(
+                                res.start,
+                                bytes,
+                                self.name(),
```
k-sareen, line 158: Just to be consistent with everything else:
```suggestion
                                self.get_name(),
```

src/policy/space.rs
```diff
@@ -293,15 +300,13 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
...
-            .common()
+        self.common()
             .metadata
-            .try_map_metadata_space(self.common().start, self.common().extent)
-            .is_err()
-        {
-            // TODO(Javad): handle meta space allocation failure
-            panic!("failed to mmap meta memory");
-        }
+            .try_map_metadata_space(self.common().start, self.common().extent, self.name())
```
k-sareen, line 305: Same as above
```suggestion
            .try_map_metadata_space(self.common().start, self.common().extent, self.get_name())
```

src/util/heap/layout/mmapper.rs
```diff
@@ -49,7 +50,13 @@ pub trait Mmapper: Sync {
     // NOTE: There is a monotonicity assumption so that only updates require lock
```
k-sareen, line 50: Need to update the doc comment to mention the new argument

src/util/heap/layout/mmapper.rs
```diff
@@ -37,6 +37,7 @@ pub trait Mmapper: Sync {
         start: Address,
         pages: usize,
         strategy: MmapStrategy,
+        anno: &MmapAnnotation,
```
k-sareen, line 40: Need to update the doc comment to mention the new argument

src/util/heap/layout/mmapper.rs
```diff
@@ -157,6 +168,7 @@ impl MapState {
         state_slices: &[&[Atomic<MapState>]],
         mmap_start: Address,
         strategy: MmapStrategy,
+        anno: &MmapAnnotation,
```
k-sareen, line 171: Need to update the doc comment to mention the new argument

src/util/memory.rs
```diff
@@ -144,24 +233,58 @@ pub fn dzmmap_noreplace(start: Address, size: usize, strategy: MmapStrategy) ->
...
+    _anno: &MmapAnnotation,
 ) -> Result<()> {
     let ptr = start.to_mut_ptr();
     let prot = strategy.prot.into_native_flags();
     wrap_libc_call(
         &|| unsafe { libc::mmap(start.to_mut_ptr(), size, prot, flags, -1, 0) },
         ptr,
     )?;
+
+    #[cfg(target_os = "linux")]
```
k-sareen, line 261:
```suggestion
    #[cfg(any(target_os = "linux", target_os = "android"))]
```

src/util/metadata/side_metadata/global.rs
```diff
@@ -1409,9 +1425,19 @@ impl SideMetadataContext {
     /// * `start` - The starting address of the source data.
     /// * `size` - The size of the source data (in bytes).
     /// * `no_reserve` - whether to invoke mmap with a noreserve flag (we use this flag to quarantine address range)
```
k-sareen, line 1427: Need to update the doc comment to mention the new argument

k-sareen (commented):
LGTM except some minor stuff in review.

src/util/memory.rs
```diff
@@ -15,6 +16,14 @@ const MMAP_FLAGS: libc::c_int = libc::MAP_ANON | libc::MAP_PRIVATE | libc::MAP_F
 // MAP_FIXED is used instead of MAP_FIXED_NOREPLACE (which is not available on macOS). We are at the risk of overwriting pre-existing mappings.
 const MMAP_FLAGS: libc::c_int = libc::MAP_ANON | libc::MAP_PRIVATE | libc::MAP_FIXED;
 
+/// This static variable controls whether we annotate mmapped memory region using `prctl`. It can be
+/// set via `Options::mmap_annotation` or the `MMTK_MMAP_ANNOTATION` environment variable.
+///
+/// FIXME: Since it is set via `Options`, it is in theory a decision per MMTk instance. However, we
+/// currently don't have a good design for multiple MMTk instances, so we use static variable for
+/// now.
+pub(crate) static MMAP_ANNOTATION: AtomicBool = AtomicBool::new(true);
```
qinsoon, line 25: This looks bad. Though there is a `FIXME` above, we should avoid introducing this at first place.

Both `Plan` and `Space` have references to the options. You can let plan/space to check options and create a `MmapAnnotation`. If the runtime option is not enabled, we can use `Option<MmapAnnotation>` or `MmapAnnotation::Omit/None`.

src/util/memory.rs
```diff
@@ -84,6 +93,76 @@ pub enum HugePageSupport {
...
+/// memory range is shared between different spaces. Spaces may use `mmap` to map new chunks, but
+/// the same chunk may later be reused by other spaces. The annotation only applies when `mmap` is
+/// called for a chunk for the first time, which reflects which space first attempted the mmap, not
+/// which space is currently using the chunk.  Use `crate::policy::space::print_vm_map` to print a
+/// more accurate mapping between address ranges and spaces.
+///
+/// On 32-bit architecture, side metadata are allocated in a chunked fasion.  One single `mmap`
+/// region will contain many different metadata.  In that case, we simply annotate the whole region
+/// with a `MmapAnnotation::SideMeta` where `meta` is `"all"`.
+pub enum MmapAnnotation<'a> {
```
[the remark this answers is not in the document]
wks: Probably not adding `MmapAnnotation` as part of `MmapStrategy`.  Currently we define a few `MmapStrategy` constant instances, such as `INTERNAL_MEMORY`, `SIDE_METADATA` and `TEST`, and use `INTERNAL_MEMORY` for all side metadata.  But when annotating mmap, we need a different annotation for each concrete space or side metadata.

I considered making a super struct `MmapOption(MmapStrategy, MmapAnnotation)`, That can simplify `CommonSpace::mmap_strategy` a bit by letting it return a `MmapOption` directly, with the `MmapAnnotation` being `MmapAnnotation::Space`.  But that's only for spaces.  When doing mmap for side metadata, we will still need different annotations for different metadata.  And having a `MmapOption` returned from `CommonSpace` seems to encourage the developers to reuse that instance for all mmap use case.  That's just the opposite of my intention.  I intend to let every call site of mmap be annotated specifically for its purpose.

src/util/memory.rs
wks: Yes.  That looks bad.

Although it is possible to let plans and spaces check options, and optionally pass `MmapAnnotation::Omit` or a `None` for `Option<MmapAnnotation>`, it is not ideal.

One important part of my intention is that *the annotation cannot be omitted*.  I'd like the function signatures to enforce that *an annotation must be provided for every call to mmap*.  If we give the caller the option to pass `None` or `Omit`, developers will do it if they think it is not worth annotating.  Instead, the decision of whether to actually annotate the mmap should be postponed until `mmap` is actually called and is successful, and should be controlled by the user via an option rather than at call sites.

And checking at every call site is repetitive and error prone.  Developers may forget to do the check or erroneously omit the annotation.

Ideally, functions related to memory mapping should be wrapped into a struct, maybe named `MmapSupport` or `CommonMapper`.  It will be created per `MMTK` instance, like `Mmapper` and `VMMap` (which are currently both global singletons), and referenced by plans, spaces, SFT, etc., whatever needs to use mmap.  The `mmap_annotation: bool` should be an instance variable of `CommonMapper`, and `map_fixed` should be an instance method of `MmapSupport`.  That may involve much refactoring, including moving `dzmmap`, `dzmmap_noreplace` and `mmap_noreserve` into `MmapSupport`.  For now, creating a global boolean variable seems to be the easiest solution.

In the short term, we may introduce `MmapSupport` as a global singleton, like `Mmapper` and `VMMap`.  In the long term, we can make them specific to the `MMTK` instance.

qinsoon: > One important part of my intention is that the annotation cannot be omitted. I'd like the function signatures to enforce that an annotation must be provided for every call to mmap. 

We could enforce the users to supply `MmapAnnotation`, and turn it into `Option<MmapAnnotation>` before we send it to mmap functions. Plus the problem of having a global var outweighs the benefits of 'enforcing annotations'.

Alternatively, we can have an instance for `memory` which holds the boolean value from the option. I think they are both worth trying before adopting the current implementation. We have arg structs for creating plans and spaces. Creating a `Memory` instance and letting plan/space to hold it shouldn't be too hard. https://github.com/mmtk/mmtk-core/blob/8640ab89f43f33af99e61b9ec8ff2164ed2cfab7/src/plan/global.rs#L372

> It will be created per MMTK instance, like Mmapper and VMMap (which are currently both global singletons), and referenced by plans, spaces, SFT, etc., whatever needs to use mmap. 

I think it is still arguable whether `Mmapper` and `VMMap` should be global or per instance. We do have a hacky `vm_layout()` that should be per instance, and is global at the moment. But we should avoid introducing more globals like this. The more we have, the more difficult and the more resistance we will face when we try to support multiple instances.

wks: > Alternatively, we can have an instance for `memory` which holds the boolean value from the option. I think they are both worth trying before adopting the current implementation. We have arg structs for creating plans and spaces. Creating a `Memory` instance and letting plan/space to hold it shouldn't be too hard.
> 
> https://github.com/mmtk/mmtk-core/blob/8640ab89f43f33af99e61b9ec8ff2164ed2cfab7/src/plan/global.rs#L372

Yes.  This is my intended way to do it.  But this may involve too many changes and I think it is better to do the refactoring separately.

The introduced instance (let's call it `MmapSupport` for now) should contain `dzmap`, `dzmap_noreplace`, `mmap_noreserve` and `mmap_fixed`.  I grepped the code and found that the following places will be affected.

-   `MapState`: `MapState::transition_to_*` will call those functions directly, as if the `MapState` can transition its own state without the help of `Mmapper` or the `MmapSupport` instance.  This needs to be refactored so that `MapState` will become a state holder, and `Mmapper` implementations (`FragmentedMapper` and `ByteMapMmapper`) can call the mmap functions in `MmapSupport` to do the transition.
-   `LockFreeImmortalSpace`: It skipped the `Mmapper` and directly calls `dzmmap_noreplace`.  I think this is a bug because `Mmapper` should be aware of the mapping of spaces.
-   `RawMemoryFreeList`: It calls `dzmmap_noreplace` to map its memory.  Not sure if it should go through `Mmapper`, but probably it should because a `RawMemoryFreeList` is part of a space using it.

In the end, we may have all mmap invocations going through `Mmapper`, and an `Mmapper` holds a reference to `MmapSupport`.

> We could enforce the users to supply `MmapAnnotation`, and turn it into `Option<MmapAnnotation>` before we send it to mmap functions.

It is good to enforce the users to supply `MmapAnnotation`.  The difficulty is drawing a border between the "user" and the "mmap functions".  And another difficulty is that we lost access to `Options` before reaching "mmap functions".

Take the call chain `SideMetadataContext::map_metadata_internal` -> `try_mmap_contiguous_metadata_space` -> `MMAPPER.ensure_mapped` for example.  

-   In `SideMetadataContext::map_metadata_internal`, we have enough information (space name and side metadata name) to construct the annotation.  We construct `MmapAnnotation` here.  But since `SideMetadataContext` does not reference any`Space` or `Plan`, it cannot access `Options`.
-   `try_mmap_contiguous_metadata_space` is a top-level function.  It has access to neither space name nor `Option`.
-   `MMAPER.ensure_mapped` is a function that is more related to mmap.  It may be refactored to hold a boolean variable `MMAP_ANNOTATION`, or refer to `MmapSupport` I mentioned above.

So the availability of `Options` or the `MMAP_ANNOTATION` variable does not match the distinction between user and mmap.  I think it's easier to refactor the code and just let "mmap functions" decide whether to annotate.

> Plus the problem of having a global var outweighs the benefits of 'enforcing annotations'.

Having global var is bad, but I think it is worse to let the users check `Options::mmap_annotation` before constructing the `Option<MmapAnnotation>` because it is repetitive and error-prone.  Having a global variable is the easiest workaround before we refactor `Mmapper`, `MapState`, etc.

If you think having a global variable is unacceptible, I can remove `Options::mmap_annotation` and make the annotation controlled by a Cargo feature instead, or leave it always on.  I think it is OK to leave it always on because there will be strictly fewer `prctl` than `mmap`, and we don't call `mmap` very often, either.

qinsoon: > If you think having a global variable is unacceptible, I can remove Options::mmap_annotation and make the annotation controlled by a Cargo feature instead, or leave it always on.

Right. We dont really need the option anyway. We can always annotate the mmaps.

The reason that I am against having a global var is that we know it's wrong and this kind of small issues will accumulate over time and eventually make it hard to get rid of them. We should try to avoid introducing them in first place.

src/mmtk.rs
wks: In the end, I removed the option `options.mmap_annotation`.

src/policy/space.rs
wks: Done.

src/policy/space.rs
wks: Done.

src/util/heap/layout/mmapper.rs
wks: Done.

src/util/heap/layout/mmapper.rs
wks: Done.

src/util/heap/layout/mmapper.rs
wks: Done.

src/util/metadata/side_metadata/global.rs
wks: Done.

src/util/memory.rs
wks: I removed the option, and added a Cargo feature `no_mmap_annotation` just in case anyone needs to disable it for any reason.

src/util/memory.rs
k-sareen: Tell me about it. My fork of MMTk core is riddled with this stuff so that it works with Android as well. I honestly don't get it either.

src/policy/lockfreeimmortalspace.rs
```diff
@@ -241,15 +242,22 @@ impl<VM: VMBinding> LockFreeImmortalSpace<VM> {
...
+        space
             .metadata
-            .try_map_metadata_space(start, aligned_total_bytes)
-            .is_err()
-        {
-            // TODO(Javad): handle meta space allocation failure
-            panic!("failed to mmap meta memory");
-        }
+            .try_map_metadata_space(start, aligned_total_bytes, space.get_name())
+            .unwrap_or_else(|e| {
```
qinsoon, line 257: I don't see improvement from this refactoring. This refactoring seems based on a personal coding style, rather than improvement for the code, which is not a reason to refactor existing code base. We have style checks and lint checks in CI, and any code that passed the checks is considered acceptable. We should avoid refactoring existing code purely based on personal coding style. Alternatively, a better approach is to introduce new lints (and justify why we need the new lints) so we adopt new styles as our 'standard' styles.

Besides, here we just need to know if the call returns an error, and we do not really need the value of the result. Using `is_err` is okay, or even better than `unwrap` to reflect what we are doing. 

I don't request the refactoring to be reverted (it doesn't worth the extra efforts), but refactoring purely based on personal coding style should be discouraged.

src/policy/marksweepspace/malloc_ms/metadata.rs
```diff
@@ -99,16 +99,20 @@ fn map_active_chunk_metadata(chunk_start: Address) {
         chunk_start + (size / 2)
     );
 
-    assert!(
-        CHUNK_METADATA.try_map_metadata_space(start, size).is_ok(),
-        "failed to mmap meta memory"
-    );
+    CHUNK_METADATA
+        .try_map_metadata_space(start, size, space_name)
```
qinsoon, line 103: This is good refactoring with clear improvement. The old code put the actual action performed in the `assert!` which is bad.

src/util/memory.rs
qinsoon: This global var can be removed now. I am suprised that we did not get any warning from the compiler for this, this seems to be an unused variable.

qinsoon (commented):
Generally looks good. The global var `MMAP_ANNOTATION` can be removed.

src/policy/lockfreeimmortalspace.rs
wks: The intention is showing the error in the panic message.  There are several things that can go wrong in `mmap_fixed`, namely `mmap`, `madvise` and the newly introduced `prctl`.  Here we don't really need the value of the result for the correct functioning of the program (because we just panic).  But if we omit the error, it will simply print "failed to mmap meta memory" in the panic message, and the stack trace will not include the exact place that returned the error.  That will leave the user or developer wondering what went wrong.  Showing the error message at least will provide the user some hints about the error.

I considered letting `prctl` return `Err`, too, because it does fail in some cases, for example, the name string is too long.  But since it cannot be distinguished from the case when it is executed on a kernel older than 5.17, I chose to log the error rather than returning `Err` or panicking.

qinsoon (approved):
Looks good

wks:
V8 failed for HTTP 404.  I think it might be a transient failure if the Ubuntu server is updating itself and happens to be in an inconsistent state when we ran the test.  See: https://github.com/mmtk/mmtk-core/actions/runs/12078239098/job/33686682978?pr=1236

JikesRVM failed for the classic "instruction ... not in RVM space" error.  Interestingly, the dumped maps file showed that our annotation is working.  See: https://github.com/mmtk/mmtk-core/actions/runs/12078239098/job/33686683158?pr=1236

I'll rerun those tests.

PR #1238, opened 2024-11-25
Collect live bytes per space, and report by space

qinsoon:
The current `count_live_bytes_in_gc` feature adds the size of all live objects and compare with the used pages reported by the plan. There are two issues with the feature: 
1. VM space is not included in the used pages reported by the plan, but the live objects include objects in the VM space. So the reported fragmentation/utilization is wrong when the VM space is in use.
2. Spaces/policies have very different fragmentation ratio. Reporting the fragmentation for the entire heap is not useful.

This PR refactors the current `count_live_bytes_in_gc` feature so we collect live bytes per space, and report by space.

### DIFF
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -12,7 +12,6 @@ use crate::util::address::ObjectReference;
 use crate::util::analysis::AnalysisManager;
 use crate::util::finalizable_processor::FinalizableProcessor;
 use crate::util::heap::gc_trigger::GCTrigger;
-#[cfg(feature = "count_live_bytes_in_gc")]
 use crate::util::heap::layout::heap_parameters::MAX_SPACES;
 use crate::util::heap::layout::vm_layout::VMLayout;
 use crate::util::heap::layout::{self, Mmapper, VMMap};
@@ -28,7 +27,6 @@ use crate::util::statistics::stats::Stats;
 use crate::vm::ReferenceGlue;
 use crate::vm::VMBinding;
 use std::cell::UnsafeCell;
-#[cfg(feature = "count_live_bytes_in_gc")]
 use std::collections::HashMap;
 use std::default::Default;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -531,7 +529,6 @@ impl<VM: VMBinding> MMTK<VM> {
         })
     }
 
-    #[cfg(feature = "count_live_bytes_in_gc")]
     /// Aggregate a hash map of live bytes per space with the space stats to produce
     /// a map of live bytes stats for the spaces.
     pub(crate) fn aggregate_live_bytes_in_last_gc(
--- a/src/mmtk.rs
+++ b/src/mmtk.rs
@@ -26,6 +26,8 @@ use crate::util::statistics::stats::Stats;
 use crate::vm::ReferenceGlue;
 use crate::vm::VMBinding;
 use std::cell::UnsafeCell;
+#[cfg(feature = "count_live_bytes_in_gc")]
+use std::collections::HashMap;
 use std::default::Default;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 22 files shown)


wks:
Using space name as the key may be inefficient.  Can we use `VMMap::get_descriptor_for_address` to get the `SpaceDescriptor`, and then use `SpaceDescriptor::get_index()` to get an integer ordinal for the space, and use it to index a `Vec<LiveBytesStats>` (or `[LiveBytesStats; MAX_SPACES]`) instead?

qinsoon:
> Using space name as the key may be inefficient. Can we use `SpaceDescriptor::get_index()` to get an integer ordinal for the space, and use it to index a `Vec<LiveBytesStats>` (or `[LiveBytesStats; MAX_SPACES]`) instead?

Sounds good. I have made the change.

We might be able to use a runtime option for collecting live bytes data, and might be able to leave it enabled by default. I will run benchmarks first to know the overhead. Probably let's postpone the review for this PR until I get the benchmark results.

qinsoon:
The following graph shows 3 runs: 1. `count_live_bytes_in_gc` as a feature and the feature is disabled, 2. `count_live_bytes_in_gc` as a runtime option and it is turned off, and 3. `count_live_bytes_in_gc` as a runtime option and it is turned on. The overhead from runtime option is little (0.7% GC time when it is turned off, and 1% GC time when it is turned on). So I made it as a runtime option.

livebytes

[plotty](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|mole-2024-11-29-Fri-035235&benchmark^build^invocation^iteration&time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark&build;jdk-mmtk-live-bytes-disabled|41&Histogram%20(with%20CI)^build^benchmark&)

src/scheduler/worker.rs
```diff
@@ -54,21 +54,32 @@ impl<VM: VMBinding> GCWorkerShared<VM> {
...
-    pub(crate) fn get_and_clear_live_bytes(&self) -> usize {
-        self.live_bytes.swap(0, Ordering::SeqCst)
+    pub(crate) fn increase_live_bytes(
+        live_bytes_per_space: &mut [usize; MAX_SPACES],
+        object: ObjectReference,
+    ) {
+        use crate::mmtk::VM_MAP;
+        use crate::vm::object_model::ObjectModel;
+
+        // The live bytes ofr the object
```
wks, line 70:
```suggestion
        // The live bytes of the object
```

src/scheduler/worker.rs
```diff
@@ -54,21 +54,32 @@ impl<VM: VMBinding> GCWorkerShared<VM> {
...
+        let space_descriptor = VM_MAP.get_descriptor_for_address(object.to_raw_address());
+        let space_index = space_descriptor.get_index();
+        debug_assert!(
+            space_index < MAX_SPACES,
+            "Space index {} is not in the range of [0, {})",
+            space_index,
+            MAX_SPACES
+        );
+        // Accumulate the live bytes for the index
+        live_bytes_per_space[space_descriptor.get_index()] += bytes;
```
wks, line 82:
```suggestion
        live_bytes_per_space[space_index] += bytes;
```

src/scheduler/scheduler.rs
```diff
@@ -553,22 +553,16 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
-            info!(
-                "Live objects = {} bytes ({:04.1}% of {} used pages)",
-                live_bytes,
-                live_bytes as f64 * 100.0 / used_bytes as f64,
-                mmtk.get_plan().get_used_pages()
-            );
+        if *mmtk.get_options().count_live_bytes_in_gc {
+            for (space_name, &stats) in mmtk.state.live_bytes_in_last_gc.borrow().iter() {
+                info!(
+                    "{} = {} bytes ({:.1}% of {} used pages)",
```
wks, line 559: In the previous `info!` log, we print "End of GC" and the number of pages.  But here we print the number of bytes.  To be consistent, and for the convenience of the users, I suggest we print both number of pages and the number of bytes in both log statements.  The result will look like the following:

```
[INFO] End of GC (1000/1024 pages, 4000/4096 KiB, took 100 ms)
[INFO]   nursery = 0/400 pages, 0/1600 KiB (0% of 1000 used pages)
[INFO]   immixspace = 500/600 pages, 2000/2400 KiB (50% of 1000 used pages)
[INFO]   ...
```

And it should be OK to use KiB (1 KiB = 1024 B) instead of bytes, or KB for short, although in SI units 1 KB = 1000 B.  Since we count at page granularity, the result will always be a integer multiple of 4096 bytes.

src/mmtk.rs
```diff
@@ -526,4 +528,34 @@ impl<VM: VMBinding> MMTK<VM> {
...
+        live_bytes_per_space: [usize; MAX_SPACES],
+    ) -> HashMap<&'static str, crate::LiveBytesStats> {
+        use crate::policy::space::Space;
+        let mut ret = HashMap::new();
+        self.get_plan().for_each_space(&mut |space: &dyn Space<VM>| {
+            let space_name = space.get_name();
+            let space_idx = space.get_descriptor().get_index();
+            let used_pages = space.reserved_pages();
+            if used_pages != 0 {
+                let used_bytes = used_pages << crate::util::constants::LOG_BYTES_IN_PAGE;
```
wks, line 545: We have conversion functions.

```suggestion
                let used_bytes = conversions::pages_to_bytes(used_pages);
```

src/scheduler/scheduler.rs
qinsoon: I changed the log a bit, but I did not log bytes for everything -- I think it would be too verbose, and it is redundant info (pages and bytes can be computed from each other anyway).

wks: > I changed the log a bit, but I did not log bytes for everything -- I think it would be too verbose, and it is redundant info (pages and bytes can be computed from each other anyway).

I think you missed the point.  The log information is for *humans* to read when the program is executing.  Although pages and bytes can be converted to each other, humans are not that good at multiplying a number by 4 (to get KiB) or by 4096 (to get bytes).  The log should let the user know the amount of used/total bytes and used/total pages in a glance.  That's what developers usually do when debugging.  They look at the program output and want to know the number immediately without having to open a Python shell and type something like (`500 * 4096`) to compute the number of bytes, or `1050456 / 4096` to get the number of pages.

qinsoon: The percentage is what matters. Human only need to read the percentage. Other info just shows how the percentage is calculated. Presenting live bytes as pages, or presenting used pages as bytes can be misleading.

We can do this:
```
[INFO] End of GC (1000/1024 pages, took 100 ms)
[INFO]   nursery = 400 pages (0% live)
[INFO]   immixspace = 600 pages (50% live)
```

wks (approved):
LGTM

PR #1240, opened 2024-11-28
Remove env_var and command_line from options

wks:
Removed `from_env_var` and `from_command_line` from `MMTkOption` because now all options can be set in both ways.

Replaced `Options::set_from_command_line` and `Options::set_from_env_var` with a single `Options::set_from_string`.

Moved several methods of `Options` out of the `options!` macro if they don't need to be generated via macro expansion.  Now only two methods are generated by the macro.

-   `set_from_string_inner`: Requires matching field names against user-provided strings.
-   `new`: Initializing fields using default values.

In order to replicate the behavior that `Options::read_env_var_settings` silently ignores unrecognized option keys in `MMTK_*` environment variables, we introduced an error type `SetOptionByStringError` which `Options::set_from_string_inner` returns so that `Options::read_env_var_settings` and `Options::set_bulk_from_string` can behave differently when encountering invalid keys.


qinsoon (approved):
LGTM

wks:
The extended tests passed except for JikesRVM which failed for the well-known "xxxx is not in RVM space" error.  I just added a change to the API migration log, and this should be ready to merge now.

PR #1241, opened 2024-11-29
Add misc debugging helpers and feature to poison dead memory

k-sareen:
We add some misc debugging helpers to help binding implementers in debugging tricky memory corruption bugs. We now can also poison dead memory in each space's `release` method. This feature supersedes "immix_zero_on_release".

### DIFF
--- a/src/policy/copyspace.rs
+++ b/src/policy/copyspace.rs
@@ -188,6 +188,11 @@ impl<VM: VMBinding> CopySpace<VM> {
     }
 
     pub fn release(&self) {
+        #[cfg(feature = "poison_on_release")]
+        for (start, size) in self.pr.iterate_allocated_regions() {
+            crate::util::memory::set(start, 0xbc, size);
+        }
+
         for (start, size) in self.pr.iterate_allocated_regions() {
             // Clear the forwarding bits if it is on the side.
             if let MetadataSpec::OnSide(side_forwarding_status_table) =
--- a/src/policy/immix/block.rs
+++ b/src/policy/immix/block.rs
@@ -248,8 +248,8 @@ impl Block {
                         holes += 1;
                     }
 
-                    #[cfg(feature = "immix_zero_on_release")]
-                    crate::util::memory::zero(line.start(), Line::BYTES);
+                    #[cfg(feature = "poison_on_release")]
+                    crate::util::memory::set_pattern(line.start(), (0xdeadbeef ^ line.start().as_usize()) & !0xff, Line::BYTES);
 
                     // We need to clear the pin bit if it is on the side, as this line can be reused
                     #[cfg(feature = "object_pinning")]
--- a/src/util/memory.rs
+++ b/src/util/memory.rs
@@ -105,6 +105,46 @@ pub fn set(start: Address, val: u8, len: usize) {
     }
 }
 
+/// Set a range of memory to the given pattern. Similar to C++'s std::fill.
+pub fn set_pattern(start: Address, pattern: usize, len: usize) {
+    debug_assert!(std::mem::size_of::<usize>() <= len);
+    debug_assert!(len % std::mem::size_of::<usize>() == 0);
+
+    let end_addr = (start + len).to_mut_ptr::<usize>();
+    let mut current = start.to_mut_ptr::<usize>();
+    while current < end_addr {
+        unsafe {
+            *current = pattern;
+            current = current.add(1);
+        }
+    }
+}
+
+/// Dump RAM around a given address. Note that be careful when using this function as it may
+/// segfault for unmapped memory. ONLY use it for locations that are KNOWN to be broken AND
+/// allocated by MMTk.
+pub fn dump_ram_around_address(addr: Address, bytes: usize) -> String {
+    let mut string: String = String::new();
+    let end_addr = (addr + bytes).to_ptr::<usize>();
+    let mut current = (addr - bytes).to_ptr::<usize>();
+    while current < end_addr {
+        unsafe {
+            if current == addr.to_ptr::<usize>() {
+                string.push_str(" | ");
+            } else {
+                string.push_str(" ");
+            }
+            let s = unsafe { current.read() };
+            #[cfg(target_pointer_width = "64")]
+            string.push_str(format!("{:#018x}", s).as_str());
+            #[cfg(target_pointer_width = "32")]
+            string.push_str(format!("{:#010x}", s).as_str());
+            current = current.add(1);
+        }
+    }
+    string
+}
+
 /// Demand-zero mmap:
 /// This function mmaps the memory and guarantees to zero all mapped memory.
 /// This function WILL overwrite existing memory mapping. The user of this function
--- a/src/policy/largeobjectspace.rs
+++ b/src/policy/largeobjectspace.rs
@@ -288,8 +288,15 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
         let sweep = |object: ObjectReference| {
             #[cfg(feature = "vo_bit")]
             crate::util::metadata::vo_bit::unset_vo_bit(object);
+            if self.common.needs_log_bit {
+                VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC
+                    .clear::<VM>(object, Ordering::SeqCst);
+            }
+            let start = object.to_object_start::<VM>();
+            #[cfg(feature = "poison_on_release")]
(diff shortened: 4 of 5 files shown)


k-sareen:
The failing style checks are unrelated to the change. Seems like new Rust stable is a bit smarter.

src/policy/copyspace.rs
```diff
@@ -188,6 +188,11 @@ impl<VM: VMBinding> CopySpace<VM> {
     }
 
     pub fn release(&self) {
+        #[cfg(feature = "poison_on_release")]
+        for (start, size) in self.pr.iterate_allocated_regions() {
+            crate::util::memory::set(start, 0xbc, size);
```
qinsoon, line 193: Instead of using an arbitrary value, you can define a mapping between space and its poison value, or use something from the space (such as the SpaceDescriptor index, or the space name).

I would suggest creating a module `crate::util::poison_on_release`. Define the poison values there, and define functions to poison a memory region.

src/policy/immix/block.rs
```diff
@@ -248,8 +248,13 @@ impl Block {
                         holes += 1;
                     }
 
-                    #[cfg(feature = "immix_zero_on_release")]
-                    crate::util::memory::zero(line.start(), Line::BYTES);
+                    // Cheeky trick to encode the released line into the poisoned value
+                    #[cfg(feature = "poison_on_release")]
+                    crate::util::memory::set_pattern(
+                        line.start(),
+                        (0xdeadbeef ^ line.start().as_usize()) & !0xff,
```
qinsoon, line 255: Why poison as a pattern rather than a `u8` value as other spaces? What is the pattern here? How do we supposed to use the pattern?

qinsoon: This removes the existing code for `immix_zero_on_release`. So immix will not zero lines even if `immix_zero_on_release` is enabled.

You should have an assertion to make sure that `immix_zero_on_release` and `poison_on_release` should not co-exist, as they conflict with each other for Immix space, and you should zero/poison the memory based on which feature is enabled.

src/policy/copyspace.rs
qinsoon: I am not sure if we need the different values to tell which space it is, especially after Kunshan's change in https://github.com/mmtk/mmtk-core/pull/1236. But if we do want to use different values, there needs to be a place where we can look at and know what the value means.

qinsoon (changes requested):
Aside the in-line comments, this PR is missing implementation for `poison_on_release` for spaces other than "copy/immix/largeobject" spaces. Favorably the PR should implement the feature for all the spaces, and minimally the PR should at least panick if the feature is enabled and spaces in use do not have an implementation for `poison_on_release`.

src/policy/immix/block.rs
k-sareen: This feature supersedes zero on release. The pattern here is helpful since it encodes the line that was released into the pattern. So you can XOR with `0xdeadbeef` to get the exact line that was released and where the object comes from. It was useful to me while debugging ART.

k-sareen: `immix_zero_on_release` was a debugging feature to catch untraced objects. But using `0x0` can actually make it harder to debug (in my and others' experience) since you don't know if it's a null pointer or something that was released early. So previously some of the students and I used an obvious nonsense value like `0xab` as the pattern. This is refinement of the same since it encodes the line into the pattern.

k-sareen:
> Favorably the PR should implement the feature for all the spaces, and minimally the PR should at least panick if the feature is enabled and spaces in use do not have an implementation for poison_on_release

You can't implement this for native mark-sweep space since it uses the cell to thread a free-list. You can't exactly do it with mark-compact space as well. You could only do it with a completely released 4MB chunk. And the rest of the spaces don't release.

src/policy/immix/block.rs
qinsoon: If you think "poison_on_release" is better than "zero_on_release", just remove the `immix_zero_on_release` feature, and make sure that `poison_on_release` can be a replacement for `immix_zero_on_release`.

qinsoon:
> > Favorably the PR should implement the feature for all the spaces, and minimally the PR should at least panick if the feature is enabled and spaces in use do not have an implementation for poison_on_release
> 
> You can't implement this for native mark-sweep space since it uses the cell to thread a free-list. You can't exactly do it with mark-compact space as well. You could only do it with a completely released 4MB chunk. And the rest of the spaces don't release.

Native mark-sweep only uses the first address in a cell as the freelist, and the rest of the cell can be poisoned. Mark compact also releases chunks/pages.

The issue is that you need to define what `poison_on_release` means, and all the policies should behave consistently to that (or give clear errors messages if a policy does not behave properly). Otherwise when a user turns on the feature, what would he expect?

src/policy/immix/block.rs
qinsoon: > So you can XOR with 0xdeadbeef to get the exact line that was released and where the object comes from. 

I am not following here. Can you get the line by aligning down from the faulty address? Why is XOR needed?

k-sareen: The problem isn't the line itself but what object was referring to it. In my case I did not have a debugger available because crashes were happening at Android boot time. So I had to be creative to debug memory corruption errors (hence the `dump_ram_around_address` function as well).

So consider the case that a line has been overwritten when an object inside it was live.

Object = `0x12000040`
Line = `0x12000000`
Encoded pattern = `0xCCADBE00`

An object referring to `0x12000040` would maybe load a field at an offset `0x8`. So it loads the address `0x12000048`. If we wrote just `0x0`s into the line then we would get a SIGSEGV but then no way to figure out what address was loaded. But if we wrote `0xCCADBE00` to it then the SIGSEGV output would list say `0xCCADBE00` and then we can XOR with `0xDEADBE00` and get back `0x12000000` which was the line that was overwritten.

k-sareen: This is I guess unnecessary in the general case if you have a debugger, but I definitely found it useful.

qinsoon: I see. Without the explanation above, I wouldn't know why you encode the value in that way, and how I am supposed to use the encoded value. Those explanation should go into the comments and the docs in the code so others won't have the same question as I did when they encounter the code and when they use this feature.

Besides, I suggest separating it into a different features:
* `poison_on_release`: this just generally poisons released memory (or dead objects) with an identifiable value.
* `immix_poison_with_line_address_on_release` (or something like this): it poisons the memory with an encoded line address.

src/util/memory.rs
```diff
@@ -105,6 +105,48 @@ pub fn set(start: Address, val: u8, len: usize) {
     }
 }
 
+/// Set a range of memory to the given pattern. Similar to C++'s std::fill.
+pub fn set_pattern(start: Address, pattern: usize, len: usize) {
```
wks, line 109: I suggest using a fixed-size integer type (or a type trait) for the pattern.  At the call site, we give the pattern as a hexadecimal literal, but here we write memory at the granularity of `usize` which is platform-dependent.  For example, if `pattern` is `0xdeadbeef`, and the platform is x86_64, then the memory will be filled with `EF BE AD DE 00 00 00 00 EF BE AD DE 00 00 00 00 EF BE ...`, and I think we don't want the `00 00 00 00` in between.

```suggestion
pub fn set_pattern(start: Address, pattern: u32, num_bytes: usize) {
```

Alternatively

```suggestion
trait UnsignedIntButNotUsize: num_traits::PrimInt + num_traits::Unsigned + num_traits::ToBytes {}
impl UnsignedIntButNotUsize for u8 {}
impl UnsignedIntButNotUsize for u16 {}
impl UnsignedIntButNotUsize for u32 {}
impl UnsignedIntButNotUsize for u64 {}

pub fn set_pattern<T: UnsignedIntButNotUsize>(start: Address, pattern: T, num_bytes: usize) {
```

wks (commented):
In general, I think we should keep a way to set the poison bytes to all zero.  In some VM runtimes, such as Ruby, there are many NULL checking, and it will panic if it sees a field is NULL.  So clearing the dead object to all zero bytes will make the program fail faster in some cases, crashing on one of the NULL-checking assertions.  Overwriting the object to other poison values may slip through the assertion and fail somewhere else.

k-sareen:
> Overwriting the object to other poison values may slip through the assertion and fail somewhere else.

Sure. It's often the reverse in Java. You would add `if (object != null) { /* do stuff */ }` so you would actually miss incorrect uses.

PR #1242, opened 2024-11-29
Fix clippy warnings for Rust 1.83

wks:
Clippy 1.83 produces some new warnings:

-   `needless_lifetimes` is extended to suggest eliding `impl` lifetimes.
-   `empty_line_after_doc_comments` is added to the `suspicious` group.


qinsoon (approved):
LGTM. Thanks.

PR #1244, opened 2024-12-01
Fix warnings for lifetime in MmapAnnotation impl

qinsoon:
https://github.com/mmtk/mmtk-core/pull/1242 fixed most similar issues in the repo, but https://github.com/mmtk/mmtk-core/pull/1236 introduced `MmapAnnotation` and introduced a new warning.


wks (approved):
LGTM

PR #1248, opened 2024-12-05
Allow setting object metadata for VM space objects. Expose VO bit under a feature.

qinsoon:
This PR changes a few things for vo bit:
1. Add a function to set object metadata for an object in the VM space: `MMTK::initialize_vm_space_object`.
2. Add a feature `vo_bit_access` to expose VO bit and a binding may use it at its own risk.
3. Mark VO bit side metadata base address only avilable for 64 bits.

The second is needed for Julia. The Julia binding uses MMTk immortal allocation or VM space for a region of memory, and pop the regions with boot image objects with no clear way to identify each object. The easist workaround is to bulk set VO bit for the region. The problem from this is that MMTk cannot identify valid objects in those regions. However, Julia binding only uses VO bit for pinning objects. Objects in the immortal space or the VM space will not be moved so failing to pinning objects in those regions is benign. Currently the Julia binding duplicates a bunch of side metadata code to bulk set VO bit only using the VO bit side metadata base address. See https://github.com/mmtk/mmtk-julia/pull/200

### DIFF
--- a/src/util/metadata/side_metadata/mod.rs
+++ b/src/util/metadata/side_metadata/mod.rs
@@ -22,9 +22,3 @@ pub(crate) use helpers::*;
 #[allow(unused_imports)]
 pub(crate) use helpers_32::*;
 pub(crate) use sanity::SideMetadataSanity;
-
-/// Side metadata spec for VO bit. This is only available when the feature "vo_bit_access" is enabled.
-/// Check the comments on "vo_bit_access" in `Cargo.toml` before use. Use at your own risk.
-#[cfg(feature = "vo_bit_access")]
-pub const VO_BIT_SIDE_METADATA_SPEC: SideMetadataSpec =
-    crate::util::metadata::side_metadata::spec_defs::VO_BIT;
--- a/src/util/metadata/side_metadata/mod.rs
+++ b/src/util/metadata/side_metadata/mod.rs
@@ -23,6 +23,8 @@ pub(crate) use helpers::*;
 pub(crate) use helpers_32::*;
 pub(crate) use sanity::SideMetadataSanity;
 
+/// Side metadata spec for VO bit. This is only available when the feature "vo_bit_access" is enabled.
+/// Check the comments on "vo_bit_access" in `Cargo.toml` before use. Use at your own risk.
 #[cfg(feature = "vo_bit_access")]
 pub const VO_BIT_SIDE_METADATA_SPEC: SideMetadataSpec =
     crate::util::metadata::side_metadata::spec_defs::VO_BIT;
--- a/src/util/metadata/side_metadata/mod.rs
+++ b/src/util/metadata/side_metadata/mod.rs
@@ -22,3 +22,7 @@ pub(crate) use helpers::*;
 #[allow(unused_imports)]
 pub(crate) use helpers_32::*;
 pub(crate) use sanity::SideMetadataSanity;
+
+#[cfg(feature = "vo_bit_access")]
+pub const VO_BIT_SIDE_METADATA_SPEC: SideMetadataSpec =
+    crate::util::metadata::side_metadata::spec_defs::VO_BIT;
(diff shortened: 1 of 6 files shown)


src/util/metadata/vo_bit/mod.rs
```diff
@@ -222,3 +223,23 @@ pub fn is_internal_ptr_from_vo_bit<VM: VMBinding>(
...
 }
+
+#[cfg(feature = "vm_space")]
+use crate::MMTK;
+
+/// Set VO bit for a VM space object.
+/// Objects in the VM space are allocated/managed by the binding. This functio provides a way for
+/// the binding to set VO bit for an object in the space.
+#[cfg(feature = "vm_space")]
+pub fn set_vo_bit_for_vm_space_object<VM: VMBinding>(
```
wks, line 234: Currently the `mmtk::util::metadata::vo_bit` module is `pub(crate)`, and `set_vo_bit_for_vm_space_object` is not re-exported. So this method is still inaccessible to the VM binding although it is `pub`.

I suggest we make the `vo_bit` module `pub`, but make every `pub` items inside the `vo_bit` module `pub(crate)`, and make `set_vo_bit_for_vm_space_object` `pub`.  Then the VM binding can call this function by `mmtk::util::metadata::vo_bit::set_vo_bit_for_vm_space_object`.

I intended to do this refactoring anyway, and remove the `is_mmtk_object` feature, and move everything related to `is_mmtk_object` into `vo_bit` itself (see [this issue](https://github.com/mmtk/mmtk-core/issues/804)).

src/util/metadata/side_metadata/mod.rs
```diff
@@ -22,3 +22,9 @@ pub(crate) use helpers::*;
 #[allow(unused_imports)]
 pub(crate) use helpers_32::*;
 pub(crate) use sanity::SideMetadataSanity;
+
+/// Side metadata spec for VO bit. This is only available when the feature "vo_bit_access" is enabled.
+/// Check the comments on "vo_bit_access" in `Cargo.toml` before use. Use at your own risk.
+#[cfg(feature = "vo_bit_access")]
+pub const VO_BIT_SIDE_METADATA_SPEC: SideMetadataSpec =
```
wks, line 29: Once we make the `vo_bit` module public, we can move this into the `vo_bit` module directly, as a `pub const`.

src/util/metadata/vo_bit/mod.rs
```diff
@@ -222,3 +223,23 @@ pub fn is_internal_ptr_from_vo_bit<VM: VMBinding>(
...
+    mmtk: &'static MMTK<VM>,
+    object: ObjectReference,
+) {
+    use crate::policy::space::Space;
+    debug_assert!(
+        mmtk.get_plan().base().vm_space.in_space(object),
+        "{} is not in VM space",
+        object
+    );
+    set_vo_bit(object);
```
wks, line 244: This sets a single VO bit at a time.  Depending on the density of objects in the VM space, if this is a bottleneck, we can provide another function that allows the VM binding to provide a bitmap directly, and memcpy into our VO bit metadata (in a single pass, or chunk by chunk).

src/util/metadata/vo_bit/mod.rs
qinsoon: But I removed this function `set_vo_bit_for_vm_space_object`. I realized that a binding needs to set all the object metadata to properly initialize an object in the VM space. I added a function `MMTK::initialize_vm_space_object()` which calls `SFT::initialize_object_metadata` on `vm_space`.

src/util/metadata/vo_bit/mod.rs
qinsoon: I agree, but I feel this is out of scope for this PR.

wks (approved):
LGTM

qinsoon:
binding-refs
JULIA_BINDING_REF=use-mmtk-vo-bit-access

PR #1249, opened 2024-12-05
Use macos-15 for style check

qinsoon:
https://github.com/mmtk/mmtk-core/pull/1216 updated the test runner image from `macos-12` to `macos-15`, but I forgot to update the image for style checks. This PR updates the runner for style checks as well.


wks (approved):
LGTM

Issue #1254, opened 2025-01-07
Native MarkSweep is not marking objects atomically.


wks:
In `src/policy/marksweepspace/native_ms/global.rs`:

```rust
    fn trace_object<Q: ObjectQueue>(
        &self,
        queue: &mut Q,
        object: ObjectReference,
    ) -> ObjectReference {
        // ...
        if !VM::VMObjectModel::LOCAL_MARK_BIT_SPEC.is_marked::<VM>(object, Ordering::SeqCst) {
            VM::VMObjectModel::LOCAL_MARK_BIT_SPEC.mark::<VM>(object, Ordering::SeqCst);
            let block = Block::containing(object);
            block.set_state(BlockState::Marked);
            queue.enqueue(object);
        }
        object
    }
```

The marking is non-atomic.  If two GC workers attempt to mark the same object, it will be marked twice, and will be enqueued twice.  It has two problems.

Firstly, enqueuing twice will cause the same object to be scanned twice, resulting in **more work packets to be generated to do duplicated jobs**.

Secondly, some VM binding make **assumptions that each live object is scanned exactly once** during full-heap GC (or at most once during nursery GC).  Such behavior is usually related to the behaviors of the existing GCs of the VM.  One example is CRuby.  It attempts to clean up some data inside objects during the mark phase of a GC.  When using MMTk binding, it does so when scanning an object.  But when using MMTk with the MarkSweep plan and multiple GC workers, two GC workers will attempt to clean up the same object.  The data structure being cleaned is not thread-safe, and it will crash the VM.

I think it is a reasonable to assume that a GC enqueues each object and scans each object at most once, unless there is a compelling reason not to do so.  MarkSweep is the only plan that marks object non-atomically, and I think it is a mistake.

Related discussions: https://mmtk.zulipchat.com/#narrow/channel/313365-mmtk-ruby/topic/MMTk.20and.20marking.20id_table.20entries.2E

qinsoon:
It is intended for performance. See the discussion in https://github.com/mmtk/mmtk-core/issues/313.

However, what is strange is that for Immix, we actually atomically mark objects. In Java MMTk, marking for Immix and mark sweep is non atomic.

wks:
[...]

PR #1255, opened 2025-01-07
Unique object enqueuing option

wks:
Added a constant `VMBinding::UNIQUE_OBJECT_ENQUEUING`.  When set to true, MMTk will guarantee that each object is enqueued at most once in each GC.  This can be useful for VMs that piggyback on object scanning to visit objects during GC.

Implementation-wise, the mark bit is set atomically when `VMBinding::UNIQUE_OBJECT_ENQUEUING` is true.  This PR only affects the native MarkSweep space.  Other spaces already do this atomically.

Fixes: https://github.com/mmtk/mmtk-core/issues/1254

### DIFF
--- a/src/vm/mod.rs
+++ b/src/vm/mod.rs
@@ -79,4 +79,19 @@ where
     /// Note that MMTk does not attempt to do anything to align the cursor to this value, but
     /// it merely asserts with this constant.
     const ALLOC_END_ALIGNMENT: usize = 1;
+
+    /// When set to `true`, all plans will guarantee that during each GC, each live object is
+    /// enqueued at most once, and therefore scanned (by either [`Scanning::scan_object`] or
+    /// [`Scanning::scan_object_and_trace_edges`]) at most once.
+    ///
+    /// When set to `false`, MMTk may enqueue an object multiple times due to optimizations, such as
+    /// using non-atomic operatios to mark objects.  Consequently, an object may be scanned multiple
+    /// times during a GC.
+    ///
+    /// The default value is `false` because duplicated object-enqueuing is benign for most VMs, and
+    /// related optimizations, such as non-atomic marking, can improve GC speed. VM bindings can
+    /// override this if they need.  For example, some VMs piggyback on object-scanning to visit
+    /// objects during a GC, but may have data race if multiple GC workers visit the same object at
+    /// the same time.  Such VMs can set this constant to `true` to workaround this problem.
+    const UNIQUE_OBJECT_ENQUEUING: bool = false;
 }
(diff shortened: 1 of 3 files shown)


wks:
This PR intends to fix the immediate problem of MarkSweep enqueuing objects multiple times, which may crash the Ruby binding when there are multiple GC workers and the plan is MarkSweep.

I observed that MarkSweepSpace and ImmixSpace do object marking completely differently, and there are code fragments related to marking scattered in `ImmixSpace`, `impl VMLocalMarkBitSpec` and `impl MarkState`.  We should do a larger scale refactoring to make all spaces use the `MarkState` struct.  I plan to do it later because it may need more careful performance evaluation.

k-sareen:
Re: MarkState refactoring. Yeah I've done it in [my branch](https://github.com/k-sareen/mmtk-core/blob/58883cc2f15ee5c6499143c02f4a2b388b4fe53e/src/policy/immix/immixspace.rs#L54), but you have to be careful since you can't use it for LOS. LOS mark bits are 2-bits _per page_ not 2-bits _per valid object_.

wks:
The "Public API Check" failed.  This PR introduces a new API `pub const mmtk::vm::VMBinding::UNIQUE_OBJECT_ENQUEUING: bool`, but it has a default value `false` so it won't break existing bindings.

src/vm/mod.rs
```diff
@@ -79,4 +79,19 @@ where
...
+    /// When set to `false`, MMTk may enqueue an object multiple times due to optimizations, such as
+    /// using non-atomic operatios to mark objects.  Consequently, an object may be scanned multiple
+    /// times during a GC.
+    ///
+    /// The default value is `false` because duplicated object-enqueuing is benign for most VMs, and
+    /// related optimizations, such as non-atomic marking, can improve GC speed. VM bindings can
+    /// override this if they need.  For example, some VMs piggyback on object-scanning to visit
+    /// objects during a GC, but may have data race if multiple GC workers visit the same object at
+    /// the same time.  Such VMs can set this constant to `true` to workaround this problem.
+    const UNIQUE_OBJECT_ENQUEUING: bool = false;
```
qinsoon, line 96: We can move this to the `Scanning` trait. It seems only relevant for scanning.

qinsoon (approved):
Just a minor suggestion. Otherwise it looks good to me.

src/vm/mod.rs
wks: Agreed.  I moved it to the `Scanning` trait.

k-sareen:
Should update `may_trace_duplicate_edges` to be equal to the constant for MS

wks:
> Should update `may_trace_duplicate_edges` to be equal to the constant for MS

It's a bit hard for the current implementation because `PlanConstraint` is not parameterized with `<VM: VMBinding>`, but the feature is controlled by `Scanning::UNIQUE_OBJECT_ENQUEUING`.

PR #1259, opened 2025-01-13
Mention pitfall in `ObjectTracerContext::with_tracer` doc comment

k-sareen:


src/vm/scanning.rs
```diff
@@ -61,10 +61,16 @@ pub trait ObjectTracerContext<VM: VMBinding>: Clone + Send + 'static {
...
     ///
-    /// When the `ObjectTracer::trace_object` is called, if the traced object is first visited
-    /// in this transitive closure, it will be enqueued.  After `func` returns, the implememtation
-    /// will create work packets to continue computing the transitive closure from the newly
-    /// enqueued objects.
+    /// A typical use-case of this method is to call `ObjectTracer::trace_object` in a loop for
+    /// finalization candidates to create `ProcessEdgesWork` packets.
+    //// **Note**: Do not use this API in such a way that only one object is enqueued at a time.
+    /// This can result in creating a work packet for a single object, exhausting (physical) memory
+    /// if there are enough finalization candidates.
```
wks, line 68: I think the description is too specific to the bug you fixed recently.  It's not specific to finalization (it's also applicable to SoftReference), and it is better to provide an example of the *right* way to use it, and then describe the consequence of flipping the order of the loop and the closure.  You can leave this to me.  I plan to write a dedicated porting guide section for weak references and finalizers, and update this doc comment at the same time.

k-sareen: Okay. Thank you.

PR #1261, opened 2025-01-14
Fix a subtraction overflow in get_free_pages.

wks:
The used pages can also be greater than the total pages for the same reason as those in computing `get_available_pages`, and it can also happen if the VM binding disabled GC, in which case we may over-allocate without triggering GC.  When it overflows, `get_free_pages` will cause subtraction overflow, and will panic in debug build.

We switch to `saturating_sub` so that it will return 0 if overflow happens.  It still makes sense.  0 means there is no free pages because we are over-allocating beyond the current heap size set by the GC trigger.


wks:
Rust 1.84 added a new lint [unnecessary_map_or](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_map_or), and the style check failed for that reason.  I'll fix that separately.

qinsoon (approved):
LGTM

PR #1263, opened 2025-01-15
Always return false for collection_required for NoGC

peterzhu2118:
If we use the dynamic heap with NoGC, it will trigger a GC when `space_full` is true. This will cause a Rust panic of `entered unreachable code: GC triggered in nogc` because `collection_required` returns true.

### DIFF
--- a/src/plan/nogc/global.rs
+++ b/src/plan/nogc/global.rs
@@ -46,8 +46,8 @@ impl<VM: VMBinding> Plan for NoGC<VM> {
         &NOGC_CONSTRAINTS
     }
 
-    fn collection_required(&self, space_full: bool, _space: Option<SpaceStats<Self::VM>>) -> bool {
-        self.base().collection_required(self, space_full)
+    fn collection_required(&self, _space_full: bool, _space: Option<SpaceStats<Self::VM>>) -> bool {
+        false
     }
 
     fn base(&self) -> &BasePlan<VM> {


qinsoon:
The current behavior of NoGC with dynamic heap sizes is that MMTk panics when the heap meets the current heap size (not the maximum heap size). This is not ideal.

With this PR, NoGC will keep allocating, regardless of the heap sizes (dynamic or fixed). I don't think it is correct either. 

I think the expected behavior for NoGC with dynamic heap sizes is that MMTk keeps allocating until it reaches the maximum heap size. We need to do a bit more refactoring to allow that.

qinsoon:
Maybe we can implement a special GC trigger specifically for NoGC with dynamic heap size to handle this. We cannot really use `MemBalancer`, as `MemBalancer` needs GC stats to adjust heap sizes, and we don't have GC stats in NoGC.

wks:
I agree with @qinsoon.  If we simply let `collection_required` return `false`, it will keep allocating until running out of physical memory, bypassing the GC trigger completely, and disregarding any heap size set by the user.

And "dynamic" heap size doesn't make sense for NoGC.  There could be several solutions.

1.  Disallow any form of dynamic heap size when using NoGC.  Simply panic if the user sets plan to NoGC and the GCTrigger to dynamic.
2.  Make MemBalancer set the initial heap size to the maximum if the plan is NoGC.  This will gracefully degrade it to a fixed heap size (equal to the max heap size) anyway without crashing.

I prefer the second approach.

qinsoon:
> I agree with @qinsoon. If we simply let `collection_required` return `false`, it will keep allocating until running out of physical memory, bypassing the GC trigger completely, and disregarding any heap size set by the user.
> 
> And "dynamic" heap size doesn't make sense for NoGC. There could be several solutions.
> 
> 1. Disallow any form of dynamic heap size when using NoGC.  Simply panic if the user sets plan to NoGC and the GCTrigger to dynamic.
> 2. Make MemBalancer set the initial heap size to the maximum if the plan is NoGC.  This will gracefully degrade it to a fixed heap size (equal to the max heap size) anyway without crashing.
> 
> I prefer the second approach.

We can just check here and if the plan is NoGC, we just create a `FixedHeapSizeTrigger` with the maximum heap size with a warning.
https://github.com/mmtk/mmtk-core/blob/2f6f078b6fd7bdc23688ead55a055d825a23b863/src/util/heap/gc_trigger.rs#L43-L46

This is probably cleaner, as we do not need to 'hack' mem balancer.

peterzhu2118:
Thanks for fixing it!

PR #1264, opened 2025-01-16
Force fixed heap size when using NoGC

wks:
The dynamic heap size trigger needs GC to adjust the heap size, but NoGC can't do GC.  So it doesn't make sense to use dynamic heap size with NoGC.  Currently, if the user selects the NoGC plan and the dynamic heap size trigger, it will trigger GC at the minimum heap size and then panic immediately.

With this change, MMTk will give a warning and use fixed heap size trigger instead, using the maximum heap size specified in the dynamic trigger as the heap size.

### DIFF
--- a/src/util/heap/gc_trigger.rs
+++ b/src/util/heap/gc_trigger.rs
@@ -40,10 +40,19 @@ impl<VM: VMBinding> GCTrigger<VM> {
                 GCTriggerSelector::FixedHeapSize(size) => Box::new(FixedHeapSizeTrigger {
                     total_pages: conversions::bytes_to_pages_up(size),
                 }),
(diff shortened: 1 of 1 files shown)


wks:
Note: This is intended to fix the problem https://github.com/mmtk/mmtk-core/pull/1263 addresses.  With this PR, if CRuby user only selects the NoGC plan and doesn't specify a trigger, the default behavior will be using the fixed-size trigger with 80% physical memory (the default max heap size).  This should be big enough for experiment purpose.  I don't think it's practical to use Ruby with NoGC in production, but the user can still specify a fixed heap size if needed.

qinsoon (approved):
LGTM

PR #1265, opened 2025-01-16
Special topic chapter for finalizers and weak references

wks:
This PR adds a special topic chapter in the Porting Guide for supporting finalizers and weak references.  This topic is frequently asked and somewhat complex, and needs a dedicated chapter.

We also updated the doc comments of the `Scanning::process_weak_refs` API to add code example of the intended use case, and warn the users about potential pitfalls.


wks:
I have finished the chapter, and I invited @qinsoon and @k-sareen to review it.  There are some things I haven't done in this PR, but can be added later.

1.  Ephemerons: We currently have not implemented ephemerons in any VMs.  We need Ephemeron for V8, but that's not a priority at this moment.  We can add a subsection for ephemerons when we implemented it, or when someone needs help implementing ephemerons with MMTk.
2.  Hyperlinks between the doc comments in the code and the Porting Guide chapter.  The porting guide needs to be rendered in order to be published to https://docs.mmtk.io/, and that'll happen after this PR is merged.  We can add the URL when the new chapter becomes publicly available.

Maybe I also need to define some concepts in that chapter, such as what exactly is a finalizer or a weak reference.  I think the current state is clear enough for people with experience with GC and VM development.  But please let met know if anything is not clear.

wks:
It is still arguable whether "retain" or "resurrect" is a better word for keeping an unreachable object alive in a GC.  "Resurrect" is more intuitive because we also talk about "finalizers can resurrect dead objects".  But the word "resurrect" may imply that the object has already died once.  It is OK for finalizers because the object has already become unreachable during mutator time, which can be considered as "dead".  But for soft references, it is a bit awkward to say we "resurrect" the referent of a soft reference because that implies we consider the referent as "dead" before.  It is "dead" in the sense that it is not strongly reachable, but softly reachable objects usually live through GCs, as long as it is not emergency GC.  Maybe "retain" is better when describing soft references.

In existing code, we use the word "retain" in many places in ReferenceProcessor.  There is only one use of "resurrect" in FinalizerProcessor, and it was added by me.

docs/userguide/src/portingguide/topics/prefix.md
```diff
@@ -0,0 +1,5 @@
+# Special topics
+
+Every VM is special in some way.  Because of this, some VM bindings may use MMTk features not
```
qinsoon, line 3: If we change the title, the paragraph needs to be changed accordingly. I feel it is more reasonable that this section is dedicated to how to implement different runtime features with MMTk.

docs/userguide/src/portingguide/topics/weakref.md
```diff
@@ -0,0 +1,451 @@
+# Finalizers and Weak References
+
+Some VMs support **finalizers**.  In simple terms, finalizers are clean-up operations associated
+with an object, and are executed when the object is dead.
+
+Some VMs support **weak references**.  If an object cannot be reached from roots following only
+strong references, the object will be considered dead.  Weak references to dead objects will be
+cleared, and associated clean-up operations will be executed.  Some VMs also support more complex
```
qinsoon, line 8: The definition for weak references is incorrect.

> If an object cannot be reached from roots following only strong references, the object will be considered dead.

If an object cannot be reached from strong references, it is not "strongly reachable". It is not dead, as it can still be "weakly reachable". Sometimes "reachable" implies strongly reachable. That's fine. But this should not be used in a context that may cause confusion.

> Weak references to dead objects will be cleared, and associated clean-up operations will be executed.

This is indeed confusing. If we have a weak reference to the object, the object is not dead. It is up to the language semantics or the VM to decide whether to keep the weakly-referenced object alive or not. Only when the VM decides that the referent will not be kept alive, the object/referent is dead, otherwise it is alive.

docs/userguide/src/portingguide/topics/weakref.md
```diff
@@ -0,0 +1,451 @@
...
+During each GC, after the transitive closure is computed (i.e. after all objects reachable from
+roots have been reached), MMTk calls `Scanning::process_weak_refs` which is implemented by the VM
+binding.  Inside this function, the VM binding can do several things.
+
+-   **Query reachability**: The VM binding can query whether any given object has been reached.
+    +   Do this with `ObjectReference::is_reachable()`.
+-   **Query forwarded address**: If an object is already reached, the VM binding can further query
+    the new address of an object.  This is needed to support copying GC.
+    +   Do this with `ObjectReference::get_forwarded_object()`.
+-   **Resurrect objects**: If an object is not reached, the VM binding can optionally resurrect the
```
qinsoon, line 34: See my argument above for the definition of weak ref. We do not 'resurect' weak reference. 

It is arguable whether we should say 'resurrect' finalizable objects. GC needs to keep finalizable objects alive until they are properly finalized -- I don't want to call it 'resurrect', simply it is not dead from the GC's perspective. Also 'resurrecting objects' normally refers to the specific application behavior that an object that is being finalized and should become dead after finalization is at the language/application level kept alive. Using 'resurrect' in the GC may get things confusing.

GC handbook's definition for 'resurrection': "an action performed by a finaliser that caused the previously unreachable object to become reachable."

docs/userguide/src/portingguide/topics/weakref.md
```diff
@@ -0,0 +1,451 @@
...
+
+Some VMs, particularly the Java VM, executes finalizers during mutator time.  The dead finalizable
+objects must be brought back to life so that they can still be accessed after the GC.
+
+The `Scanning::process_weak_refs` has an parameter `tracer_context: impl ObjectTracerContext<VM>`.
+This parameter provides the necessary mechanism to resurrect objects and make them (and their
+descendants) live through the current GC.  The typical use pattern is:
+
+```rust
+impl<VM: VMBinding> Scanning<VM> for VMScanning {
```
qinsoon, line 114: It is generally preferrable to put the code into a test, and include the snippets in the test in the doc. In such cases, the code will be always checked by the tests and the CI. Maybe we cannot run `process_weak_refs` in our mock tests, but we can still make sure that it can be compiled. 

See here for examples: https://github.com/mmtk/mmtk-core/blob/051bc7470feef915c445305301e6113f86d3957b/docs/userguide/src/portingguide/perf_tuning/alloc.md?plain=1#L36

qinsoon (commented):
The doc is written with dfiferent definitions of 'weak ref', and 'resurrect objects'. I think we should refer to a well-accepted literature such as the GC handbook for such definitions, and be consistent with their definitions. In my opinion, we need a pass through the section to fix the definitions and related descriptions. I would like to do a detailed review after that.

k-sareen:
I would avoid inventing terminology and just use what is in the GC Handbook. The handbook uses resurrect for finalizers and "trace" for soft references

docs/userguide/src/SUMMARY.md
```diff
@@ -36,6 +36,8 @@
     - [Performance Tuning](portingguide/perf_tuning/prefix.md)
         - [Link Time Optimization](portingguide/perf_tuning/lto.md)
         - [Optimizing Allocation](portingguide/perf_tuning/alloc.md)
+    - [Special Topics](portingguide/topics/prefix.md)
```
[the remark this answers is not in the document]
wks: What about "VM-Specific Concerns"?

The *GC Handbook* has a chapter named *Language-Specific Concerns*, and it discusses only finalizers and weak references.  I think that's because finalizer and weak reference semantics is part of the programming language, and is visible to the programmers.  Things like conservative stack scanning, object pinning and interior pointers are not all exposed at the language level, but they are more related to VM implementations.  So "VM-Specific Concerns" may be a better title.

I don't like calling them "Runtime/Language *Features*" because "features" are things meant to be used by their users, but things like stack scanning are peculiar aspects of those runtimes/languages that their implementers should care about.

And I think "VM peculiarities" also describes what the chapters in this part are about.  But it sounds offensive to the VMs.

docs/userguide/src/portingguide/topics/weakref.md
wks: I will rewrite this part using definitions from the *GC handbook*, and probably add a "Definitions" section before the "Overview" section.

I didn't use the terms "strongly/weakly reachable" because I thought they were Java-specific.  But since the *GC Handbook* also uses that term in the language-neutral introduction section of weak references, I believe those terms should be acceptable for other VMs in general, too.

docs/userguide/src/portingguide/topics/weakref.md
wks: I consulted @eliotmoss and read the GC handbook, and they both agree with what you said.  "Resurrection" refers to the action in the finalizer that makes the object reachable from other parts of the program by, for example, assigning its reference to a global variable.  This is not what `tracer.trace_object` does.  Even if we call `trace_object(object)`, it still doesn't guarantee the `object` will "resurrect" in that sense.  If the `finalize()` function doesn't leak its reference, the object will still be collectable after `finalize()` returns.

"Retain" should be a better term, and it is already used in JikesRVM as well as our existing `ReferenceProcessor` in mmtk-core.

docs/userguide/src/SUMMARY.md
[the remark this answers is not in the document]
wks: I renamed it to "VM-specific concerns".

docs/userguide/src/portingguide/topics/prefix.md
wks: Yes. I changed this part, too.

docs/userguide/src/portingguide/topics/weakref.md
wks: I added a "Definitions" section before the "Overview" section, and I used the term "strongly reachable" and "weakly reachable".  I also added a Glossary chapter at the top level of the User Guide and linked to it from this chapter.  Currently I only added the description of "object graph" in the Glossary.

docs/userguide/src/portingguide/topics/weakref.md
wks: I now use "retain" consistently when describing what `tracer.trace_object` does.  I still mentioned "resurrect" briefly when describing the counterintuitive behavior that finalizers make an unreachable object reachable from user threads again.

docs/userguide/src/portingguide/topics/weakref.md
wks: I now use a mock test to contain the example code.  This can ensure it compiles.  But we still can't actually run it in mock tests because we can't construct a `GCWorker` object.

wks:
The tests on 1.71.1 (our MSRV) failed because the latest version of the "built" crate (v0.7.7) requires Rust 1.74 or later.

k-sareen:
If we want to update our MSRV or pin the crate version it should be a separate PR. Currently my ART port is stuck at Rust v1.74 for reasons and I can't easily update it. So I'd personally like if our MSRV doesn't get bumped too high, but I can always just maintain it in my own branch for backwards compatibility.

wks:
I have made changes since this PR was reviewed the last time.  Important changes include:

-   I renamed "Special Topics" to "VM-specific Concerns"
-   I now use the word "retain" to refer to the thing `trace_object` does.  Use "resurrect" only to describe the application-visible phenomenon, i.e. an unreachable object becoming reachable again.
-   I added a "Definitions" section and define finalizers, weak references and strongly/weakly reachable in the way the *GC Handbook* defines them.
-   I also added a "Glossary" chapter at the top level and added definition for the "object graph" there.  It should make the definition of "strongly/weakly reachable" clearer in the added Finalizer and Weak Reference chapter.
-   I clarified the behavior of `ObjectReference::is_reachable`, and emphasize that it may conservatively return `true` for objects that have not been reached, particularly mature objects during nursery GC.
-   I now use a mock test case to hold the example code to ensure it is free of compilation error.

Currently there is an error when testing with Rust version 1.71.1 (our current MSRV).  I have made [another PR](https://github.com/mmtk/mmtk-core/pull/1276) for bumping the MSRV to 1.74.1.

docs/userguide/src/SUMMARY.md
```diff
@@ -36,6 +38,8 @@
     - [Performance Tuning](portingguide/perf_tuning/prefix.md)
         - [Link Time Optimization](portingguide/perf_tuning/lto.md)
         - [Optimizing Allocation](portingguide/perf_tuning/alloc.md)
+    - [VM-specific Concerns](portingguide/concerns/prefix.md)
```
k-sareen, line 41: In the step-by-step porting guide I preferred to use the term "language runtime" or "runtime" instead of VM since MMTk is more general. I won't comment on all the instances here. Just doing it once.

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+    called again after computing the transitive closure that includes *retained objects and their
+    descendants*.  This helps handling multiple levels of weak reference strength.
+    +   Do this by returning `true` from `process_weak_refs`.
+
+The `Scanning::process_weak_refs` function also gives the VM binding a chance to perform other
+operations, including (but not limited to)
+
+-   **Do clean-up operations**: The VM binding can perform clean-up operations, or queue them to be
+    executed after GC.
+-   **update fields** that contain weak references.
```
k-sareen, line 70:
```suggestion
-   **Update fields** that contain weak references.
```

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+        copying GC.
+    -   **Clear the field**: It can clear the field if the referent has not been reached and the
+        binding decides it is unreachable.
+
+Using those primitive operations, the VM binding can support different flavors of finalizers and/or
+weak references.  We will discuss common use cases in the following sections.
+
+## Supporting finalizers
+
+Different VMs define "finalizer" differently, but they all involve performing operations when an
```
k-sareen, line 81: Technically it's the language specification (if it exists) that defines finalization, no? Same for other locations below.

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+### When to run finalizers?
+
+Depending on the finalizer semantics in different VMs, finalizers can be executed during GC or
+during mutator time after GC.
+
+The VM binding can run finalizers immediately in `Scanning::process_weak_refs` when finding a
+finalizable object unreachable.  Beware that executing finalizers can be time-consuming.  The VM
+binding can creating work packets and let each work packet process a part of all finalizable
+objects.  In this way, multiple GC workers can process finalizable objects in parallel.  The
+`Scanning::process_weak_refs` function is executed in the `VMRefClosure` stage, so the created work
```
k-sareen, line 105: Should we be using the term "stage" for work buckets?

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+```rust
+for object in objects {
+    tracer_context.with_tracer(worker, |tracer| { // This is expensive! DON'T DO THIS!
+        tracer.trace_object(object);
+    });
+}
+```
+
+Keep in mind that **tracer_context implements the `Clone` trait**.  As introduced in the *When to
+run finalizers* section, the VM binding can use work packets to parallelize finalizer processing.
```
k-sareen, line 167: Back link to section?

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+each work packet a clone of `tracer_context`.
+
+### WARNING: object resurrection
+
+If the VM binding retains an unreachable object for finalization, and the finalizer writes a
+reference of that object into a place readable by application threads, including global or static
+variable, then the previously unreachable object will become reachable by the application again.
+This phenomenon is known as **"resurrection"**, and can be surprising to the programmers.
+
+Developers of VM bindings of existing VMs may have no choice but implementing the finalizer
```
k-sareen, line 178:
```suggestion
Developers of VM bindings of existing VMs may have no choice but to implement the finalizer
```

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+semantics, we may choose to execute the clean-up operations immediately, or enqueue them to be
+executed after GC.  We may retain the unreachable referent if we need to.
+
+### Soft references
+
+Java has a special kind of weak reference: `SoftReference`.  The API allows the GC to choose between
+(1) retaining softly reachable referents, and (2) clearing references to softly reachable objects.
+When using MMTk, there are two ways to implement this semantics.
+
+The easiest way is **treating `SoftReference` as strong references in non-emergency GCs, and
```
k-sareen, line 242: Perhaps add TODO to glossary for "emergency GC"

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+binding to expand the transitive closure multiple times, each handling weak references at different
+levels of strength.
+
+Take Java as an example,  we may run `process_weak_refs` four times.
+
+1.  Visit all `SoftReference`.
+    -   If the referent has been reached, then
+        -   forward the referent field.
+    -   If the referent has not been reached, yet, then
+        -   if it is not emergency GC, then
```
k-sareen, line 279:
```suggestion
        -   if it is not an emergency GC, then
```

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+            -   retain the referent and update the referent field.
+        -   it it is emergency GC, then
```
k-sareen, line 281:
```suggestion
        -   it it is an emergency GC, then
```

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+            -   clear the referent field,
+            -   remove the `SoftReference` from the list of soft references, and
+            -   optionally enqueue it to the associated `ReferenceQueue` if it has one.
+    -   (This step may expand the transitive closure in emergency GC if any referents are retained.)
```
k-sareen, line 285:
```suggestion
    -   (This step may expand the transitive closure in an emergency GC if any referents are retained.)
```

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+[Ephemeron]: https://dl.acm.org/doi/10.1145/263700.263733
+
+To support ephemerons, the VM binding needs to identify ephemerons.  This includes ephemerons as
+individual objects, objects that contain ephemerons, and, equivalently, objects that contain
+key/value fields that have semantics similar to ephemerons.
+
+The following is the algorithm for processing ephemerons.  It gradually discovers ephemerons as we
+do the tracing.  We maintain a queue of ephemerons which is empty before the `Closure` stage.
+
+1.  In `Scanning::scan_object` and `Scanning::scan_object_and_trace_edges`, we enqueue ephemerons as
```
k-sareen, line 372: Have to phrase this a bit delicately since there are two queues here. The ephemeron queue and the scanning/tracing queue. I got tripped up by you saying "enqueuing ephemerons but not tracing them".

docs/userguide/src/portingguide/concerns/weakref.md
```diff
@@ -0,0 +1,477 @@
...
+    slots must be pointing to old objects (if not `null`).  For weak hash tables, if existing
+    entries are immutable, it is sufficient to only visit newly added entries.
+
+Implementation-wise, the VM binding can split the lists or hash tables into two parts: one for old
+entries and another for young entries.
+
+### Copying versus non-copying GC
+
+MMTk provides both copying and non-copying GC plans.  `MarkSweep` never moves any objects.
+`MarkCompact`, `SemiSpace` always moves all objects.  Immix-based plans sometimes do non-copying GC,
```
k-sareen, line 429: Ehhh. That's not true since LOS/non-moving objects are not moved. Should say normal objects there maybe?

src/util/address.rs
```diff
@@ -638,6 +638,37 @@ impl ObjectReference {
...
+    /// This method is primarily used during weak reference processing.  It can check if an object
+    /// (particularly finalizable objects and objects pointed by weak references) has been reached
+    /// by following strong references or weak references of higher strength.
+    ///
+    /// This method can also be used during tracing for debug purposes.
+    ///
+    /// When called at other times, particularly during mutator time, the behavior is specific to
+    /// the implementation of the plan and policy due to their strategies of metadata clean-up.  If
+    /// the VM needs to know if any given reference is still valid, it should instead use the valid
+    /// object bit (VO bit) metadata which is enabled by the Cargo feature "vo_bit".
```
k-sareen, line 653:
```suggestion
    /// object bit (VO-bit) metadata which is enabled by the Cargo feature "vo_bit".
```

src/vm/scanning.rs
```diff
@@ -282,64 +282,95 @@ pub trait Scanning<VM: VMBinding> {
...
+    ///
+    /// **Use `ObjectReference::get_forwarded_object()` to get the forwarded address of reachable
+    /// objects.  Only use `tracer.trace_object` for retaining unreachable objects.** If
+    /// `trace_object` is called on an already reached object, it will also return its new address
+    /// if moved. However, `tracer_context.with_tracer` has a cost, and the VM binding may
+    /// accidentally "resurrect" dead objects if failed to check `object.is_reachable()` first. If
+    /// the VM binding does not intend to retain any objects, it should completely avoid touching
+    /// `tracer_context`.
+    ///
+    /// **Clone the `tracer_context` for paralelism.**  The `ObjectTracerContext` has `Clone` as its
```
k-sareen, line 370:
```suggestion
    /// **Clone the `tracer_context` for parallelism.**  The `ObjectTracerContext` has `Clone` as its
```

k-sareen (commented):
LGTM. Thanks! The only main concern I have is to use more general language, i.e. "runtime"/"language runtime" over "VM".

docs/userguide/src/SUMMARY.md
wks: I agree.  But MMTk seems to use the term "VM" extensively, like "VM binding" instead of "runtime binding".  I think it may be a legacy from JikesRVM, where MMTk is just a component of the VM.

Personally, I think "runtime" is more accurate to describe the thing the Rust MMTk supports.  Not all languages that have GC use VM.  Go, RPython (not PyPy), GHC, ScalaNative and KotlinNative are a few examples.

I don't plan to change every occurrence of "VM" with "runtime" in the documentation now.  If we do, we should do the same for the code, too.

k-sareen: I mean the point was that we don't add more documentation that uses the VM terminology. But sure, I'm fine with it. I just thought it made sense to have the porting guide all use similar terminology.

docs/userguide/src/portingguide/concerns/weakref.md
wks: This is a bit subtle.  Some languages (Java for example) define the semantics of finalizers clearly.  For other languages or VMs, it is a bit more subtle.

For Ruby, finalizer is not part of the language.  There are two kinds of finalizers, both are VM details.
1.  `ObjectSpace.define_finalizer`: The entire `ObjectSpace` module is explicitly documented as implementation-specific and not portable.
2.  The `obj_free` function.  It is for fundamental and internal types in CRuby and its C extensions.  It is not part of the language, either.

By the way, the idiomatic way to close resources in Ruby is enclosing the resource in a block, for example `File.open(...) {|f| use(f) }`.

And even for the same language, different VM implementations may implement it slightly differently.  OpenJDK and JikesRVM are, for example, different.

And for Java, although the language spec defines finalization, the semantics of weak references is defined in neither the language spec nor the VM spec, but the API.  I still consider the API as part of the language, and the VM still needs to support the Java API anyway, including its weak reference semantics.

But I can rephrase this part.  I'll say different *languages and* VMs define finalizer differently.

docs/userguide/src/portingguide/concerns/weakref.md
wks: That's our current terminology.  We used to call it "phase" in JikesRVM.  Currently the type name is `WorkBucketStage` defined in `src/scheduler/work_bucket.rs`.

docs/userguide/src/portingguide/concerns/weakref.md
wks: Yes.  This "enqueue" refers to the ephemeron queue.

docs/userguide/src/portingguide/concerns/weakref.md
wks: Sure.  I'll make it clear objects in the LOS, ImmortalSpace, etc. don't move.

wks:
> LGTM. Thanks! The only main concern I have is to use more general language, i.e. "runtime"/"language runtime" over "VM".

@k-sareen  I have made changes according to your comments.

PR #1268, opened 2025-01-28
Clear stale line mark state

tianleq:
image
image
image
Overall, this fix does not incur significant overhead


qinsoon (approved):
LGTM. Thanks for the fix.

qinsoon:
The missing import needs to be fixed though.

tianleq:
> The missing import needs to be fixed though.

Yes, fixed

PR #1269, opened 2025-02-04
Add a function to print object info from the space for debugging

qinsoon:
This PR adds a non-mangled function to print object information for debugging. I also added a section in the doc about the function.


src/mmtk.rs
```diff
@@ -597,3 +597,13 @@ impl<VM: VMBinding> MMTK<VM> {
             .initialize_object_metadata(object, false)
     }
 }
+
+#[no_mangle]
+pub fn mmtk_debug_print_object_info(object: crate::util::ObjectReference) {
```
wks, line 602: I think this function is intended to be called from C or GDB.  In this case, the user may just supply an arbitrary number and want it to be used as an `ObjectReference`.  But `ObjectReference` is not `repr(C)`, and it has validity requirements (non-zero and word-aligned).  I think we can change this to `Address` and validate it inside this function.

```suggestion
pub fn mmtk_debug_print_object_info(object_addr: crate::util::Address) {
    let Some(object) = ObjectReference::from_raw_address(object_addr) else {
        println!("{object_addr} is not a valid object reference");
        return;
    };
```

src/mmtk.rs
```diff
@@ -597,3 +597,13 @@ impl<VM: VMBinding> MMTK<VM> {
...
+    println!(
+        "{}",
+        SFT_MAP
+            .get_checked(object.to_raw_address())
+            .debug_get_object_info(object)
+    )
```
wks, line 608: Instead of printing the result here, I suggest we let `debug_get_object_info` gradually print more and more information.  The reason is that when the user uses GDB or rr to debug a runtime, chance is high that the heap may have already been corrupted, and the metadata may have been inconsistent.  In this case, one of the steps inside `debug_get_object_info` may go wrong and panic before it returns a human-readable string.  Instead, if we let it gradually print more and more info, it can make the best effort even if it fails to finish normally

src/policy/copyspace.rs
```diff
@@ -100,6 +100,15 @@ impl<VM: VMBinding> SFT for CopySpace<VM> {
...
         self.trace_object(queue, object, self.common.copy, worker)
     }
+
+    fn debug_get_object_info(&self, object: ObjectReference) -> String {
+        format!(
+            "{}: {:?}, {:?}",
+            self.name(),
+            object_forwarding::debug_get_object_forwarding_info::<VM>(object),
+            crate::policy::sft::debug_get_object_global_info(object)
+        )
```
wks, line 110: The same is here.  We can put three `println!` statements here and print three pieces of information separately.

src/policy/sft.rs
```diff
@@ -101,6 +101,27 @@ pub trait SFT {
...
+}
+
+pub(crate) fn debug_get_object_global_info(_object: ObjectReference) -> String {
+    #[allow(unused_mut)]
+    let mut ret = String::new();
+    #[cfg(feature = "vo_bit")]
+    {
+        ret += &format!(
+            "vo bit = {}",
+            crate::util::metadata::vo_bit::is_vo_bit_set(_object)
```
wks, line 121: `is_vo_bit_set` doesn't check if the metadata is mapped or not.  We should call `is_mmtk_object` or equivalently `is_vo_bit_set_for_addr` instead.  And we should actually check it in `crate::mmtk::mmtk_debug_print_object_info` when converting the `Address` to `ObjectReference`.

src/util/object_forwarding.rs
```diff
@@ -233,3 +233,16 @@ pub(super) fn forwarding_bits_offset_in_forwarding_pointer<VM: VMBinding>() -> O
 pub(super) fn forwarding_bits_offset_in_forwarding_pointer<VM: VMBinding>() -> Option<isize> {
     unimplemented!()
 }
+
+pub(crate) fn debug_get_object_forwarding_info<VM: VMBinding>(object: ObjectReference) -> String {
+    let forwarding_bits = get_forwarding_status::<VM>(object);
```
wks, line 238: Again before actually checking the forwarding information, we should check if the metadata is mapped.  Some VMs (such as CRuby) are using side forwarding bits.  It is possible that `object` is invalid and its metadata is not mapped, yet.

src/mmtk.rs
wks: By the way, if the `vo_bit` feature is enabled, we should use the VO bit to convert the object to `ObjectReference`.  But if that fails, we may still "force" converting `object_addr` to an `ObjectReference` to provide some information with "best-effort".

wks (commented):
I have some high-level suggestions.

-   The argument of `mmtk_debug_print_object_info` should be an `Address` instead of `ObjectReference` because GDB and rr users may pass any number to that function.
-   Because the object may not be valid, we need to check if the side metadata are mapped before we attempt to read any side metadata.
-   Even with all the checking in place, it is still better to print one piece of information at a time using multiple `println!` statements.  If any step panics (or cause SIGSEGV if it reads unmapped metadata), the user can still see partial information.

qinsoon:
> The argument of mmtk_debug_print_object_info should be an Address instead of ObjectReference because GDB and rr users may pass any number to that function.

I don't agree with this. We do want the user to give us an object reference (even if it is represented as a `usize` in the debugger). The intention of this function is to print information for **an object** -- we don't want to access object metadata for an arbitrary address.

We can do our best guess in the function to rule out some possible issues, such as unmapped address/metadata, etc. But the function may crash if the heap/metadata is corrupted, or if the argument is invalid. After all this is just a utility function to help debugging.

wks:
> > The argument of mmtk_debug_print_object_info should be an Address instead of ObjectReference because GDB and rr users may pass any number to that function.
> 
> I don't agree with this. We do want the user to give us an object reference (even if it is represented as a `usize` in the debugger). The intention of this function is to print information for **an object** -- we don't want to access object metadata for an arbitrary address.
> 
> We can do our best guess in the function to rule out some possible issues, such as unmapped address/metadata, etc. But the function may crash if the heap/metadata is corrupted, or if the argument is invalid. After all this is just a utility function to help debugging.

I mean, the user SHOULD always pass an object reference.  But the user usually cannot verify if the address is valid.  They usually copy an address from a log message or a stack frame info, such as

    #3 0x00007ffff7e77090 in some_function(obj=1073741828) from /xxxx/xxx.so

The user probably don't use a calculator to check if 1073741828 is a multiple of 8 (in fact it is not) before invoking `p mmtk_debug_print_object_info(1073741828)`.  That's why the parameter type is `Address`.  `mmtk_debug_print_object_info` will verify it, and make it an `ObjectReference`.  If the parameter type is already `ObjectReference`, it will silently accept 1073741828 as the value of a `ObjectReference` and pass it to other places that assume it is aligned, and some weird error will occur.

qinsoon:
If the user passes an object reference, we print information for the object. If they don't pass an object reference, they get what they get, which may not make any sense at all.

Without VO bit, we cannot tell if an address is an object reference or not. We can make guesses and rule out some least possible cases, but we cannot tell definitely that the address is an object or not. In such a case, we can't access the object metadata in a meaningful way if we are not sure if the address is an object reference.

I think we can just require an object reference. It is the user's responsibility to pass us a valid object reference. However, we do some checks in case that the reference is obviously invalid.

src/mmtk.rs
qinsoon: Done. Now each function impl only fetches one piece of metadata at a time, and print one line at a time.

src/policy/sft.rs
qinsoon: I added a check to check if the object is mapped or not before forwarding to SFT. If the object is mapped, we can assume all the metadata can be accessed.

qinsoon: If the object is mapped, but some side metadata is not mapped, then it is a bug in MMTk core. We do not expect the function to work properly, and we do not expect the function can help us figure out the bug either.

src/mmtk.rs
```diff
@@ -597,3 +597,31 @@ impl<VM: VMBinding> MMTK<VM> {
...
+/// A non-mangled function to print object information for debugging purposes. This function can be directly
+/// called from a debugger.
+#[no_mangle]
+pub fn mmtk_debug_print_object_info(object: crate::util::ObjectReference) {
+    // If the address is unmapped, we cannot access its metadata. Just quit.
+    assert!(
+        object.to_raw_address().is_mapped(),
+        "{} is not mapped in MMTk",
+        object
+    );
```
wks, line 610: This is an assertion.  If this fails, it will panic.  It is right for normal execution, but this function is for calling in GDB.  For example,

```gdb
(gdb) p (int)abort()

Program received signal SIGABRT, Aborted.
0x00007ffff7e0174c in ?? () from /usr/lib/libc.so.6
The program being debugged was signaled while in a function called from GDB.
GDB remains in the frame where the signal was received.
To change this behavior use "set unwind-on-signal on".
Evaluation of the expression containing the function
(abort) will be abandoned.
When the function is done executing, GDB will silently stop.
(gdb) 
```

It is better to just use `println!` to print an error message, and return if it is impossible to proceed.

qinsoon: Done.

src/policy/copyspace.rs
```diff
@@ -100,6 +100,12 @@ impl<VM: VMBinding> SFT for CopySpace<VM> {
         let worker = worker.into_mut::<VM>();
         self.trace_object(queue, object, self.common.copy, worker)
     }
+
+    fn debug_print_object_info(&self, object: ObjectReference) {
+        println!("In {}", self.name());
+        object_forwarding::debug_print_object_forwarding_info::<VM>(object);
+        crate::policy::sft::debug_print_object_global_info(object);
```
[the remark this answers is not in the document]
qinsoon: I put `println!("In {}", self.name());` to the caller. I realized that we cannot put `debug_print_object_global_info` to the caller, as we need the `VM` parameter which we only have for each space. I moved the function to `CommonSpace`, and adds printing for log bits there.

qinsoon:
The above issues are fixed.

wks (approved):
LGTM

PR #1270, opened 2025-02-05
Fix julia extended tests

qinsoon:
This PR changes the extended testing workflow for Julia:
1. It now tests with `master` in the upstream Julia repo, and `master` in the binding repo by default.
2. In addition to specifying the binding version, we can also specify the Julia version to run with. This change is necessary, as we no longe record the Julia version in the binding (we record the binding version in Julia instead).


udesou (approved):
LGTM.

PR #1273, opened 2025-02-14
Add tracking functionality for memory management in allocators

playX18:
Added `track.rs` which provides simple utility functions to track memory allocated by MMTk using `crabgrind` crate. When `crabgrind` feature is not enabled all functions are no-ops.


qinsoon:
Thanks for the PR. I will look into this this week.

src/plan/global.rs
```diff
@@ -115,6 +115,8 @@ pub fn create_plan<VM: VMBinding>(
     plan.for_each_space(&mut |s| {
         sft_map.notify_space_creation(s.as_sft());
         s.initialize_sft(sft_map);
+        // after SFT is initialized, we can also initialize mempool tracking
+        s.get_page_resource().track();
```
qinsoon, line 119: I guess that you use the address of `PageResource` to register as a mempool, and the address is only fixed here. If this is the case, just make this explicit in the comments.

src/policy/largeobjectspace.rs
```diff
@@ -352,10 +353,7 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
         let sweep = |object: ObjectReference| {
             #[cfg(feature = "vo_bit")]
             crate::util::metadata::vo_bit::unset_vo_bit(object);
-            // Clear log bits for dead objects to prevent a new nursery object having the unlog bit set
-            if self.clear_log_bit_on_sweep {
-                VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.clear::<VM>(object, Ordering::SeqCst);
-            }
+            track_free(object.to_object_start::<VM>(), 0 /* TODO: Size */);
```
qinsoon, line 356: The original code was accidentally removed?

src/util/alloc/bumpallocator.rs
```diff
@@ -116,13 +117,16 @@ impl<VM: VMBinding> Allocator<VM> for BumpAllocator<VM> {
...
+            track_malloc(result, size, false);
             result
         }
     }
 
     fn alloc_slow_once(&mut self, size: usize, align: usize, offset: usize) -> Address {
         trace!("alloc_slow");
-        self.acquire_block(size, align, offset, false)
+        let block = self.acquire_block(size, align, offset, false);
+        track_malloc(block, size, false);
```
qinsoon, line 128: This doesn't look right: the change above marks each allocation/object as 'malloc', but this marks the entire thread local buffer as 'malloc'.

qinsoon: What happens if you mark overlapping memory? E.g. marking the buffer first, then marking objects in it.

src/util/heap/blockpageresource.rs
```diff
@@ -136,6 +147,7 @@ impl<VM: VMBinding, B: Region> BlockPageResource<VM, B> {
         self.block_queue.add_global_array(array);
         // Finish slow-allocation
         self.commit_pages(reserved_pages, required_pages, tls);
+        track_mempool_alloc(self, first_block, required_pages * BYTES_IN_PAGE);
```
qinsoon, line 150: This creates a chunk for valgrind that starts at `first_block` and have `required_pages`, which could be multiple blocks. But later `track_mempool_free` is called at block granularity, which means it frees a subset of what was allocated. Is this allowed for valgrind?

If not, we could just call 'track_mempool_alloc` for each block, and free each block separately.

src/util/heap/externalpageresource.rs
```diff
@@ -28,6 +28,10 @@ pub struct ExternalPages {
 }
 
 impl<VM: VMBinding> PageResource<VM> for ExternalPageResource<VM> {
+    fn track(&self) {
+        /* cannot track external pages reliably? */
```
qinsoon, line 32: Use `panic` or `warn!`.

src/util/heap/blockpageresource.rs
qinsoon: This also applies to other `PageResource` that allocates at page granularity.

src/util/heap/monotonepageresource.rs
```diff
@@ -149,7 +154,7 @@ impl<VM: VMBinding> PageResource<VM> for MonotonePageResource<VM> {
                 sync.current_chunk = chunk_align_down(sync.cursor);
             }
             self.commit_pages(reserved_pages, required_pages, tls);
-
+            track_mempool_alloc(self, rtn, conversions::pages_to_bytes(required_pages));
```
qinsoon, line 157: This is never freed. Pages can be freed when the page resource releases the pages.

src/util/metadata/vo_bit/helper.rs
```diff
@@ -184,6 +185,18 @@ pub(crate) fn on_object_forwarded<VM: VMBinding>(new_object: ObjectReference) {
 }
 
 pub(crate) fn on_region_swept<VM: VMBinding, R: Region>(region: &R, is_occupied: bool) {
+    if tracking_enabled() {
+        let mut cursor = region.start();
+        while cursor < region.end() {
+            if let Some(object) = vo_bit::is_vo_bit_set_for_addr(cursor) {
```
qinsoon, line 191: This will not work reliably, depending on how we maintain VO bits (see the code below it). This is the issue that I explained: MMTk as a GC usually does not know what objects are dead, and only knows what is alive. We may not be able to reliably tell what objects are freed, we can only tell valgrind when the page is empty, and let valgrind to automatically free those objects.

src/util/memory.rs
```diff
@@ -353,6 +353,21 @@ pub fn handle_mmap_error<VM: VMBinding>(
 /// This function is currently left empty for non-linux, and should be implemented in the future.
 /// As the function is only used for assertions, MMTk will still run even if we never panic.
 pub(crate) fn panic_if_unmapped(_start: Address, _size: usize, _anno: &MmapAnnotation) {
+    #[cfg(feature = "crabgrind")]
+    {
+        use crabgrind::memcheck::Error;
+        let result = crabgrind::memcheck::is_defined(_start.to_mut_ptr(), _size);
```
qinsoon, line 359: We may use this function for side metadata memory. Those memory is not annotated for valgrind. Will this work?

src/util/track.rs
```diff
@@ -0,0 +1,150 @@
...
+        crabgrind::run_mode() != crabgrind::RunMode::Native
+    }
+
+    #[cfg(not(feature = "crabgrind"))]
+    {
+        false
+    }
+}
+
+pub fn track_malloc(p: Address, size: usize, zero: bool) {
```
qinsoon, line 17: The arguments passed as `zero` from different call sites are a bit random:
* bump pointer: false
* freelist and large object: true

Each space defines whether the memory needs to be zeroed. You could get the value from there, or just use `true` in this function (assuming they will be zeroed).

src/util/track.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    #[cfg(not(feature = "crabgrind"))]
+    {
+        let _ = p;
+        let _ = size;
+        let _ = zero;
+    }
+}
+
+pub fn track_free(p: Address, size: usize) {
```
qinsoon, line 31: `size` can be removed. `crabgrind::memcheck::alloc::free` asks for `rz` (https://docs.rs/crabgrind/latest/crabgrind/memcheck/alloc/fn.free.html). Is it redzone size, rather than allocation size?

qinsoon (commented):
This PR annotates memory allocated by MMTk for tools like valgrind memcheck so valgrind can reason about MMTk's memory.

The following is how the PR maps MMTk's memory terms to valgrind's terms:
|MMTk|valgrind|
|-|-|
|`PageResource`|meta mem pool (with `VALGRIND_CREATE_MEMPOOL`[1])|
|allocated pages|superblocks/chunks (with `VALGRIND_MEMPOOL_ALLOC`[1])|
|objects|malloclike blocks (with `VALGRIND_MALLOCLIKE_BLOCK`[2])|

* [1] https://valgrind.org/docs/manual/mc-manual.html#mc-manual.mempools
* [2] https://valgrind.org/docs/manual/manual-core-adv.html#manual-core-adv.clientreq

This mapping looks reasonable. 

However, one obvious caveat is that, valgrind expects us to annotate freed objects (following the malloc/free pattern) while MMTk as a GC usually does not know dead objects. This means when an object is actually dead, we cannot tell valgrind promptly -- MMTk only tells valgrind after the entire region (whatever the MMTk policy defines) is empty, then valgrind automatically marks the internal objects in the region as free. But this is not a show stopper. I believe even with this limitation, valgrind can still be helpful.

PR #1276, opened 2025-02-18
Bump MSRV to 1.74.1

wks:
The current latest version of the "built" crate (v0.7.7) requires MSRV 1.74.  We bump the MSRV to 1.74.1.

Since version 0.7.6 of the "built" crate, it generates `static` items instead of `const` items for `PKG_VERSION`, `FEATURES_STR`, etc.  Our `build_info.rs` used to define `const` items that take their values.  After this change, the Rust compiler now interpret those lines as taking references of `static` items, which is unstable until Rust 1.83.  We instead replaced those `const` items in `build_info.rs` with `use` statements that create aliases of the items generated by "built".

Bumping MSRV to 1.74.1 also allows us to bump the version of the dependency "criterion" to 0.5 which also requires MSRV 1.74.  Previously, we locked the version of "criterion" to 0.4 due to its MSRV requirement.

We also updated all dependencies to their latest versions.  Among those changes, the "sysinfo" crate renamed several `new` methods to `nothing`.  We make changes accordingly.

We also use `usize::div_ceil` which was introduced in Rust 1.73.  This fixes a clippy warning.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,12 +25,12 @@ atomic-traits = "0.4.0"
 bytemuck = { version = "1.14.0", features = ["derive"] }
 cfg-if = "1.0"
 crossbeam = "0.8.1"
-delegate = "0.12.0"
-downcast-rs = "1.1.1"
+delegate = "0.13.2"
+downcast-rs = "2.0.1"
 enum-map = "2.4.2"
 env_logger = { version = "0.11.3", optional = true }
 is-terminal = "0.4.7"
-itertools = "0.12.0"
+itertools = "0.14.0"
 jemalloc-sys = { version = "0.5.3", features = ["disable_initial_exec_tls"], optional = true }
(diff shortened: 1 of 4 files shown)


wks:
The CI detected public API changes, including

1.  `mmtk::build_info::MMTK_FEATURES` and `mmtk::build_info::MMTK_PKG_VERSION` are changed from `const` to `static`.  This is the result of a change in "built" v0.7.6.  We now create aliases of the items "built" generated.
2.  The "downcast-rs" crate now generates implementations of the `DowncastSend` trait for types annotated by the `impl_downcast!` macro.

Both are expected.

wks:
> Was the consensus in the meeting to bump MSRV or pin dependency version? It wasn't clear to me what the resolution was.

I said pinning dependency version is a bad thing, and nobody objected.

PR #1277, opened 2025-02-26
Remove dead trace_object methods.

wks:
Now that we are using the `#[derive(HasSpaces, PlanTraceObject)]` derive macros to generate the trace_object methods, we can remove the manually-written dead code.


qinsoon (approved):

wks:
There is a new Clippy warning in the stable release.  We'll fix that separately.

PR #1283, opened 2025-03-17
Optimize object forwarding for single threaded GC

k-sareen:


wks:
1.74.1 failed to compile because icu4x started to require rust 1.81.  Let's fix that separately.

wks (commented):
Two high-level comments:

Firstly, we can skip the `BEING_FORWARDED` state without affecting the callers in CopySpace and ImmixSpace.  After calling `attempt_to_forward`, the caller always calls `state_is_forwarded_or_being_forwarded` to test if the return value is `FORWARDING_NOT_TRIGGERED_YET` or other states (`BEING_FORWARDED` or `FORWARDED`), but in fact it can never observe `BEING_FORWARDED` even we set the state to `BEING_FORWARDED` in `attempt_to_forward`.  That's because if it is `FORWARDING_NOT_TRIGGERED_YET`, the (only) GC worker will immediately try to forward the object, and it will either successfully copy the object and change the state to `FORWARDED`, or decide not to copy the object and revert the state to `FORWARDING_NOT_TRIGGERED_YET`. Then a subsequent invocation of `attempt_to_forward` will either observe `FORWARDED` or  `FORWARDING_NOT_TRIGGERED_YET`.  So we can simply omit the `store` in `attempt_to_forward`.

Secondly, I think the "early exit" style is more readable than the current `if-else` style.

k-sareen:
> Secondly, I think the "early exit" style is more readable than the current if-else style.

The problem is then I think the code for the multi-threaded object forwarding will still be in the generated binary leading to code bloat. I don't know if the Rust/LLVM compiler will remove that code if the `if` statement is always true. Might need to actually look at the assembly code to confirm this.

k-sareen:
`cfg!` doesn't remove code, only `#[cfg]` does: https://doc.rust-lang.org/rust-by-example/attribute/cfg.html

caizixian:
The main difference is that code under `cfg!` needs to be syntactically and semantically correct (no missing imports, type checks, etc.) even if the condition always evaluates to false. The dead code elimination is just an optimization.

While `[cfg]` actually removes the code before compilation.

PR #1284, opened 2025-03-17
Fixing MSRV-breaking dependencies

wks:
Some dependencies started to require MSRV above our current MSRV.

The build dependency `built` transitively depends on some crates from the ICU4X project that recently started to depend on Rust 1.81, such as `litemap`.  The dependency on ICU4X is completely unnecessary, and we removed it from our dependency tree by forcing the use of a particular version of `idna_adapter`.  See: https://docs.rs/crate/idna_adapter/1.2.0

The dev dependency `criterion` depends on `ciborium` which depends on the `half` crate.  Since v2.5.0, `half` started to depend on Rust 1.81.  `ciborium` needs to be fixed because its MSRV is 1.58 and shouldn't depend on a crate that requires Rust 1.81.  We lock the version of the `half` to 2.4.1.  This kind of problem should be properly addressed with the new MSRV-aware dependency resolver introduced in Rust 1.84.  See: https://doc.rust-lang.org/cargo/reference/resolver.html#rust-version

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,13 +57,13 @@ paste = "1.0.8"
 rand = "0.9.0"
 rand_chacha = "0.9.0"
 criterion = "0.5"
-# Note: `half` v2.5.0 started to require Rust 1.8.1.
+# Note: `half` v2.5.0 started to require Rust 1.81.
 # `criterion` transitively depends on `half` via the dependency chain:
 #     criterion -> ciborium -> ciborium_ll -> half
 # The MSRV of `ciborium` v0.2.2 is declared as 1.58, and is declared to depend on `half` v2.2.
-# The Ciborium project should either bump its MSRV or select a version of `half` that satisfies its current MSRV.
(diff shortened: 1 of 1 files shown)


wks:
@k-sareen This is related to the failure of the PR you created today.  See https://github.com/mmtk/mmtk-core/actions/runs/13888780197/job/38857183528?pr=1283

k-sareen (approved):
LGTM. Thanks! PS: do you know why the URL resolving is failing in the docs? HTTP 403 is forbidden error code. I would have expected HTTP 429 if we were being rate limited.

wks:
> LGTM. Thanks! PS: do you know why the URL resolving is failing in the docs? HTTP 403 is forbidden error code. I would have expected HTTP 429 if we were being rate limited.

I think ACM's web site is frustrated by so many requests sent from robots, and it blocked the GitHub CI.  Maybe we can skip "*.acm.org" when doing link checks.

k-sareen:
Makes sense 👍

PR #1285, opened 2025-03-18
Make work packet buffer size configurable from one location

### LINKED ISSUE #1281: Make work packet sizes easily configurable
You'd think that changing the work packet size [here](https://github.com/mmtk/mmtk-core/blob/4ca8812607bfb0b398278e01cbd3755959c7d010/src/scheduler/gc_work.rs#L559) would be enough, but actually we need to change the size in [VectorQueue](https://github.com/mmtk/mmtk-core/blob/4ca8812607bfb0b398278e01cbd3755959c7d010/src/plan/tracing.rs#L28) as well as in [the binding code](https://github.com/mmtk/mmtk-openjdk/blob/1abbc89254a06e9b35750faf0316dc3ff2059327/mmtk/src/scanning.rs#L15) if it has hardcoded the value. This needs to be fixed with a global constant.

See [here](https://mmtk.zulipchat.com/#narrow/channel/262673-mmtk-core/topic/Changing.20work.20packet.20size.20requires.20changing.20.3E.202.20locations) for more discussion

k-sareen:
Closes #1281

### DIFF
--- a/src/scheduler/gc_work.rs
+++ b/src/scheduler/gc_work.rs
@@ -10,6 +10,9 @@ use crate::*;
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 
+/// Buffer size for [`ProcessEdgesWork`] work packets.
+pub const EDGES_WORK_BUFFER_SIZE: usize = 4096;
+
 pub struct ScheduleCollection;
 
 impl<VM: VMBinding> GCWork<VM> for ScheduleCollection {
@@ -556,7 +559,7 @@ pub trait ProcessEdgesWork:
     /// Higher capacity means the packet will take longer to finish, and may lead to
     /// bad load balancing. On the other hand, lower capacity would lead to higher cost
     /// on scheduling many small work packets. It is important to find a proper capacity.
-    const CAPACITY: usize = util::constants::BUFFER_SIZE;
+    const CAPACITY: usize = EDGES_WORK_BUFFER_SIZE;
     /// Do we update object reference? This has to be true for a moving GC.
     const OVERWRITE_REFERENCE: bool = true;
     /// If true, we do object scanning in this work packet with the same worker without scheduling overhead.
--- a/src/scheduler/gc_work.rs
+++ b/src/scheduler/gc_work.rs
@@ -556,7 +556,7 @@ pub trait ProcessEdgesWork:
     /// Higher capacity means the packet will take longer to finish, and may lead to
     /// bad load balancing. On the other hand, lower capacity would lead to higher cost
     /// on scheduling many small work packets. It is important to find a proper capacity.
-    const CAPACITY: usize = 4096;
+    const CAPACITY: usize = util::constants::BUFFER_SIZE;
     /// Do we update object reference? This has to be true for a moving GC.
     const OVERWRITE_REFERENCE: bool = true;
     /// If true, we do object scanning in this work packet with the same worker without scheduling overhead.
--- a/src/scheduler/gc_work.rs
+++ b/src/scheduler/gc_work.rs
@@ -10,10 +10,12 @@ use crate::*;
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 
-/// Buffer size for [`ProcessEdgesWork`] work packets. This constant is exposed to users so that
-/// they can use this value for places in their binding that interface with the work packet system,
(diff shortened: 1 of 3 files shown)


src/util/constants.rs
```diff
@@ -22,6 +22,9 @@ pub const LOG_BYTES_IN_KBYTE: u8 = 10;
 /// The number of bytes in a kilobyte
 pub const BYTES_IN_KBYTE: usize = 1 << LOG_BYTES_IN_KBYTE;
 
+/// Work packet buffer size
+pub const BUFFER_SIZE: usize = 4096;
```
wks, line 26: Since this is part of the public API, we should be more careful about naming.

This constant is a bit different from others in the `util::constants` module, such as `{LOG_,}BYTES_IN_{K,M,G}BYTE`.  `BUFFER_SIZE` is related to the GC work packet design, and we chose 4096 arbitrarily (and may not be the optimal value).  On the contrary, others are mathematical constants.

And the name "buffer" alone does not make it clear what kind of buffer it is referring to.

I suggest we add a `pub mod gc_work` and split `BUFFER_SIZE` into two constants:

-   `NODES_PACKET_SIZE`: the number of nodes in a ScanObjects packet, and
-   `EDGES_PACKET_SIZE`: the number of edges (represented as `Slot` now, but may be represented as `ObjectReference` if we think appropriate for MarkSweep) in a `ProcessEdgesWork` work packet.

We can define both as 4096 so we don't change the semantics, but we may, in the future, make them different.

k-sareen: I agree with the naming, but I'm not so sure if we want to split the constant into two, at least in this PR.

k-sareen: > (represented as Slot now, but may be represented as ObjectReference if we think appropriate for MarkSweep)

I don't know what you mean here. An edge is a slot, no? Regardless of the GC algorithm used.

wks: > > (represented as Slot now, but may be represented as ObjectReference if we think appropriate for MarkSweep)
> 
> I don't know what you mean here. An edge is a slot, no? Regardless of the GC algorithm used.

You can find "Node-ObjRef", "Edge-ObjRef" and "Edge-Slot" in [Angus' paper](https://www.steveblackburn.org/pubs/theses/atkinson-2023.pdf)).  In that paper, node and edge refer to the queuing strategy, while objref and slot refer to the thing it enqueues.  For MarkSweep, since we don't need to update any slot, we can load from the slots immediately while we scan the object, and enqueue the referents directly.

In the current mmtk-core, `Slot` represents a slot, that is, something that holds a reference, and can be updated.  But "edge" is more about the queuing strategy.  The `ProcessEdgesWork` work packet does process edges, in the sense that (1) it provides the `trace_object` method which in theory traces an edge in the object graph and returns how the edge should be updated, and (2) its elements (slots) are created when scanning objects without looking at the child objects.  In MarkSweep, we can in theory replace `ProcessEdgesBase::slots` with a `Vec<ObjectReference>` (i.e. the values in the slots).  We can also replace it with a `Vec<(Slot, ObjectReference)>` (slots and their values) which is the "Edge-Tuple" strategy in Angus' paper.

qinsoon: I feel it doesn't matter too much whether we split it into two constants. At least for now, we do not have plans to have different values for edges vs nodes. 

The document for the constant should include why we expose this constant to the user, and how users should use this constant. 

> I suggest we add a pub mod gc_work

The constant may be put to the `scheduler` module. We tend to expose constants from their own modules now, rather than putting to `util::constants`.

k-sareen: Done

src/scheduler/gc_work.rs
```diff
@@ -10,6 +10,12 @@ use crate::*;
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 
+/// Buffer size for [`ProcessEdgesWork`] work packets. This constant is exposed to users so that
+/// they can use this value for places in their binding that interface with the work packet system,
+/// specifically the transitive closure via `ProcessEdgesWork` work packets such as roots gathering
+/// code or weak reference processing.
```
qinsoon, line 16: I think the key statement is missing: we expect users to split their workload by this buffer size for process-edges-related work, so each work packet should include at most this many items in the buffer.

k-sareen: Done.

qinsoon (approved):
LGTM

PR #1287, opened 2025-03-20
implement mutator builder

tianleq:
https://mmtk.zulipchat.com/#narrow/channel/262673-mmtk-core/topic/create.20plan.2Fmutator/near/504996362
Introducing `MutatorBuilder` to simplify customising `Mutator`, adding/removing fields of a certain plan's `Mutator` no longer affects other plan as `MutatorBuilder` provides default values for `Mutator`

### DIFF
--- a/src/plan/mutator_context.rs
+++ b/src/plan/mutator_context.rs
@@ -12,6 +12,8 @@ use crate::vm::VMBinding;
 
 use enum_map::EnumMap;
 
+use super::barriers::NoBarrier;
+
 pub(crate) type SpaceMapping<VM> = Vec<(AllocatorSelector, &'static dyn Space<VM>)>;
 
 /// A place-holder implementation for `MutatorConfig::prepare_func` that should not be called.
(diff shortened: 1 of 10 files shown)


wks:
@tianleq Could you edit this post https://github.com/mmtk/mmtk-core/pull/1287#issue-2933753721 to include introductions of the purpose of this commit?  In our current CI, the content of the merge request will become the commit message of the squashed commit.

src/plan/mutator_context.rs
```diff
@@ -78,6 +80,48 @@ impl<VM: VMBinding> std::fmt::Debug for MutatorConfig<VM> {
     }
 }
 
+/// Used to build a mutator struct
+pub struct MutatorBuilder<VM: VMBinding> {
+    allocators: Allocators<VM>,
```
qinsoon, line 85: I think we can do this:
```rust
pub struct MutatorBuilder<VM: VMBinding> {
    barrier: Box<dyn Barrier<VM>>,
    /// The mutator thread that is bound with this Mutator struct.
    mutator_tls: VMMutatorThread,
    mmtk: &'static MMTK<VM>,
    config: MutatorConfig<VM>,
}
```

The constructing of `Allocators` will be done when we call `build()`.
```rust
    pub fn build(self) -> Mutator<VM> {
        Mutator {
            allocators: Allocators::<VM>::new(self.mutator_tls, self.mmtk, &self.config.space_mapping),
            barrier: self.barrier,
            mutator_tls: self.mutator_tls,
            plan: self.mmtk.get_plan(),
            config: self.config,
        }
    }
```

qinsoon (commented):
LGTM. We can create `Allocators` when we build `Mutator` (instead of `MutatorBuilder`).

qinsoon (approved):

PR #1288, opened 2025-03-20
Fix to bytemuck_derive 1.8.1

qinsoon:
`bytemuck` uses `bytemuck_derive ^1.4.1`. Recent `bytemuck_derive` versions are not compatible with our MSRV 1.74.1 (1.9.1 requires Rust 1.84+, 1.9.0 requires edition 2024 which is stablized in 1.84). So fix `bytemuck_derive` to the last version before 1.9.0.


wks (approved):
LGTM

PR #1289, opened 2025-03-24
Check if an object is in our heap before using VM map during counting live bytes

qinsoon:
This PR fixes a bug in `increase_live_bytes`. If an object in the VM space is scanned, the method will be called with an object reference into the VM space, which may not be in our heap range. Then we get a space descriptor from the VM map, and we will see panics in the following code.
https://github.com/mmtk/mmtk-core/blob/129362d4f39a1191562d276c42f5b03e52aa287a/src/util/heap/layout/map64.rs#L209
https://github.com/mmtk/mmtk-core/blob/129362d4f39a1191562d276c42f5b03e52aa287a/src/util/heap/layout/map32.rs#L258

This PR works around the problem by checking if the object is in our heap range first.

### DIFF
--- a/src/scheduler/worker.rs
+++ b/src/scheduler/worker.rs
@@ -69,17 +69,20 @@ impl<VM: VMBinding> GCWorkerShared<VM> {
 
         // The live bytes of the object
         let bytes = VM::VMObjectModel::get_current_size(object);
-        // Get the space index from descriptor
-        let space_descriptor = VM_MAP.get_descriptor_for_address(object.to_raw_address());
-        let space_index = space_descriptor.get_index();
-        debug_assert!(
-            space_index < MAX_SPACES,
-            "Space index {} is not in the range of [0, {})",
-            space_index,
-            MAX_SPACES
-        );
-        // Accumulate the live bytes for the index
-        live_bytes_per_space[space_index] += bytes;
+        // Check if the object is in our heap. If it is not in our heap (e.g. in the VM space), we cannot use VM map to get its descriptor.
+        if crate::util::heap::vm_layout::vm_layout().in_available_range(object.to_raw_address()) {
+            // Get the space index from descriptor
+            let space_descriptor = VM_MAP.get_descriptor_for_address(object.to_raw_address());
+            let space_index = space_descriptor.get_index();
+            debug_assert!(
+                space_index < MAX_SPACES,
+                "Space index {} is not in the range of [0, {})",
+                space_index,
+                MAX_SPACES
+            );
+            // Accumulate the live bytes for the index
+            live_bytes_per_space[space_index] += bytes;
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


wks:
Even if the address is in the available range, its space index (Map64) or chunk index (Map32) may still map to a `descriptor_map` entry that is `SpaceDescriptor::UNINITIALIZED`.  It just happened that `increase_live_bytes` always calls `VMMap::get_descriptor_for_address` with the raw address of valid `ObjectReference` values.

In my opinion, `VMMap::get_descriptor_for_address(address: Address)` should return `Option<SpaceDescriptor>`  because the argument `address` is never required to be a valid object address, or an address within a space.  And for Map32, some chunks in the heap range may well be un-occupied by any space, and it is natural for `get_descriptor_for_address` to return `None`.  Then in `increase_live_bytes`, we just skip objects that map to `SpaceDescriptor::UNINITIALIZED`. 

Note: If the performance of `Option<SpaceDescriptor>` is an issue, we can redefine `SpaceDescriptor` as `NonZeroUsize` because it can never be zero.  If we don't want to do this refactoring, we can let `get_descriptor_for_address` return `SpaceDescriptor::UNINITIALIZED` if the argument `address` is not in the heap range.  But I strongly recommend using `Option<SpaceDescriptor>`.

qinsoon:
I changed `get_descriptor_for_address`. It will return `SpaceDescriptor::UNINITIALIZED` if the address is not in the heap range. This change should have minimal impact on performance. In the current code, we do `unwrap` for 64 bits, and we have bounds check for 32 bits. This change just makes those explicit and returns an alternative value.

Using `Option<SpaceDescriptor>` is fine. But that requires larger changes, such as changing the related tables to `Option<SpaceDescriptor>`. Otherwise, the function may return `None` or `Some(SpaceDescriptor::UNINITIALIZED)` which basically means the same thing.

k-sareen:
FYI I wanted to mention that @steveblackburn wanted us to have the same mechanism for address lookup for all our spaces, i.e. VM space should not be special cased. We discussed this in a recent group meeting when I went through the issues I was facing in ART. So arguably this is a bit of a hack still.

wks (approved):
LGTM

wks:
> FYI I wanted to mention that @steveblackburn wanted us to have the same mechanism for address lookup for all our spaces, i.e. VM space should not be special cased. We discussed this in a recent group meeting when I went through the issues I was facing in ART. So arguably this is a bit of a hack still.

It is nice to have a unified mechanism to get the space index for any given address (or `None` if not in any space).  But the current `SpaceDescriptor` mechanism has its restrictions.  For example, on Map32, spaces occupy whole chunks; on Map64, spaces occupy contiguous regions of log size `log_space_extent`.  It should be possible to extend `SpaceDescriptor` to VM spaces if VM spaces also follow such restrictions, but I doubt if this is true.  Alternatively, we can implement it like this:  We first look up the `descriptor_map` and, if it returns `None` or `UNINITIALIZED`, we fall back to looking up VM spaces by searching all registered VM space ranges.

PR #1291, opened 2025-03-25
Make `EDGES_WORK_BUFFER_SIZE` constant public to bindings

k-sareen:


src/scheduler/mod.rs
```diff
@@ -1,5 +1,13 @@
 //! A general scheduler implementation. MMTk uses it to schedule GC-related work.
 
+/// Buffer size for [`ProcessEdgesWork`] work packets. This constant is exposed to binding
+/// developers so that they can use this value for places in their binding that interface with the
+/// work packet system, specifically the transitive closure via `ProcessEdgesWork` work packets
+/// such as roots gathering code or weak reference processing. In order to have better load
+/// balancing, it is recommended that binding developers use this constant to split work up into
+/// different work packets.
+pub const EDGES_WORK_BUFFER_SIZE: usize = 4096;
```
qinsoon, line 9: You can just re-export it. See the `pub use` below.

k-sareen: I thought it'd be better to actually have all constants together in `mod.rs` like we do for many other cases like in `immixspace/mod.rs` for example.

qinsoon: The title suggested that you just wanted to make the constant public to bindings. I was suggesting `pub use` is the straightforward way to achieve that.

k-sareen: Right. But I thought it made sense to move the constant so that if we ever have other scheduler constants, they are not scattered around the different files. Up to you. I can move it back and re-export

qinsoon: It looks okay. I think the difference is insigifnicant whether we put the constants in the top-level modules or lower-level submodules. There is no convention to require either way so it is your choice.

qinsoon (approved):

PR #1292, opened 2025-03-25
Aggregate live bytes info in on_gc_finished

qinsoon:
This PR moves the call to `aggregate_live_bytes_in_last_gc` from the `Release` work packet to `on_gc_finished`. Spaces may do release work in parallel with the `Release` work packet, and `aggregate_live_bytes_in_last_gc` accesses reserved pages of spaces. This leads to a data race. We observed the reserved pages in the collected live bytes stats to be larger than the actual reserved pages at the end of a GC, as some pages were not released yet.

### DIFF
--- a/src/scheduler/scheduler.rs
+++ b/src/scheduler/scheduler.rs
@@ -559,10 +559,10 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
                 .scheduler
                 .worker_group
                 .get_and_clear_worker_live_bytes();
-            let mut stats = mmtk.state.live_bytes_in_last_gc.borrow_mut();
-            *stats = mmtk.aggregate_live_bytes_in_last_gc(live_bytes);
+            let mut live_bytes_in_last_gc = mmtk.state.live_bytes_in_last_gc.borrow_mut();
+            *live_bytes_in_last_gc = mmtk.aggregate_live_bytes_in_last_gc(live_bytes);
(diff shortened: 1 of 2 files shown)


src/scheduler/scheduler.rs
```diff
@@ -554,7 +554,15 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
-            for (space_name, &stats) in mmtk.state.live_bytes_in_last_gc.borrow().iter() {
+            // Aggregate the live bytes
+            let live_bytes = mmtk
+                .scheduler
+                .worker_group
+                .get_and_clear_worker_live_bytes();
+            let mut stats = mmtk.state.live_bytes_in_last_gc.borrow_mut();
+            *stats = mmtk.aggregate_live_bytes_in_last_gc(live_bytes);
+            // Logging
+            for (space_name, &stats) in stats.iter() {
```
wks, line 565: There are two different variables with the same name `stats`, one defined on line 562 by the `let mut`, and the other on line 565 by the `&stats` in the `for` loop.  It is a bit confusing.  I suggest renaming the former to `live_bytes_in_last_gc`, `map` or `guard`, whatever makes sense.

qinsoon: Thanks for spotting this.

wks (approved):
LGTM

Issue #493, opened 2021-11-02
Stat log buffer overflow


wenyuzhao:
https://github.com/mmtk/mmtk-core/blob/533d19f7a8b59978ad6c92f2c46e12881ded8083/src/util/statistics/stats.rs#L14

When the log buffer overflow happens, instead of simply presenting a warning, we should either panic directly or log extra message to break the plotty parser.

PR #1295, opened 2025-03-31
Fix broken stats gathering code if no. of GCs > `MAX_PHASES`

### LINKED ISSUE #1074: Stats: number of GC stuck at 2047
When using `harness_begin` and `harness_end` to enable statistics collection, the number of GC will be stuck at 2047 if GC is triggered more than that number.  It will print an error message in STDERR:

```
Warning: number of GC phases exceeds MAX_PHASES
```

The message is printed `Stats::start_gc`:

```rust
    pub fn start_gc(&self) {
        self.gc_count.fetch_add(1, Ordering::SeqCst);
        if !self.get_gathering_stats() {
            return;
        }
        if self.get_phase() < MAX_PHASES - 1 {
            let counters = self.counters.lock().unwrap();
            for counter in &(*counters) {
                counter.lock().unwrap().phase_change(self.get_phase());
            }
            self.shared.increment_phase();
        } else if !self.exceeded_phase_limit.load(Ordering::SeqCst) {
            eprintln!("Warning: number of GC phases exceeds MAX_PHASES");
            self.exceeded_phase_limit.store(true, Ordering::SeqCst);
        }
    }
```

And the number of GC printed in `harness_end` is actually `self.get_phase() / 2`, and it comes from the following function:

```
    pub fn print_stats<VM: VMBinding>(&self, mmtk: &'static MMTK<VM>) {
        println!(
            "============================ MMTk Statistics Totals ============================"
        );
        let scheduler_stat = mmtk.scheduler.statistics();
        self.print_column_names(&scheduler_stat);  // This prints "GC  time.other   time.stw  ....."
        print!("{}\t", self.get_phase() / 2); // This is the number under "GC"
        let counter = self.counters.lock().unwrap();
        // ...
```

Because the phase no longer increments when reached `MAX_PHASES` which is defined as `1 << 12` which is 4096, the number of GC is printed is 2047.

When running `lusearch` from DaCapo Chopin, if the plan is `GenCopy` or `GenImmix` and the heap size is set at 2.5x min heap size, the number of GC will exceed 2047.  This may happen to other benchmarks, too.  For example, if this happens when running `xalan`, the error message `Warning: number of GC phases exceeds MAX_PHASES` printed in STDERR will cause the benchmark validation to fail.

There is a comment near the `MAX_PHASES` constant:

```
// TODO: Increasing this number would cause JikesRVM die at boot time. I don't really know why.
// E.g. using 1 << 14 will cause JikesRVM segfault at boot time.
pub const MAX_PHASES: usize = 1 << 12;
```

In each `EventCounter`, there is an embedded array of `MAX_PHASES` elements:

```rust
pub struct EventCounter {
    name: String,
    pub implicitly_start: bool,
    merge_phases: bool,
    count: Box<[u64; MAX_PHASES]>, // This one
    current_count: u64,
    running: bool,
    stats: Arc<SharedStats>,
}
```

It looks like it intends to record the number for each individual STW and other phase.  This works fine if the benchmark never triggers more than 2047 GCs.  But actually it often does, especially with smaller heap and generational plan.  After the number of phases exceeded the threshold, subsequent counter values (such as `time.stw` and `time.other`) are ignored.  This explains some of the anomalies in https://github.com/mmtk/mmtk-core/pull/1062.  From the un-normalized plots, we can see the number of GC for GenCopy and GenImmix for `lusearch` is stuck at 2047.  This means the `time.other` and `time.stw` for subsequent GCs are ignored.  This explains the hard-to-explain increment and decrement in mutator time when run on different machines.

# What should we do?

We never look at the STW time and "other" time for individual phases.  We use the statistics counters mainly for printing the "GC", "time.other", "time.stw" when calling `harness_end`.  That only needs to record the accumulated time for all "STW" phases and all "other" phases.  It is more practical to only keep two numbers for each counter: the aggregated value for all odd phases and the aggregated value for all even phases.

If we want to inspect each phase (e.g. find the number of work packets executed), we have a better tool now.  We can use the eBPF tracing tool for that.

k-sareen:
Closes #1074 and #493.

### DIFF
--- a/src/util/statistics/counter/event_counter.rs
+++ b/src/util/statistics/counter/event_counter.rs
@@ -1,5 +1,5 @@
 use super::*;
-use crate::util::statistics::stats::{SharedStats, MAX_PHASES};
+use crate::util::statistics::stats::{SharedStats, DEFAULT_NUM_PHASES};
 use std::sync::Arc;
 
 /**
@@ -10,7 +10,7 @@ pub struct EventCounter {
     name: String,
     pub implicitly_start: bool,
     merge_phases: bool,
-    count: Box<[u64; MAX_PHASES]>,
+    count: Vec<u64>,
     current_count: u64,
     running: bool,
     stats: Arc<SharedStats>,
@@ -27,7 +27,7 @@ impl EventCounter {
             name,
             implicitly_start,
             merge_phases,
-            count: Box::new([0; MAX_PHASES]),
+            count: Vec::with_capacity(DEFAULT_NUM_PHASES),
             current_count: 0,
             running: false,
             stats,
@@ -76,13 +76,15 @@ impl Counter for EventCounter {
             return;
         }
         debug_assert!(self.running);
-        self.count[self.stats.get_phase()] = self.current_count;
+        self.count.push(self.current_count);
+        debug_assert_eq!(self.count[self.stats.get_phase()], self.current_count);
         self.running = false;
     }
 
     fn phase_change(&mut self, old_phase: usize) {
         if self.running {
-            self.count[old_phase] = self.current_count;
+            self.count.push(self.current_count);
+            debug_assert_eq!(self.count[old_phase], self.current_count);
             self.current_count = 0;
         }
     }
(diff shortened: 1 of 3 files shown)


wks (approved):
LGTM.

wks:
The V8 binding tests failed due to some problems in the CI script.  I am fixing it separately. https://github.com/mmtk/mmtk-v8/pull/90

k-sareen:
Wait to run V8 tests or merge @wks?

wks:
> Wait to run V8 tests or merge @wks?

Just merge it.  The V8 binding only uses NoGC at this time, so there is nothing for statistics.

PR #1296, opened 2025-04-07
Fix incongruent defaults for copying GCs

k-sareen:
We change the max non-LOS allocation size to 8 KB.

### DIFF
--- a/src/plan/plan_constraints.rs
+++ b/src/plan/plan_constraints.rs
@@ -56,7 +56,7 @@ impl PlanConstraints {
             max_non_los_default_alloc_bytes: MAX_INT,
             max_non_los_copy_bytes: MAX_INT,
             // As `LAZY_SWEEP` is true, needs_linear_scan is true for all the plans. This is strange.
-            // https://github.com/mmtk/mmtk-core/issues/1027 trackes the issue.
+            // https://github.com/mmtk/mmtk-core/issues/1027 tracks the issue.
             needs_linear_scan: crate::util::constants::SUPPORT_CARD_SCANNING
                 || crate::util::constants::LAZY_SWEEP,
             needs_concurrent_workers: false,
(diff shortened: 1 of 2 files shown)


k-sareen:
See [#mmtk-core > SemiSpace/MarkCompact have incongruent defaults @ 💬](https://mmtk.zulipchat.com/#narrow/channel/262673-mmtk-core/topic/SemiSpace.2FMarkCompact.20have.20incongruent.20defaults/near/498647959)

qinsoon:
The Block size is the minimal thread local buffer we get each time.
The max non-LOS allocation size is the max object size that is allowed in a non-LOS space.

Are you saying two constants are somehow related?

wks:
Could you post some performance data, such as execution time and/or wasted spaces?  Since this PR is changing the default values, we should have evidence to support that it solves the problems in some pathological cases.

k-sareen:
Wasted space for lusearch ~2500 SemiSpace GCs:
image

I'll run an experiment later today.

qinsoon:
> Could you post some performance data, such as execution time and/or wasted spaces? Since this PR is changing the default values, we should have evidence to support that it solves the problems in some pathological cases.

We probably also need to see if this changes performance. I am not sure if we use those values in the bindings (especially hard-coded values), as the binding may need to check against the value to decide which allocator to use..

k-sareen:
Just had a meeting with @steveblackburn and he mentioned that LOS threshold should be in the order of magnitude of 8 KB for copying plans. Immix is supposed to be special since it operates on blocks. I'll run an experiment with multiple thresholds.

k-sareen:
~10% reduction in total time
image

~20% reduction in # of GCs
image

Up to 8% reduction in mutator time for benchmarks with a lot of large object allocations
image

~35% reduction in STW time
image

~15% reduction in STW time per-GC
image

This was ran with SemiSpace with 1.5x SemiSpace minheap. I have also ran 1.25x (much better results because more GCs) and will also run 2.0x.

It's very safe to say that this is a good idea.

tianleq (approved):
LGTM

PR #1298, opened 2025-04-08
Allow objects to span multiple contiguous blocks for bump allocator

k-sareen:


wks (approved):
LGTM

k-sareen:
Did you want me to run a perf analysis of this or no?

k-sareen:
It's not necessarily a 100% because it could be a new chunk. But yes, close to 100% for single-threaded GC (or mutator for that matter). I just meant since it's a change to the slow-path, to evaluate the perf implications of it.

qinsoon:
This is just an optimization/fix for a pathological case where there is only one allocating thread and the object frequently cannot fit into the rest of the current allocation buffer. When you have multiple threads but only one is actively allocating, we may also see an improvement.

The improvement may not be from allocation throughput, as with the patch, we just save some slow path allocations. But we could get improvement from better memory utilization.

k-sareen:
Well yes. I have measured improvements for lusearch with single-threaded GC. But the above was the multi-threaded GC case.

PR #1299, opened 2025-04-11
Fix contiguous chunk iterator in monotonic PR

wks:
Fixed a bug where the iterator defined in `MonotonicPageResource::iterate_allocated_regions` erronious thinks the cursor is within the current contiguous region, while the cursor is actually at the end of the previous chunk.  This bug will cause side metadata (such as VO-bits) to be left over after releasing a CopySpace.


k-sareen (approved):
LGTM. Thanks for catching this!

PR #1301, opened 2025-04-14
Properly handle multiple plans in GITHUB-CI in CI scripts

qinsoon:
The PR fixes an issue in our CI script. We sometimes want to run a test with multiple plans, such as `GITHUB-CI: MMTK_PLAN=NoGC,SemiSpace`. The CI script does not parse `MMTK_PLAN`, and It instead just set the entire string as the env var (`MMTK_PLAN=NoGC,SemiSpace`). This results in an invalid value for `MMTK_PLAN`, and the default plan is used.


wks (approved):
LGTM

PR #1305, opened 2025-04-17
Allow configuring each Immix space to be non moving

qinsoon:
We plan to use non-moving Immix as a non-moving space. In this case, we will have two Immix spaces, one allows moving, and the other does not. We currently use global constants to control whether Immix moves objects or not, and we cannot have two Immix spaces that are configured differently.

This PR removes those constants, and adds a flag `never_move_objects` in `ImmixSpaceArgs` to indicate whether this Immix space disallows moving.

### DIFF
--- a/src/plan/immix/global.rs
+++ b/src/plan/immix/global.rs
@@ -140,7 +140,7 @@ impl<VM: VMBinding> Immix<VM> {
                 unlog_object_when_traced: false,
                 #[cfg(feature = "vo_bit")]
                 mixed_age: false,
-                never_move_objects: cfg!(feature = "immix_non_moving"),
+                never_move_objects: false,
             },
         )
     }
--- a/src/plan/immix/global.rs
+++ b/src/plan/immix/global.rs
@@ -38,7 +38,7 @@ pub struct Immix<VM: VMBinding> {
 
 /// The plan constraints for the immix plan.
 pub const IMMIX_CONSTRAINTS: PlanConstraints = PlanConstraints {
-    moves_objects: crate::policy::immix::DEFRAG,
+    moves_objects: true,
     // Max immix object size is half of a block.
     max_non_los_default_alloc_bytes: crate::policy::immix::MAX_IMMIX_OBJECT_SIZE,
     needs_prepare_mutator: false,
@@ -51,7 +51,8 @@ impl<VM: VMBinding> Plan for Immix<VM> {
     }
 
     fn last_collection_was_exhaustive(&self) -> bool {
-        ImmixSpace::<VM>::is_last_gc_exhaustive(self.last_gc_was_defrag.load(Ordering::Relaxed))
+        self.immix_space
+            .is_last_gc_exhaustive(self.last_gc_was_defrag.load(Ordering::Relaxed))
     }
 
     fn constraints(&self) -> &'static PlanConstraints {
@@ -139,6 +140,7 @@ impl<VM: VMBinding> Immix<VM> {
                 unlog_object_when_traced: false,
                 #[cfg(feature = "vo_bit")]
                 mixed_age: false,
+                never_move_objects: cfg!(feature = "immix_non_moving"),
             },
         )
     }
--- a/src/plan/immix/global.rs
+++ b/src/plan/immix/global.rs
@@ -38,7 +38,8 @@ pub struct Immix<VM: VMBinding> {
 
 /// The plan constraints for the immix plan.
 pub const IMMIX_CONSTRAINTS: PlanConstraints = PlanConstraints {
-    moves_objects: true,
+    // If we disable moving in Immix, this is a non-moving plan.
+    moves_objects: !cfg!(feature = "immix_non_moving"),
     // Max immix object size is half of a block.
     max_non_los_default_alloc_bytes: crate::policy::immix::MAX_IMMIX_OBJECT_SIZE,
     needs_prepare_mutator: false,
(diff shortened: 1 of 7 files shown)


src/plan/immix/global.rs
```diff
@@ -139,6 +140,7 @@ impl<VM: VMBinding> Immix<VM> {
                 unlog_object_when_traced: false,
                 #[cfg(feature = "vo_bit")]
                 mixed_age: false,
+                never_move_objects: cfg!(feature = "immix_non_moving"),
```
wks, line 143: This line `never_move_objects: cfg!(feature = "immix_non_moving")` is repeated three times for Immix, GenImmix and StickyImmix.  But the semantics of the Cargo feature "immix_non_moving" is disallowing *any* ImmixSpace instance to move objects.  Repeating `cfg!(feature = "immix_non_moving")` imposes a danger that in the future, as we add more spaces or plans, we may miss one or two of this, and accidentally assign `true` to `never_move_objects` when the "immix_non_moving" feature is enabled.

I suggest we allow plans to set `never_move_objects` according to the requirements in the "normal" scenario (i.e. the main ImmixSpace is movable, but the dedicated non-moving space is non-movable), and let `ImmixSpace::new` override its value according to `cfg!(feature = "immix_non_moving")`.  In this way, we guarantee `never_move_objects == false` if  the "immix_non_moving" feature is enabled.

Suggestion:
```suggestion
                never_move_objects: true,
```

But override it in `ImmixSpace::new`.

src/policy/immix/immixspace.rs
```diff
@@ -278,12 +280,13 @@ impl<VM: VMBinding> ImmixSpace<VM> {
         args: crate::policy::space::PlanCreateSpaceArgs<VM>,
         space_args: ImmixSpaceArgs,
     ) -> Self {
-        #[cfg(feature = "immix_non_moving")]
-        info!(
-            "Creating non-moving ImmixSpace: {}. Block size: 2^{}",
-            args.name,
-            Block::LOG_BYTES
-        );
+        if space_args.never_move_objects {
```
wks, line 283: We can make `space_args` mutable

```diff
-space_args: ImmixSpaceArgs,
+mut space_args: ImmixSpaceArgs,
```

and override `never_move_objects` here if `cfg!(feature = "immix_non_moving")` is true.  We can also print a log here, telling the user that this is overridden.

src/policy/immix/immixspace.rs
```diff
@@ -278,12 +280,13 @@ impl<VM: VMBinding> ImmixSpace<VM> {
...
+            info!(
+                "Creating non-moving ImmixSpace: {}. Block size: 2^{}",
+                args.name,
+                Block::LOG_BYTES
+            );
```
wks, line 288: This existing `info!` log here was intended to warn the user that the user is forcing the ImmxSpace not to move objects, which may hurt the performance.  We can still print this log if the user overrides `never_move_objects` with `cfg!(feature = "immix_non_moving")`.  Maybe raise it to the `warn!` level.

But if the current `ImmixSpace` is the dedicated non-moving space, it may not be worth logging this at the `info!` level because it is a normal thing.

qinsoon:
The following is the performance evaluation of this PR. The PR has no measurable impact on performance.

Immix
nonmoving-ix

StickyImmix
nonmoving-stickyix

GenImmix
nonmoving-genix

qinsoon:
@wks I have addressed your comments, and posted performance results for the PR. Can you review it again?

wks (dismissed):
LGTM

src/plan/immix/global.rs
```diff
@@ -38,7 +38,7 @@ pub struct Immix<VM: VMBinding> {
 
 /// The plan constraints for the immix plan.
 pub const IMMIX_CONSTRAINTS: PlanConstraints = PlanConstraints {
-    moves_objects: crate::policy::immix::DEFRAG,
+    moves_objects: true,
```
wks, line 41: This can be false if the "immix_non_moving" feature is enabled.

src/plan/sticky/immix/global.rs
```diff
@@ -41,7 +40,7 @@ pub struct StickyImmix<VM: VMBinding> {
 
 /// The plan constraints for the sticky immix plan.
 pub const STICKY_IMMIX_CONSTRAINTS: PlanConstraints = PlanConstraints {
-    moves_objects: crate::policy::immix::DEFRAG || crate::policy::immix::PREFER_COPY_ON_NURSERY_GC,
+    moves_objects: true,
```
wks, line 43: This can be false if both the "immix_non_moving" and the "sticky_immix_non_moving_nursery" are enabled

wks (changes requested):
One minor issue is that `PlanConstraints::moves_objects` may conditionally be true if some Cargo features are enabled.

From one hand, we may leave it as `true` so that the VM will conservatively consider `Immix` and `StickyImmix` are moving plans.  This is OK for testing object copying on VMs that plans to supported copying GC.

But I think there may be some VMs that intentionally only supports non-copying GCs. They can reject any plans labelled with `PlanConstrant::moves_objects == false`.  In this case, those VM bindings will always enable `immix_non_moving` and `sticky_immix_non_moving_nursery` and expect Immix and StickyImmix to have `PlanConstrant::moves_objects ==  true`.  I am not sure how the performance may be (because of fragmentation), but I think setting `PlanConstrant::moves_objects` accurately could be helpful.

src/plan/sticky/immix/global.rs
qinsoon: We should only check against `immix_non_moving`, which disables all kinds of moving. If we don't have `immix_non_moving`, and we have `sticky_immix_non_moving_nursery`, we may still do defrag.

wks (approved):
LGTM

PR #1308, opened 2025-04-23
Choose Non-moving Policy based on features

qinsoon:
This PR uses an Immix space as the default non moving space, and adds two features to use mark sweep or immortal as the non moving space.
* `Mutator` now includes 2 Immix allocators (one for Immix as the default space, and the other for Immix as the non moving space). 
* Call prepare/release/end_of_gc for the non moving space.
* Add `common_prepare_func` and `common_release_func` for mutators.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ name = "main"
 harness = false
 
 [features]
-default = ["builtin_env_logger", "immortal_as_nonmoving"]
+default = ["builtin_env_logger"]
 
 # Built-in env_logger.  This feature is enabled by default.
 # The user can disable this default feature to remove `env_logger` from the dependencies.
@@ -234,7 +234,6 @@ malloc_mark_sweep = []
 
 # Group:nonmovingspace
 immortal_as_nonmoving = []
-immix_as_nonmoving = []
 marksweep_as_nonmoving = []
 
 # If there are more groups, they should be inserted above this line
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mmtk"
-version = "0.30.0"
+version = "0.31.0"
 authors = ["The MMTk Developers <>"]
 edition = "2021"
 license = "MIT OR Apache-2.0"
@@ -39,7 +39,7 @@ log = { version = "0.4", features = ["max_level_trace", "release_max_level_off"]
 memoffset = "0.9"
 mimalloc-sys = { version = "0.1.6", optional = true }
 # MMTk macros - we have to specify a version here in order to publish the crate, even though we use the dependency from a local path.
-mmtk-macros = { version="0.30.0", path = "macros/" }
+mmtk-macros = { version="0.31.0", path = "macros/" }
 num_cpus = "1.8"
 num-traits = "0.2"
 pfm = { version = "0.1.1", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ name = "main"
 harness = false
 
 [features]
-default = ["builtin_env_logger"]
+default = ["builtin_env_logger", "immortal_as_nonmoving"]
 
 # Built-in env_logger.  This feature is enabled by default.
 # The user can disable this default feature to remove `env_logger` from the dependencies.
@@ -224,9 +224,6 @@ malloc_jemalloc = ["dep:jemalloc-sys"]
 # is not compiled in default builds.
 malloc_native_mimalloc = []
 
-# If there are more groups, they should be inserted above this line
-# Group:end
-
 # Group:marksweepallocation
 # default is native allocator with lazy sweeping
 eager_sweeping = []
@@ -234,4 +231,11 @@ eager_sweeping = []
 # normal heap range, we will have to use chunk-based SFT table. Turning on this feature will use a different SFT map implementation on 64bits,
 # and will affect all the plans in the build. Please be aware of the consequence, and this is only meant to be experimental use.
 malloc_mark_sweep = []
+
+# Group:nonmovingspace
+immortal_as_nonmoving = []
+immix_as_nonmoving = []
+marksweep_as_nonmoving = []
+
+# If there are more groups, they should be inserted above this line
 # Group:end
--- a/src/plan/global.rs
+++ b/src/plan/global.rs
@@ -7,6 +7,8 @@ use crate::plan::tracing::ObjectQueue;
 use crate::plan::Mutator;
 use crate::policy::immortalspace::ImmortalSpace;
 use crate::policy::largeobjectspace::LargeObjectSpace;
+use crate::policy::marksweepspace::native_ms::MarkSweepSpace;
+use crate::policy::immix::ImmixSpace;
 use crate::policy::space::{PlanCreateSpaceArgs, Space};
 #[cfg(feature = "vm_space")]
 use crate::policy::vmspace::VMSpace;
@@ -553,7 +555,7 @@ pub struct CommonPlan<VM: VMBinding> {
     pub los: LargeObjectSpace<VM>,
     // TODO: We should use a marksweep space for nonmoving.
     #[space]
-    pub nonmoving: ImmortalSpace<VM>,
+    pub nonmoving: ImmixSpace<VM>,
     #[parent]
     pub base: BasePlan<VM>,
 }
@@ -571,12 +573,17 @@ impl<VM: VMBinding> CommonPlan<VM> {
                 args.get_space_args("los", true, false, VMRequest::discontiguous()),
                 false,
             ),
-            nonmoving: ImmortalSpace::new(args.get_space_args(
+            nonmoving: ImmixSpace::new(args.get_space_args(
                 "nonmoving",
                 true,
                 false,
                 VMRequest::discontiguous(),
-            )),
+            ), crate::policy::immix::ImmixSpaceArgs {
+                unlog_object_when_traced: false,
+                #[cfg(feature = "vo_bit")]
+                mixed_age: false,
+                never_move_objects: true,
+            }),
             base: BasePlan::new(args),
         }
     }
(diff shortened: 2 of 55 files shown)


qinsoon:
binding-refs
OPENJDK_BINDING_REF=update-mmtk-core-1308

Cargo.toml
```diff
@@ -224,14 +224,17 @@ malloc_jemalloc = ["dep:jemalloc-sys"]
...
 # Group:marksweepallocation
 # default is native allocator with lazy sweeping
 eager_sweeping = []
 # Use library malloc as the freelist allocator for mark sweep. This will makes mark sweep slower. As malloc may return addresses outside our
 # normal heap range, we will have to use chunk-based SFT table. Turning on this feature will use a different SFT map implementation on 64bits,
 # and will affect all the plans in the build. Please be aware of the consequence, and this is only meant to be experimental use.
 malloc_mark_sweep = []
+
+# Group:nonmovingspace
+immortal_as_nonmoving = []
```
qinsoon, line 236: This is a bit awkward. Ideally, we should have 3 features `immix_as_nonmoving`, `immortal_as_nonmoving` and `marksweep_as_nonmoving`, and we use `immix_as_nonmoving` as a default feature. But we need to ensure that for every build, exactly one feature from those 3 features is enabled. This also applies to the mark sweep features above, which should be 3 features `lazy_sweeping`, `eager_sweeping` and `malloc_mark_sweep`, and we need to choose exactly one.

So basically we need mutually exclusive features, and we need to make sure the features are mandatory that you need to use one of them. We cannot do this with cargo. In our test script, we try to achieve this when we cover all the builds with different features, but our test script only supports optional mutually exclusive features. So for now, we assume the default (without any feature to control it), and make the other two features optional and mutually exclusive.

I would like to do this refactoring in a separate PR.

k-sareen:
We can do something like this: https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features

Cargo.toml
k-sareen: We can do something like this: https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features

wks (commented):
I refactored the code a bit to make `NonMovingSpace` a trait.  By doing so, we greatly reduced the amount of `cfg` attributes used.

https://github.com/qinsoon/mmtk-core/compare/nonmoving-space-options...wks:mmtk-core:feature/nonmoving-space-options-trait?expand=1

If you think that is appropriate, you can cherry-pick it into this PR.

wks:
OK. I have pushed more changes to

https://github.com/qinsoon/mmtk-core/compare/nonmoving-space-options...wks:mmtk-core:feature/nonmoving-space-options-trait?expand=1

@qinsoon You can merge it into your `nonmoving-space-options` branch to make it part of this PR if you think it is appropriate.

I am also trying to further remove the cfg in `create_allocator_mapping` (see [here](https://github.com/qinsoon/mmtk-core/commit/d731b4c2b9b4836c01973e45c5237945092f5a78)), but without success.  I basically wanted to use the "visitor pattern" to remove the cascading if-else.  But the `ChosenNonMovingSpace` type still has a `<VM>` type parameter, while the `create_allocator_mapping` doesn't have that.  I think there is a deeper problem that `create_allocator_mapping` should really visit each space in the `Plan` and let each space reserve its desired allocator.  But that requires more refactoring, and may be better done in a separate PR.

src/plan/global.rs
```diff
@@ -542,6 +546,15 @@ impl<VM: VMBinding> BasePlan<VM> {
...
 }
 
+#[cfg(feature = "immortal_as_nonmoving")]
+pub type NonMovingSpace<VM> = crate::policy::immortalspace::ImmortalSpace<VM>;
+
+#[cfg(not(any(feature = "immortal_as_nonmoving", feature = "marksweep_as_nonmoving")))]
+pub type NonMovingSpace<VM> = crate::policy::immix::ImmixSpace<VM>;
+
+#[cfg(feature = "marksweep_as_nonmoving")]
+pub type NonMovingSpace<VM> = crate::policy::marksweepspace::native_ms::MarkSweepSpace<VM>;
```
wks, line 556: If you think it is not worth introducing the `trait NonMovingSpace` as I suggested in https://github.com/mmtk/mmtk-core/pull/1308#issuecomment-2862128839, we can still simplify this part of the code with `cfg_if::cfg_if!`.  The `else` branch can save you from having to type `not(any(feature = "immortal_as_nonmoving", feature = "marksweep_as_nonmoving"))`.

src/plan/global.rs
```diff
@@ -610,9 +626,51 @@ impl<VM: VMBinding> CommonPlan<VM> {
         &self.los
     }
 
-    pub fn get_nonmoving(&self) -> &ImmortalSpace<VM> {
+    pub fn get_nonmoving(&self) -> &NonMovingSpace<VM> {
         &self.nonmoving
     }
+
+    fn new_nonmoving_space(args: &mut CreateSpecificPlanArgs<VM>) -> NonMovingSpace<VM> {
```
wks, line 633: And we can use `cfg_if` in these functions, too.

src/plan/mutator_context.rs
```diff
@@ -481,8 +530,18 @@ pub(crate) fn create_allocator_mapping(
...
+            feature = "immortal_as_nonmoving",
+            feature = "marksweep_as_nonmoving"
+        ))) {
+            reserved.add_immix_allocator()
+        } else if cfg!(feature = "marksweep_as_nonmoving") {
+            reserved.add_free_list_allocator()
+        } else if cfg!(feature = "immortal_as_nonmoving") {
+            reserved.add_bump_pointer_allocator()
+        } else {
+            panic!("No policy selected for nonmoving space")
```
wks, line 543: Actually the first branch `cfg!(not(any(feature = "immortal_as_nonmoving", feature = "marksweep_as_nonmoving")))` covers all other cases, so this `else` is unreachable.  But I suggest we move `reserved.add_immix_allocator()` into this `else` branch and removt the first `if cfg!(not(any(..., ...)))` branch.

src/plan/mutator_context.rs
```diff
@@ -534,9 +593,19 @@ pub(crate) fn create_space_mapping<VM: VMBinding>(
...
+            if cfg!(feature = "marksweep_as_nonmoving") {
+                reserved.add_free_list_allocator()
+            } else if cfg!(not(any(
+                feature = "immortal_as_nonmoving",
+                feature = "marksweep_as_nonmoving"
+            ))) {
+                reserved.add_immix_allocator()
+            } else if cfg!(feature = "immortal_as_nonmoving") {
+                reserved.add_bump_pointer_allocator()
+            } else {
```
wks, line 606: Same here.

qinsoon:
> I refactored the code a bit to make `NonMovingSpace` a trait. By doing so, we greatly reduced the amount of `cfg` attributes used.
> 
> https://github.com/qinsoon/mmtk-core/compare/nonmoving-space-options...wks:mmtk-core:feature/nonmoving-space-options-trait?expand=1
> 
> ~If you think that is appropriate, you can cherry-pick it into this PR.~ Please tell me if it is appropriate, as I am still adding comments and making other changes.

I considered introducing a `struct NonMovingSpace`, but I gave up the idea. I think it is inconsistent with the design. We have policies which are GC algorithms, and we use those policies for certain spaces, such as non moving, or nursery. It would be strange to have something like `NonMovingSpace`, which confuses GC algorithms and its actual use.

Besides, the methods in the proposed `ChosenNonMovingSpace` are just methods that should have been in the `Space`. I think we could just add methods like `prepare`, `release`, `end_of_gc`, etc, to `Space`. Some policies may need extra parameters like the defrag stats for Immix, we can invoke a separate method in the `ImmixSpace` to set the parameter, and make `prepare/release/end_of_gc` have the same parameters for all the spaces.

src/plan/global.rs
qinsoon: See my comments in https://github.com/mmtk/mmtk-core/pull/1308#discussion_r2078718672. I would like to introduce a feature `immix_as_nonmoving` as a default feature. That would solve most of the issues here with annoying `cfg` (we also have the same issue with the features like `malloc_jemalloc`, and `eager_sweep`).

Cargo.toml
wks: Default features can be disabled (see https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature).  mmtk-core currently has one default feature: `builtin_env_logger`, and VM bindings that use their own logging frameworks (such as OpenJDK) could disable default features and wire Rust logs into their own.

If the user disables default features, there will be neither `immix_as_nonmoving`, `immortal_as_nonmoving` nor `marksweep_as_nonmoving`.  So we need to consider the case that there is no non-moving space.

Actually I think the default should be "no non-moving space".  Not all VMs need that.  Neither OpenJDK, JikesRVM nor Ruby need it.  We can make all of them optional, and use the trick described in https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features to reject multiple non-moving spaces.

qinsoon:
I have made changes according to the comments about the  `cfg_if`.

src/plan/mutator_context.rs
```diff
@@ -36,7 +54,38 @@ pub(crate) fn unreachable_release_func<VM: VMBinding>(
...
+        unsafe {
+            mutator
+                .allocators
+                .get_allocator_mut(mutator.config.allocator_mapping[AllocationSemantics::NonMoving])
+        }
+        .downcast_mut::<FreeListAllocator<VM>>()
+        .unwrap()
+        .release();
+    }
+    #[cfg(not(any(feature = "immortal_as_nonmoving", feature = "marksweep_as_nonmoving")))]
```
wks, line 73: We can use `cfg_if!`.  Although "immortal_as_nonmoving" needs no action in this function. I still think it is more explicit and more readable to have something like:

```rust
cfg_if::cfg_if! {
    if #[cfg(feature = "immortal_as_nonmoving")] {
        // We don't need to do anything for ImmortalSpace.
    } else if #[cfg(feature = "mark_sweep_as_nonmoving")] {
        // ...
    } else {
        // ...
    }
}
```

At least we won't miss any cases if we write this way.

wks (changes requested):
Two more things to change:

1.  Some `#[cfg]` can be re-written with `cfg_if`.  I admit that without visitor pattern or generalizing the `prepare/release/end_of_gc`, chained if-else may be the best thing we can get for now.
2.  If we use the default value of `needs_prepare_mutator`, we can omit that line because `..PlanConstraints::default()` will fill in default values.

wks (approved):
LGTM.

There is an opportunity to remove some `unsafe`, but we can leave it to another PR.

PR #1310, opened 2025-04-25
(Don't merge) Test binding test using OpenJDK 21

wks:
This PR is for checking if the CI script for binding tests work for OpenJDK 21.

When we are ready for OpenJDK 21, we will update the `master` branch of the https://github.com/mmtk/mmtk-openjdk repo so that we don't need change on the mmtk-core side.


wks:
Closed.  The OpenJDK-side PR is https://github.com/mmtk/mmtk-openjdk/pull/307.  We don't need a PR in mmtk-core for the OpenJDK update.

PR #1312, opened 2025-05-01
Use ChunkMap in MallocSpace

qinsoon:
This PR changes `MallocSpace` to remove its own global side metadata `MS_ACTIVE_CHUNK`, and use `ChunkMap` instead. This PR resolves the discussion in https://github.com/mmtk/mmtk-core/pull/1304#discussion_r2055114236.


k-sareen (approved):
LGTM. Thanks!

PR #1315, opened 2025-05-09
Move prepare/release/end_of_gc to trait Space

qinsoon:


src/policy/immix/immixspace.rs
```diff
@@ -188,6 +188,123 @@ impl<VM: VMBinding> Space<VM> for ImmixSpace<VM> {
...
+                if let MetadataSpec::OnSide(side) = *VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC {
+                    for chunk in self.chunk_map.all_chunks() {
+                        side.bzero_metadata(chunk.start(), Chunk::BYTES);
+                    }
+                }
+            }
+
+            // Prepare defrag info
+            if self.is_defrag_enabled() {
+                let plan_stats = arg.unwrap().downcast::<StatsForDefrag>().unwrap();
```
wks, line 212: This means `ImmixSpace` is expecting a `StatesForDefrag` argument.

src/policy/copyspace.rs
```diff
@@ -138,6 +139,39 @@ impl<VM: VMBinding> Space<VM> for CopySpace<VM> {
     fn enumerate_objects(&self, enumerator: &mut dyn ObjectEnumerator) {
         object_enum::enumerate_blocks_from_monotonic_page_resource(enumerator, &self.pr);
     }
+
+    fn prepare(&mut self, _major_gc: bool, arg: Option<Box<dyn Any>>) {
+        let is_from_space = arg.unwrap().downcast::<bool>().unwrap();
```
wks, line 144: This means `CopySpace` expects a `bool`

wks (commented):
After a second thought, I think it is *not* a good idea to "unify" the `prepare`, `release` and `end_of_gc` methods into the `Space` trait.

The obvious problem is the `arg` argument in `prepare`.  Actually each space expects a specific arg type.  Specifically, `CopySpace` expects a `bool`, while `ImmixSpace` expects a `StatsForDefrag`.  Making it a `Option<Box<dyn std::any::Any>>` simply makes it dynamically types, and errors when passing the argument cannot be detected until run time.  On the contrary, when using existing way, the Plan knows the concrete types of its spaces, and calls the `prepare` methods of each space differently, providing exactly what type of argument required.  And that's type-checked by the Rust compiler.

The purpose of this PR is to support the polymorphic non-moving space (https://github.com/mmtk/mmtk-core/pull/1308).  The `prepare` method is still called differently depending on the concrete space type.

In [my refactoring](https://github.com/qinsoon/mmtk-core/compare/nonmoving-space-options...wks:mmtk-core:feature/nonmoving-space-options-trait?expand=1) I made a `trait NonMovingSpace` and its `prepare_nonmoving_space` takes only a `full_heap` parameter (passed to `major_gc` in `prepare` of some spaces).  I admit that it is not general, either.  It only works coincidentally because `full_heap` happens to be the union of all arguments of `ImmortalSpace`, `MarkSweepSpace` and `ImmixSpace`.

qinsoon:
> After a second thought, I think it is not a good idea to "unify" the prepare, release and end_of_gc methods into the Space trait.

I am okay with closing this PR. That means we understand that most spaces have `prepare/release/end_of_gc`, but we allow those functions to have different signatures, and we do not intend to call those functions in a unified way.

qinsoon:
See the discussion in this PR and in https://github.com/mmtk/mmtk-core/issues/1316. We do not want to do this change for now.

PR #1317, opened 2025-05-10
Fix `CopySpace::is_live`

k-sareen:
This commit fixes the `CopySpace::is_live` function to check if a to-space object's address is actually below the current page resource cursor. By definition, a to-space object cannot be live if it is beyond the cursor.

### DIFF
--- a/src/policy/copyspace.rs
+++ b/src/policy/copyspace.rs
@@ -30,7 +30,11 @@ impl<VM: VMBinding> SFT for CopySpace<VM> {
     }
 
     fn is_live(&self, object: ObjectReference) -> bool {
-        !self.is_from_space() || object_forwarding::is_forwarded::<VM>(object)
+        if !self.is_from_space() {
+            object.to_raw_address() < self.pr.cursor()
+        } else {
+            object_forwarding::is_forwarded::<VM>(object)
+        }
     }
 
     #[cfg(feature = "object_pinning")]


k-sareen:
Ran into this bug when I was using fixed extent semi-spaces. Since we don't zero pages on release, we could have stale data there from the last GC wherein the (dead) object is beyond the cursor but it was forwarded so both `!self.is_from_space()` and `is_live` will return true. The fix now makes it clearer that it needs to be mutually exclusive, i.e. we can't have forwarded objects in to-space.

qinsoon:
> Ran into this bug when I was using fixed extent semi-spaces. Since we don't zero pages on release, we could have stale data there from the last GC wherein the (dead) object is beyond the cursor but it was forwarded so both `!self.is_from_space()` and `is_forwarded` will return true. The fix now makes it clearer that it needs to be mutually exclusive, i.e. we can't have forwarded objects in to-space.

The function `is_live` requires an object reference. But in this case, what gets passed to the function is not a valid object reference.

k-sareen:
Okay sure -- I'm using it in a bit of a non-standard way since I'm using it to linear scan the heap.

wks:
> Okay sure -- I'm using it in a bit of a non-standard way since I'm using it to linear scan the heap.

The standard way to linearly scan the heap uses the VO bit.  But right after tracing, some metadata may happen to be set precisely at valid objects.  It is the mark bits for ImmixSpace.  But for CopySpace, maybe it should be both the forwarding bits and the cursor.

You may make a customized linear-scanning function with a parameterized "is_valid_object" function, similar to `slice::sort_by` with the comparing function as a parameter.

k-sareen:
> But right after tracing, some metadata may happen to be set precisely at valid objects.

Right. I'm only doing it right after a GC so it should be all the valid objects. I just don't use the VO bit because I don't want to add it to the allocation fast-path in the general case. The above is mainly only for creating the bootimage.

PR #1320, opened 2025-05-14
Remove dead field in `VMMap`

k-sareen:


qinsoon (approved):
LTGM.

The field was used for triggering stress GC. I think it is not used any more. It was replaced by `allocation_bytes` in `GlobalState` for triggering stress GCs.

PR #1321, opened 2025-05-15
Move `inside_harness` to `GlobalState`

k-sareen:
This can be useful in case we want to be able to measure timing in parts of the code where we don't have access to the `MMTK` struct.


qinsoon (approved):
LGTM

PR #1324, opened 2025-05-20
Remove the immix_stress_copying feature. Replace with separate options.

qinsoon:
`immix_stress_copying` may cause confusion with the stress GC and the stress factor, as `immix_stress_copying` does not imply stress GCs. This PR removes `immix_stress_copying`, and replaces with separate options to control the specific behavior.

### DIFF
--- a/src/util/options.rs
+++ b/src/util/options.rs
@@ -902,6 +902,10 @@ options! {
     transparent_hugepages:  bool                    [|v: &bool| !v || cfg!(target_os = "linux")] = false,
     /// Count live bytes for objects in each space during a GC.
     count_live_bytes_in_gc: bool                    [always_valid] = false,
+    /// Make every GC a defragment GC. (for debugging)
+    immix_stress_defrag: bool                       [always_valid] = false,
+    /// Mark every allocated block as defragmentation source before GC. (for debugging)
+    immix_defrag_every_block: bool                  [always_valid] = false,
     /// Percentage of heap size reserved for defragmentation.
     /// According to [this paper](https://doi.org/10.1145/1375581.1375586), Immix works well with
     /// headroom between 1% to 3% of the heap size.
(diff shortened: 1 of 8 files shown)


src/util/options.rs
```diff
@@ -901,7 +901,15 @@ options! {
...
-    count_live_bytes_in_gc: bool                    [always_valid] = false
+    count_live_bytes_in_gc: bool                    [always_valid] = false,
+    /// Make every GC a defragment GC. (for debugging)
+    immix_stress_defrag: bool                       [always_valid] = false,
+    /// Mark every allocated block as defragmentation source before GC. (for debugging)
+    immix_defrag_every_block: bool                  [always_valid] = false,
+    /// Percentage of heap size reserved for defragmentation.
+    /// According to [this paper](https://doi.org/10.1145/1375581.1375586), Immix works well with
+    /// headroom between 1% to 3% of the heap size.
+    immix_defrag_headroom_percent: usize            [|v: &usize| *v <= 50] = 2
```
wks, line 912: Merely setting `immix_defrag_every_block = true` will make it run out of headroom before copying every live object.  We should mention, in either `immix_defrag_every_block` or `immix_defrag_headroom_percent` or both, that we need to set `immix_defrag_headroom_percent = 50` if we want to stress copying.  We can also provide a link to the porting guide.

qinsoon: This is a good point. So setting `immix_defrag_every_block` to true alone does not make sense.

What if we set the defrag headroom to 50% but we do not set `immix_defrag_every_block`? I assume the defrag also works, as it will still defrag enough blocks (which should be all the blocks) to use the headroom.

I am wondering if we really need an option for `immix_defrag_every_block`.

qinsoon:
I have made changes to address the reviews. Can you take a look again? @wks

wks (approved):
LGTM.

The mmtk-ruby binding currently has a feature named `immix_stress_copying` which is just a wrapper of the mmtk-core feature fo the same name.  I'll remove it from mmtk-ruby once this PR is merged.

PR #1326, opened 2025-05-26
Bump Ruby binding test image to 24.04

wks:
CRuby now requires an existing CRuby 3.1.0 installation to build itself, but Ubuntu 22.04 only provides CRuby 3.0.  We upgrade to 24.04.

### DIFF
--- a/.github/workflows/extended-tests-bindings.yml
+++ b/.github/workflows/extended-tests-bindings.yml
@@ -154,6 +154,8 @@ jobs:
           ./.github/scripts/ci-style.sh
 
   extended-tests-ruby:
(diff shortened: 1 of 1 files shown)


qinsoon (approved):

.github/workflows/extended-tests-bindings.yml
```diff
@@ -154,7 +154,7 @@ jobs:
           ./.github/scripts/ci-style.sh
 
   extended-tests-ruby:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
qinsoon, line 157: Maybe add some comments here to explain why Ruby uses a different image.

PR #1332, opened 2025-07-07
Make scan_non_zero_values_fast scan entirely forwards

no-defun-allowed:
Presently the `forwards` argument to `ranges::break_bit_range` is false, causing the function to scan the last bits, then the middle bytes, then the first bits. The argument should be true, so that the function scans the first bits first.


k-sareen (approved):
LGTM. Thanks!

PR #1333, opened 2025-07-07
Fix style check for Rust 1.88

qinsoon:
Mostly https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args. Rust allows inlining identifiers for format args, such as `println!("a = {a}")`, clippy by default deney uninlined format args if they can be inlined (`println!("a = {}", a)`) is not allowed).

### DIFF
--- a/benches/mock_bench/mod.rs
+++ b/benches/mock_bench/mod.rs
@@ -27,7 +27,7 @@ pub fn bench(c: &mut Criterion) {
             "alloc" => alloc::bench(c),
             "internal_pointer" => internal_pointer::bench(c),
             "sft" => sft::bench(c),
-            _ => panic!("Unknown benchmark {:?}", bench),
+            _ => panic!("Unknown benchmark {bench:?}"),
         },
         Err(_) => panic!("Need to name a benchmark by the env var MMTK_BENCH"),
     }
--- a/src/global_state.rs
+++ b/src/global_state.rs
@@ -147,9 +147,7 @@ impl GlobalState {
         let old = self.scanned_stacks.fetch_add(1, Ordering::SeqCst);
         debug_assert!(
             old < n_mutators,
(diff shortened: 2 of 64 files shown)


k-sareen:
I feel like this lint is silly. It doesn't really help with readability specially if the string gets too long/there is a mix of the arguments. I'm inclined to say that we just disable `uninlined_format_args` globally.

qinsoon:
> I feel like this lint is silly. It doesn't really help with readability specially if the string gets too long/there is a mix of the arguments. I'm inclined to say that we just disable `uninlined_format_args` globally.

I don't quite like this lint either, but it does help in some cases.

Like, our current code has this (this is the required format by `cargo fmt`).
```rust
trace!(
    "Immix nursery object {} is being traced without moving",
    object
);
```
When the arg gets inlined, it becomes one line:
```rust
trace!("Immix nursery object {object} is being traced without moving");
```

wks:
The `uninlined_format_args` lint has caused much controversy in the past.  It happened in 1.67.0 (https://github.com/rust-lang/rust-clippy/issues/10087), and Rust 1.67.1 downgraded it back to "pedantic" (https://releases.rs/docs/1.67.1/, https://github.com/rust-lang/rust-clippy/pull/10265).  And they are upgrading it again in 1.88, and someone has already reported it as an issue: https://github.com/rust-lang/rust-clippy/issues/15151

In my opinion, inlining format arguments is not really more readable than not inlining them, and mixing inlined and un-inlined args makes it even less readable.  I suggest disabling that lint completely.

k-sareen:
Ah. The rust-analyzer reason is even more relevant. I didn't realize that use-case would break.

qinsoon:
This PR can be closed without merging, as we don't need to adapt the code base to the `uninlined_format_args` lint.

PR #1335, opened 2025-07-09
Move lint config to Cargo.toml. Adapt to Rust 1.88

qinsoon:
This PR fixes issues for Rust 1.88 style check.
* Move lint configurations to Cargo.toml from `lib.rs` (Lint configs in `Cargo.toml` is effective for the `bench` source code, but `lib.rs` does not effect `bench`)
* Disable `uninlined_format_args`.
* Allow one more case of `mut_from_ref`, identified by clippy.
* Set `enum-map` to `2.7.3` (We use `EnumMap::from_fn`. [`2.7.3`](https://docs.rs/enum-map/latest/enum_map/struct.EnumMap.html#method.from_fn) has this function, but [`2.6.3`](https://docs.rs/enum-map/2.6.3/enum_map/struct.EnumMap.html#) does not have this function)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,9 +88,9 @@ idna_adapter = "=1.1.0"
 # not worth it.
 upper_case_acronyms = "allow"
 # Temporarily allow this -- 9 July 2025. Clippy suggests that we should always inline format args if possible.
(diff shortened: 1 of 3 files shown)


wks (approved):
LGTM

Issue #1336, opened 2025-07-09
Inefficient locking in `FragmentedMapper::get_or_optionally_allocate_slab_table`


wks:
`FragmentedMapper::get_or_optionally_allocate_slab_table(addr, allocate)` finds the slab table for an address, or optionally allocates a slab table for it if not found.  It is essentially a hash map.  Its algorithm is:

1.  Compute the hash value of the base address derived form `addr`, and use the hash value as the current `index`.
2.  Look up the `slab_map` from the `index`.
3.  If it hits, return the `stab_table` entry.
4.  Otherwise, **acquire the `lock`**.
5.  If the entry at `index` is free, either return `None` or allocate a slab table there and return it.
6.  Othrwise the entry must be occupied by the slab table of another address.
7.  Increment `index` by one.
8.  **Release the `lock`**.
9.  Go to step 2

It will work normally if step 2 hits immediately.  But if the index happens to be used for another address, it will go through step 4 to 9 and repeat from 2, linearly scanning the hash table until it finds the matching or vacant entry.  During this process, it will frequently acquire and release the mutex `FragmentedMapper::lock`, causing significant slowdown.

The effect is amplified by frequently calling `is_mmtk_object` which checks if an arbitrary address points to a valid MMTk object.  In the [mmtk-ruby binding](https://github.com/mmtk/mmtk-ruby), I added [an aggressive assertion](https://github.com/mmtk/ruby/blob/3051b4d4f3808e7301351e0fc9621cd162d8e0b3/gc/default/default.c#L6587) at every write barrier to check if both the source and the target objects are valid MMTk objects.  It normally costs very little compared to the rest of the VM.  But remember that MMTk must check if an address is mapped before accessing the VO bits, and `FragmentedMapper::get_or_optionally_allocate_slab_table` is called when checking if a given metadata address is mapped.  If the metadata address happens to have a conflict in the `FragmentedMapperInner::slab_map`, it will reach the pathological case I described above.  When running the [optcarrot](https://github.com/mame/optcarrot), I observed almost 50% drop in FPS when choosing certain GC plans.  The problem goes away when I disable the assertion.

# Solution

We may argue that we should remove such assertions when measuring performance.  But I still believe the current implementation of `FragmentedMapper::get_or_optionally_allocate_slab_table` should be improved to eliminate the pathological cases.

[...]

PR #1337, opened 2025-07-09
Add MultiLevelMapper

wks:
**DRAFT**:
-   [ ] Needs more correctness testing
-   [ ] Needs performance evaluation

This is a replacement for `FragmentedMapper`.  It does not use hash tables and is lock-free when looking up or lazily initializing the slabs table.  It is intended to eliminate the pathological case of `FragmentedMapper` where a hash collision can result in many locking operations to find the matching or vacant slab table entry.

Fixes: https://github.com/mmtk/mmtk-core/issues/1336

### DIFF
--- a/src/util/rust_util/atomic_box.rs
+++ b/src/util/rust_util/atomic_box.rs
@@ -1,0 +1,48 @@
+use std::sync::atomic::{AtomicPtr, Ordering};
+
+/// A lazily initialized box.  Similar to an `Option<Box<T>>`, but can be initialized atomically.
+pub struct OnceOptionBox<T> {
+    inner: AtomicPtr<T>,
+}
+
+impl<T> OnceOptionBox<T> {
+    pub fn new() -> OnceOptionBox<T> {
+        Self {
+            inner: AtomicPtr::new(std::ptr::null_mut()),
+        }
+    }
+
+    pub fn get(&self, order: Ordering) -> Option<&T> {
+        let ptr = self.inner.load(order);
+        unsafe { ptr.as_ref() }
+    }
+
+    pub fn get_or_init(
+        &self,
+        order_load: Ordering,
+        order_store: Ordering,
+        init: impl FnOnce() -> T,
+    ) -> Option<&T> {
+        if let Some(get_result) = self.get(order_load) {
+            return Some(get_result);
+        }
+
+        let new_inner = Box::into_raw(Box::new(init()));
+        let cas_result = self.inner.compare_exchange(
+            std::ptr::null_mut(),
+            new_inner,
+            order_store,
+            Ordering::Relaxed,
+        );
+        match cas_result {
+            Ok(old_inner) => {
+                debug_assert_eq!(old_inner, std::ptr::null_mut());
+                unsafe { new_inner.as_ref() }
+            }
+            Err(old_inner) => {
+                drop(unsafe { Box::from_raw(new_inner) });
+                unsafe { old_inner.as_ref() }
+            }
+        }
+    }
+}
--- a/src/util/heap/layout/mod.rs
+++ b/src/util/heap/layout/mod.rs
@@ -6,6 +6,7 @@ pub use self::mmapper::Mmapper;
 mod byte_map_mmapper;
 #[cfg(target_pointer_width = "64")]
 mod fragmented_mapper;
+mod multi_level_mapper;
 
 mod map;
 pub(crate) use self::map::CreateFreeListResult;
@@ -37,7 +38,8 @@ pub fn create_mmapper() -> Box<dyn Mmapper + Send + Sync> {
 #[cfg(target_pointer_width = "64")]
 pub fn create_mmapper() -> Box<dyn Mmapper + Send + Sync> {
     // TODO: ByteMapMmapper for 39-bit or less virtual space
-    Box::new(fragmented_mapper::FragmentedMapper::new())
+    // Box::new(fragmented_mapper::FragmentedMapper::new())
+    Box::new(multi_level_mapper::MultiLevelMapper::new())
 }
 
 use crate::util::Address;
--- a/src/util/heap/layout/multi_level_mapper.rs
+++ b/src/util/heap/layout/multi_level_mapper.rs
@@ -1,0 +1,503 @@
+use super::mmapper::MapState;
+use super::Mmapper;
+use crate::util::constants::BYTES_IN_PAGE;
+use crate::util::conversions;
+use crate::util::heap::layout::vm_layout::*;
+use crate::util::memory::{MmapAnnotation, MmapStrategy};
+use crate::util::rust_util::atomic_box::OnceOptionBox;
+use crate::util::Address;
+use atomic::{Atomic, Ordering};
+use std::fmt;
+use std::io::Result;
+use std::sync::Mutex;
+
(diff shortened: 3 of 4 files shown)


wks:
This PR is intended as a quick fix for the pathological case of `FragmentedMapper`  The newly added `two_level_mmapper.rs` inherits most of the code from `FragmentedMapper`, including both `impl Mmapper for TwoLevelMmapper` and the test cases.  The `impl Mmapper` part should be as correct (or incorrect) as the `FragmentedMapper`, but I found that those parts have program logics of slab operations intermixed with (single-chunk or bulk) memory protection.  I will refactor that part in another PR, and probably make `TwoLevelMmapper` share code with `ByteMapMmapper`.

wks:
With this PR applied, the `optcarrot` emulator can run at 58-60 FPS even with the aggressive VO bit assertion on WB enabled.  It means VO bit checks still have a cost, but it is much cheaper than 30 FPS (50% total time overhead).

I also ran some benchmarks on VO bit linear scanning for internal pointers.  The data is shown in https://github.com/mmtk/mmtk-core/pull/1341#issuecomment-3072047098.  In short, this PR greatly improves the performance of internal pointer checks.

src/util/heap/layout/two_level_mmapper.rs
```diff
@@ -4,68 +4,71 @@ use crate::util::constants::BYTES_IN_PAGE;
...
-unsafe impl Sync for FragmentedMapper {}
+/// The slab type.  Each slab holds the `MapState` of multiple chunks.
+type Slab = [Atomic<MapState>; 1 << LOG_MMAP_CHUNKS_PER_SLAB];
 
-struct InnerFragmentedMapper {
-    free_slab_index: usize,
-    free_slabs: Vec<Option<Box<Slab>>>,
-    slab_table: Vec<Option<Box<Slab>>>,
-    slab_map: Vec<Address>,
+/// A two-level implementation of `Mmapper`.
```
qinsoon, line 55: Could have a little bit more description about the implementation. E.g. what are the two levels.

src/util/rust_util/atomic_box.rs
```diff
@@ -0,0 +1,103 @@
...
+///
+/// It is designed for implementing shared data.  Therefore, methods with `&self`, namely
+/// [`OnceOptionBox::get`] and the [`OnceOptionBox::get_or_init`] methods, only return shared
+/// references to the content (`&T`).  The user should use types that support multi-threaded
+/// accesses, such as mutexes or atomic types, if the inner type is supposed to be modified
+/// concurrently.
+///
+/// Once initialized, this object will own its content.  THe content is allocated in the heap, and
+/// will be dropped and deallocated when this instance is dropped.
+pub struct OnceOptionBox<T> {
```
qinsoon, line 21: Can you explain why we need a new type like this, instead of something like `Once<Atomic<Option<T>>>` (T is the inner type, such as `usize` for `AtomicUsize`, or `MapState` for `Atomic<MapState>`)?

qinsoon (approved):
LGTM. The performance results look fine. On average there is no obvious performance difference. The minor different we observed might be just noise.

As you added a bench, did you check how it performs with the two-level mmaper and the old fragmented mmapper?

src/util/heap/layout/two_level_mmapper.rs
wks: I added comments to explain the design of the `TwoLevelMmapper`.

src/util/rust_util/atomic_box.rs
wks: I added comments that compare `OnceOptionBox` with `std::sync::OnceLock`

k-sareen:
Wait, I don't understand why we don't just preserve the old `FragmentedMmapper`? Is there a particular reason why we don't want to have the old one?

src/util/heap/layout/two_level_mmapper.rs
```diff
@@ -1,71 +1,85 @@
...
+///
+/// It is essentially a lazily initialized array of [`Atomic<MapState>`].  Because it is designed to
+/// govern a large address range, and the array is sparse, we use a two-level design.  The higher
+/// level holds a vector of slabs, and each slab holds an array of [`Atomic<MapState>`].  Each slab
+/// governs an aligned region of [`MMAP_CHUNKS_PER_SLAB`] chunks.  Slabs are lazily created when the
+/// user intends to write into one of its `MapState`.
+pub struct TwoLevelMmapper {
+    /// Lock for transitioning map states.
+    ///
+    /// FIXME: We only needs the lock when transitioning map states.
```
k-sareen:
```suggestion
    /// FIXME: We only need this lock when transitioning map states.
```

src/util/heap/layout/two_level_mmapper.rs
```diff
@@ -230,164 +244,59 @@ impl Mmapper for FragmentedMapper {
...
-        Self::slab_align_down(addr) + MMAP_SLAB_EXTENT
+        Self::slab_align_down(addr) + MMAP_SLAB_BYTES
     }
 
-    /**
-     * @param slab Address of the slab
-     * @param addr Address within a chunk (could be in the next slab)
-     * @return The index of the chunk within the slab (could be beyond the end of the slab)
-     */
+    /// Return the index of the chunk that contains `addr` within the slab starting at `slab`
```
k-sareen:
```suggestion
    /// Return the index of the chunk that contains `addr` within the slab starting at `slab`.
```

src/util/rust_util/atomic_box.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// A lazily initialized box.  Similar to an `Option<Box<T>>`, but can be initialized atomically.
+///
+/// It is designed for implementing shared data.  Therefore, methods with `&self`, namely
+/// [`OnceOptionBox::get`] and the [`OnceOptionBox::get_or_init`] methods, only return shared
+/// references to the content (`&T`).  The user should use types that support multi-threaded
+/// accesses, such as mutexes or atomic types, if the inner type is supposed to be modified
+/// concurrently.
+///
+/// Once initialized, this object will own its content.  THe content is allocated in the heap, and
```
k-sareen:
```suggestion
/// Once initialized, this object will own its content.  The content is allocated in the heap, and
```

k-sareen (commented):
Minor typos

wks:
> As you added a bench, did you check how it performs with the two-level mmaper and the old fragmented mmapper?

Here are the results after running on my laptop:

| bench | master | pr |
|---|---|---|
| is_mapped_regular | 3.5942 | 1.2196 |
| is_mapped_large | 64.757 | 1.2077 |
| is_mapped_low | 9.4018 | 1.3892 |
| is_mapped_high | 9.2861 | 1.3844 |
| is_mapped_seq | 245.42 | 23.722 |

The `FragmentedMapper` in the master branch hit the hash conflict in `is_mapped_large` (with search distance of 9) and `is_mapped_seq` (with search distances ranging between 2 to 10).

The `FragmentedMapper` always has to acquire the lock if a slab is not created for an address, even if the hash hits an empty bucket at the first index.  This explains why `is_mapped_low` and `is_mapped_high` is significantly slower than `is_mapped_regular` in the master branch.  For `is_mapped_seq`, the `FragmentedMapper` also frequently hits the case of addresses without a created slab, which makes it slower than it should be.

PR #1340, opened 2025-07-15
Add a stop-the-world, serial Compressor

no-defun-allowed:
This PR adds a [Compressor](https://dl.acm.org/doi/10.1145/1133255.1134023)-esque garbage collector, which is based on MarkCompact but uses a mark bitmap for forwarding metadata like the Compressor.

### DIFF
--- a/src/plan/compressor/global.rs
+++ b/src/plan/compressor/global.rs
@@ -1,15 +1,12 @@
 use super::gc_work::CompressorWorkContext;
 use super::gc_work::{
-    CalculateForwardingAddress,
-    ForwardingProcessEdges,
-    MarkingProcessEdges,
-    Compact,
-    UpdateReferences
+    CalculateForwardingAddress, Compact, ForwardingProcessEdges, MarkingProcessEdges,
+    UpdateReferences,
 };
-use crate::plan::global::{BasePlan, CommonPlan};
+use crate::plan::compressor::mutator::ALLOCATOR_MAPPING;
 use crate::plan::global::CreateGeneralPlanArgs;
 use crate::plan::global::CreateSpecificPlanArgs;
-use crate::plan::compressor::mutator::ALLOCATOR_MAPPING;
+use crate::plan::global::{BasePlan, CommonPlan};
 use crate::plan::AllocationSemantics;
 use crate::plan::Plan;
 use crate::plan::PlanConstraints;
@@ -47,19 +44,19 @@ impl<VM: VMBinding> Plan for Compressor<VM> {
     fn constraints(&self) -> &'static PlanConstraints {
         &COMPRESSOR_CONSTRAINTS
     }
-    
+
     fn collection_required(&self, space_full: bool, _space: Option<SpaceStats<Self::VM>>) -> bool {
         self.base().collection_required(self, space_full)
     }
-    
+
     fn common(&self) -> &CommonPlan<VM> {
         &self.common
     }
-    
+
     fn base(&self) -> &BasePlan<VM> {
         &self.common.base
     }
-    
+
     fn base_mut(&mut self) -> &mut BasePlan<Self::VM> {
         &mut self.common.base
     }
@@ -86,7 +83,7 @@ impl<VM: VMBinding> Plan for Compressor<VM> {
         // TODO use schedule_common once it can work with the Compressor
         // The main issue there is that we need to ForwardingProcessEdges
         // in FinalizableForwarding.
-        
+
         // Stop & scan mutators (mutator scanning can happen before STW)
         scheduler.work_buckets[WorkBucketStage::Unconstrained]
             .add(StopMutators::<CompressorWorkContext<VM>>::new());
@@ -95,11 +92,13 @@ impl<VM: VMBinding> Plan for Compressor<VM> {
         scheduler.work_buckets[WorkBucketStage::Prepare]
             .add(Prepare::<CompressorWorkContext<VM>>::new(self));
 
-        scheduler.work_buckets[WorkBucketStage::CalculateForwarding]
-            .add(CalculateForwardingAddress::<VM>::new(&self.compressor_space));
+        scheduler.work_buckets[WorkBucketStage::CalculateForwarding].add(
+            CalculateForwardingAddress::<VM>::new(&self.compressor_space),
+        );
         // do another trace to update references
         scheduler.work_buckets[WorkBucketStage::SecondRoots].add(UpdateReferences::<VM>::new(self));
-        scheduler.work_buckets[WorkBucketStage::Compact].add(Compact::<VM>::new(&self.compressor_space));
+        scheduler.work_buckets[WorkBucketStage::Compact]
+            .add(Compact::<VM>::new(&self.compressor_space));
 
         // Release global/collectors/mutators
         scheduler.work_buckets[WorkBucketStage::Release]
@@ -166,8 +165,7 @@ impl<VM: VMBinding> Plan for Compressor<VM> {
     }
 
     fn get_used_pages(&self) -> usize {
-        self.compressor_space.reserved_pages()
-            + self.common.get_used_pages()
+        self.compressor_space.reserved_pages() + self.common.get_used_pages()
     }
 }
 
--- a/src/plan/compressor/global.rs
+++ b/src/plan/compressor/global.rs
@@ -1,0 +1,196 @@
+use super::gc_work::CompressorWorkContext;
+use super::gc_work::{
+    CalculateForwardingAddress,
+    ForwardingProcessEdges,
+    MarkingProcessEdges,
+    Compact,
+    UpdateReferences
+};
+use crate::plan::global::{BasePlan, CommonPlan};
+use crate::plan::global::CreateGeneralPlanArgs;
+use crate::plan::global::CreateSpecificPlanArgs;
+use crate::plan::compressor::mutator::ALLOCATOR_MAPPING;
+use crate::plan::AllocationSemantics;
+use crate::plan::Plan;
+use crate::plan::PlanConstraints;
+use crate::policy::compressor::CompressorSpace;
+use crate::policy::space::Space;
+use crate::scheduler::gc_work::*;
+use crate::scheduler::GCWorkScheduler;
+use crate::scheduler::WorkBucketStage;
+use crate::util::alloc::allocators::AllocatorSelector;
+use crate::util::heap::gc_trigger::SpaceStats;
+#[allow(unused_imports)]
+use crate::util::heap::VMRequest;
+use crate::util::metadata::side_metadata::SideMetadataContext;
+use crate::util::opaque_pointer::*;
+use crate::vm::VMBinding;
+use enum_map::EnumMap;
+use mmtk_macros::{HasSpaces, PlanTraceObject};
+
+#[derive(HasSpaces, PlanTraceObject)]
+pub struct Compressor<VM: VMBinding> {
+    #[parent]
+    pub common: CommonPlan<VM>,
+    #[space]
+    pub compressor_space: CompressorSpace<VM>,
+}
+
+/// The plan constraints for the no gc plan.
+pub const COMPRESSOR_CONSTRAINTS: PlanConstraints = PlanConstraints {
+    moves_objects: true,
+    needs_forward_after_liveness: true,
+    ..PlanConstraints::default()
+};
+
+impl<VM: VMBinding> Plan for Compressor<VM> {
+    fn constraints(&self) -> &'static PlanConstraints {
+        &COMPRESSOR_CONSTRAINTS
+    }
+    
+    fn collection_required(&self, space_full: bool, _space: Option<SpaceStats<Self::VM>>) -> bool {
+        self.base().collection_required(self, space_full)
+    }
+    
+    fn common(&self) -> &CommonPlan<VM> {
+        &self.common
+    }
+    
+    fn base(&self) -> &BasePlan<VM> {
+        &self.common.base
+    }
+    
+    fn base_mut(&mut self) -> &mut BasePlan<Self::VM> {
+        &mut self.common.base
+    }
+
+    fn prepare(&mut self, tls: VMWorkerThread) {
+        self.common.prepare(tls, true);
+        self.compressor_space.prepare();
+    }
+
+    fn release(&mut self, tls: VMWorkerThread) {
+        self.common.release(tls, true);
+        self.compressor_space.release();
+    }
+
+    fn end_of_gc(&mut self, tls: VMWorkerThread) {
+        self.common.end_of_gc(tls);
+    }
+
+    fn get_allocator_mapping(&self) -> &'static EnumMap<AllocationSemantics, AllocatorSelector> {
+        &ALLOCATOR_MAPPING
+    }
+
+    fn schedule_collection(&'static self, scheduler: &GCWorkScheduler<VM>) {
+        // TODO use schedule_common once it can work with the Compressor
+        // The main issue there is that we need to ForwardingProcessEdges
+        // in FinalizableForwarding.
+        
+        // Stop & scan mutators (mutator scanning can happen before STW)
+        scheduler.work_buckets[WorkBucketStage::Unconstrained]
+            .add(StopMutators::<CompressorWorkContext<VM>>::new());
+
+        // Prepare global/collectors/mutators
+        scheduler.work_buckets[WorkBucketStage::Prepare]
+            .add(Prepare::<CompressorWorkContext<VM>>::new(self));
+
+        scheduler.work_buckets[WorkBucketStage::CalculateForwarding]
+            .add(CalculateForwardingAddress::<VM>::new(&self.compressor_space));
+        // do another trace to update references
+        scheduler.work_buckets[WorkBucketStage::SecondRoots].add(UpdateReferences::<VM>::new(self));
+        scheduler.work_buckets[WorkBucketStage::Compact].add(Compact::<VM>::new(&self.compressor_space));
+
+        // Release global/collectors/mutators
+        scheduler.work_buckets[WorkBucketStage::Release]
+            .add(Release::<CompressorWorkContext<VM>>::new(self));
+
+        // Reference processing
+        if !*self.base().options.no_reference_types {
+            use crate::util::reference_processor::{
+                PhantomRefProcessing, SoftRefProcessing, WeakRefProcessing,
+            };
+            scheduler.work_buckets[WorkBucketStage::SoftRefClosure]
+                .add(SoftRefProcessing::<MarkingProcessEdges<VM>>::new());
+            scheduler.work_buckets[WorkBucketStage::WeakRefClosure]
+                .add(WeakRefProcessing::<VM>::new());
+            scheduler.work_buckets[WorkBucketStage::PhantomRefClosure]
+                .add(PhantomRefProcessing::<VM>::new());
+
+            use crate::util::reference_processor::RefForwarding;
+            scheduler.work_buckets[WorkBucketStage::RefForwarding]
+                .add(RefForwarding::<ForwardingProcessEdges<VM>>::new());
+
+            use crate::util::reference_processor::RefEnqueue;
+            scheduler.work_buckets[WorkBucketStage::Release].add(RefEnqueue::<VM>::new());
+        }
+
+        // Finalization
+        if !*self.base().options.no_finalizer {
+            use crate::util::finalizable_processor::{Finalization, ForwardFinalization};
+            // finalization
+            // treat finalizable objects as roots and perform a closure (marking)
+            // must be done before calculating forwarding pointers
+            scheduler.work_buckets[WorkBucketStage::FinalRefClosure]
+                .add(Finalization::<MarkingProcessEdges<VM>>::new());
+            // update finalizable object references
+            // must be done before compacting
+            scheduler.work_buckets[WorkBucketStage::FinalizableForwarding]
+                .add(ForwardFinalization::<ForwardingProcessEdges<VM>>::new());
+        }
+
+        // VM-specific weak ref processing
+        scheduler.work_buckets[WorkBucketStage::VMRefClosure]
+            .set_sentinel(Box::new(VMProcessWeakRefs::<MarkingProcessEdges<VM>>::new()));
+
+        // VM-specific weak ref forwarding
+        scheduler.work_buckets[WorkBucketStage::VMRefForwarding]
+            .add(VMForwardWeakRefs::<ForwardingProcessEdges<VM>>::new());
+
+        // VM-specific work after forwarding, possible to implement ref enququing.
+        scheduler.work_buckets[WorkBucketStage::Release].add(VMPostForwarding::<VM>::default());
+
+        // Analysis GC work
+        #[cfg(feature = "analysis")]
+        {
+            use crate::util::analysis::GcHookWork;
+            scheduler.work_buckets[WorkBucketStage::Unconstrained].add(GcHookWork);
+        }
+        #[cfg(feature = "sanity")]
+        scheduler.work_buckets[WorkBucketStage::Final]
+            .add(crate::util::sanity::sanity_checker::ScheduleSanityGC::<Self>::new(self));
+    }
+
+    fn current_gc_may_move_object(&self) -> bool {
+        true
+    }
+
+    fn get_used_pages(&self) -> usize {
+        self.compressor_space.reserved_pages()
+            + self.common.get_used_pages()
+    }
+}
+
+impl<VM: VMBinding> Compressor<VM> {
+    pub fn new(args: CreateGeneralPlanArgs<VM>) -> Self {
+        let mut plan_args = CreateSpecificPlanArgs {
+            global_args: args,
+            constraints: &COMPRESSOR_CONSTRAINTS,
+            global_side_metadata_specs: SideMetadataContext::new_global_specs(&[]),
+        };
+
+        let res = Compressor {
+            compressor_space: CompressorSpace::new(plan_args.get_space_args(
+                "compressor_space",
+                true,
+                false,
+                VMRequest::discontiguous(),
+            )),
+            common: CommonPlan::new(plan_args),
+        };
+
+        res.verify_side_metadata_sanity();
+
+        res
+    }
+}
--- a/src/plan/compressor/mutator.rs
+++ b/src/plan/compressor/mutator.rs
@@ -1,0 +1,60 @@
+use crate::plan::mutator_context::common_prepare_func;
+use crate::plan::mutator_context::Mutator;
+use crate::plan::mutator_context::MutatorBuilder;
+use crate::plan::mutator_context::MutatorConfig;
+use crate::plan::mutator_context::{
+    create_space_mapping, ReservedAllocators,
+};
+use crate::plan::compressor::Compressor;
+use crate::plan::AllocationSemantics;
+use crate::util::alloc::BumpAllocator;
+use crate::util::alloc::allocators::AllocatorSelector;
+use crate::util::{VMMutatorThread, VMWorkerThread};
+use crate::vm::VMBinding;
+use crate::MMTK;
+use enum_map::{enum_map, EnumMap};
+
+const RESERVED_ALLOCATORS: ReservedAllocators = ReservedAllocators {
+    n_bump_pointer: 1,
+    ..ReservedAllocators::DEFAULT
+};
+
+lazy_static! {
+    pub static ref ALLOCATOR_MAPPING: EnumMap<AllocationSemantics, AllocatorSelector> = enum_map! {
+        _ => AllocatorSelector::BumpPointer(0),
+    };
+}
+
+pub fn create_compressor_mutator<VM: VMBinding>(
+    mutator_tls: VMMutatorThread,
+    mmtk: &'static MMTK<VM>,
+) -> Mutator<VM> {
+    let plan = mmtk.get_plan().downcast_ref::<Compressor<VM>>().unwrap();
+    let config = MutatorConfig {
+        allocator_mapping: &ALLOCATOR_MAPPING,
+        space_mapping: Box::new({
+            let mut vec = create_space_mapping(RESERVED_ALLOCATORS, false, plan);
+            vec.push((AllocatorSelector::BumpPointer(0), &plan.compressor_space));
+            vec
+        }),
+        prepare_func: &common_prepare_func,
+        release_func: &compressor_mutator_release,
+    };
+
+    let builder = MutatorBuilder::new(mutator_tls, mmtk, config);
+    builder.build()
+}
+
+pub fn compressor_mutator_release<VM: VMBinding>(mutator: &mut Mutator<VM>, _tls: VMWorkerThread) {
+    // reset the thread-local allocation bump pointer
+    let bump_allocator = unsafe {
+        mutator
+            .allocators
+            .get_allocator_mut(mutator.config.allocator_mapping[AllocationSemantics::Default])
+    }
+    .downcast_mut::<BumpAllocator<VM>>()
+    .unwrap();
+    bump_allocator.reset();
+    // This doesn't do anything sensible.
+    //common_release_func(mutator, _tls);
+}
--- a/src/plan/compressor/mutator.rs
+++ b/src/plan/compressor/mutator.rs
@@ -1,14 +1,12 @@
+use crate::plan::compressor::Compressor;
 use crate::plan::mutator_context::common_prepare_func;
 use crate::plan::mutator_context::Mutator;
 use crate::plan::mutator_context::MutatorBuilder;
 use crate::plan::mutator_context::MutatorConfig;
-use crate::plan::mutator_context::{
-    create_space_mapping, ReservedAllocators,
-};
-use crate::plan::compressor::Compressor;
+use crate::plan::mutator_context::{create_space_mapping, ReservedAllocators};
 use crate::plan::AllocationSemantics;
-use crate::util::alloc::BumpAllocator;
 use crate::util::alloc::allocators::AllocatorSelector;
+use crate::util::alloc::BumpAllocator;
 use crate::util::{VMMutatorThread, VMWorkerThread};
 use crate::vm::VMBinding;
 use crate::MMTK;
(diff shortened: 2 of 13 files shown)


src/plan/compressor/global.rs
```diff
@@ -0,0 +1,194 @@
...
+
+#[derive(HasSpaces, PlanTraceObject)]
+pub struct Compressor<VM: VMBinding> {
+    #[parent]
+    pub common: CommonPlan<VM>,
+    #[space]
+    pub compressor_space: CompressorSpace<VM>,
+}
+
+/// The plan constraints for the no gc plan.
```
k-sareen: Update comments here?

src/plan/compressor/mutator.rs
```diff
@@ -0,0 +1,58 @@
...
+use enum_map::{enum_map, EnumMap};
+
+const RESERVED_ALLOCATORS: ReservedAllocators = ReservedAllocators {
+    n_bump_pointer: 1,
+    ..ReservedAllocators::DEFAULT
+};
+
+lazy_static! {
+    pub static ref ALLOCATOR_MAPPING: EnumMap<AllocationSemantics, AllocatorSelector> = enum_map! {
+        _ => AllocatorSelector::BumpPointer(0),
```
k-sareen, line 25: This should only override `AllocationSemantics::Default`. Currently this means you are not using a LOS.

src/plan/compressor/mutator.rs
```diff
@@ -0,0 +1,58 @@
...
+    let bump_allocator = unsafe {
+        mutator
+            .allocators
+            .get_allocator_mut(mutator.config.allocator_mapping[AllocationSemantics::Default])
+    }
+    .downcast_mut::<BumpAllocator<VM>>()
+    .unwrap();
+    bump_allocator.reset();
+    // This doesn't do anything sensible.
+    //common_release_func(mutator, _tls);
```
k-sareen:
```suggestion
```

src/plan/compressor/global.rs
```diff
@@ -0,0 +1,194 @@
...
+#[allow(unused_imports)]
+use crate::util::heap::VMRequest;
+use crate::util::metadata::side_metadata::SideMetadataContext;
+use crate::util::opaque_pointer::*;
+use crate::vm::VMBinding;
+use enum_map::EnumMap;
+use mmtk_macros::{HasSpaces, PlanTraceObject};
+
+#[derive(HasSpaces, PlanTraceObject)]
+pub struct Compressor<VM: VMBinding> {
```
k-sareen, line 33: Could you add the paper reference here as well?

src/policy/compressor/compressorspace.rs
```diff
@@ -0,0 +1,368 @@
...
+                });
+            for (region_start, size) in self.pr.iterate_allocated_regions() {
+                crate::util::metadata::vo_bit::bzero_vo_bit(region_start, size);
+            }
+        }
+
+        self.forwarding
+            .scan_marked_objects(start, end, &mut |obj: ObjectReference| {
+                let copied_size = VM::VMObjectModel::get_size_when_copied(obj);
+                debug_assert!(copied_size == VM::VMObjectModel::get_current_size(obj));
```
k-sareen, line 324: You should add a comment here or print in the assertion why this is required. This is required, to my understanding, because the end-bit has been set for the current size.

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,165 @@
...
+
+pub struct ForwardingMetadata<VM: VMBinding> {
+    mark_bit_spec: SideMetadataSpec,
+    pub(crate) first_address: Address,
+    calculated: AtomicBool,
+    block_offsets: Vec<AtomicUsize>,
+    vm: PhantomData<VM>,
+}
+
+const BLOCK_SIZE: usize = 512;
```
k-sareen: Could you add a comment here defining "block size"

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    pub fn mark_end_of_object(&self, object: ObjectReference) {
+        use crate::util::metadata::side_metadata::{address_to_meta_address, meta_byte_lshift};
+        let end_of_object =
+            object.to_raw_address() + VM::VMObjectModel::get_current_size(object) - MIN_OBJECT_SIZE;
+        let a1 = address_to_meta_address(&self.mark_bit_spec, object.to_raw_address());
+        let s1 = meta_byte_lshift(&self.mark_bit_spec, object.to_raw_address());
+        let a2 = address_to_meta_address(&self.mark_bit_spec, end_of_object);
+        let s2 = meta_byte_lshift(&self.mark_bit_spec, end_of_object);
```
k-sareen: You can put these behind a `#[cfg(debug_assertions)] {}`. Though hopefully the compiler is smart enough to remove them in the release build.

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,165 @@
...
+        debug_assert!((a1, s1) < (a2, s2));
```
k-sareen: Could you add a comment or print in the assertion to mention why this is required. It might not be clear to someone who's unfamiliar with the metadata layout in MMTk.

src/plan/compressor/global.rs
no-defun-allowed: The tutorial doesn't mention that there's a "no gc" too :)

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,177 @@
...
+            first_address: start,
+            calculated: AtomicBool::new(false),
+            block_offsets,
+            vm: PhantomData,
+        }
+    }
+
+    pub fn mark_end_of_object(&self, object: ObjectReference) {
+        let end_of_object =
+            object.to_raw_address() + VM::VMObjectModel::get_current_size(object) - MIN_OBJECT_SIZE;
```
qinsoon: Why do we need to minus `MIN_OBJECT_SIZE`? It looks like you try to compute an address that is guaranteed inside the object.

Also the start of the allocation should be` object.to_object_start::<VM>()` rather than `object.to_raw_address()` if you want to add the size to it.

src/policy/compressor/compressorspace.rs
```diff
@@ -0,0 +1,372 @@
...
+        _worker: &mut GCWorker<VM>,
+    ) -> ObjectReference {
+        debug_assert!(
+            KIND != TRACE_KIND_TRANSITIVE_PIN,
+            "Compressor does not support transitive pin trace."
+        );
+        if KIND == TRACE_KIND_MARK {
+            self.trace_mark_object(queue, object)
+        } else if KIND == TRACE_KIND_FORWARD {
+            self.trace_forward_object(queue, object)
```
qinsoon: This looks like a second transitive closure. I assume this only happens for roots. Right? If so, I suggest making it much clearer (e.g. `TRACE_KIND_FORWARD` -> `TRACE_KIND_FORWARD_ROOT`, `self.trace_forward_object` -> `self.forward_root`, etc), and making the code more distinguishable from mark compact.

no-defun-allowed:
> You can cherry pick this commit to the PR: [qinsoon@ac90d4b](https://github.com/qinsoon/mmtk-core/commit/ac90d4b95e990f6417a5f9f9465e8d202b3fb6e3).

Thanks!

> It runs mock VM with side metadata. Compressor currently fails for tests that are not using contiguous space.

Right. Should we skip such tests when running with the Compressor?

src/policy/compressor/forwarding.rs
k-sareen: You need to set the bit at the end of the object. Consider an object at `0x10000000` of size `0x10`, then we want the mark bits to be set for `0x10000000` and `0x1000000f`, not `0x10000000` and `0x10000010`. Since each machine word (8-byte on 64-bits and 4-byte on 32-bits) maps down to a single bit, it's sufficient to subtract the minimum object size to set the bit there.

src/plan/compressor/mutator.rs
```diff
@@ -0,0 +1,68 @@
...
+use crate::MMTK;
+use enum_map::{enum_map, EnumMap};
+
+const RESERVED_ALLOCATORS: ReservedAllocators = ReservedAllocators {
+    n_bump_pointer: 1,
+    ..ReservedAllocators::DEFAULT
+};
+
+lazy_static! {
+    /// When nogc_multi_space is disabled, force all the allocation go to the default allocator and space.
```
k-sareen:
```suggestion
    /// When using `compressor_single_space`, force all the allocations to go to the default allocator and space.
```

src/policy/compressor/forwarding.rs
k-sareen: Notably, this requires the start and end of the object to map to different bits. Hence the debug assertion in the function.

src/policy/compressor/compressorspace.rs
no-defun-allowed: Right and right.

src/policy/compressor/forwarding.rs
qinsoon: > You need to set the bit at the end of the object. Consider an object at 0x10000000 of size 0x10, then we want the mark bits to be set for 0x10000000 and 0x1000000f, not 0x10000000 and 0x10000010. 

Then why not just use `-1`?

> Since each machine word (8-byte on 64-bits and 4-byte on 32-bits) maps down to a single bit, it's sufficient to subtract the minimum object size to set the bit there.

It seems you are assuming the min object size is one word, which might not be always true in the future. You could also use `-word_size`.

k-sareen: > It seems you are assuming the min object size is one word, which might not be always true in the future.

Yes this is a strong assumption the code is making, hence the debug assertion. I was the one who recommended @no-defun-allowed to use the predefined mark-bit, but I guess if we want to be more explicit, we could use a new metadata (like `COMPRESSOR_MARK_BIT_SPEC`) where the region size is set to machine word size. This could then also allow Compressor to work without the side mark-bits since it maintains its own metadata. What do you think @no-defun-allowed @qinsoon?

k-sareen:
> The block offset vector is currently a Rust Vec<AtomicUsize> which I do the address-to-block calculation assuming a contiguous heap; I should be using some other metadata mechanism for that anyway.

For what its worth, ART also uses a [`uint32_t*`](https://android.googlesource.com/platform/art/+/6484611fd45e69db9f33f98bfd6864014b030ecf/runtime/gc/collector/mark_compact.h#914) for the offset vector calculation.

src/policy/compressor/forwarding.rs
qinsoon: > > It seems you are assuming the min object size is one word, which might not be always true in the future.
> 
> Yes this is a strong assumption the code is making, hence the debug assertion. I was the one who recommended @no-defun-allowed to use the predefined mark-bit, but I guess if we want to be more explicit, we could use a new metadata (like `COMPRESSOR_MARK_BIT_SPEC`) where the region size is set to machine word size. This could then also allow Compressor to work without the side mark-bits since it maintains its own metadata. What do you think @no-defun-allowed @qinsoon?

I see. The mark bit is per 'min object size', thus you have to minus the min object size.

Yeah, using a separate side metadata might be cleaner. For compressor, you also mark both the start and the end of an object -- this is not normal mark bits anyway.

no-defun-allowed:
Now the block offset vector and the marks are in their own local metadata. I'm still undecided on if and how to support discontiguous spaces, since the forwarding algorithm in the Compressor paper seems very tied to having a contiguous space.

qinsoon:
The failures for 32 bits was caused by this: https://github.com/mmtk/mmtk-core/pull/1342. However, you reverted the side metadata commit for mock VM, so the bug won't appear in the PR any way.

src/util/metadata/side_metadata/spec_defs.rs
```diff
@@ -94,6 +94,10 @@ define_side_metadata_specs!(
     MS_LOCAL_FREE   = (global: false, log_num_of_bits: LOG_BITS_IN_ADDRESS, log_bytes_in_region: crate::policy::marksweepspace::native_ms::Block::LOG_BYTES),
     // First cell of thread free list in block for native mimalloc
     MS_THREAD_FREE  = (global: false, log_num_of_bits: LOG_BITS_IN_ADDRESS, log_bytes_in_region: crate::policy::marksweepspace::native_ms::Block::LOG_BYTES),
+    // Start and end marks by Compressor
+    COMPRESSOR_MARK = (global: false, log_num_of_bits: 0, log_bytes_in_region: LOG_MIN_OBJECT_SIZE as usize),
```
k-sareen: You can say `LOG_BYTES_IN_WORD` to be more specific. You can also mention this in the comment above (1 bit per machine word).

no-defun-allowed:
I added a `MMTK_PLAN=discontiguous` set to the CI script, for tests which require discontiguous heaps. The script also removes the Compressor from `all` on i686, since i686 defaults to a discontiguous heap too; some tests like `test_vm_layout_default` should still be tested on x86_64 but will panic on i686.

k-sareen:
I've made an issue for the commit that fixes the space mapping since it's a very subtle bug that I think can be checked with assertions (https://github.com/mmtk/mmtk-core/pull/1340/commits/cd3bde3504dfdbe45f68590c82cecdc8269e8c0f): https://github.com/mmtk/mmtk-core/issues/1343

src/plan/compressor/mutator.rs
k-sareen: Actually, this is my bad. I think we need to call the `common_release_func` because otherwise the Compressor will not work with [non-moving Immix as the non-moving space otherwise](https://github.com/mmtk/mmtk-core/blob/31a78a41f02fc7228780b501c4944ba750e32ee4/src/plan/mutator_context.rs#L64-L69). This only works because the default currently is immortal.

no-defun-allowed: Yeah, `common_release_func` probably didn't work before because I messed up initialising the common spaces, so I removed the call then. I put the call back.

src/util/metadata/side_metadata/spec_defs.rs
k-sareen: @no-defun-allowed Could you make this more explicit? Like Yi said, the min object size is subject to change. Just say `LOG_BYTES_IN_WORD`.

Cargo.toml
```diff
@@ -197,6 +197,9 @@ extreme_assertions = []
 # Enable multiple spaces for NoGC, each allocator maps to an individual ImmortalSpace.
 nogc_multi_space = []
 
+# Disable multiple spaces for Compressor.
```
qinsoon, line 200: Is there any reason that we want to keep this feature? Would it be useful to anyone?

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,165 @@
...
+use std::sync::atomic::AtomicBool;
+
+/// A finite-state machine which processes the positions of mark bits,
+/// and accumulates the size of live data that it has seen.
+///
+/// The Compressor caches the state of the transducer at the start of
+/// each block by serialising the state using [`Transducer::encode`];
+/// the state can then be deserialised using [`Transducer::decode`].
+#[derive(Debug)]
+struct Transducer {
```
qinsoon, line 23: It is a bit hard to understand how this type works. More comments would help. For example, explain what each fields mean. Also you could comment `step` to say that `address` is always either the start of an object, or the end of an object, etc.

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub fn new() -> Self {
+        Self {
+            live: 0,
+            last_bit_seen: Address::ZERO,
+            in_object: false,
+        }
+    }
+    pub fn step(&mut self, address: Address) {
+        if self.in_object {
+            self.live += address - self.last_bit_seen + MIN_OBJECT_SIZE;
```
qinsoon: Why do we still need `MIN_OBJECT_SIZE` here?

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,165 @@
...
+            first_address: start,
+            calculated: AtomicBool::new(false),
+            vm: PhantomData,
+        }
+    }
+
+    pub fn mark_end_of_object(&self, object: ObjectReference) {
+        let end_of_object = object.to_object_start::<VM>()
+            + VM::VMObjectModel::get_current_size(object)
+            - MIN_OBJECT_SIZE;
```
qinsoon: If the object size is exactly `MIN_OBJECT_SIZE`, then you would get the start of the object rather than the end of the object.

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,165 @@
...
+        #[cfg(debug_assertions)]
+        {
+            use crate::util::metadata::side_metadata::{address_to_meta_address, meta_byte_lshift};
+            // We require to be able to iterate upon start and end bits in the
+            // same bitmap. Therefore the start and end bits cannot be the
+            // same, else we would only encounter one of the two bits.
+            let a1 = address_to_meta_address(&MARK_SPEC, object.to_raw_address());
```
qinsoon: Could add a function to `SideMetadataSpec`, such as `fn are_separate_bits(&self, addr1, addr2) -> bool`. In that case, we don't need to expose `meta_byte_lshift` as `pub(crate)`.

qinsoon (commented):
I found the `Transducer` implementation is not easy to understand. A bit more comments would be helpful.

We probably also need a PR for OpenJDK that supports Compressor.

qinsoon:
I also requested reviews from @wks

Cargo.toml
k-sareen: This will be the most space-efficient an algorithm can be. I believe Steve mentioned it.

src/policy/compressor/forwarding.rs
no-defun-allowed: The size of an object is the difference between its end and its start; the last bit is the previous word from the end, so the distance between the two mark bits is off by one word. I'll comment this too.

src/policy/compressor/forwarding.rs
no-defun-allowed: Yes, an object has to be at least two words by my reading of the Compressor paper. The debug assertion checks that the bits for the first and last word of the object are different; following from [my comment on the off-by-one in `step`](https://github.com/mmtk/mmtk-core/pull/1340#discussion_r2218295927) I think I should call it the last word rather than the end too, to be clearer about the off-by-one.

src/policy/compressor/forwarding.rs
k-sareen: I think you would need to replace `MIN_OBJECT_SIZE` instances with `BYTES_IN_WORD` to be consistent given your metadata has changed its definition. Then it would clear up the confusions with `MIN_OBJECT_SIZE`.

qinsoon: > The size of an object is the difference between its end and its start; the last bit is the previous word from the end, so the distance between the two mark bits is off by one word. I'll comment this too.

So if it is currently 'in_object', the address would be the 'end address', and the 'last_bit_seen' would be the start of the object. The diff plus 'MIN_OBJECT_SIZE' (or whatever you use to calculate the end address) would be the object size. Right?

You can make this more clear with code. It is more verbose, but clearer.
```rust
if self.in_object {
  let obj_end = address;
  let obj_start = self.last_bit_seen;
  let obj_size = obj_end - obj_start + X;
  self.live += obj_size;
}
```

Cargo.toml
k-sareen: @no-defun-allowed Just mention in the comment there that this is configuration will move all objects. Don't use if you can't tolerate that. This is primarily for understanding how space-efficient we can be.

src/policy/compressor/forwarding.rs
no-defun-allowed: > Right?

Right. I like this idea.

src/policy/compressor/forwarding.rs
no-defun-allowed: Added in 1ef15cf

src/policy/compressor/forwarding.rs
no-defun-allowed: I've done this in 1ef15cf

Cargo.toml
qinsoon: Right. Also the non-moving semantic can't be used with this feature.

It would be better if we can panic when the non-moving semantic is ever used. This probably would need a 'fake allocator' that panics when used, and the non-moving semantic is mapped to it. I am not sure if it is worth it.

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,176 @@
...
+            - BYTES_IN_WORD;
+        #[cfg(debug_assertions)]
+        {
+            // We require to be able to iterate upon first and last bits in the
+            // same bitmap. Therefore the first and last bits cannot be the
+            // same, else we would only encounter one of the two bits.
+            // This requirement implies that objects must be at least two words
+            // large.
+            debug_assert!(
+                MARK_SPEC.are_different_metadata(object.to_raw_address(), last_word_of_object),
```
k-sareen:
```suggestion
                MARK_SPEC.are_different_metadata(object.to_object_start::<VM>(), last_word_of_object),
```
to be consistent with the calculation above.

src/util/metadata/side_metadata/global.rs
```diff
@@ -131,6 +131,15 @@ impl SideMetadataSpec {
         );
     }
 
+    #[cfg(debug_assertions)]
+    pub(crate) fn are_different_metadata(&self, addr1: Address, addr2: Address) -> bool {
```
k-sareen:
```suggestion
    pub(crate) fn are_different_metadata_bits(&self, addr1: Address, addr2: Address) -> bool {
```
Just to be more precise. It's a bit confusing otherwise.

k-sareen:
I'm happy to approve this PR unless @wks you have some concerns?

qinsoon:
> I'm happy to approve this PR unless @wks you have some concerns?

I think we want to have an OpenJDK PR to support and test Compressor as well (I assume that is already done).

k-sareen:
mmtk-openjdk PR: https://github.com/mmtk/mmtk-openjdk/pull/310

qinsoon:
binding-refs
OPENJDK_BINDING_REF=add-compressor-tests

wks:
> I'm happy to approve this PR unless @wks you have some concerns?

Sorry about the delay.  I'll review this PR now.

wks:
A high-level question:  Will this algorithm work if some objects have stricter alignment requirements than others?  For example, object A is 8-byte aligned, and object B is 16-byte aligned, but A is 40 bytes in size, not a multiple of 16 bytes.  They are laid out like this.  `A` and `B` are bits in the object. `.` is the gap between objects.  `m` is the mark bit.

```
0        8        16       24       32       40       48       56       64       
AAAAAAAA AAAAAAAA AAAAAAAA AAAAAAAA AAAAAAAA ........ BBBBBBBB BBBBBBBB
m                                   m                 m        m
```

The algorithm described in `forwarding::Transducer` will think there are 40 live bytes before object B.  But since B requires 16-byte alignment, it cannot be moved to offset 40.  It must be moved to offset 48 instead.

FYI, the existing mark compact space adds padding bits for the alignment when calculating forwarding addresses.

```rust
// MarkCompactSpace::calculating_forwarding_pointer
            for obj in self
                .linear_scan_objects(from_start..from_end)
                .filter(Self::to_be_compacted)
            {
                let copied_size =
                    VM::VMObjectModel::get_size_when_copied(obj) + Self::HEADER_RESERVED_IN_BYTES;
                let align = VM::VMObjectModel::get_align_when_copied(obj);
                let offset = VM::VMObjectModel::get_align_offset_when_copied(obj);
                // move to_cursor to aliged start address
                to_cursor = align_allocation_no_fill::<VM>(to_cursor, align, offset);
                // move to next to-block if there is no sufficient memory in current region
```

If Compressor cannot handle this case and it is hard to fix, we can temporarily check `VMBinding::MAX_ALIGNMENT` when initializing the `Compressor` plan and panic if we cannot support that VM.

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,179 @@
...
+        self.calculated.store(false, Ordering::Relaxed);
+    }
+
+    pub fn forward(&self, address: Address) -> Address {
+        debug_assert!(
+            self.calculated.load(Ordering::Relaxed),
+            "forward() should only be called when we have calculated an offset vector"
+        );
+        let block_number = (address - self.first_address) / BLOCK_SIZE;
+        let block_address = self.first_address + (block_number * BLOCK_SIZE);
```
wks: Looks like `block_address` is just `address.align_down(BLOCK_SIZE)`.  But you can also define a `Block` type, as I mentioned in another comment.

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,179 @@
...
+        MARK_SPEC.fetch_or_atomic(last_word_of_object, GC_MARK_BIT_MASK, Ordering::SeqCst);
+    }
+
+    pub fn calculate_offset_vector(&self, pr: &MonotonePageResource<VM>) {
+        let mut state = Transducer::new();
+        let last_block = (pr.cursor() - self.first_address) / BLOCK_SIZE;
+        debug!("calculating offset of {last_block} blocks");
+        for block in 0..last_block {
+            let block_start = self.first_address + (block * BLOCK_SIZE);
+            let block_end = block_start + BLOCK_SIZE;
```
wks: There is a lot of address computation involved.  You can define a `Block(Address)` type specific to the `CompressorSpace`, let it implement `crate::util::linear_scan::Region`, and you can use the `RegionIterator` type.  It will simplify things a lot.

no-defun-allowed:
My code doesn't check alignment at all, and I'm not sure if there is a good way to implement variable alignment; we don't get much freedom in computing forwarding pointers from the mark bitmap. (The same applies to supporting discontiguous heaps with one Compressor setup/without regions.)

Is the size of an object always a multiple of its alignment? If not, I probably should take the alignment into account when determining where the last word of an object is.

src/policy/compressor/compressorspace.rs
```diff
@@ -0,0 +1,334 @@
...
+        &self,
+        _queue: &mut Q,
+        object: ObjectReference,
+    ) -> ObjectReference {
+        self.forward(object, true)
+    }
+
+    pub fn test_and_mark(object: ObjectReference) -> bool {
+        let old = forwarding::MARK_SPEC.fetch_or_atomic(
+            object.to_raw_address(),
```
wks, line 238: Here you set the mark bit on the raw address.  But the raw address may not be the starting address of an object (unless `UNIFIED_OBJECT_REFERENCE_ADDRESS` is true).

src/policy/compressor/forwarding.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub fn scan_marked_objects(
+        &self,
+        start: Address,
+        end: Address,
+        f: &mut impl FnMut(ObjectReference),
+    ) {
+        let mut in_object = false;
+        MARK_SPEC.scan_non_zero_values::<u8>(start, end, &mut |addr: Address| {
+            if !in_object {
+                let object = ObjectReference::from_raw_address(addr).unwrap();
```
wks, line 186: Note that the starting address may not be the raw address of `ObjectReference` (unless `UNIFIED_OBJECT_REFERENCE_ADDRESS` is true).  If the mark bits are set on the starting and the ending words of an object, we will not be able to obtain `ObjectReference` from the starting address alone.  But if we also use the VO bits, we can scan the VO bits to identify the `ObjectReference`.

Note that the existing MarkCompactSpace uses the VO bits as a local metadata even when we don't enable global VO bits.

Also note that we have `ObjectModel::get_reference_when_copied_to(from: ObjectReference, to: Address) -> ObjectReference;`.  It doesn't work here.  It needs a `from` reference which points to the object before copying in order to get the type information and compute the distance between the starting address and the `ObjectReference`.

src/policy/compressor/forwarding.rs
no-defun-allowed: That is much nicer, thanks.

wks (commented):
In MMTk, the starting address of an object is not necessarily the raw address of the `ObjectReference`.  See the doc comment of `ObjectReference` in `src/util/address.rs`.  Currently, JikesRVM may add a hash code in front of an object to implement address-based hashing.  CRuby currently prepends a word to store the object size (which is a workaround and is supposed to be removed in the future).

But some VMs guarantee that the starting address is always the `ObjectReference`'s raw address.  They will set `ObjectModel::UNIFIED_OBJECT_REFERENCE_ADDRESS` to `true`, and it will simplify some algorithms inside mmtk-core.  Currently the OpenJDK guarantees this.

So you may check if `UNIFIED_OBJECT_REFERENCE_ADDRESS` is true and reject unsupported bindings when using Compressor.

You can also try to fix the code to support VMs where `ObjectReference` is not necessarily the starting address.  That may involve setting the mark bits at the starting address instead of at `ObjectReference`, and scanning the VO bits in addition to the mark bits to find the `ObjectReference` when copying.

wks:
> > I'll assert `UNIFIED_OBJECT_REFERENCE_ADDRESS` for now.
> 
> But then I can't run mock VM tests with the Compressor, and adding `const UNIFIED_OBJECT_REFERENCE_ADDRESS: bool = true;` to MockVM produces this test failure:
> 
> ```
> ---- vm::tests::mock_tests::mock_test_internal_ptr_normal_object::interior_pointer_in_normal_object stdout ----
> start = 0x20000000000, end = 0x20000000010, obj = 0x20000000008
> thread 'vm::tests::mock_tests::mock_test_internal_ptr_normal_object::interior_pointer_in_normal_object' panicked at src/util/address.rs:627:9:
> The binding claims unified object reference address, but for object reference 0x20000000008, ref_to_object_start() returns 0x20000000000
> ```

Oops.  I have some test cases for the cases where the starting address is not equal to the `ObjectReference`.

Maybe we can use Cargo features to conditionally enable compressor-related tests and set `UNIFIED_OBJECT_REFERENCE_ADDRESS` to true, and conditionally disable those tests that conflict with those options.

k-sareen:
Maybe this can be a `warn!()` about if `UNIFIED_OBJECT_REFERENCE_ADDRESS` is not true. Might let us side-step this issue in the short-term.

qinsoon:
So currently the compressor has the following requirements/limitations:
* Requires contiguous space
* Requires object references to be the same as the object start addresses
* Ignores object alignment (assumes objects are aligned to word-size?)

I feel it is worth discussing over these points in a meeting, and see what are intrinsic to the algorithm, and what are introduced due to the implementation.

qinsoon:
The following is my understanding about the limitations:

---
>     * Requires contiguous space

Supporting discontiguous space seems doable, but it complicates things, especially the forwarding part.

First, encoding should be fine with discontiguous space, as the offset vector itself does not include any address. During computing offset vector, the current code uses the next block's start address to calculate the remaining live bytes in the previous block -- this assumes contiguous spaces. We could visit the block end address once to calculate the remaining live bytes. This should be easy to fix.
https://github.com/mmtk/mmtk-core/blob/58458361d76b75ed3848ca04c77c1ac4e801f0d0/src/policy/compressor/forwarding.rs#L61

Then, forwarding. The forwarding address is computed as below:
https://github.com/mmtk/mmtk-core/blob/58458361d76b75ed3848ca04c77c1ac4e801f0d0/src/policy/compressor/forwarding.rs#L174
To support discontiguous space, we cannot use the start of the space, and we need some extra information (e.g. a table that maps block index to block start) to know the current block start, then apply the offset internal to a block to it -- `state.live % block_size`.

---
>     * Requires object references to be the same as the object start addresses

This should be purely an implementation limitation. In theory, the implementation could always use `object.to_object_start()` and never use `object.to_raw_address()`. And when it needs to return an object referencefrom a forwarded address, use `ObjectModel::get_reference_when_copied_to`.

---
>     * Ignores object alignment (assumes objects are aligned to word-size?)

It is also possible to properly support object alignment.

When we encounter an object during computing offset vector, we can compute the required alignment in the same way as mark compact. The diff between the aligned to-cursor and the previous to-cursor is the alignment padding that can be added to the live bytes prior to the object.

https://github.com/mmtk/mmtk-core/blob/31a78a41f02fc7228780b501c4944ba750e32ee4/src/policy/markcompactspace.rs#L377-L380

---
>     * Objects should not grow size during copying.

if we calculate object size through the mark bit, it will be the original size before copying. But we can query `ObjectModel::get_size_when_copied`, and apply the diff to the size when we calculating `live` bytes for the offset vector.

src/plan/compressor/global.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub common: CommonPlan<VM>,
+    #[space]
+    pub compressor_space: CompressorSpace<VM>,
+}
+
+/// The plan constraints for the Compressor plan.
+pub const COMPRESSOR_CONSTRAINTS: PlanConstraints = PlanConstraints {
+    moves_objects: true,
+    needs_forward_after_liveness: true,
+    ..PlanConstraints::default()
```
wks, line 45: The default `max_non_los_default_alloc_bytes` is `MAX_INT` (0x7fff_ffff).  That's too large.  If we still intend to use LOS, we should give it a smaller value, such as the same value chosen by `MarkCompact` (`MAX_NON_LOS_ALLOC_BYTES_COPYING_PLAN` which is 8KB).

no-defun-allowed:
> When we encounter an object during computing offset vector

> apply the diff to the size when we calculating live bytes for the offset vector

The issue is that we don't only use the offset vector - we also scan the mark bitmap between the start of the block containing an object and that object. So we need to be able to encode the space taken in the mark bitmap. I think this is possible for alignment/padding, as there should be enough space in the mark bitmap to adjust the first/last word marks, but there might not be with objects which grow when copied.

no-defun-allowed:
>  To support discontiguous space, we cannot use the start of the space, and we need some extra information (e.g. a table that maps block index to block start) to know the current block start, then apply the offset internal to a block to it -- state.live % block_size.

The forwarding loop doesn't account for any chunk boundaries either. I'm undecided on if preventing objects from crossing boundaries can be done with the mark bitmap or not.

src/plan/compressor/global.rs
no-defun-allowed: I've done just that (and got the LOS working at all).

k-sareen:
@qinsoon Should we skip the Compressor for MockVM tests for the time-being then?

k-sareen:
@no-defun-allowed Can you just temporarily set `ALL_PLANS` to `ALL_DISCONTIGUOUS_PLANS` in the test script and add a comment there? That should work I think. Then we can merge this PR.

qinsoon (approved):
LGTM

PR #1341, opened 2025-07-15
LOS: Reduce is_mapped calls during VO bit scan

wks:
Like in `find_last_non_zero_bit_in_metadata_bytes`, we should only call `is_mapped` when we reached a new chunk because the granularity of `MapState` is per chunk.

### DIFF
--- a/src/policy/largeobjectspace.rs
+++ b/src/policy/largeobjectspace.rs
@@ -106,15 +106,25 @@ impl<VM: VMBinding> SFT for LargeObjectSpace<VM> {
         max_search_bytes: usize,
     ) -> Option<ObjectReference> {
         use crate::util::metadata::vo_bit;
+        use crate::util::heap::vm_layout::MMAP_CHUNK_BYTES;
+
+        // We need to check if metadata address is mapped or not. But we only check at chunk granularity.
+        // This records the start of a chunk that is tested to be mapped.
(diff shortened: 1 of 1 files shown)


wks:
This optimization is orthogonal to https://github.com/mmtk/mmtk-core/pull/1337.  While https://github.com/mmtk/mmtk-core/pull/1337 reduces the overhead of each individual `is_mapped` invocation, this PR reduces the number of calls of `is_mapped` for the LOS.

(Note that the linear searching algorithm for non-LOS spaces already take `MapState`'s chunk granularity into consideration.)

I ran `benches/mock_bench/internal_pointer.rs` on several revisions, and here are the performance comparison.  Numbers are the average iteration time, and the unit is nanosecond.  1337 and 1341 are PR numbers.  1337 and 1341 is a branch with both PRs merged.

| space | master | 1337 | 1341 | 1337 + 1341 |
|---|---|---|---|---|
| normal | 101.66 | 83.862 | ~~94.948~~ | ~~78.371~~ |
| LOS | 294.80 | 30.533 | 94.320 | 27.795 |

Note that PR 1341 should not affect normal spaces, and the performance improvements are likely to be noises.

With either or both of PR 1337 and PR 1341 applied, the LOS searches faster (for the same object size 16KB) than other spaces because it searches at page granularity.  This is the expected behavior.

PR #1342, opened 2025-07-18
Deduplicate specs in `SideMetadataSanity::get_all_specs`

qinsoon:
For 32 bits chunked side metadata, we check the total size of side metadata per chunk to make sure it won't get too large. 
https://github.com/mmtk/mmtk-core/blob/1dc681561097688ac37382b1e23e56f2d160b7c7/src/util/metadata/side_metadata/sanity.rs#L112

However, the function `get_all_specs` collects all the side metadata specs from all the policies, and different policies may use the same local side metadata. This results in duplicate entries in the vec, and results in incorrect total size calculated based on the vec. This PR fixes this.


wks (approved):
LGTM

Issue #1344, opened 2025-07-19
Mmapper granularity: chunk or page?


wks:
Implementations of `Mmapper` keep a map from chunks to `MapState` values.  Obviously it works at the granularity of chunks internally.

But some methods of `Mmapper` take the number of pages as arguments.  Concretely:

-   `quarantine_address_range(start, pages, strategy, anno)`
-   `ensure_mapped(start, pages, strategy, anno)`
-   `protect(start, pages)`

I don't think `Mmapper` can maintain `MapState` at the granularity of pages because that would be too fine-grained.  To handle large address ranges up to 48 bits, we would need to switch to a different data structure, such as [segment tree](https://en.wikipedia.org/wiki/Segment_tree) which is slower to query than chunk-grained arrays.

If we want `Mmapper` to keep working at chunk granularity, we should change the public interface in order **not to give its user a false impression that it may work at page granularity**.  We should remove all mentions of "page" and use one of the following strategies:

1.  `quarantine_address_ranges(start: Address, bytes: usize, ...)` while requiring both `start` and `bytes` to be chunk-aligned.
2.  `quarantine_address_ranges(start_chunk_index: usize, num_chunks: usize, ...)` where both `start_chunk_index` and `num_chunks` are chunk indices counted from address 0.  That is, `chunk_index = address >> LOG_BYTES_IN_CHUNK`.
3.  `quarantine_address_ranges(chunk_range: ChunkRange, ...)` where `chunk_range` is a dedicated data structure that refers to whole chunks.

I personally prefer method 3.  Regardless of the internal representation of `ChunkRange` (whether it uses aligned addresses or chunk indices), there is no way the user can supply an unaligned range.

qinsoon:
This might not be an issue -- a mmapper guarantees the required pages to be mmapped, but it internally mmaps at the granularity of chunks.

> we should change the public interface in order not to give its user a false impression that it may work at page granularity

The users should not assume how the memory is mmapped -- they just know the pages they require will be mapped.

wks:
> This might not be an issue -- a mmapper guarantees the required pages to be mmapped, but it internally mmaps at the granularity of chunks.
> 
> > we should change the public interface in order not to give its user a false impression that it may work at page granularity
> 
> The users should not assume how the memory is mmapped -- they just know the pages they require will be mapped.

[...]

Issue #587, opened 2022-05-05
Cleaning up OOP-style inheritance and the impls of Deref


wks:
TL;DR: Several types in MMTk core use inheritance in OOP.  This is not idiomatic in Rust.  In the long term, we should switch to a more idiomatic style, such as *composition instead of inheritance*.

In practice, we should tread lightly, and do refactoring in small steps. Not all such cases need to be eliminated, because there may not always be a better solution.

# Examples:

The following types use inheritance.

-   Plan: concrete plans extend `CommonPlan` extends `BasePlan`
-   Space: concrete spaces extend `CommonSpace`
-   ProcessEdgesWork: Any concrete ProcessEdgesWork always contains a `ProcessEdgesBase`
-   PageResources: Any concrete page resource contains a `ComonPageResource`

# Analysis

## Inheritance vs composition

An instance of a concrete type contains an instance of an abstract type because the abstract type holds common information shared by multiple concrete types.  For example, all spaces should have name.

Conversely, inheritance also means *each subclass customises the abstract class in a certain way*.  For example, all `ProcessEdgesWork` have an incoming queue edges to be processed, and an outgoing queue of objects to be scanned, as shown in `ProcessEdgesBase`, but each concrete `ProcessEdgesWork` customises the way how to process each edge.

If we view the "inheritance" relation as a form of "customisation", we may think that *there is only one kind of ProcessEdgesWork*, but each instance allow some behaviour to be customised.  And there are many different ways of customisation besides OOP-style inheritence.  One possibility is composition.  If we consider `trace_object` as a responsibility of `Plan`, then all ProcessEdgesWork are the same.  ProcessEdgesWork only calls into Plan to do the actual tracing.

Both https://github.com/mmtk/mmtk-core/pull/575 and  https://github.com/mmtk/mmtk-core/pull/570 attempt to eliminate plan-specific ProcessEdgesWork implementations.  They are examples of replacing inheritance with composition.

## Methods for MMTK accidentally added to `BasePlan`

`BasePlan` is special.  It comes from the `Plan` abstract class in the JikesRVM version of MMTk.  Because Java MMTk didn't have the global instance type `MMTK`, globally shared data fields are fields of the Plan class.

Most fields in `BasePlan` are not really plan-specific.  For example,

-   `BasePlan::initialized` indicates if MMTk is initialised.  No matter what plan we use, MMTk always needs to be initialised.
[...]

PR #1345, opened 2025-07-19
Make Mmapper a struct, and existing impls storages

wks:
We make Mmapper a struct instead of trait.  It will be solely responsible for transitioning the map states.

Existing implementations, including `ByteMapMmapper` and `TwoLevelMmapper`, become the storage backends of `Mmapper`, namely `ByteMapStateStorage` and `TwoLevelStateStorage`.  They are only responsible for maintaining the (one-level or two-level) array of `Atomic<MapState>`, and provide high-level functions for bulk-setting and bulk-transitioning map states.

Methods of `MapState` are moved to `Mmapper`.  All bulk-setting operations can call `mmap` or `mprotect` on contiguous range of chunks instead of one chunk at a time, regardless whether using `ByteMapStateStorage` or `TwoLevelStateStorage`.

**DRAFT**: This is a very aggressive refactoring over the existing `Mmapper` framework.  We need to
-   [ ] Fix https://github.com/mmtk/mmtk-core/issues/1344 in this PR if not already addressed by another PR.
-   [ ] Make sure it works on both 32-bit and 64-bit machines.
-   [ ] Remove unused methods.
-   [ ] Rename variables, especially those that mention "mmapper", "fragmented", etc.
-   [ ] Performance evaluation, especially whether we need a specialized fast path for single-chunk modification, compared to the iterator-based "bulk transition" mechanism.
-   [ ] Find a way to allow dynamic selection of `ByteMapMmapper` and `TwoLevelMmapper`.  This may involve a level of `dyn` somewhere.  Make sure the performance is OK.  It should be because it is supposed to be much cheaper than `mmap` and `mprotect` system calls.


wks:
If this refactoring is successful, it may shed some light on the old issue https://github.com/mmtk/mmtk-core/issues/587.  It demonstrates a way to transform Java-style OOP with abstract classes and inheritance into Rust-style data structure composition, where former sub-classes become "backends" (`ByteMapStateStorage` and `TwoLevelStateStorage`) of the "main data type" (`Mmapper`) that performs the actual work.  We may refactor other types in this way, such as

-   `VMMap` and the backends `Map32` and `Map64`
-   `FreeList` and the backends `IntArrayFreeList` and `RawMemoryFreeList`
-   `Space` as a struct (with most fields from `BaseSpace`), and make each concrete space a "backend".

PR #1349, opened 2025-07-23
Only quarantine side metadata memory for the accessible space extent

qinsoon:
We currently quarantine side metadata address range for the whole extent of each space (2TB per space for 64 bits). Though performance-wise there is little difference, this wastes virtual memory address. This PR changes it to only quarantine the max heap size for each space, as each space should never grow larger than the max heap size.

### DIFF
--- a/src/policy/space.rs
+++ b/src/policy/space.rs
@@ -671,8 +671,14 @@ impl<VM: VMBinding> CommonSpace<VM> {
         }
 
         // For contiguous space, we know its address range so we reserve metadata memory for its range.
+        // The accessible range for a single space never needs to be larger than the max heap size. We only need to reserve the accessible range.
+        let max_heap_size = conversions::raw_align_up(
+            conversions::pages_to_bytes(rtn.gc_trigger.policy.get_max_heap_size_in_pages()),
+            BYTES_IN_CHUNK,
+        );
+        let accessible_extent = usize::min(extent, max_heap_size);
         rtn.metadata
-            .try_map_metadata_address_range(rtn.start, rtn.extent, rtn.name)
+            .try_map_metadata_address_range(rtn.start, accessible_extent, rtn.name)
             .unwrap_or_else(|e| {
                 // TODO(Javad): handle meta space allocation failure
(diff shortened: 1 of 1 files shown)


k-sareen:
The problem is likely to do with the `side_after` etc. calculations which bottoms out [here](https://github.com/mmtk/mmtk-core/blob/37f7228ffbc19b3fb1067fa1e97c36e826d277dd/src/util/metadata/side_metadata/helpers.rs#L222-L225) for example. We may need to fix some of the calculations.

wks:
> The problem is likely to do with the `side_after` etc. calculations which bottoms out [here](https://github.com/mmtk/mmtk-core/blob/37f7228ffbc19b3fb1067fa1e97c36e826d277dd/src/util/metadata/side_metadata/helpers.rs#L222-L225) for example. We may need to fix some of the calculations.

The current implementation of `side_after` gives each metadata an address range that is able to cover the entire 47 bit address space (128TB).

```rust
pub(super) const fn metadata_address_range_size(metadata_spec: &SideMetadataSpec) -> usize {
    1usize << (VMLayout::LOG_ARCH_ADDRESS_SPACE - log_data_meta_ratio(metadata_spec))
}
```

For metadata with a data-meta-ratio of 64:1, such as mark bit and VO bit, it will be `1<<41` bits, i.e. 2TB.

But since MMTk only has `MAX_SPACES` (16) spaces, and each space is `max_space_extent()` (2TB), a global side metadata needs to cover a range of `max_space_extent() * MAX_SPACES` (`2TB * 16 = 32TB`), while a local side metadata only needs to cover `max_space_extent()` (2TB).  For a 64:1 side metadata, it will be 512GB and 32GB, respectively.

There are some exceptions, though.

-   **malloc mark-sweep space**: Since we can't determine where malloc allocates memory (unless we dig into the malloc library), we have to let both local and global side metadata cover 47 bit address space (128TB).
-   **VM space**: If we allow the VM binding to place the VM space anywhere, it will be the same.  If we can force it to be located in designated areas within MMTk's heap, we can reduce its range, too.

PR #1350, opened 2025-07-23
Workaround eBPF errors

wks:
The kernel is raising errors related to the variable `@every`.  We workaround this problem by replacing it with an integer literal (expanded from Python template variable `$EVERY`) instead.

The USDT `process_root_nodes` may not exist in the compiled .so file of the VM binding because the methods that contain this USDT are generic methods that are instantiated on demand.  VMs that do not use object-enqueuing tracing (such as OpenJDK) may not have this USDT.  We added the `--no-root-nodes` option to skip that USDT.

Added some debugging options to `capture.py`, including `-d` and `-v`.


qinsoon (approved):

caizixian:
Thanks for reporting this to the upstream @wks

PR #1360, opened 2025-07-31
Enable reference processing in the core by default

### LINKED ISSUE #1359: Both `no_finalizer` and `no_reference_types` should have the same value
Currently `no_finalizer` is `false` and `no_reference_types` is `true`. This leads to a really odd configuration. Finalization and reference processing are interlinked, so it makes no sense to be able to finalize without dealing with weak references, at least in the context of OpenJDK. Further, if we are comparing against G1 or any stock Hotspot collector, this configuration is just not possible there so comparisons need to be very careful of this inconsistency.

k-sareen:
Closes #1359.

### DIFF
--- a/src/policy/compressor/compressorspace.rs
+++ b/src/policy/compressor/compressorspace.rs
@@ -38,7 +38,13 @@ impl<VM: VMBinding> SFT for CompressorSpace<VM> {
     }
 
     fn get_forwarded_object(&self, object: ObjectReference) -> Option<ObjectReference> {
-        Some(self.forward(object, false))
+        // Check if forwarding addresses have been calculated before attempting
+        // to forward objects
+        if self.forwarding.has_calculated_forwarding_addresses() {
+            Some(self.forward(object, false))
+        } else {
+            None
+        }
     }
 
     fn is_live(&self, object: ObjectReference) -> bool {
--- a/.github/workflows/micro-bm.yml
+++ b/.github/workflows/micro-bm.yml
@@ -66,7 +66,7 @@ jobs:
         with:
           repository: mmtk/ci-perf-kit
           token: ${{ secrets.GITHUB_TOKEN }}
-          ref: "0.8.2"
+          ref: "0.8.3"
           path: ci-perf-kit
           submodules: true
       # Use rust-toolchain in the trunk (it doesnt matter much - if the toolchains defined in the trunk and the branch are different, we cant run anyway)
--- a/.github/workflows/perf-baseline.yml
+++ b/.github/workflows/perf-baseline.yml
@@ -42,7 +42,7 @@ jobs:
         with:
           token: ${{ secrets.CI_ACCESS_TOKEN }}
           repository: mmtk/ci-perf-kit
-          ref: "0.8.2"
+          ref: "0.8.3"
           path: ci-perf-kit
           submodules: true
       # setup
@@ -94,7 +94,7 @@ jobs:
         with:
           token: ${{ secrets.CI_ACCESS_TOKEN }}
           repository: mmtk/ci-perf-kit
-          ref: "0.8.2"
+          ref: "0.8.3"
           path: ci-perf-kit
           submodules: true
       # setup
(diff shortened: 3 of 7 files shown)


k-sareen:
Hm. I thought we had already tested Compressor with weak references before 🤔 I guess we can temporarily disable it for Compressor, unless it's a quick fix.

k-sareen:
@wks The issue is the weak reference buckets are opened before we do the forwarding calculation in the Compressor (and also MarkCompact). How is this even working for MarkCompact? I think we should move those buckets to be near the end. I think it will fix this issue.

qinsoon:
> @wks The issue is the weak reference buckets are opened before we do the forwarding calculation in the Compressor (and also MarkCompact). How is this even working for MarkCompact? I think we should move those buckets to be near the end. I think it will fix this issue.

I think that's expected. We need to trace all the objects before calculating forwarding (the same for both Compressor and MarkCompact). The issue is that when we trace weak refs, we should only trace objects rather than rather than attempting to forward them (e.g. marking objects only for mark compact, setting mark bits for compressor). Forwarding for weak refs is done in a separate bucket after calculating forwarding.

k-sareen:
Perhaps add a `bool: forward_references` to the work packets where it's `true` for everything else and `false` for Compressor and MarkCompact which will do the forwarding at a later stage.

qinsoon:
> Perhaps add a `forward_references: bool` to the work packets where it's `true` for everything else and `false` for Compressor and MarkCompact which will do the forwarding at a later stage.

I think the issue is `get_forwarded_object` for compressor.
https://github.com/mmtk/mmtk-core/blob/54b5e152e43af3ea18d6d90768690cd89adcc7b8/src/policy/compressor/compressorspace.rs#L40-L42

It should not always return `Some`. When it has not forwarded the object, it should return `None`. Even if the policy has calculated forwarding, but hasn't really moved the object, it should return `None`.

Returning `Some` basically means the object is not at its original location, please look at the new location.

k-sareen:
There are likely performance implications for this solution since we have to load an `AtomicBool` for each object, as opposed to a static `bool` which is passed during work packet creation time.

src/policy/compressor/compressorspace.rs
```diff
@@ -38,7 +38,13 @@ impl<VM: VMBinding> SFT for CompressorSpace<VM> {
     }
 
     fn get_forwarded_object(&self, object: ObjectReference) -> Option<ObjectReference> {
-        Some(self.forward(object, false))
+        // Check if forwarding addresses have been calculated before attempting
+        // to forward objects
+        if self.forwarding.has_calculated_forwarding_addresses() {
```
qinsoon, line 43: This is still not exactly correct. When `forwarding` calculates the offsets but does not yet move the object to the new location, what this function returns is an invalid object reference. Currently no function calls `get_forwarded_object` during that period of time, so it appears to be fine.

k-sareen: Why do we care about the actual content when we're forwarding references?

k-sareen: Or do you mean this is a generic function in SFT and that if a binding/VM queries this function then it may get incorrect information.

qinsoon: > Why do we care about the actual content when we're forwarding references?

This question is irrelevant. The function returns an object reference, and the object reference it returns may not be valid at all. There is no restriction on how callers may do with the object reference -- they could call this function at any time, and do anything with the object reference. I am simply pointing out that this code works under the assumption that the function should not be called at certain point.

The more assumptions there are, the more brittle the code will be. There could be in the future someone calls this function at the time that does not match your assumption, sees a segfault out of nowhere, and spends hours debugging it.

With code that clearly has flaws, there is no assertion to avoid this, there is no comment to mention this possibility. It will become a headache for next person who stumbles on the issue and discover all these issues that we have discussed here.

k-sareen: Right sure. But this is exactly what happens in MarkCompact as well. The `Compact` bucket is near the end. So your argument extends to MarkCompact as well. I'm not sure if we can solve this easily for GCs that have a separate compact phase.

k-sareen:
@qinsoon What should I do for this PR? Update the doc comment for `get_forwarded_object`?

k-sareen:
It's timing out the extended OpenJDK benchmarks though

k-sareen:
binding-refs
OPENJDK_BINDING_REF=enable-reference-processing

k-sareen:
@qinsoon Happy to merge this PR then?

Issue #1256, opened 2025-01-08
Systematic approach to clear/set side metadata


wks:
TL;DR: No scattering side metadata clearing/setting operations everywhere!  Don't blindly insert those operations before GC / before nursery GC / after GC / when allocating / etc., and per-space / per-chunk / per-block / per-object / etc.  Do it systematically.  Make a framework for it.

This issue is only about side metadata.  Because free spaces are always zeroed, in-header metadata are always 0 upon allocation/copying of an object, and we don't need to worry about bulk-clearing.  But we may start worrying about cyclic mark bits if we start to support in-header mark bits for ImmixSpace.  (No.  We still don't support in-header mark bits now!)

# Problem

Take ImmixSpace as an example.

-   In `PrepareBlockState` (only in major GCs), it
    -   bulk clears mark bits (if on the side)
    -   bulk clears unlog bits (if on the side)
-   In `SweepChunk`, it
    -   iterate through every block and
        -   clears the forwarding bits (if on the side, and if the block is a defrag source)
        -   iterate through every line in the block (or just every block if `BLOCK_ONLY`) and
            -   bulk clear VO bits (if the line is free)
            -   (should) bulk clear unlog bits (if the line is free and if unlog bits are on the side)  See https://github.com/mmtk/mmtk-core/pull/1169
            -   (should) bulk clear pin bits (if the line is free and if the pin bits are on the side)  See https://github.com/mmtk/mmtk-core/issues/1118#issuecomment-2060502398

Some metadata have undergone several refactoring.  The forwarding bits, for example, was
-   cleared in `prepare` for every block in the beginning of every major GC when the forwarding bits were first allowed to be on the side (https://github.com/mmtk/mmtk-core/pull/753)
-   then "optimized" to clear only forwarding bits of defrag sources (https://github.com/mmtk/mmtk-core/pull/777)
-   then the "optimization" was reverted because minor GCs can move objects, too.  (https://github.com/mmtk/mmtk-core/pull/974)
-   then changed to be cleared after GC because young objects may be allocated over free spaces with stale forwarding bits, and we don't clear forwarding bits before minor GCs.  (https://github.com/mmtk/mmtk-core/pull/1138)

[...]

PR #1364, opened 2025-08-08
Refactor log bits

qinsoon:
This PR refactors the log bits: the current implementation assumes log bits are used for generational plans, and they carry generation semantics. This PR removes such assumptions.

With this PR, log bits are global metadata, and its semantic is defined by the plans. A plan may directly manipulate log bits, and may tell policies how to deal with log bits. A policy should never assume the semantics of log bits, and they just do whatever they are instructed to.

Some policies assume the log bit may be in the side, and this PR does not try to address this issue.

### DIFF
--- a/.github/workflows/binding-tests-openjdk.yml
+++ b/.github/workflows/binding-tests-openjdk.yml
@@ -45,3 +45,6 @@ jobs:
             cd mmtk-openjdk
             export RUST_BACKTRACE=1
             ./.github/scripts/${{ inputs.test-script }}
+        - name: Setup tmate session
+          if: ${{ failure() }}
+          uses: mxschmitt/action-tmate@v3
--- a/.github/workflows/binding-tests-openjdk.yml
+++ b/.github/workflows/binding-tests-openjdk.yml
@@ -45,6 +45,3 @@ jobs:
             cd mmtk-openjdk
             export RUST_BACKTRACE=1
             ./.github/scripts/${{ inputs.test-script }}
-        - name: Setup tmate session
-          if: ${{ failure() }}
-          uses: mxschmitt/action-tmate@v3
--- a/src/plan/generational/global.rs
+++ b/src/plan/generational/global.rs
@@ -74,9 +74,6 @@ impl<VM: VMBinding> CommonGenPlan<VM> {
     /// Release Gen. This should be called by a single thread in GC release work.
     pub fn release(&mut self, tls: VMWorkerThread) {
         let full_heap = !self.is_current_gc_nursery();
-        if VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.is_on_side() {
-            self.nursery.clear_side_log_bits();
-        }
         self.common.release(tls, full_heap);
         self.nursery.release();
     }
--- a/src/plan/generational/global.rs
+++ b/src/plan/generational/global.rs
@@ -74,6 +74,9 @@ impl<VM: VMBinding> CommonGenPlan<VM> {
     /// Release Gen. This should be called by a single thread in GC release work.
     pub fn release(&mut self, tls: VMWorkerThread) {
         let full_heap = !self.is_current_gc_nursery();
+        if VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.is_on_side() {
+            self.nursery.clear_side_log_bits();
+        }
         self.common.release(tls, full_heap);
         self.nursery.release();
     }
--- a/src/plan/generational/global.rs
+++ b/src/plan/generational/global.rs
@@ -41,7 +41,7 @@ pub struct CommonGenPlan<VM: VMBinding> {
 impl<VM: VMBinding> CommonGenPlan<VM> {
     pub fn new(mut args: CreateSpecificPlanArgs<VM>) -> Self {
         let nursery = CopySpace::new(
-            args.get_space_args("nursery", true, false, VMRequest::discontiguous()),
+            args.get_nursery_space_args("nursery", true, false, VMRequest::discontiguous()),
             true,
         );
         let full_heap_gc_count = args
--- a/src/plan/generational/immix/global.rs
+++ b/src/plan/generational/immix/global.rs
@@ -250,10 +250,14 @@ impl<VM: VMBinding> GenImmix<VM> {
                 crate::plan::generational::new_generational_global_metadata_specs::<VM>(),
         };
         let immix_space = ImmixSpace::new(
-            plan_args.get_space_args("immix_mature", true, false, VMRequest::discontiguous()),
+            plan_args.get_mature_space_args(
+                "immix_mature",
+                true,
+                false,
+                VMRequest::discontiguous(),
+            ),
             ImmixSpaceArgs {
                 // In GenImmix, young objects are not allocated in ImmixSpace directly.
-                #[cfg(feature = "vo_bit")]
                 mixed_age: false,
                 never_move_objects: false,
             },
--- a/src/plan/generational/immix/global.rs
+++ b/src/plan/generational/immix/global.rs
@@ -130,6 +130,7 @@ impl<VM: VMBinding> Plan for GenImmix<VM> {
         self.gen.prepare(tls);
         if full_heap {
             if VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.is_on_side() {
+                self.gen.common.clear_side_log_bits();
                 self.immix_space.clear_side_log_bits();
             }
             self.immix_space.prepare(
--- a/src/plan/generational/immix/global.rs
+++ b/src/plan/generational/immix/global.rs
@@ -130,7 +130,6 @@ impl<VM: VMBinding> Plan for GenImmix<VM> {
         self.gen.prepare(tls);
         if full_heap {
             if VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.is_on_side() {
-                self.gen.common.clear_side_log_bits();
                 self.immix_space.clear_side_log_bits();
             }
             self.immix_space.prepare(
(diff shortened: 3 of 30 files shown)


src/plan/plan_constraints.rs
```diff
@@ -45,6 +45,10 @@ pub struct PlanConstraints {
     /// `MutatorConfig::prepare_func`).  Those plans can set this to `false` so that the
     /// `PrepareMutator` work packets will not be created at all.
     pub needs_prepare_mutator: bool,
+    /// Should a policy unlog newly allocated objects?
+    pub unlog_allocated_object: bool,
```
qinsoon: I think this field needs to be defined per space, rather than per plan. In our generational plans, this value should be `false` for the nursery space and any space that allows mixed generation (such as LOS, or sticky immix's immix space). And the value should be `true` for other spaces.

k-sareen: ~~Hm. It should also be `false` for LOS since it has a nursery. What spaces do we directly allocate into that would need this for generational GCs? Immortal space, I guess. Because LOS and nursery space should be the main spaces where new objects are allocated into. Again, I'm only talking about generational GCs. I'm aware that SATB requires this.~~
Nvm. I misread what you said, sorry. Yes I agree with what you said.

qinsoon: > What spaces do we directly allocate into that would need this for generational GCs?

Immortal, non moving, code, VM space (we don't allocate into VM space, but we ask the binding to call our post_alloc after their allocation), etc. Also if we allow pretenure allocation, we may directly allocate into a mature space and also unlog the objects.

.github/workflows/binding-tests-openjdk.yml
```diff
@@ -45,3 +45,6 @@ jobs:
             cd mmtk-openjdk
             export RUST_BACKTRACE=1
             ./.github/scripts/${{ inputs.test-script }}
+        - name: Setup tmate session
```
qinsoon: This was used for debugging. I will revert this.

src/policy/immix/immixspace.rs
```diff
@@ -877,6 +876,10 @@ impl<VM: VMBinding> ImmixSpace<VM> {
         if !super::MARK_LINE_AT_SCAN_TIME {
             self.mark_lines(object);
         }
+        if self.common.unlog_traced_object {
```
wks, line 886: If the `object` is already coped, then the `VMObjectModel::copy` defined by the VM binding will have already called `CopySpaceCopyContext::post_copy` which calls `ImmixSpace::post_copy` which will also set the unlog bit.  Therefore when we reach here and `object` is already copied, the unlog bit will have already been set.

This is benign because Immix seldom copies object, and it is idempotent to set the unlog bit twice.

If we want to remove the redundant setting, we can modify `ImmixSpace::trace_object_with_opportunistic_copy` and move the invocation of `self.unlog_object_if_needed(new_object);` to the non-moving branch of the `if` statement, and assert that the unlog bit is already set on the moving branch.

```diff
diff --git a/src/policy/immix/immixspace.rs b/src/policy/immix/immixspace.rs
index 58dc969fac..940a015e67 100644
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -699,6 +699,8 @@ impl<VM: VMBinding> ImmixSpace<VM> {
                     self.mark_lines(object);
                 }
 
+                self.unlog_object_if_needed(object);
+
                 object
             } else {
                 // We are forwarding objects. When the copy allocator allocates the block, it should
@@ -709,6 +711,14 @@ impl<VM: VMBinding> ImmixSpace<VM> {
                     semantics,
                     copy_context,
                     |_new_object| {
+                        // `ObjectModel::post_copy` should have set the unlog bit
+                        // if `unlog_traced_object` is true.
+                        debug_assert!(
+                            !self.common.unlog_traced_object
+                                || VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC
+                                    .is_unlogged::<VM>(object, Ordering::Relaxed)
+                        );
+
                         #[cfg(feature = "vo_bit")]
                         vo_bit::helper::on_object_forwarded::<VM>(_new_object);
                     },
@@ -721,7 +731,6 @@ impl<VM: VMBinding> ImmixSpace<VM> {
 
             queue.enqueue(new_object);
             debug_assert!(new_object.is_live());
-            self.unlog_object_if_needed(new_object);
             new_object
         }
     }
```

src/plan/global.rs
```diff
@@ -621,6 +732,18 @@ impl<VM: VMBinding> CommonPlan<VM> {
...
+        self.immortal.clear_side_log_bits();
+        self.los.clear_side_log_bits();
+        self.base.clear_side_log_bits();
+    }
+
+    pub fn set_side_log_bits(&self) {
+        self.immortal.set_side_log_bits();
+        self.los.set_side_log_bits();
+        self.base.set_side_log_bits();
+    }
```
wks, line 745: It looks like neither `clear_side_log_bits` nor `set_side_log_bits` are methods of any traits, but methods of specific spaces that support log bits.  We may extract a trait and make it a trait method.

```rust
pub trait SupportsLogBits: Space {
    fn clear_side_log_bits(&self);
    fn set_side_log_bits(&self);
}
```

Then when needed, we can simply iterate through all spaces and call that method.

```rust
fn prepare(&self) {
            self.for_each_space(|space| {
                if let Some(s) = space.downcast::<SupportsUnlogBits>() {
                    s.clear_side_log_bits();
                }
            });
}
```

src/policy/immix/immixspace.rs
```diff
@@ -529,6 +513,21 @@ impl<VM: VMBinding> ImmixSpace<VM> {
         did_defrag
     }
 
+    pub fn clear_side_log_bits(&self) {
```
wks: Now this method is called in both `GenImmix::prepare` and `StickyImmix::prepare`, but not within `ImmixSpace::prepare` or `ImmixSpace::release`.  This is not a problem because the plan may need more control about which space to prepare and which metadata to clear.  But we may need to take a note because we want a systematic approach to clearing metadata.  https://github.com/mmtk/mmtk-core/issues/1256

src/plan/generational/immix/global.rs
```diff
@@ -129,6 +129,9 @@ impl<VM: VMBinding> Plan for GenImmix<VM> {
         let full_heap = !self.gen.is_current_gc_nursery();
         self.gen.prepare(tls);
         if full_heap {
+            if VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.is_on_side() {
+                self.immix_space.clear_side_log_bits();
+            }
```
wks, line 134: We are not calling `self.gen.common.clear_side_log_bits()`, but we do in `StickyImmix::prepare`.  Is this intentional?  I think when using `GenImmix`, we may still directly allocate large objects into LOS and they may become unlogged if they survive a GC.

wks:
@qinsoon  Among the comments I have given above, only https://github.com/mmtk/mmtk-core/pull/1364#discussion_r2272692497 can be a result of potential bugs.  Others are aesthetic comments which may be fixed or may be left as is, depending you your taste.

The style check failure in the CI should be fixed by https://github.com/mmtk/mmtk-core/pull/1367

src/policy/immix/immixspace.rs
qinsoon: I have made changes as suggested.

src/plan/global.rs
qinsoon: I added those methods to `trait Space`. All the spaces should allow plans to manipulate log bits on them. However, for some policies, we are not using those methods for the current plans. I leave them as `unimplemented!()`.

src/policy/immix/immixspace.rs
qinsoon: Yes. That is the intention of this PR --  to allow plans to directly manipulate log bits through spaces. I saw you have edited https://github.com/mmtk/mmtk-core/issues/1256.

src/plan/generational/immix/global.rs
qinsoon: I removed `clear_side_log_bits` in `StickyImmix`. We should deal with log bits for common/base spaces in the same way for the 3 plans, `GenImmix`, `StickyImmix` and `GenCopy`. This should be the same behavior as before the refactoring.

k-sareen:
I think we need to ensure that the bug we found and fixed in https://github.com/mmtk/mmtk-core/pull/1169 is not re-introduced.

qinsoon:
> I think we need to ensure that the bug we found and fixed in #1169 is not re-introduced.

We can only ensure 1. a PR passes the current tests, 2. a PR passes code review. You can leave comments if you believe some changes in the PR may break https://github.com/mmtk/mmtk-core/pull/1169.

src/plan/generational/global.rs
```diff
@@ -74,6 +74,9 @@ impl<VM: VMBinding> CommonGenPlan<VM> {
     /// Release Gen. This should be called by a single thread in GC release work.
     pub fn release(&mut self, tls: VMWorkerThread) {
         let full_heap = !self.is_current_gc_nursery();
+        if VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.is_on_side() {
+            self.nursery.clear_side_log_bits();
```
k-sareen: Why do we need to clear the log bits in the nursery? The nursery in a generational plan should never have log bits set. If you're saying that plans need to inform spaces how to maintain log bits, then I'm not sure why this is necessary since the plan knows that the log bits are never set for the nursery in STW generational plans.

qinsoon: I agree it is unnecessary, I will remove it.

I added this code because we call `self.nursery.release()`, and there used to be side log bits clearing in `CopySpace::release`.

wks:
There are new methods introduced in this PR that return `PlanCreateSpaceArgs<VM>`, which Clippy warns about.  We should change them to `PlanCreateSpaceArgs<&'_, VM>`.

qinsoon:
We saw a test failure in https://github.com/mmtk/mmtk-core/actions/runs/16955407376/job/48133843550?pr=1364. It is probably not related with this PR. I opened an issue for it: https://github.com/mmtk/mmtk-openjdk/issues/317.

Issue #1365, opened 2025-08-11
The x86_64-apple-darwin toolchain is demoted


wks:
Rust 1.89.0 is [Demoting `x86_64-apple-darwin` to Tier 2 with host tools](https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/#demoting-x86-64-apple-darwin-to-tier-2-with-host-tools).  

> GitHub will soon [discontinue](https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down) providing free macOS x86_64 runners for public repositories. Apple has also announced their [plans](https://en.wikipedia.org/wiki/Mac_transition_to_Apple_silicon#Timeline) for discontinuing support for the x86_64 architecture.

This means the `x86_64-apple-darwin` toolchain will be [guaranteed to build, but no longer guaranteed to work](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-2-with-host-tools).

The direct impact will be GitHub runners.  Currently we do minimal tests with the following configurations:

```
  minimal-tests-core:
    needs: setup-test-matrix
    strategy:
      fail-fast: false
      matrix:
        target:
          - { os: ubuntu-22.04, triple: x86_64-unknown-linux-gnu }
          - { os: ubuntu-22.04, triple: i686-unknown-linux-gnu }
          - { os: macos-15, triple: x86_64-apple-darwin }
        rust: ${{ fromJson(needs.setup-test-matrix.outputs.rust )}}

    name: minimal-tests-core/${{ matrix.target.triple }}/${{ matrix.rust }}
    runs-on: ${{ matrix.target.os }}
```

[...]

PR #1366, opened 2025-08-12
Run CI tests with aarch64-apple-darwin

wks:
Apple will stop supporting Intel Mac, and Rust 1.89 has demoted the x86_64-apple-darwin toolchain to Tier 2.  We instead run CI tests against aarch64-apple-darwin which is used by Mac computers with Apple silicon.

Fixes: https://github.com/mmtk/mmtk-core/issues/1365


Cargo.toml
```diff
@@ -37,7 +37,7 @@ lazy_static = "1.1"
 libc = "0.2"
 log = { version = "0.4", features = ["max_level_trace", "release_max_level_off"] }
 memoffset = "0.9"
-mimalloc-sys = { version = "0.1.6", optional = true }
+libmimalloc-sys = { version = "0.1.43", features = ["extended"], optional = true }
```
k-sareen: Sorry to be pedantic, but would it be possible to maintain alphabetical order?

wks: Sure.  I'll make the change.

PR #1369, opened 2025-08-14
Refactor both Mmapper impls into ChunkStateMmapper

wks:
The two existing `Mmapper` implementations, namely `ByteMapMmapper` and `TwoLevelMmapper` have much duplicated code.

We introduce `ChunkStateMmapper` which implements a chunk-grained `Mmapper` backed by a logical array of `MapState` elements, one for each chunk.  It is responsible for actually calling `mmap` on memory regions.  The `ByteMapMmapper` and `TwoLevelMmapper` become `ByteMapStateStorate` and `TwoLevelStateStorage`, respectively, and they are only responsible for holding the per-chunk states and iterating over ranges of memory to help `ChunkStateMmapper`.  We retains the trait `Mmapper` so that we can still implement other memory mappers that work at different granularities.

`MapState` is now a pure data type, and state-transitioning methods previously implemented in `impl MapState` are inlined into methods of `ChunkStateMmapper`.

We also removed the constant `MMAP_CHUNK_BYTES`. It was defined as `BYTES_IN_CHUNK`, but it seemed to convey the message that the "mmap chunk" is special and can have a size that is different from "chunk".  We instead added a function `Mmapper::granularity()` which returns the actual granularity of the mmapper.

### DIFF
--- a/src/util/heap/layout/mmapper/csm/two_level_storage.rs
+++ b/src/util/heap/layout/mmapper/csm/two_level_storage.rs
@@ -227,27 +227,6 @@ impl TwoLevelStateStorage {
             low = high;
         }
     }
-
-    fn chunk_index_to_address(base: Address, chunk: usize) -> Address {
-        base + (chunk << LOG_BYTES_IN_CHUNK)
-    }
-
-    /// Align `addr` down to slab size.
-    fn slab_align_down(addr: Address) -> Address {
-        addr.align_down(MMAP_SLAB_BYTES)
-    }
-
-    /// Get the base address of the next slab after the slab that contains `addr`.
-    fn slab_limit(addr: Address) -> Address {
-        Self::slab_align_down(addr) + MMAP_SLAB_BYTES
-    }
-
-    /// Return the index of the chunk that contains `addr` within the slab starting at `slab`.
-    /// If `addr` is beyond the end of the slab, the result could be beyond the end of the slab.
-    fn chunk_index(slab: Address, addr: Address) -> usize {
-        let delta = addr - slab;
-        delta >> LOG_BYTES_IN_CHUNK
-    }
 }
 
 impl Default for TwoLevelStateStorage {


wks:
I ran lusearch on mole.moma, with GenImmix, with two heap sizes (2.4x and 3.0x w.r.t. G1), with and without compressed oops, 10 invocations, 5 iterations.  Results show there is no obvious difference compared to the master branch.

The numbers in the plotty plots are normalized to master (>1 means slowdown and <1 means speedup).

https://squirrel.anu.edu.au/plotty/wks/noproject/#0|mole-2025-08-27-Wed-111727&benchmark^build^hfac^invocation^iteration^nocp&GC^time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark^hfac^nocp&build;build1|60&hfac&nocp|70&hfac-nocp^hfac-nocp|10&build^2^build1|40&Histogram%20(with%20CI)^hfac-nocp^benchmark&

src/util/heap/layout/mmapper/csm/byte_map_storage.rs
```diff
@@ -0,0 +1,124 @@
...
+impl fmt::Debug for ByteMapStateStorage {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "ByteMapStateStorage({})", MMAP_NUM_CHUNKS)
+    }
+}
+
+impl MapStateStorage for ByteMapStateStorage {
+    fn get_state(&self, chunk: Address) -> MapState {
+        let index = chunk >> LOG_BYTES_IN_CHUNK;
+        let Some(slot) = self.mapped.get(index) else {
```
qinsoon, line 45: I notice that the `get` method for `ByteMapStateStorage` and `TwoLevelStateStorage` is more conservative. If we query an address that is outside the range supported by the mmapper, it returns `Unmapped`. But if we set an address that is not supported, it panics. Is this intended?

src/util/heap/layout/mmapper/csm/mod.rs
```diff
@@ -0,0 +1,466 @@
+use crate::util::constants::LOG_BYTES_IN_PAGE;
+use crate::util::conversions::raw_is_aligned;
+use crate::util::heap::layout::vm_layout::*;
+use crate::util::heap::layout::Mmapper;
+use crate::util::memory::*;
+use crate::util::Address;
+use bytemuck::NoUninit;
+use std::{io::Result, sync::Mutex};
+
+mod byte_map_storage;
```
qinsoon, line 10: Can we add `#[cfg(target_pointer_width = "32")]` for this type? Clearly it only works for 32 bits addresses.

src/util/heap/layout/mmapper/csm/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+/// The back-end storage of [`ChunkStateMmapper`].  It is responsible for holding the states of each
+/// chunk (eagerly or lazily) and transitioning the states in bulk.
+trait MapStateStorage {
+    /// Return the state of a given `chunk` (must be aligned).
+    ///
+    /// Note that all chunks are logically `MapState::Unmapped` before the states are stored.  They
+    /// include chunks outside the mappable address range.
+    fn get_state(&self, chunk: Address) -> MapState;
+
+    /// Set all chunks within `range` to `state`.
```
qinsoon, line 82: Probably mention that this function only set the state as instructed, and does not do anything with the underlying memory. A user usually should use `bulk_transition_state` instead, where they can define how to transition between memory state.

src/util/heap/layout/mmapper/csm/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+    /// group_state)` for each contiguous chunk range `group_range` that have the same state
+    /// `group_state`.  `transformer` can take actions accordingly and return one of the three
+    /// values:
+    /// -   `Err(err)`: Stop visiting and return `Err(err)` from `bulk_transition_state`
+    ///     immediately.
+    /// -   `Ok(None)`: Continue visiting the next chunk range without changing chunk states.
+    /// -   `Ok(Some(new_state))`: Set the state of all chunks within `group_range` to `new_state`.
+    ///
+    /// Return `Ok(())` if finished visiting all chunks normally.
+    fn bulk_transition_state<F>(&self, range: ChunkRange, transformer: F) -> Result<()>
```
qinsoon: I suggest renaming `transformer` to something more clear, like `update_fn`, `transitioner`, `visitor`, etc.

src/util/heap/layout/mmapper/csm/two_level_storage.rs
```diff
@@ -0,0 +1,253 @@
...
+use crate::util::heap::layout::vm_layout::*;
+use crate::util::rust_util::atomic_box::OnceOptionBox;
+use crate::util::rust_util::rev_group::RevisitableGroupByForIterator;
+use crate::util::rust_util::zeroed_alloc::new_zeroed_vec;
+use crate::util::Address;
+use atomic::{Atomic, Ordering};
+use std::fmt;
+use std::io::Result;
+
+/// Logarithm of the address space size a user-space program is allowed to use.
```
qinsoon: We have a value defined in `VMLayout::log_address_space`. 

This `LOG_MAPPABLE_BYTES` defines how many chunks this storage allows, and it should be strictly larger than the value defined in `VMLayout::log_address_space`.

src/util/heap/layout/mmapper/csm/byte_map_storage.rs
wks: Yes.  It is intended.  `Unmapped` literally means the memory is not mapped by the current `Mmapper`, so returning `Unmapped` makes sense.  For example, if `is_mapped_address(addr)` is passed an `addr` that is outside the governed address range of the `Mmapper`, it is literally unmapped, and it should return false.

But if the user wants to change the mmap state, it must be within the address range that the `Mmapper` can manage.  Therefore it will panic.

src/util/heap/layout/mmapper/csm/mod.rs
wks: Actually it works for 64-bit machines if MMTk only does mmap within the lowest 32-bit range.  It was designed to handle up to something like 39 bit address space, but I limited it to 32 bits for simplicity.  It is supposed to be temporary.  In the future, it should be configurable which `Mmapper` implementation to use at start time.

Another reason why I left it enabled on 64-bit targets is that it is easier to debug this way.  If I changed some global things (such as the `trait MapStateStorage`), the IDE and the compile will immediately complain that something within `byte_map_storage.rs` won't compile.

src/util/heap/layout/mmapper/csm/mod.rs
wks: Actually no methods in the `MapStateStorage` do anything to the underlying memory, including `mmap()`.  `MapStateStorage` only manages the state, and it is the responsibility of `struct ChunkStateMmapper` to actually call `mmap()`.

src/util/heap/layout/mmapper/csm/mod.rs
wks: Yes.  I think `update_fn` is a better name.  I'll make the change.

src/util/heap/layout/mmapper/csm/two_level_storage.rs
wks: Yes.  I'll change the comment, and add an assertion that `VMLayout::log_address_space` cannot exceed `LOG_MAPPABLE_BYTES` when using `TwoLevelStateStorage`.

wks:
@qinsoon I have fixed the problems you mentioned in the comments.

qinsoon (approved):
LGTM

wks:
The merge test failed once due to https://github.com/mmtk/mmtk-core/issues/1378.  I'll retry merging.

PR #1371, opened 2025-08-14
Use `bytemuck::Zeroable` for zeroable types.

wks:
Our utility function `zeroed_alloc::new_zeroed_vec<T>` now requires the element type `<T>` to implement `bytemuck::Zeroable`, a marker trait which means the type can be safely initialized with all zero bits. `new_zeroed_vec` is no longer `unsafe`.

`SpaceDescriptor` now implements `Zeroable` because a zero value represents `SpaceDescriptor::UNINITIALIZED`.

`Map32` now uses `new_zeroed_vec` for allocating the `descriptor_map` because when used on 64-bit machines (usually with compressed pointers), the number of chunks can still be large.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ doctest = false
 atomic = "0.6.0"
 atomic_refcell = "0.1.7"
 atomic-traits = "0.4.0"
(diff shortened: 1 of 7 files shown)


qinsoon (approved):
LGTM

PR #1372, opened 2025-08-20
Add a parallel/regional Compressor

no-defun-allowed:
This PR adds regions to the Compressor. The benefits are that the collector compacts each region in parallel, and the collector supports the use of discontiguous heaps (as required for compressed oops in OpenJDK).


qinsoon:
> Two issues come to mind:
> 
> In the original PR for the Compressor, I modified the CI script to skip some tests on garbage collectors which don't support discontiguous heaps (i.e. only the Compressor). There are no such collectors now, so should I remove that special casing?

Yeah. If no other code needs 'discontiguous', preferably it can be cleaned up.

> I added a `RegionPageResource` to allow the `BumpAllocator` to allocate from a region-structured heap. All allocations take a lock, which I'm somewhat unhappy with, as other page resources appear to be lock-free until they need to grow the heap. But the Compressor still has the best mutator time on DaCapo, so I'm not sure if it is worthwhile to remove the lock.

If there is no measurable performance issue, I would think it is fine for now.

Just to clarify, how is a `RegionPageResource` different from other existing page resources? To my understanding, it returns the required arbitrary number of pages to the space, but it 'internally' organizes all the used memory as regions of the same sizes. `BlockPageResource` only returns blocks (which are regions of the same sizes), and other page resource does not organize used memory as regions of the same sizes. Is this understanding correct?

no-defun-allowed:
> Just to clarify, how is a RegionPageResource different from other existing page resources?

I think your understanding is right. To elaborate, the `RegionPageResource` structures the heap into regions, and has a bump allocator for each region. The `RegionPageResource` can handle requests for any number of pages smaller than a region; but requests probably should be smaller than the region size by a decent margin, to avoid fragmentation. The regions in the Compressor are 1MiB, though that size was picked arbitrarily, and I don't know if that size is a good size for any idea of "good".

`RegionPageResource` allows the GC design to assume that objects never span multiple regions when they are allocated, which is necessary to compact each region separately; and that the objects in a region will be contiguously allocated, which we utilise in only scanning for objects between the start and allocation cursor of a region (though I haven't measured how much this matters).

src/util/heap/regionpageresource.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::util::linear_scan::Region;
+use crate::util::object_enum::ObjectEnumerator;
+use crate::util::Address;
+use crate::util::VMThread;
+use crate::vm::VMBinding;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::RwLock;
+
+/// A region in a RegionPageResource and its allocation cursor.
+pub struct RegionAllocator<R: Region> {
```
qinsoon: I am slightly concerned that the name `RegionAllocator` could cause some confusion. The name looks similar to allocators in `util::alloc`, but this one does not implement `util::alloc::Allocator` and they are not the same kind of allocators. But I don't have a strong opinion on this.

@wks What do you think?

src/util/heap/regionpageresource.rs
```diff
@@ -0,0 +1,194 @@
...
+        space_descriptor: SpaceDescriptor,
+        reserved_pages: usize,
+        required_pages: usize,
+        tls: VMThread,
+    ) -> Result<PRAllocResult, PRAllocFail> {
+        let mut b = self.sync.write().unwrap();
+        let succeed = |start: Address, new_chunk: bool| {
+            Result::Ok(PRAllocResult {
+                start,
+                pages: Self::TLAB_PAGES,
```
qinsoon: Why does this return the fixed number of 8 pages, instead of what is requested? If the code really assumes the reserved pages is exactly 8 pages, you can add an assertion to check that.

src/util/heap/regionpageresource.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    fn get_available_physical_pages(&self) -> usize {
+        self.mpr.get_available_physical_pages()
+    }
+}
+
+impl<VM: VMBinding, R: Region + 'static> RegionPageResource<VM, R> {
+    // Same as crate::util::alloc::bumpallocator::BLOCK_SIZE
+    const TLAB_PAGES: usize = 8;
+    const TLAB_BYTES: usize = Self::TLAB_PAGES * BYTES_IN_PAGE;
```
qinsoon: See my comment on `TLAB_PAGES` in `alloc`. I am not sure why we need these constants. But if you do need them and assume they are the same as `bumpallocator::BLOCK_SIZE`, they should be computed from `bumpallocator::BLOCK_SIZE`.

src/util/heap/regionpageresource.rs
```diff
@@ -0,0 +1,194 @@
...
+            })
+        };
+        let bytes = reserved_pages * BYTES_IN_PAGE;
+        // First try to reuse a region.
+        while b.next_region < b.all_regions.len() {
+            let cursor = b.next_region;
+            if let Option::Some(address) =
+                self.allocate_from_region(&mut b.all_regions[cursor], bytes)
+            {
+                self.commit_pages(reserved_pages, required_pages, tls);
```
qinsoon, line 118: This commits `reserved_pages`, but `succeed` returns 8 pages.

qinsoon:
Just out of curiosity, what is performance like for this PR? It makes the offset calculation and the actual copying parallelized. I would expect it to improve STW time.

src/util/heap/regionpageresource.rs
no-defun-allowed: Indeed we don't need them.

src/util/heap/regionpageresource.rs
no-defun-allowed: It used to assume that, but there's absolutely no need, and I seem to have gotten confused somewhere whilst refactoring that assumption away.

src/util/heap/regionpageresource.rs
k-sareen: What about `AllocatedRegion`? I think that is semantically closer to what it represents as well

qinsoon: `AllocatedRegion` sounds good.

src/util/heap/regionpageresource.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::util::Address;
+use crate::util::VMThread;
+use crate::vm::VMBinding;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::RwLock;
+
+/// A region in a RegionPageResource and its allocation cursor.
+pub struct RegionAllocator<R: Region> {
+    pub region: R,
+    cursor: AtomicUsize,
```
k-sareen: I'm pretty sure `Atomic<Address>` also works. It'll simplify the back-and-forth you do with `usize`. Though hopefully `Atomic<Address>` doesn't do anything silly on other platforms.

src/util/heap/regionpageresource.rs
no-defun-allowed: Renamed in 238f33c.

no-defun-allowed:
> Just out of curiosity, what is performance like for this PR? It makes the offset calculation and the actual copying parallelized. I would expect it to improve STW time.

I don't have apples-to-apples results handy - I have benchmark results with compressed oops for the regional Compressor, but none without compressed oops.

k-sareen:
> > Just out of curiosity, what is performance like for this PR? It makes the offset calculation and the actual copying parallelized. I would expect it to improve STW time.
> 
> I don't have apples-to-apples results handy - I have benchmark results with compressed oops for the regional Compressor, but none without compressed oops.

The comparisons against SemiSpace, Immix, etc. would still be interesting to see. I don't think Yi would have seen those results.

no-defun-allowed:
> > > Just out of curiosity, what is performance like for this PR? It makes the offset calculation and the actual copying parallelized. I would expect it to improve STW time.
> > 
> > 
> > I don't have apples-to-apples results handy - I have benchmark results with compressed oops for the regional Compressor, but none without compressed oops.
> 
> The comparisons against SemiSpace, Immix, etc. would still be interesting to see. I don't think Yi would have seen those results.

Right. I have such comparisons on plotty: http://squirrel.anu.edu.au/plotty/hayleyp/plots/p/Z9KkGZ

src/util/heap/regionpageresource.rs
no-defun-allowed: Fixed in de59733

wks:
Is the main reason for introducing `RegionPageResource` the ability to enumerate regions?  It holds a `Vec<RegionAllocator<R>>` so that you can iterate through it and process multiple regions in parallel.

But if `CompressorSpace` is backed by a `MonotonicPageResource`, you can already iterate using `iterate_allocated_regions`.  Each "allocated region" is one chunk obtained from `VMMap` (which can be either `Map32` or `Map64`), and it will always be exactly one chunk because no spaces except LOS can allocate objects larger than a chunk.  All chunks will be almost full, and the last chunk will be filled up to `MonotonicPageResourceSync::cursor`.  Because each chunk is exactly 4MB, you can always divide it into integer multiples of "regions" chosen by `CompressorSpace` (in the case where each `CompressorRegion` is 1MB, each chunk can be split into 4 regions.  Then you can wrap the iterator implemented in `MonotonicPageResource::iterate_allocated_regions` to make your "region iterator".

In this way, you don't really need the `RegionPageResource` or a `Vec<RegionAllocator<R>>`.  You can just use `MonotonicPageResource`, and iterate over 1MB "virtual" regions.

no-defun-allowed:
>  It never moves objects between regions.

> So we need one cursor per region to identify where we can continue allocating.

Exactly.

src/util/heap/regionpageresource.rs
no-defun-allowed: Changed in a0c1d30

qinsoon:
The test fail in Julia should be unrelated with this PR.

qinsoon (approved):
LGTM

k-sareen:
We should make a separate PR for the OpenJDK binding to remove the `-XX:-UseCompressedOops` from the CI for the Compressor.

k-sareen:
binding-refs
OPENJDK_BINDING_REF=enable-compressed-oops-compressor

PR #1373, opened 2025-08-22
Rename Workbucket's active flag to open, add enable flag

qinsoon:
This PR refactors work bucket to allow disabling a bucket and to support concurrent GC.
* Add a field `WorkBucket::enabled`: if a bucket is disabled, it behaves the same as if the bucket/stage does not exist for scheduling, except that users can add work to a disabled bucket. There are mostly two use cases for this: 1. a plan can disable buckets that the plan does not use, and 2. a concurrent plan may push concurrent work to the disabled concurrent bucket during STW, and enable the bucket at the end of a STW.
* Add `WorkBucketStage::Concurrent`: this is unused in this PR.
* Clean up most checks about stages in `scheduler.rs`, and move them to `work_bucket.rs`.
* Add a reference to the scheduler in `BasePlan`.

### DIFF
--- a/src/util/finalizable_processor.rs
+++ b/src/util/finalizable_processor.rs
@@ -153,19 +153,31 @@ impl<E: ProcessEdgesWork> GCWork<E::VM> for Finalization<E> {
         }
 
         let mut finalizable_processor = mmtk.finalizable_processor.lock().unwrap();
+        let num_candidates_begin = finalizable_processor.candidates.len();
+        let num_ready_for_finalize_begin = finalizable_processor.ready_for_finalize.len();
         debug!(
             "Finalization, {} objects in candidates, {} objects ready to finalize",
-            finalizable_processor.candidates.len(),
-            finalizable_processor.ready_for_finalize.len()
+            num_candidates_begin, num_ready_for_finalize_begin
         );
 
         let mut w = E::new(vec![], false, mmtk, WorkBucketStage::FinalRefClosure);
         w.set_worker(worker);
         finalizable_processor.scan(worker.tls, &mut w, is_nursery_gc(mmtk.get_plan()));
+
+        let num_candidates_end = finalizable_processor.candidates.len();
+        let num_ready_for_finalize_end = finalizable_processor.ready_for_finalize.len();
+
         debug!(
             "Finished finalization, {} objects in candidates, {} objects ready to finalize",
-            finalizable_processor.candidates.len(),
-            finalizable_processor.ready_for_finalize.len()
+            num_candidates_end, num_ready_for_finalize_end
+        );
+        probe!(
+            mmtk,
+            finalization,
+            num_candidates_begin,
+            num_candidates_end,
+            num_ready_for_finalize_begin,
+            num_ready_for_finalize_end
         );
     }
 }
--- a/src/util/reference_processor.rs
+++ b/src/util/reference_processor.rs
@@ -146,7 +146,7 @@ pub struct ReferenceProcessor {
     allow_new_candidate: AtomicBool,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone, Copy)]
 pub enum Semantics {
     SOFT,
     WEAK,
@@ -371,13 +371,26 @@ impl ReferenceProcessor {
             .filter_map(|reff| self.process_reference::<VM>(*reff, &mut enqueued_references))
             .collect();
 
+        let num_old = sync.references.len();
+        let num_new = new_set.len();
+        let num_enqueued = enqueued_references.len();
+
         debug!(
             "{:?} reference table from {} to {} ({} enqueued)",
-            self.semantics,
-            sync.references.len(),
-            new_set.len(),
-            enqueued_references.len()
+            self.semantics, num_old, num_new, num_enqueued,
         );
+
+        let semantics_int = self.semantics as usize;
+
+        probe!(
+            mmtk,
+            reference_scanned,
+            semantics_int,
+            num_old,
+            num_new,
+            num_enqueued
+        );
+
         sync.references = new_set;
         sync.enqueued_references.extend(enqueued_references);
 
@@ -400,6 +413,10 @@ impl ReferenceProcessor {
             sync.references
         );
 
+        let num_refs = sync.references.len();
+        let mut num_live = 0usize;
+        let mut num_retained = 0usize;
+
         for reference in sync.references.iter() {
             trace!("Processing reference: {:?}", reference);
 
@@ -408,13 +425,17 @@ impl ReferenceProcessor {
                 // following trace. We postpone the decision.
                 continue;
             }
+            num_live += 1;
             // Reference is definitely reachable.  Retain the referent.
             if let Some(referent) = <E::VM as VMBinding>::VMReferenceGlue::get_referent(*reference)
             {
                 Self::keep_referent_alive(trace, referent);
+                num_retained += 1;
                 trace!(" ~> {:?} (retained)", referent);
             }
         }
+
+        probe!(mmtk, reference_retained, num_refs, num_live, num_retained,);
 
         debug!("Ending ReferenceProcessor.retain({:?})", self.semantics);
     }
--- a/tools/tracing/README.md
+++ b/tools/tracing/README.md
@@ -58,6 +58,18 @@ Currently, the core provides the following tracepoints.
 -   `mmtk:alloc_slow_once_end()`: the allocation slow path ends.
 -   `mmtk:plan_end_of_gc_begin()`: before executing `Plan::end_of_gc`.
 -   `mmtk:plan_end_of_gc_end()`: after executing `Plan::end_of_gc`.
+-   `mmtk:finalization(cb: int, ce: int, rb: int, re: int)`: a `Finalization` work packet.  The
+    arguments are the number of candidates at the beginning and the end of the work packet, and the
+    number of ready-to-finalize objects at the beginning and the end of the work packet.
+-   `mmtk:reference_scanned(semantics: int, old: int, new: int, enqueued: int)`: An invocation of
+    `ReferenceProcessor::scan`.  `semantics` is the semantics.  `old` and `new` are the number of
+    references of this semantics before and other this invocation, and `eneueue` is the number of
+    references enqueued.
+-   `mmtk:reference_retained(num_refs: int, num_live: int, num_retained: int)`: An invocation of
+    `ReferenceProcessor::retain`.  `num_refs` is the total number of reference objects visited.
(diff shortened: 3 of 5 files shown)


src/scheduler/work_bucket.rs
```diff
@@ -297,8 +383,48 @@ pub enum WorkBucketStage {
...
+    pub const fn is_sequentially_activated(&self) -> bool {
+        self.is_stw() && !self.is_first_stw_stage()
+    }
+
+    /// Is this stage a stop-the-world stage?
+    pub const fn is_stw(&self) -> bool {
+        !self.is_concurrent()
+    }
+
+    /// Is this stage concurrent (which may be executedd during mutator time)?
```
wks:
```suggestion
    /// Is this stage concurrent (which may be executed during mutator time)?
```

src/scheduler/work_bucket.rs
```diff
@@ -111,12 +176,16 @@ impl<VM: VMBinding> WorkBucket<VM> {
     }
 
     pub fn is_drained(&self) -> bool {
-        self.is_activated() && self.is_empty()
+        self.is_disabled() || (self.is_activated() && self.is_empty())
     }
 
     /// Disable the bucket
```
wks: This comment says "disable the bucket"

src/scheduler/work_bucket.rs
```diff
@@ -39,13 +39,55 @@ impl<VM: VMBinding> BucketQueue<VM> {
...
+            for work in items {
+                self.queue.push(work);
+            }
+        }
+
+        f.debug_struct("BucketQueue")
+            .field("queue", &debug_items)
+            .finish()
+    }
+}
```
wks: This part drains the bucket to print the packets, and re-fill the packets.  If this happens while some GC workers are fetching packets, the other worker may erroneously think the bucket is already drained.

If we really want to debug work buckets in such a detail, we should add a dedicated method, such as `BucketQueue::debug_dump_packets`.  The `Debug` trait should be side-effect-free.

src/scheduler/scheduler.rs
```diff
@@ -326,7 +353,7 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
             }
         }
         if let Some(id) = error_example {
-            panic!("Some active buckets (such as {:?}) are not empty.", id);
+            panic!("Some open buckets (such as {:?}) are not empty.", id);
```
wks, line 350: This says "open".

src/scheduler/scheduler.rs
```diff
@@ -260,10 +282,15 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
+            if id.is_always_activated() {
                 continue;
             }
             let bucket = &self.work_buckets[id];
+            if bucket.is_disabled() {
+                debug!("Work bucket {:?} is disabled. Skip.", id);
+                continue;
+            }
+            debug!("Checking if {:?} can be opened...", id);
             let bucket_opened = bucket.update(self);
```
wks, line 288: This line also says "opened".

src/scheduler/work_bucket.rs
```diff
@@ -59,20 +101,43 @@ pub struct WorkBucket<VM: VMBinding> {
...
+    pub(crate) fn new(stage: WorkBucketStage, active: bool, monitor: Arc<WorkerMonitor>) -> Self {
         Self {
             active: AtomicBool::new(active),
+            stage,
             queue: BucketQueue::new(),
             prioritized_queue: None,
             monitor,
             can_open: None,
             sentinel: Mutex::new(None),
+            disable: AtomicBool::new(false),
```
wks: It is confusing to have a negative property.  "enabled" should be better.

src/scheduler/work_bucket.rs
```diff
@@ -59,20 +101,43 @@ pub struct WorkBucket<VM: VMBinding> {
...
         }
     }
 
+    pub fn set_as_enabled(&self) {
+        self.disable.store(false, Ordering::SeqCst)
+    }
+
+    pub fn set_as_disabled(&self) {
+        self.disable.store(true, Ordering::SeqCst)
+    }
```
wks: We can implement them with a single method

```rust
pub fn set_enabled(&self, enabled: bool)
```

Or we can make the names of the two methods shorter: "disable" and "enable".

src/scheduler/work_bucket.rs
```diff
@@ -59,20 +101,43 @@ pub struct WorkBucket<VM: VMBinding> {
...
+
+    pub fn is_disabled(&self) -> bool {
+        self.disable.load(Ordering::Relaxed)
+    }
```
wks: Like the field name, I think "is_enabled" is better.

src/scheduler/work_bucket.rs
```diff
@@ -39,13 +39,55 @@ impl<VM: VMBinding> BucketQueue<VM> {
...
+            .field("queue", &debug_items)
+            .finish()
+    }
+}
+
 pub type BucketOpenCondition<VM> = Box<dyn (Fn(&GCWorkScheduler<VM>) -> bool) + Send>;
 
 pub struct WorkBucket<VM: VMBinding> {
+    /// Whether this bucket has been opened. Work from an open bucket can be fetched by workers.
     active: AtomicBool,
```
wks: Change this to "open".

src/scheduler/work_bucket.rs
```diff
@@ -297,8 +383,48 @@ pub enum WorkBucketStage {
...
+
+    /// Is this the first stop-the-world stage? See [`Self::FIRST_STW_STAGE`].
+    pub const fn is_first_stw_stage(&self) -> bool {
+        matches!(self, &WorkBucketStage::FIRST_STW_STAGE)
+    }
+
+    /// Is this stage always activated?
+    pub const fn is_always_activated(&self) -> bool {
+        matches!(self, WorkBucketStage::Unconstrained)
+    }
```
wks:
```suggestion
    pub const fn is_always_open(&self) -> bool {
        matches!(self, WorkBucketStage::Unconstrained)
    }
```

src/scheduler/work_bucket.rs
```diff
@@ -297,8 +383,48 @@ pub enum WorkBucketStage {
...
+
+    /// Is this stage activated by default?
+    pub const fn is_activated_by_default(&self) -> bool {
```
wks:
```suggestion
    pub const fn is_open_by_default(&self) -> bool {
```

src/scheduler/work_bucket.rs
```diff
@@ -297,8 +383,48 @@ pub enum WorkBucketStage {
...
+        matches!(
+            self,
+            WorkBucketStage::Unconstrained | WorkBucketStage::Concurrent
+        )
+    }
+
+    /// Is this stage disabled by default?
+    pub const fn is_disabled_by_default(&self) -> bool {
+        matches!(self, WorkBucketStage::Concurrent)
+    }
```
wks:
```suggestion
    pub const fn is_enabled_by_default(&self) -> bool {
        !matches!(self, WorkBucketStage::Concurrent)
    }
```

src/scheduler/work_bucket.rs
```diff
@@ -59,20 +101,43 @@ pub struct WorkBucket<VM: VMBinding> {
...
+    /// except that users can add work to a disabled bucket, and enable it later to allow those
+    /// work to be scheduled.
+    disable: AtomicBool,
 }
 
 impl<VM: VMBinding> WorkBucket<VM> {
-    pub(crate) fn new(active: bool, monitor: Arc<WorkerMonitor>) -> Self {
+    pub(crate) fn new(stage: WorkBucketStage, active: bool, monitor: Arc<WorkerMonitor>) -> Self {
         Self {
             active: AtomicBool::new(active),
```
wks: If a stage knows if the bucket at that stage is open by default, we can just initialize it here.

```suggestion
    pub(crate) fn new(stage: WorkBucketStage, monitor: Arc<WorkerMonitor>) -> Self {
        Self {
            active: AtomicBool::new(stage.is_open_by_default()),
```

wks (changes requested):
The algorithm should be OK because we haven't actually added concurrent GC, and I think this PR is compatible with existing plans.

One problem is that the newly introduced terms "enabled/disabled" is somewhat confusing when used together with "active/inactive" or "activated/deactivated".  The doc comment of `WorkBucket::deactivate` even says "disable the bucket".  I suggest **renaming "active" to "open"**.  Then "open/closed" means if the bucket is open during a GC as a result of, for example, draining previous buckets.  Meanwhile, "enabled/disabled" means whether we hide a bucket from mutators as if the bucket doesn't exist.

qinsoon:
@wks I have made changes as suggested. Can you review again?

src/scheduler/work_bucket.rs
```diff
@@ -37,15 +37,55 @@ impl<VM: VMBinding> BucketQueue<VM> {
...
+    open: AtomicBool,
+    /// The stage name of this bucket.
+    stage: WorkBucketStage,
     queue: BucketQueue<VM>,
     prioritized_queue: Option<BucketQueue<VM>>,
     monitor: Arc<WorkerMonitor>,
+    /// The open condition for a bucket. If this is `Some`, the bucket will be activated
+    /// when the condition is met. If this is `None`, the bucket needs to be activated manually.
     can_open: Option<BucketOpenCondition<VM>>,
     /// After this bucket is activated and all pending work packets (including the packets in this
```
wks:
```suggestion
    /// After this bucket is opened and all pending work packets (including the packets in this
```

src/scheduler/work_bucket.rs
```diff
@@ -59,23 +99,43 @@ pub struct WorkBucket<VM: VMBinding> {
     /// recursively, such as ephemerons and Java-style SoftReference and finalizers.  Sentinels
     /// can be used repeatedly to discover and process more such objects.
     sentinel: Mutex<Option<Box<dyn GCWork<VM>>>>,
+
+    /// Whether this bucket is enabled.
+    /// A disabled work bucket will behave as if it does not exist in terms of scheduling,
+    /// except that users can add work to a disabled bucket, and enable it later to allow those
+    /// work to be scheduled.
+    enabled: AtomicBool,
```
wks: Consider moving this field just after (or just before) the `open` field.  These two fields are closely related, although different.

qinsoon:
> There are still other places in the comments that mention "activated buckets". We should grep the word "activate" and change it to "open" or "opened", depending on the context.
> 
> We should update the pull request description [#1373 (comment)](https://github.com/mmtk/mmtk-core/pull/1373#issue-3344150675), too, as it will become the commit log.

Done.

wks (approved):
LGTM.

I also pushed some pedantic changes and synchronized with master.  You can merge this PR when you think it is appropriate.

PR #1374, opened 2025-08-22
eBPF probes for finalization and reference processing

wks:
We add some eBPF probes to show more details of finalization and reference processing in the MMTk core, including the number of finalizable objects processed and the number of references scanned or retained in various work packets.


qinsoon (approved):
LGTM

PR #1386, opened 2025-09-05
Add `AllInSet` affinity kind for pinning threads to all cores in a set

k-sareen:


src/util/options.rs
```diff
@@ -881,10 +910,22 @@ options! {
...
-    /// example: 0,5,8-11 specifies that cores 0,5,8,9,10,11 should be used for pinning threads.
+    /// scheduler.
+    ///
+    /// There are two ways cores can be allocated to threads:
+    ///  1. round-robin, wherein each GC thread is allocated exactly one core to run
+    ///     on in a round-robin fashion; and
+    ///  2. "all in set", wherein each GC thread is allocated all the cores in the provided
+    ///     CPU set.
+    ///
+    /// The method can selected by specifying "`RoundRobin:<core ids>`" or "`AllInSet:<core ids>`"
```
qinsoon:
```suggestion
    /// The method can be selected by specifying "`RoundRobin:<core ids>`" or "`AllInSet:<core ids>`"
```

src/util/options.rs
```diff
@@ -367,6 +370,28 @@ impl AffinityKind {
             return Ok(AffinityKind::OsDefault);
         }
 
+        let mut all_in_set = false;
+        let kind_split: Vec<&str> = cpulist.splitn(2, ':').collect();
```
qinsoon, line 385: Some comments are needed for this section of code to explain that this is dealing with two kinds of string patterns: one that specifies an affinity kind, and the other that omits the affinity kind. It is a bit hard to understand what it tries to do here, unless we look at the comments on the option first. 

We could also always ask users to explicitly specify the affinity kind. This would be more consistent with other options.

k-sareen: I just wanted to maintain backwards compatibility. I'll add the comments.

PR #1392, opened 2025-09-18
Warn the user about stroboscopic effect

wks:
If the user chooses an even number for the `-e` option of `capture.py`, it may get misleading results if the GCs alternate between two different behaviors, such as nursery GC and full-heap GC.  We now warn the user if the value of `-e` is an even number.

### DIFF
--- a/tools/tracing/timeline/README.md
+++ b/tools/tracing/timeline/README.md
@@ -19,16 +19,21 @@ Suite.
 Run the following command with a **normal** user (*not* as `root` or using `sudo`):
 
 ```shell
-./capture.py -e 50 -m /path/to/libmmtk_openjdk.so --no-root-nodes
+./capture.py -e 47 -m /path/to/libmmtk_openjdk.so --no-root-nodes
 ```
 
-`-e 50` means we only capture one GC in every 50 GCs because otherwise it will have to print too
+`-e 47` means we only capture one GC in every 47 GCs because otherwise it will have to print too
 much log.  (Note: Printing in bpftrace is done via a fixed-size user/kernel space buffer, therefore
 excessive printing will overrun the buffer and cause events to be dropped.  The `-e` option helps
 reducing the volume of log, thereby reducing the likelihood of buffer overrun and the time for
 post-processing.  If one single GC still produces too much log and overruns the buffer, the user
 should consider setting the `BPFTRACE_PERF_RB_PAGES` environment variable.  See the man page of
-`bpftrace`.)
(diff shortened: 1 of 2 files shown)


k-sareen:
The broad term would be sampling bias.

tools/tracing/timeline/README.md
```diff
@@ -19,16 +19,21 @@ Suite.
...
 
-`-e 50` means we only capture one GC in every 50 GCs because otherwise it will have to print too
+`-e 47` means we only capture one GC in every 47 GCs because otherwise it will have to print too
 much log.  (Note: Printing in bpftrace is done via a fixed-size user/kernel space buffer, therefore
 excessive printing will overrun the buffer and cause events to be dropped.  The `-e` option helps
 reducing the volume of log, thereby reducing the likelihood of buffer overrun and the time for
 post-processing.  If one single GC still produces too much log and overruns the buffer, the user
 should consider setting the `BPFTRACE_PERF_RB_PAGES` environment variable.  See the man page of
-`bpftrace`.)
+`bpftrace`.)  We choose a large prime number, such as 47, because some GCs may exhibit periodic
```
qinsoon, line 31: I don't think a prime number helps much. It is just sampling. GC behavior isn’t perfectly periodic (full GCs can occur irregularly, nursery/full ratios vary), so the only guarantee is that -e N reduces logging by about an N-fold factor. That’s the practical benefit. 

'stroboscopic effect' doesn't precisely describe the issue. Better just remove it to avoid any confusion.

tools/tracing/timeline/capture.py
```diff
@@ -94,9 +97,14 @@ def main():
 
         command_line = ["sudo", args.bpftrace] + extra_options + ["--unsafe", tmp.name]
 
+        if args.every % 2 == 0:
```
qinsoon, line 100: The check is unnecessary.

tools/tracing/timeline/README.md
qinsoon: According to what we discussed in the meeting, it is helpful for certain cases to use prime numbers and avoid using powers of 2.

tools/tracing/timeline/capture.py
wks: Why? This is the heart of this PR, i.e. warn the user of potential problems.

qinsoon: I commented based on https://github.com/mmtk/mmtk-core/pull/1392#discussion_r2366500472. Now I am persuaded that such a check is helpful, and I marked this comment as 'resolved'.

qinsoon (approved):

PR #1393, opened 2025-09-19
Fix LargeObjectSpace::initialize_object_metadata

wks:
It did not set the VO bit when allocating as live (by ConcurrentImmix). We fixed that.  It's always set as long as VO bit is enabled.

We simplified the logic for setting the mark/nursery bits.  We no longer load from the bits.

We always assert the global unlog bit is unset as long as the LOS needs unlog bits.


qinsoon (approved):
LGTM

PR #1394, opened 2025-09-19
Fix Clippy 1.90 warning

wks:
It now warns that `SFTProcessEdges` is unused.  But we still provide the `SFTProcessEdges` work packet so that in the future we can create other plans that use it.

We now define `OpaquePointer::UNINITIALIZED` using `std::ptr::null_mut::<c_void>()`.


qinsoon (approved):
LGTM

PR #1400, opened 2025-10-09
Use boolean allocation options

wks:
We remove `OnAllocationFail` and add boolean fields to `AllocationOptions`:

-   `allow_overcommit`: whether we allow overcommit
-   `at_safepoint`: whether this allocation is at a safepoint
-   `allow_oom_call`: whether to call `Collection::out_of_memory`

Now `Space::acquire` always polls before trying to get new pages.  Particularly, when `allow_overcommit == true`, polling and over-committing will happen in one allocation attempt.  If we also set `at_safepoint == false`, the current mutator will be able to allocate normally in this allocation, but block for GC at the nearest safepoint.  This is useful for certain VMs.

### DIFF
--- a/src/vm/tests/mock_tests/mock_test_allocate_no_gc_oom_on_acquire_allow_oom_call.rs
+++ b/src/vm/tests/mock_tests/mock_test_allocate_no_gc_oom_on_acquire_allow_oom_call.rs
@@ -7,7 +7,6 @@ use crate::AllocationSemantics;
 /// It will call `Collection::out_of_memory` and return null.
 #[test]
 pub fn allocate_no_gc_oom_on_acquire_allow_oom_call() {
-    //let mut out_of_memory_called = false;
     // 1MB heap
     with_mockvm(
         || -> MockVM {
@@ -41,44 +40,6 @@ pub fn allocate_no_gc_oom_on_acquire_allow_oom_call() {
             // out_of_memory should be called.
             read_mockvm(|mock| {
                 assert!(mock.out_of_memory.is_called());
-            });
-        },
-        no_cleanup,
-    )
-}
-
-/// This test will allocate an object that is larger than the heap size. The call will fail by
-/// returning null.
-#[test]
-pub fn allocate_no_gc_oom_on_acquire_no_oom_call() {
-    // 1MB heap
-    with_mockvm(
-        default_setup,
-        || {
-            const KB: usize = 1024;
-            let mut fixture = MutatorFixture::create_with_heapsize(KB);
-
-            // Attempt to allocate an object that is larger than the heap size.
-            let addr = memory_manager::alloc_with_options(
-                &mut fixture.mutator,
-                1024 * 10,
-                8,
-                0,
-                AllocationSemantics::Default,
-                AllocationOptions {
-                    at_safepoint: false,
-                    allow_oom_call: false,
-                    ..Default::default()
-                },
-            );
-            // We should get zero.
-            assert!(addr.is_zero());
-            // block_for_gc and out_of_memory won't be called.
-            read_mockvm(|mock| {
-                assert!(!mock.block_for_gc.is_called());
-            });
-            read_mockvm(|mock| {
-                assert!(!mock.out_of_memory.is_called());
             });
         },
         no_cleanup,
--- a/src/vm/tests/mock_tests/mock_test_allocate_no_gc_oom_on_acquire_no_oom_call.rs
+++ b/src/vm/tests/mock_tests/mock_test_allocate_no_gc_oom_on_acquire_no_oom_call.rs
@@ -1,0 +1,42 @@
+use super::mock_test_prelude::*;
+
+use crate::util::alloc::allocator::AllocationOptions;
+use crate::AllocationSemantics;
+
+/// This test will allocate an object that is larger than the heap size. The call will fail by
+/// returning null.
+#[test]
+pub fn allocate_no_gc_oom_on_acquire_no_oom_call() {
+    // 1MB heap
+    with_mockvm(
+        default_setup,
+        || {
+            const KB: usize = 1024;
+            let mut fixture = MutatorFixture::create_with_heapsize(KB);
+
+            // Attempt to allocate an object that is larger than the heap size.
+            let addr = memory_manager::alloc_with_options(
+                &mut fixture.mutator,
+                1024 * 10,
+                8,
+                0,
+                AllocationSemantics::Default,
+                AllocationOptions {
+                    at_safepoint: false,
+                    allow_oom_call: false,
+                    ..Default::default()
+                },
+            );
+            // We should get zero.
+            assert!(addr.is_zero());
+            // block_for_gc and out_of_memory won't be called.
+            read_mockvm(|mock| {
+                assert!(!mock.block_for_gc.is_called());
+            });
+            read_mockvm(|mock| {
+                assert!(!mock.out_of_memory.is_called());
+            });
+        },
+        no_cleanup,
+    )
+}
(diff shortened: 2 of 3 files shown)


src/util/alloc/allocator.rs
```diff
@@ -28,44 +28,124 @@ pub enum AllocationError {
...
+    /// **The default is `true`**.
+    ///
+    /// If `true`, the allocation will let the GC trigger poll before acquiring pages from the page
+    /// resource, giving the GC trigger a chance to schedule a collection.
+    ///
+    /// If `false`, the allocation will not notify the GC trigger *before* acquiring pages from the
+    /// page resource.  Note that if the allocation is at a safepoint (i.e. [`Self::at_safepoint`]
+    /// is true), it will still poll and force a GC *after* failing to get pages from the page
+    /// resource due to physical memory exhaustion.
+    pub eager_polling: bool,
```
qinsoon: From users' point of view, they just call `alloc` and they don't see eager poll or lazy poll -- they either want their `alloc` call to poll or not poll. As we discussed, we can remove this option, or replace with `allow_polling` which affects both polling in `acquire`.

src/vm/tests/mock_tests/mock_test_allocate_overcommit.rs
```diff
@@ -1,6 +1,8 @@
+// GITHUB-CI: MMTK_PLAN=NoGC
```
qinsoon: Why is this test only for NoGC?

wks: Well, it doesn't have to.  The test is about the behavior of allocation after exceeding the heap size, and it is not about the GC.  So it doesn't really matter which plan it is.  But I changed it to "all" just in case any plan triggers GC differently (mainly ConcurrentImmix).

src/util/alloc/allocator.rs
```diff
@@ -28,44 +28,116 @@ pub enum AllocationError {
...
+    /// is triggered by the polling.
+    ///
+    /// If `false` the allocation will not try to get pages from page resource as long as GC is
+    /// triggered.
+    ///
+    /// Note that MMTk lets the GC trigger poll before trying to acquire pages from the page
+    /// resource.  This gives the GC trigger a chance to trigger GC if needed.  `allow_overcommit`
+    /// does not disable polling, but only controls whether to try acquiring pages when GC is
+    /// triggered.
+    pub allow_overcommit: bool,
```
qinsoon: I think the doc is way more detailed than what is necessary.

None of these is related with 'over commit':
* MMTk will acquire pages
* MMTk uses page resources

'Overcommit' only means one thing: MMTk may go beyond the specified heap size, in order to satisfy this allocation request. Additionally, MMTk still triggers GC when it overcommits memory.

src/util/alloc/allocator.rs
```diff
@@ -28,44 +28,116 @@ pub enum AllocationError {
...
+    /// -   The GC trigger polled and triggered a GC before the allocation tries to get more pages
+    ///     from the page resource, and the allocation does not allow over-committing.
+    /// -   The allocation tried to get more pages from the page resource, but failed.  In this
+    ///     case, it will force a GC.
+    ///
+    /// If `false`, the allocation will immediately return a null address if the allocation cannot
+    /// be satisfied without a GC.  It will never block for GC, never force a GC, and never call
+    /// [`Collection::out_of_memory`].  Note that the VM can always force a GC by calling
+    /// [`crate::MMTK::handle_user_collection_request`] with the argument `force` being `true`.
+    pub at_safepoint: bool,
```
qinsoon, line 53: Same here. `at_safepoint` means MMTk may block the thread in this allocation request for GC. These are not related with `at_safepoint`:
* MMTk may call `out_of_memory`. MMTk calls `out_of_memory` when it runs out of memory, which is unrelated with `at_safepoint`. Before we have a specific flag like `allow_oom_call`, there is no definition when MMTk may call `out_of_memory` -- it is an implementation detail.
* The reasons why MMTk may block for GC are implementation details.
* `handle_user_collection_request` is unrelated. It is a separate API, and is not related with `alloc_with_options`.

qinsoon (approved):
The PR looks good to me.

There are just some minor points about the documentation -- I think it currently over-specifies the behavior of the new boolean flags. It’s good to be specific and precise, but the docs shouldn’t expose internal implementation details or define behaviors that are not controlled by the flag.

src/util/alloc/allocator.rs
wks: Yes.  I'll simplify the description and make the point that "when over-committing, it may allocate beyond the heap size".

src/util/alloc/allocator.rs
wks: We still have a method `allow_oom_call`.  It previously returns `true` only for `OnAllocationFail::RequestGC`.  Maybe we should add another option `AllocationOptions::allow_oom_call` for that.

`handle_user_collection_request` is relevant because if `alloc_with_option(at_safepoint=false)` cannot force a GC, and it can't satisfy the allocation request without a GC, the user would trigger a GC manually instead.  Currently, it mimics the behavior of `OnAllocationFail::RequestGC` and it forces a GC when it is at safepoint, and GC is initialized.

Alternatively, we can make "forcing GC" (i.e. the second `poll()` invocation with `space_full = true`) unstoppable, too.  That is, as long as we fail to get pages from the page resource, it will force a GC.  But it may return `null` if it is not at a safepoint.  This will slightly change the control flow.  One concern is that what will happen if GC is not initialized, and we failed to get pages from the PR, and it is not at safepoint?  Should it panic immediately, or should it return null?  @qinsoon what do you think of this?

qinsoon: My opinions on this are:
1. Binding users do not need to know the 'forced GC'. There is little point mentioning it.
2. `at_safepoint` only makes sure the thread will not be blocked in this call for GC. It does not do anything more than that.
3. Whether we 'force' GC after a failed allocation is an implementation detail.
4. I personally think we probably still want to 'force' the GC after a failed allocation, but do not block the current thread. In this case, `at_safepoint` only changes the behavior of blocking, and does not change the behavior of GC triggering. 

>  One concern is that what will happen if GC is not initialized, and we failed to get pages from the PR, and it is not at safepoint? Should it panic immediately, or should it return null? 

With `at_safepoint=true`, the behavior is panic. We could keep that behavior with `at_safepoint=false`. `at_safepoint` doesn't need to change that behavior.

wks: I updated the documentation so that `at_safepoint` no longer guarantees anything other than not blocking for GC.

I also made `allow_oom_call` another option.  But the current behavior of `allow_oom_call` is quite inconsistent.  It only controls `Space::handle_obvious_oom_request` and `LockFreeImmortalSpace::acquire`, but not `Allocator::alloc_slow_inline` or `util::memory::handle_oom_error` both of which call `Collection::out_of_memory`.  I didn't change the current behavior, and I am leaving it to another pull request.

src/util/alloc/allocator.rs
```diff
@@ -28,39 +28,51 @@ pub enum AllocationError {
...
+    /// If `false`, the allocation will immediately return a null address if the allocation cannot
+    /// be satisfied without a GC.
+    pub at_safepoint: bool,
+
+    /// Whether the allocation is allowed to call [`Collection::out_of_memory`].
+    ///
+    /// **The default is `true`**.
+    ///
+    /// If `true`, the allocation will call [`Collection::out_of_memory`] when out of memory and
+    /// return null.
```
qinsoon: Calling `out_of_memory` and returning null are independent. We could call `out_of_memory`, trigger a GC, then succeed in the allocation and return a valid address. Same for the `false` case.

PR #1401, opened 2025-10-10
Refactor Space::acquire

wks:
We extract portions of that function into separate functions or lambda expressions in order to make the main decision tree logic terse.

We do not intend to change the logic of this function, but we fix some obvious bugs in this commit.

-   It is a runtime error if we need GC but GC is not initialized.  We use `panic!` instead of `assert!`.
-   Now the value passed to `on_pending_allocation` always includes metadata size.

### DIFF
--- a/src/policy/space.rs
+++ b/src/policy/space.rs
@@ -131,7 +131,7 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
         } else {
             debug!("Collection not required");
 
-            if let Some(addr) = self.get_new_pages_and_initialize(pr, pages_reserved, pages, tls) {
+            if let Some(addr) = self.get_new_pages_and_initialize(tls, pages, pr, pages_reserved) {
                 addr
             } else {
                 self.not_acquiring(tls, alloc_options, pr, pages_reserved, true);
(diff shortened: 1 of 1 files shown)


src/policy/space.rs
```diff
@@ -148,139 +160,137 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
...
-                        .policy
-                        .on_pending_allocation(pages_reserved);
-
-                    VM::VMCollection::block_for_gc(VMMutatorThread(tls)); // We asserted that this is mutator.
-                    Address::ZERO
-                }
+            if let Some(addr) = self.get_new_pages_and_initialize(pr, pages_reserved, pages, tls) {
+                addr
+            } else {
+                on_gc_required(true);
```
qinsoon: This is a bit confusing: GC is not required at this point, and `on_gc_required` will actually require GC.

Maybe you could just do something like below:
```
if let Some(addr) = self.get_new_pages_and_initialize(pr, pages_reserved, pages, tls) {
    addr
} else {
    if alloc_options.on_fail.allow_gc() {
        // We thought we had memory to allocate, but somehow failed the allocation. Will force a GC.
        let gc_performed = self.get_gc_trigger().poll(true, Some(self.as_space()));
        debug_assert!(gc_performed, "GC not performed when forced.");
    
        on_gc_required();   
    }    
    Address::ZERO
}
```

src/policy/space.rs
```diff
@@ -112,33 +112,45 @@ pub trait Space<VM: VMBinding>: 'static + SFT + Sync + Downcast {
...
 
         trace!("Reserving pages");
         let pr = self.get_page_resource();
         let pages_reserved = pr.reserve_pages(pages);
         trace!("Pages reserved");
         trace!("Polling ..");
 
-        if should_poll && self.get_gc_trigger().poll(false, Some(self.as_space())) {
+        // Handle the case that GC is required.
+        let on_gc_required = |attempted_allocation_and_failed: bool| {
```
qinsoon: This can be extracted as a new method, like what you did with `get_new_pages_and_initialize`.

Issue #1398, opened 2025-10-07
`is_collection_enabled` is not thread-safe


qinsoon:
https://github.com/mmtk/mmtk-core/pull/1075 removed `disable/enable_collection` and replaced it with `Collection::is_collection_enabled`. However, it seems a binding cannot implement enable/disable collection in a thread-safe way with the latter API.

Consider a scenario like this:
1. GC is enabled by default.
2. Thread 1 tries to allocate and polls for GC. The poll is allowed, thus MMTk schedules work packets for the GC. However, none of the work packets has been executed yet (including `StopMutators` and `Collection::stop_all_mutators`).
3. Thread 2 tries to disable GC. It sets a binding-side flag, and future `Collection::is_collection_enabled` will return this flag value. Thread 2 further invokes a GC safe point to make sure there is no ongoing GC request yet. As `stop_all_mutators` has not been called yet, the binding does not know there is a GC and the safe point simply returns.
4. Thread 2 starts to execute code that is supposed to run only when GC is disabled.
5. `StopMutators` gets executed by MMTk, and it asks the binding to stop mutators.
6. Thread 2 runs to a GC safe point, and gets stopped at a place where GC is not supposed to happen.

Though [the document for `is_collection_enabled`](https://docs.mmtk.io/api/mmtk/vm/collection/trait.Collection.html#method.is_collection_enabled) states that 'any synchronization involving enabling and disabling collections by mutator threads should be implemented by the VM', there seems to be no way for the VM to do this.

The core issue is that there exists a window between when MMTk polls for GC and when it executes stop_all_mutators.
During this window:
* MMTk already considers a GC in progress.
* The binding, however, cannot detect that GC has been triggered.

If the binding disables GC in that interval, the two states diverge -- leading to GC happening when the VM assumes it cannot.

The previous `enable/disable_collection` API did not have this problem. When the binding called these methods, MMTk could internally check whether a GC was ongoing and block the call until the GC completed, ensuring proper synchronization.

PR #1405, opened 2025-10-20
Remove GCRequester, move related code to GCTrigger

qinsoon:
This PR moves GC triggering related code to `GCTrigger`, including:
* The old `GCRequester` which handles the request for GCs.
* The check for `is_collection_enabled` in `memory_manager` and `Space::acquire`.
* Code for user collection request and internal collection request in `MMTK`.

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -543,7 +543,7 @@ pub fn gc_poll<VM: VMBinding>(mmtk: &MMTK<VM>, tls: VMMutatorThread) {
         "gc_poll() can only be called by a mutator thread."
     );
 
-    if VM::VMCollection::is_collection_enabled() && mmtk.gc_trigger.poll(false, None) {
+    if mmtk.gc_trigger.poll(false, None) {
         debug!("Collection required");
         if !mmtk.state.is_initialized() {
             panic!("GC is not allowed here: collection is not initialized (did you call initialize_collection()?).");


qinsoon:
This PR was originally created to support a fix for https://github.com/mmtk/mmtk-core/issues/1398. By consolidating all GC-triggering logic into `GCTrigger`, it becomes easier (or possible) to apply locks or atomics to address the data race when enabling collection.

I’ve also prototyped a thread-safe “enable collection” API, but it turned out to be difficult to use from Julia due to the issues described in https://github.com/mmtk/mmtk-julia/issues/278. Since Julia is the main reason we need these APIs, it doesn’t make much sense to introduce them if they don’t integrate well there. As a result, a proper fix for https://github.com/mmtk/mmtk-core/issues/1398 may be postponed.

Given that said, I think this PR on its own is still a good cleanup, even if there’s no immediate follow-up to fix https://github.com/mmtk/mmtk-core/issues/1398.

src/util/heap/gc_trigger.rs
```diff
@@ -72,6 +77,32 @@ impl<VM: VMBinding> GCTrigger<VM> {
...
+
+    /// Clear the "GC requested" flag so that mutators can trigger the next GC.
+    /// Called by a GC worker when all mutators have come to a stop.
+    pub fn clear_request(&self) {
+        self.request_flag.store(false, Ordering::Relaxed);
+    }
+
+    fn get_plan(&self) -> &dyn Plan<VM = VM> {
+        unsafe { self.plan.assume_init() }
+    }
```
wks: This method is identical to `GCTrigger::plan`.  Keep one of them and move the other.

wks (changes requested):
Just one minor problem.  Other parts look good.

src/util/heap/gc_trigger.rs
qinsoon: Done. I didn't realize we had that function.

wks (dismissed):
~~LGTM~~ See comment below

wks:
I grepped the code and found there are still comments mentioning the `GCRequester` type.  We should update those comments, too.

And the eBPF tools `capture.bt` and `visualize.py` also contains the USDT name `gcrequester_request`.  It can be renamed (e.g. simply change it to `gc_requested`)

Issue #1223, opened 2024-11-01
Introduce NonZeroAddress and write more idiomatic Rust code


wks:
TL;DR: Stop using `Address::ZERO` for erroneous cases.  Use more idiomatic Rust style, such as `Option<NonZeroAddress>` or `Result<NonZeroAddress, ErrorType>`.

Unlike `ObjectReference` where 0 and misaligned addresses can't refer to any valid object, there is no such constraints for `Address`.  The zero address is actually meaningful sometimes.  But in practice, we often use `Address::ZERO` for representing special cases where a meaningful address doesn't exist.  It is most often used in the code path of allocation, where a failed allocation returns `Address::ZERO`.  It doesn't mean the object is allocated at address 0x00000000, but it is not allocated anywhere.

While this works, it is not idiomatic in Rust.  https://github.com/mmtk/mmtk-core/issues/1043 described several problems with nullable `ObjectReference`, and many of them apply to `Address`, too.  Notable ones include:

-   Since we don't know if an `Address` can be zero or not, we sometimes do redundant checks, and sometimes miss necessary checks.
-   Some functions guarantee a returned `Address` is never zero, but the caller can't take advantage of it.  Notably, we can turn a `NonZeroAddress` safely to an `ObjectReference` and turn an `Address` safely to an `Option<ObjectReference>`, but converting `Address` to `ObjectReference` requires either an `Option<ObjectReference>::unwrap()` or an unsafe operation, even if we statically know the `Address` cannot be zero.

Similar to `ObjectReference`, we can define `NonZeroAddress`, backed by `NonZeroUsize`.  Using `#[repr(transparent)]`, we can make `Option<NonZeroAddress>` have the same layout as `Address`.

Using `Option<NonZeroAddress>` will also allow us to use Rust's `?` operator which returns immediately if it is `None`.

```rust
fn bar() -> Option<NonZeroAddress> {...}

fn foo() -> Option<NonZeroAddress> {
    let x: NonZeroAddress = bar()?;  // Returns None if it is None.
    do_something_with(x);
    x
}
```

And one typical use case is allocation

```rust
fn acquire_block(&self, size: usize) -> Option<NonZeroAddress> {
    let acquired_start: NonZeroAddress = space.acquire(bytes_to_pages_up(size))?; // Return None if allocation failed.

    self.set_limit(acquired_start, acquired_start + block_size);
    self.alloc(size) // This may still return None if allocation failed for other reasons.
}
```

[...]

PR #1408, opened 2025-10-24
Allow MockVM to run GC

qinsoon:
This PR made some changes to `MockVM` to allow us to run GC in the tests.
* Add features `mock_test_side_metadata` and `mock_test_header_metadata`. We only test with `mock_test_side_metadata` (as some plans don't run with header metadata).
* No longer use lock for the `MockVM` instance (otherwise it will cause deadlock when we reentry `MockVM`)
* Statically initialize a `MMTK` instance for mock tests.
* Mock tests now use `mock_api` to create a mutator. This registers the mutator to mock VM, and can be blocked for GC.
* Mock tests now properly initialize GC threads, and handle panics in GC threads.
* Minor changes to print more information for mmap failures.


wks:
`std::sync::ReentrantLock` is an unstable feature.  But a third-party crate [parking_lot](https://docs.rs/parking_lot/latest/parking_lot/) has an implementation of [`ReentrantMutex`](https://docs.rs/parking_lot/latest/parking_lot/type.ReentrantMutex.html).  If that is helpful, we can add the `parking_lot` as a dependency when the `mock_test` feature is enabled.

src/util/heap/monotonepageresource.rs
```diff
@@ -261,15 +261,19 @@ impl<VM: VMBinding> MonotonePageResource<VM> {
      }
      }*/
 
-    pub fn reset_cursor(&self, top: Address) {
+    // top might be Address::ZERO
```
wks, line 264: This is where I think `Option<NonZeroAddress>` can be useful.  See https://github.com/mmtk/mmtk-core/issues/1223

src/util/linear_scan.rs
```diff
@@ -51,6 +51,10 @@ impl<VM: VMBinding, S: LinearScanObjectSize, const ATOMIC_LOAD_VO_BIT: bool> std
 
     fn next(&mut self) -> Option<<Self as Iterator>::Item> {
         while self.cursor < self.end {
+            if !self.cursor.is_aligned_to(ObjectReference::ALIGNMENT) {
+                self.cursor += VM::MIN_ALIGNMENT;
```
wks: Currently `ObjectReference` is required to be aligned to `ObjectReference::ALIGNMENT`.  So even if `VM::MIN_ALIGNMENT` is smaller than `ObjectReference::ALIGNMENT`, VO bits can only be set at addresses which are multiples of `ObjectReference::ALIGNMENT`.  I think the algorithm here is wrong (or just outdated).  It is pointless to increment the cursor by `VM::MIN_ALIGNMENT`.  It should simply align the cursor up to the next multiple of `ObjectReference::ALIGNMENT`.  And I wonder whether `find_last_non_zero_bit_in_metadata_bytes` could be more efficient (needs to search forward instead of backward).

qinsoon: This is resolved in https://github.com/mmtk/mmtk-core/pull/1408/commits/040cfc9593f37a25dad60934153f3cb4d27687a9.

qinsoon:
I turned this PR back to a draft for now. I will wait for https://github.com/mmtk/mmtk-core/pull/1601 (aarch64 linux) and https://github.com/mmtk/mmtk-core/pull/1599 (aarch64 macOS) to merge first, then update this PR to test on those platforms.

PR #1409, opened 2025-10-27
Update doc comment for Slot

wks:
We make it clear that a `Slot` points to a slot instead of being the slot itself.  We also emphasize that `Slot` can be copied, and the copied `Slot` points to the same slot.

We also reorganize the doc comment and explain the relation between slots in general and the `Slot` trait in MMTk.

We also use `SimpleSlot` and `OffsetSlot` as examples of what an implementation of `Slot` can contain.

### DIFF
--- a/.github/scripts/ci-common.sh
+++ b/.github/scripts/ci-common.sh
@@ -11,6 +11,7 @@ dummyvm_toml=$project_root/docs/dummyvm/Cargo.toml
 
 # Pin certain deps for our MSRV
 cargo update -p home@0.5.11 --precise 0.5.5 # This can be removed once we move to Rust 1.81 or newer
+cargo update -p home@0.5.12 --precise 0.5.5 # This requires Rust edition 2024
 
 # Repeat a command for all the features. Requires the command as one argument (with double quotes)
 for_all_features() {


src/vm/slot.rs
```diff
@@ -11,44 +11,60 @@ use atomic::Atomic;
 use crate::util::constants::{BYTES_IN_ADDRESS, LOG_BYTES_IN_ADDRESS};
 use crate::util::{Address, ObjectReference};
 
-/// A `Slot` represents a slot in an object (a.k.a. a field), on the stack (i.e. a local variable)
-/// or any other places (such as global variables).  A slot may hold an object reference. We can
-/// load the object reference from it, and we can update the object reference in it after the GC
-/// moves the object.
+/// A `Slot` value *points to* a slot, and is intended for loading and updating the reference held
```
qinsoon: > A `Slot` value *points to* a slot
> Note that a `Slot` is not the slot itself.

When people are reading this comment, they probably don't know what a MMTK's `Slot` is, and they may or may not be sure about what a slot is for their runtime. You are trying to differentiate the subtle difference between a `Slot` and a slot -- I feel this makes it more confusing. I would go with something like this, and then use more examples to explain it:
```
`Slot` is an abstraction for MMTk to load and update object references in memory.
```

Besides, a `Slot` could be the slot itself (if we don't require `Slot` to be `Copy`).
```rust
struct MyObjectLayout {
  header: MyHeader,
  field1: Arc<MySlot>,
}

impl mmtk::vm::Slot for Arc<MySlot> {
  ...
}
```

src/vm/slot.rs
```diff
@@ -11,44 +11,60 @@ use atomic::Atomic;
...
+/// slot may sometimes not hold an object reference.  For example, in JVM, a slot can hold a special
+/// `null` pointer which does not point to any object; in some VMs, including CRuby and some
+/// JavaScript engines, a slot can hold a tagged non-reference value, such as small integers and
+/// special values such as `true`, `false`, `null` (a.k.a. "none", "nil", etc.), `undefined`, etc.
 ///
-/// This intends to abstract out the differences of reference field representation among different
-/// VMs.  If the VM represent a reference field as a word that holds the pointer to the object, it
-/// can use the default `SimpleSlot` we provide.  In some cases, the VM need to implement its own
-/// `Slot` instances.
+/// The `Slot` trait is intended to abstract out the differences of reference field representation
```
qinsoon: > The `Slot` trait is intended to abstract out the differences of reference field representation
> (compressed, tagged, offsetted, etc.) among different VMs.

I feel this is also a good definition for `Slot`. It is just an abstraction. It does not really define what a `Slot` would be -- whether it is a value that points to the slot, or is the slot itself. It doesn't really matter, as long as it allows MMTk to access the object reference in the memory.

src/vm/slot.rs
InnocentZero: I'd prefer to add something like this as well, apart from the line about `Slot` being an abstraction, "In case of most runtimes, a `Slot` is a pointer to a field in the object."

InnocentZero: Also, I'd prefer if we use 'fields' instead of 'slots' everywhere, there is a good scope of confusion between `Slot` and 'slot'.

k-sareen: The point is that a slot encompasses more than a field inside an object. A slot can be a stack slot. Or a global root. These are not fields.

wks: > When people are reading this comment, they probably don't know what a MMTK's `Slot` is, and they may or may not be sure about what a slot is for their runtime. You are trying to differentiate the subtle difference between a `Slot` and a slot -- I feel this makes it more confusing. I would go with something like this, and then use more examples to explain it:
> 
> ```
> `Slot` is an abstraction for MMTk to load and update object references in memory.
> ```

Yes.  I think this sentence is clear and conveys the main intention of the `Slot` trait.  I edited the comment and used this as the first sentence of the doc comment of `Slot`.

But I think the difference between slots in general and MMTk's `Slot` trait is important.  I reorganized the doc and used one section to compare the two.

> Besides, a `Slot` could be the slot itself (if we don't require `Slot` to be `Copy`).
> 
> ```rust
> struct MyObjectLayout {
>   header: MyHeader,
>   field1: Arc<MySlot>,
> }
> 
> impl mmtk::vm::Slot for Arc<MySlot> {
>   ...
> }
> ```

Although this is possible with some tricks, it is not the intended way to use `Slot`.  MMTk will internally make use of the fact that `Slot` is trivially copy-able.  I added a section to emphasized that `Slot` should have pointer semantics and implement the `Copy` trait.

qinsoon: Do we actually have to copy `Slots`? Or is it just convenient if `Slot` is `Copy`?

wks: > I'd prefer to add something like this as well, apart from the line about `Slot` being an abstraction, "In case of most runtimes, a `Slot` is a pointer to a field in the object."

I inlined the code of `SimpleSlot` as an example of `Slot` implementation, and I also added `OffsetSlot` as another example to show a case where a `Slot` implementation can have more than a pointer.

> Also, I'd prefer if we use 'fields' instead of 'slots' everywhere, there is a good scope of confusion between `Slot` and 'slot'.

The *[GC Handbook]* uses the term *field* and *slot* interchangeably as you suggested, which is unfortunate.  The problem is, as @k-sareen mentioned, that in MMTk and real-world VMs (including JikesRVM), a slot can also refer to stack slots (local variables) and global roots (including global variables, slots in JNI handle tables, and any other global tables).  We use the term "field" specifically for heap objects, and "slot" more generally.

[GC Handbook]: https://gchandbook.org/

src/vm/slot.rs
wks: I used this in the "# Slots and the `Slot` trait" section to describe what the `Slot` trait is.

src/vm/slot.rs
wks: > Do we actually have to copy `Slots`? Or is it just convenient if `Slot` is `Copy`?

It is intended to have pointer semantics, and it implies it should be copy-able.  Of course if we talk about "have to", we can always use `Clone` instead and cope with `.clone()`.  But that's very inconvenient.

k-sareen: I think the pointer semantics are important because otherwise we're sort of lying to ourselves/making life inconvenient. Consider a slot that we get from a write barrier modbuf vs a slot that we get from scanning an object. They both might be the same. But if we prevent `Copy`, we sort of imply that they are not the same. At least in my head, it makes much more sense that `Slot`s are `Copy`.

src/vm/slot.rs
```diff
@@ -11,44 +11,91 @@ use atomic::Atomic;
...
+/// the slot.
+///
+/// A `Slot` can be [copied](std::marker::Copy), and the copied `Slot` instance points to the same
+/// slot.
+///
+/// # How to implement `Slot`?
+///
+/// If a reference field of a VM is just a word that holds the pointer to an object, and uses the 0
+/// word as the null pointer, it can use the default [`SimpleSlot`] we provide.  It simply contains
+/// a pointer to a memory location that holds an address.
```
k-sareen: Importantly, `SimpleSlot` loads `usize`-sized values from the `Address`. If you have something like ART wherein you still have just a simple `Address` there, but have to load `u32`-sized values from the Address as ART object references are 32-bit. You already mention compressed pointers below, but maybe just explicitly say `usize` or word-size here.

src/vm/slot.rs
```diff
@@ -11,44 +11,91 @@ use atomic::Atomic;
...
+/// information. The `OffsetSlot` example below also contains an offset which can be used when
+/// decoding the pointer. See `src/vm/tests/mock_tests/mock_test_slots.rs` for more concrete
+/// examples.
 ///
-/// When loading, `Slot::load` shall decode its internal representation to a "regular"
-/// `ObjectReference`.  The implementation can do this with any appropriate operations, usually
-/// shifting and masking bits or subtracting offset from the address.  By doing this conversion,
-/// MMTk can implement GC algorithms in a VM-neutral way, knowing only `ObjectReference`.
+/// ```rust
+/// pub struct OffsetSlot {
```
k-sareen, line 72: Note that a `Slot` like this will not be word-sized so may not have the best performance. It should be flagged.

k-sareen (approved):
LGTM. Thanks! Just some minor comments

src/vm/slot.rs
wks: That's right.  `SimpleSlot` is supposed to be simple and does not support compressed oops.  There is a `CompressedOopSlot` in `mock_test_slots.rs` which the user can use as an example.

src/vm/slot.rs
wks: Yes.  I think this is why the OpenJDK binding uses tagged word instead of Rust `enum` to distinguish between narrow and wide slots.

src/vm/slot.rs
k-sareen: Yes. I just meant make it explicit that `SimpleSlot` operates on word-sized pointers.

src/vm/slot.rs
qinsoon: `Copy` only means it is easy to obtain multiple copies of the `Slots`. I don't see why we need to copy slots, except that it might be just a coincidence or a convenience. It sounds possible to me that when we are given a `Slot`, we just keep use that instance without copying or cloning it. In that case, we should not extend the doc to say that it has to be `Copy` (because it does not have to be `Copy`).

wks: > `Copy` only means it is easy to obtain multiple copies of the `Slots`. I don't see why we need to copy slots, except that it might be just a coincidence or a convenience. It sounds possible to me that when we are given a `Slot`, we just keep use that instance without copying or cloning it. In that case, we should not extend the doc to say that it has to be `Copy` (because it does not have to be `Copy`).

If it is not `Copy` or `Clone`, it will follow Rust's ownership rules.  Types that use ownership rules usually have resource management concerns.  For example, `Box<T>` uniquely owns the underlying malloc memory, and cloning a `Box<T>` will make a deep-copy.  `Arc` maintains reference count.  It is `Clone` but not `Copy` because cloning an `Arc` has non-trivial cost (atomically incrementing ref count), and dropping an `Arc` may free the object.  A `File` owns a file.  It cannot be cloned, and `File::try_clone` has semantics like the `dup` system call.

But a `Slot` is not like these.  It simply points to a slot, and it doesn't have other "resources".  It doesn't "own" the slot, and it doesn't release any handles when we drop a `Slot`.  And it doesn't grant MMTk unique access to the slot, either.  `Scanning::scan_objects` will generate `Slot` instances, and repeated calling of `Scanning::scan_objects` can generate multiple `Slot` instances that point to the same slot.  And it should be trivial to copy the data of a `Slot` instances (which is a pointer most of the time), for performance reasons.  Implementing `Copy` will emphasize that it should be trivial to copy.

Maybe in some distributed GC systems, we can have `Slot` that refer to a slot in a remote instance, and closing a `Slot` will release the network connection.  But MMTk is not designed for that.  If a `Slot` is a heavy-weight handle, maybe MMTk will need to be redesigned.

k-sareen: Right exactly. My intuition is that a `Slot` has pointer semantics, which is unsafe or potentially dangerous, sure (i.e. it's not enforcing any mutability or aliasing requirements), but we have to be disciplined with it.

qinsoon: Just for the sake of argument, I just don't get why you keep saying `Slot` has pointer semantics. 

To me, `Slot` is just an abstraction for MMTk to access object reference from memory. MMTk really doesn't care how `Slot` is implemented in the binding side. It could be pointers, pointers plus offsets, the slot itself, or just an ID as long as the binding know how to access the slot with the ID. What MMTk cares is to access the memory, as for 'how' to access the memory, that's bindings' issues.

As for `Slot` requiring the `Copy` trait, to my understanding, it is just a convenience and a coincidence -- it was just implemented like this when it was first introduced. I don't see why we need to copy `Slots` around: when the binding gives us `Slots`, we can pass it around, or take references of it and use the reference instead. I don't see a reason that it has to be `Copy`.

Also we do implement `Slot` usually as a pointer. But again that's bindings' choice. Why does MMTk core care about this?

wks: > Just for the sake of argument, I just don't get why you keep saying `Slot` has pointer semantics.

Here are some historical details.  In the original MMTk in JikesRVM, and earlier versions of Rust MMTk, we used `Address` to point to slots directly.  Then entire MMTk was designed with the assumption that `Address` is a value type and copy-able.  Rust doesn't have the concept of "value type" because everything is a value.  When we introduced the `Slot` trait (was called `Edge` trait before), we mimicked the properties of `Address` to the extent that `Address` itself can be an implementation of `Slot` (and it currently still is, for backward compatibility reasons).

> To me, `Slot` is just an abstraction for MMTk to access object reference from memory. MMTk really doesn't care how `Slot` is implemented in the binding side. It could be pointers, pointers plus offsets, the slot itself, or just an ID as long as the binding know how to access the slot with the ID. What MMTk cares is to access the memory, as for 'how' to access the memory, that's bindings' issues.

Yes.  `Slot` can be implemented in many ways, but it still has pointer semantics.  We can copy the "pointer plus offset" and it still points to the same slot and knows the same offset.  We can also copy the handle and it still refers to the same indirection table entry.  I originally conceptualized implementing `Slot` as a struct that contains the slot value itself (for use in CRuby).  We can still load from such a `Slot` (by simply returning the value), but we can't update it.  We can still copy that because any VM slot that uses such a `Slot` implementation cannot be updated, so it doesn't matter whether we copy its pointer or its value.  They are both conceptually "pointers to the slot" in the sense that we can "load" its value.

> As for `Slot` requiring the `Copy` trait, to my understanding, it is just a convenience and a coincidence -- it was just implemented like this when it was first introduced. I don't see why we need to copy `Slots` around: when the binding gives us `Slots`, we can pass it around, or take references of it and use the reference instead. I don't see a reason that it has to be `Copy`.

If you mean it has its heritage from the `Address` type MMTk was originally using, we may call it a coincidence.  And it's really convenient and I think it is good to be convenient.

I can give you an example why we may copy slots.  For sanity GC, we may "cache the roots" so that we may scan them later.  This will copy the `Slot` instances into the SanityChecker, while the main work packets for computing the transitive closure use their original copies.  

> Also we do implement `Slot` usually as a pointer. But again that's bindings' choice. Why does MMTk core care about this?

"Pointer semantics" doesn't mean it has to be (or contain) a pointer.  It's just an interface.  It's about how MMTk can use the thing.

"Pointer semantics" is the opposite of "`Slot` being the slot itself".  If `Slot` has that semantics, MMTk will use `Slot` very differently, such as passing `&mut Slot` or `*mut Slot` around within MMTk core for enqueuing.  This will require much changes from a system that originally used `Address` for the same purpose.  It's not impossible, but it's just not how MMTk has been working.

qinsoon: > I can give you an example why we may copy slots. For sanity GC, we may "cache the roots" so that we may scan them later.

This is a good reason why we want to copy slots.

k-sareen: > Also we do implement Slot usually as a pointer. But again that's bindings' choice. Why does MMTk core care about this?

This matters because MMTk internally has to interface with the `Slot` type. A lot. For example for write barrier modbufs or for the initial root set, etc. So any changes to `Slot` does affect MMTk core. We should be careful with it.

PR #1410, opened 2025-10-29
Allow internal pointers to point to the end of an object

qinsoon:


wks:
If there are two object A and B allocated adjacent to each other, A is before B, and A has a `char buffer[0]` as the last field.  If `&A.buffer` happens to equal the address of `&B`, would `&A.buffer` be considered an internal pointer of A, or an internal pointer of B?  This may affect `find_object_from_internal_pointer` which is currently the only use case of `is_internal_ptr`.

And if this can happen in a VM, if the stack contains a pointer to `&B`, would it also keep the object A alive because it could potentially be an interior pointer to `&A.buffer` as well?  Which object would `find_object_from_internal_pointer(&B, LIMIT)` return?

qinsoon:
> &A.buffer == &B

If B has a header, and `&B` (the object reference) is pointing to the payload (not the header), then you wouldn't have the situation of `&A.buffer == &B`.

If there is no header, or the object reference points to the start of the object, you would have `&A.buffer == &B`. But in this case, not only MMTk but also the runtime cannot tell if the pointer is for `A` or for `B`. To me, that sounds like a design issue for the runtime.

> But maybe we should edit the doc comment of find_object_from_internal_pointer to warn the user about this corner case.

Will do that. I just encountered a case in Julia: for the value `(true, ())`, it is 9 bytes (including a 8 bytes header), and the runtime uses an address like `0x100009` for the second field. I am still evaluating if this change is correct for Julia.

Issue #517, opened 2022-01-11
Proper implementation of the treadmill algorithm


caizixian:
The current `TreadMill` is a misnomer, as it uses `HashSet`s rather than linked lists.
It was used during our prototyping phase as a "drop-in" replacement of a real treadmill.

So,
1. We should implement the actual treadmill algorithm with linked lists
2. Evaluate the best approach as to scalability of large object allocation. @wenyuzhao has found that our page allocator doesn't scale well with the increasing parallelism on modern CPUs. For certain workload, the scalability of large object allocation might matter, and a simple lock protecting the whole treadmill might not be the best solution

cc @steveblackburn

qinsoon:
> 1. We should implement the actual treadmill algorithm with linked lists

What do you mean by 'the actual treadmill algorithm'? Are you referring the treadmill in Java MMTk/JikesRVM, or the Baker's treadmill in his paper?

To my understanding, treadmill in MMTk core approximately equals treadmill in JikesRVM, and they both are different from the Baker's treadmill. They both use multiple lists/sets to mimic a 'logical' treadmill rather than using an actual cyclic doubly linked list (Baker's). They both allow objects of different sizes in the same treadmill, and allocation is done by actually allocating memory rather than reusing free nodes in the treadmill (Baker's). The differences between MMTk core and JikesRVM are: 1. JikesRVM uses doubly linked list, and mmtk-core uses hashset, and 2. JikesRVM stores links along with payload, and mmtk-core only puts object references in the set (payload is separated from the metadata).

> 2. Evaluate the best approach as to scalability of large object allocation. @wenyuzhao has found that our page allocator doesn't scale well with the increasing parallelism on modern CPUs. For certain workload, the scalability of large object allocation might matter, and a simple lock protecting the whole treadmill might not be the best solution

The page allocator is in the page resource implementation, rather than the treadmill. If we have scalability issue with page allocator, fixing treadmill won't directly help.

PR #1412, opened 2025-11-03
Fix LOS object enumeration

wks:
We fix a bug when enumerating objects in LOS.  During Prepare of ConcurrentImmix, we flip the from and to space of the TreadMill, and then enumerate objects in the LOS to set their unmark bits.  However, the object enumeration code only enumerates objects in the to-space. Because we have flipped all objects to the from-space, object enumeration visits no objects at all.  Consequently, no objects in LOS have unlog bits during concurrent marking, and the SATB barriers will not be applied to any objects in the LOS.

Now `TreadMill::enumerate_objects` take an additional argument `all: bool`. When it is set to true, we enumerate all objects in the `Treadmill` (which are all objects in the LOS).  We do this when setting and clearing unlog bits, and fix the bug.

We also refactored the `TreadMill` so that all of its `HashSet` fields are protected by one single `Mutex`, because it is not safe to enumerate one `HashSet` while another thread modifies another.

Fixes: https://github.com/mmtk/mmtk-openjdk/issues/334

### DIFF
--- a/src/util/treadmill.rs
+++ b/src/util/treadmill.rs
@@ -159,7 +159,12 @@ impl TreadMill {
             visit_objects(&sync.from_space);
         }
 
-        debug!("Enumerated {enumerated} objects in LOS.  all: {all}");
+        debug!("Enumerated {enumerated} objects in LOS.  all: {all}.  from_space: {fs}, to_space: {ts}, collect_nursery: {cn}, alloc_nursery: {an}",
+            fs=sync.from_space.len(),
+            ts=sync.to_space.len(),
+            cn=sync.collect_nursery.len(),
+            an=sync.alloc_nursery.len(),
+        );
     }
 }
 


src/util/treadmill.rs
```diff
@@ -6,111 +6,167 @@ use crate::util::ObjectReference;
...
+/// The synchronized part of [`TreadMill`]
+#[derive(Default)]
+struct TreadMillSync {
+    nursery: FromToSpace,
+    mature: FromToSpace,
+}
+
+/// A pair of from and two spaces.
+#[derive(Default)]
+struct FromToSpace {
```
k-sareen: Can we rename this to `TreadMillSpacePair` or something? The `FromToSpace` name is a bit confusing.

wks: > Can we rename this to `TreadMillSpacePair` or something? The `FromToSpace` name is a bit confusing.

I changed it to just `SpacePair` because it is private in the `treadmill` module.

src/util/treadmill.rs
```diff
@@ -6,111 +6,192 @@ use crate::util::ObjectReference;
...
-    to_space: Mutex<HashSet<ObjectReference>>,
-    collect_nursery: Mutex<HashSet<ObjectReference>>,
-    alloc_nursery: Mutex<HashSet<ObjectReference>>,
+    sync: Mutex<TreadMillSync>,
+}
+
+/// The synchronized part of [`TreadMill`]
+#[derive(Default)]
+struct TreadMillSync {
+    nursery: SpacePair,
```
qinsoon: We normally think `nursery` is the from space and `mature` is the to space. This code now uses two 'logical spaces' (from space and to space) for both nursery and mature -- that is 4 logical spaces. I don't know how that makes sense.

Besides, `collect_nursery` is supposed to hold objects that are allocated during a GC (e.g. an over-sized object is copied to LOS from GenImmix nursery). However, we don't do that any more. Large objects are allocated directly into LOS, and there is no copy from other spaces to LOS. I think you can assert if that is true. If it is true, then the code becomes `nursery + from + to` which is similar to `GenCopy` -- that makes much more sense.

src/util/treadmill.rs
```diff
@@ -6,111 +6,192 @@ use crate::util::ObjectReference;
...
+/// It is divided into the nursery and the mature space, and each of them is further divided into
+/// the from-space and the to-space.
+///
+/// All operations are protected by a single mutex [`TreadMill::sync`].
 pub struct TreadMill {
-    from_space: Mutex<HashSet<ObjectReference>>,
-    to_space: Mutex<HashSet<ObjectReference>>,
-    collect_nursery: Mutex<HashSet<ObjectReference>>,
-    alloc_nursery: Mutex<HashSet<ObjectReference>>,
+    sync: Mutex<TreadMillSync>,
```
qinsoon, line 13: This `Mutex` can be pushed to the user of `TreadMill`. They can define `Mutex<Treadmill>`.

src/util/treadmill.rs
```diff
@@ -6,111 +6,192 @@ use crate::util::ObjectReference;
...
+/// A pair of from and two spaces.
+#[derive(Default)]
+struct SpacePair {
+    from_space: HashSet<ObjectReference>,
+    to_space: HashSet<ObjectReference>,
+}
+
+/// Used by [`TreadMill::enumerate_objects`] to determine what to do to objects in the from-spaces.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub(crate) enum FromSpacePolicy {
```
qinsoon: I don't think we need a type so the caller can describe what they need to `TreadMill`. They can just do
```rust
self.treadmill.enumerate_objects(...);
self.treadmill.enumerate_from_space_objects(...);
assert!(self.treadmill.is_from_space_empty());
```

qinsoon (changes requested):

src/util/treadmill.rs
qinsoon: It turns out that the original 'treadmill' collector does have 4 spaces: https://cofault.com/treadmill.html. We can move to something closer to that. It is still fine to use 4 sets, instead of 1 doubly linked list.

qinsoon: BTW I don't think we have to do any major refactoring to `TreadMill` to support object enumeration. But if you do plan for any major refactoring (such as `[nursery, mature] x [from, to]`), I don't think it is a good idea to move further away from the original 'treadmill' concept.

wks: I think there are two dimensions.

1. Whether the object is newly allocated after the previous GC (young) or before the previous GC (old).
2. Whether the object is determined to be live (to) or has unknown liveness (from).

But what you said also makes sense because during GC, we are always going to evacuate the nursery, so it is always a from-space during GC.  If we don't allocate any large objects during GC and label it as young (which makes no sense because even over-sized objects moved from ImmixSpace are old by definition after this GC) , the "nursery to-space" will always be empty during GC, too.  So it makes sense to only keep three space, i.e. nursery + from + to.

src/util/treadmill.rs
wks: I'd like to hide the synchronization mechanism inside TreadMill.  We can reimplement TreadMill to make it lock-free, like JikesRVM's implementation based on linked-lists.  If we do that, the interface of TreadMill will not need to be changed.  See https://github.com/mmtk/mmtk-core/issues/517

src/util/treadmill.rs
wks: The problem is that enumerating objects should be mutually exclusive with other operations.  Currently we do it by holding the mutex in one function and visiting multiple hash sets.  We may do what you suggested if we move the mutex one level up, but I don't want to do that because we may change the synchronization mechanism in the future.  But if you think it is bad to expose a new type to the user of treadmill, I can change the interface of TreadMill to expose multiple functions, for enumerating all objects, live objects, and for mutator, respectively.

src/util/treadmill.rs
wks: > BTW I don't think we have to do any major refactoring to `TreadMill` to support object enumeration. But if you do plan for any major refactoring (such as `[nursery, mature] x [from, to]`), I don't think it is a good idea to move further away from the original 'treadmill' concept.

I don't plan to do a major refactoring for now.  I just needed to make the mutex more coarse-grained which requires some structural change.  I ended up using just `nursery + from + to` which should be the simplest we can have to support our current LOS usage.

src/util/treadmill.rs
wks: It turned out that two boolean parameters are sufficient to distinguish all cases.  I removed the `FromSpacePolicy` type and replaced it with `all: bool` and `assert_empty: bool`.

src/util/treadmill.rs
wks: Oh wait.  There is one catch.  If we enumerate objects during mutator time, we should enumerate both `nursery` and `to_space` because the nursery can contain newly allocated objects which should be returned, too.  But the "nursery from-space" should be empty.  Because it is only an assertion and it has undefined behavior if the mutator attempts to enumerate objects during GC, I'll reduce the precision of the assertion for now.

wks: I kept the `nursery + from + to` structure, but reduced the amount of assertion.

src/util/treadmill.rs
wks: I changed the parameters to `nursery: bool` and `from: bool` so that

- Compressor enumerates `to_space`, only.
- mutators enumerate `to_space` + `nursery`.
- ConcurrentImmix enumereate `to_space` + `nursery` + `from_space`.

src/util/treadmill.rs
wks: > It turns out that the original 'treadmill' collector does have 4 spaces: https://cofault.com/treadmill.html. We can move to something closer to that. It is still fine to use 4 sets, instead of 1 doubly linked list.

After reading that paper, I found that the `Treadmill` in JikesRVM MMTk (and therefore the Rust MMTk) is completely unrelated to the Treadmill algorithm described in that paper.

The four colors in Baker's Treadmill algorithm contains four pointers separating four cell colors, namely ecru, grey, black and white.  White cells are unallocated; ecru objects are allocated but not enqueued; grey objects are enqueued but not marked; and black objects are marked.  This organization makes it easy to free garbage objects in bulk by moving a few pointers so that ecru (un-enqueued objects) objects become white (free) and black (marked) objects become ecru (un-enqueued).  The Treadmill algorithm is not generational.

But the `Treadmill` data structure in JikesRVM MMTk keeps four lists of objects: allocNursery, collectNursery, toSpace and fromSpace.  They match the `[nursery, mature] X [from, to]` semantics I mentioned above, and they are used as such.

I dug a bit into the history of JikesRVM.

- The `Treadmill` was [initially](https://github.com/JikesRVM/JikesRVM/blob/4b4eca08/rvm/src/vm/memoryManagers/JMTk/utility/DoublyLinkedList.java) just `DoublyLinkedList`, where each node contains three pointers: the forward link, the backward link, and a pointer to the "treadmill" which is the `DoublyLinkedList` instance itself.
- The `DoublyLinkedList` type is immediately renamed to `Treadmill` in [the next commit](https://github.com/JikesRVM/JikesRVM/blob/2ff10c05/MMTk/src/org/mmtk/utility/Treadmill.java), and it then contained two separate doubly linked lists, one for the from-space, and the other for the to-space.  The third word of a node now points to the `Treadmill` instance instead of the `DoublyLinkedList` instance it is in. 
- [Several years later](https://github.com/JikesRVM/JikesRVM/blob/446f8cdb/MMTk/src/org/mmtk/utility/Treadmill.java), a third linked list "nursery" was added.
- [Half a year later](https://github.com/JikesRVM/JikesRVM/blob/aa79192c62d5d25dbb2d81f55020512a28dd6a97/MMTk/src/org/mmtk/utility/DoublyLinkedList.java), the third word (the pointer to the "treadmill") is removed.
- And [several months later](https://github.com/JikesRVM/JikesRVM/blob/75d5a71a/MMTk/src/org/mmtk/utility/Treadmill.java), the nursery is split into the "collection nursery" and the "allocation nursery" for the purpose of "in preparation for concurrent/incremental collection".

From this history, we see that the `Treadmill` in JikesRVM MMTk was never related to Baker's Treadmill algorithm.  It is named as "treadmill" because it maintains cyclic linked lists which resemble the tread of a treadmill in the real world.

And the separation of "allocation nursery" and "collection nursery" had a purpose, i.e. "concurrent/incremental collection".  We haven't implemented concurrent or incremental plans before, so we see a single `nursery` is sufficient.  But since we have just started to implement ConcurrentImmix, we may eventually find cases where we may need two nurseries.

Given this, I'll revert my refactoring on the naming and structure of the `Treadmill` type (except for making the mutex coarser) and keep the names we inherited from the JikesRVM MMTk.  We may do a refactoring on Treadmill in the future, maybe renaming it so that we don't confuse it with Baker's Treadmill algorithm.

k-sareen: Yes I remember this discussion on how the treadmill we've implemented is very different from the actual one: https://github.com/mmtk/mmtk-core/issues/517

src/policy/largeobjectspace.rs
```diff
@@ -364,12 +388,20 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
...
         } else {
-            for object in self.treadmill.collect() {
+            for object in self.treadmill.collect_mature() {
                 sweep(object)
             }
         }
     }
 
+    /// Enumerate objects for forwarding references.  Used by certain mark-compact plans.
+    pub(crate) fn enumerate_objects_for_forwarding(&self, enumerator: &mut dyn ObjectEnumerator) {
```
qinsoon: I suggest avoid using the term 'forwarding' here. LOS does not know what forwarding means -- it depends on the plan, like when the function will be called, and what object it should list, etc. 

It should be the plan that instructs LOS to iterate objects in to-space and to assert there is no nursery object (it is a hack that currently the compressor space does that instead of the plan).

src/util/treadmill.rs
qinsoon: The JikesRVM implementation is not lock free. It is a bit strange to have a struct that its only field is a `Mutex` and every method to the struct will acquire the lock. But it is fine.

PR #1414, opened 2025-11-14
Trigger Immix defrag if space utilization is low

qinsoon:


qinsoon:
I don't intend to merge this any time soon. But this heuristic can be useful if space utilization and fragmentation is a higher priority than throughput.

Issue #961, opened 2023-09-25
Windows support


qinsoon:
This is one of our goals, but it is low priority for now. It will probably be driven by one of our supported bindings when we try to upstream our changes in the language, and the language requires Windows to be a platform that needs to be supported.

caizixian:
There's one comment in #745 mentioned

> adjust the implementation of RawMemoryFreeList to fix windows system

PR #1418, opened 2025-11-26
feat: add Windows platform support

sepcnt:
This pull request fix https://github.com/mmtk/mmtk-core/issues/961, adds initial Windows support to the project, focusing on enabling CI, memory management, thread affinity, and malloc functionality for Windows targets. The changes include updates to CI scripts and workflows to handle Windows-specific cases, platform-specific Rust code for memory and thread management, and a new implementation for malloc using Windows APIs.

**Windows platform support**

* Added Windows-specific dependencies to `Cargo.toml` and implemented platform-specific code for thread affinity and CPU detection in `src/scheduler/affinity.rs`, using Windows APIs for processor count and thread affinity. [[1]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R57-R65) [[2]](diffhunk://#diff-44d71e5544160c1957dd1f6a086747341121ef4760958bc29dfcc3e49fe06e69L23-R33) [[3]](diffhunk://#diff-44d71e5544160c1957dd1f6a086747341121ef4760958bc29dfcc3e49fe06e69L62-R83) [[4]](diffhunk://#diff-44d71e5544160c1957dd1f6a086747341121ef4760958bc29dfcc3e49fe06e69L82-R118)
* Implemented a Windows-specific malloc backend in `src/util/malloc/library.rs`, using `HeapAlloc`, `HeapFree`, and related APIs; updated allocation logic in `src/util/malloc/malloc_ms_util.rs` to support Windows. [[1]](diffhunk://#diff-ec589e2f4936d5815dcc54d681529f1150c587d13b07c6363558ea3616013080L5-R16) [[2]](diffhunk://#diff-ec589e2f4936d5815dcc54d681529f1150c587d13b07c6363558ea3616013080L46-R57) [[3]](diffhunk://#diff-ec589e2f4936d5815dcc54d681529f1150c587d13b07c6363558ea3616013080R75-R113) [[4]](diffhunk://#diff-7eb3ea2724933a6a5a1c15a06fce0ec26801a7b475d7cb8dc6541e115665791cR7-R10) [[5]](diffhunk://#diff-7eb3ea2724933a6a5a1c15a06fce0ec26801a7b475d7cb8dc6541e115665791cL74-R98) [[6]](diffhunk://#diff-7eb3ea2724933a6a5a1c15a06fce0ec26801a7b475d7cb8dc6541e115665791cR109)

**CI and workflow updates**

* Modified `.github/scripts/ci-common.sh` and `.github/workflows/minimal-tests-core.yml` to add Windows jobs, set up bash shell defaults, and skip incompatible setup steps for Windows runners; also excluded `malloc_jemalloc` on Windows in feature initialization. [[1]](diffhunk://#diff-f46b82925d438c43e367cfed2e9667e55de4fee73ea40a189324fb18e4dadbe1L70-R75) [[2]](diffhunk://#diff-f46b82925d438c43e367cfed2e9667e55de4fee73ea40a189324fb18e4dadbe1R126-R130) [[3]](diffhunk://#diff-279d2145e8f81e794f0fe7edef2d79ec044eeee81b472d3e83e4f2cd1b856326R40-R49) [[4]](diffhunk://#diff-279d2145e8f81e794f0fe7edef2d79ec044eeee81b472d3e83e4f2cd1b856326R77) [[5]](diffhunk://#diff-279d2145e8f81e794f0fe7edef2d79ec044eeee81b472d3e83e4f2cd1b856326R97-R106) [[6]](diffhunk://#diff-279d2145e8f81e794f0fe7edef2d79ec044eeee81b472d3e83e4f2cd1b856326R134)

**Memory management abstraction**

* Refactored `src/util/memory.rs` to abstract memory protection, mapping, and error handling for Windows, including platform-specific flag translation and error codes, and using Windows APIs for memory operations. [[1]](diffhunk://#diff-68e0bc9c71e86736439c4423d8a513b834a0b58ac39cb4ba28ceb76f9151d36eR6) [[2]](diffhunk://#diff-68e0bc9c71e86736439c4423d8a513b834a0b58ac39cb4ba28ceb76f9151d36eR18-R21) [[3]](diffhunk://#diff-68e0bc9c71e86736439c4423d8a513b834a0b58ac39cb4ba28ceb76f9151d36eL67-R91) [[4]](diffhunk://#diff-68e0bc9c71e86736439c4423d8a513b834a0b58ac39cb4ba28ceb76f9151d36eR177-R189) [[5]](diffhunk://#diff-68e0bc9c71e86736439c4423d8a513b834a0b58ac39cb4ba28ceb76f9151d36eR220-R227) [[6]](diffhunk://#diff-68e0bc9c71e86736439c4423d8a513b834a0b58ac39cb4ba28ceb76f9151d36eR243-R251) [[7]](diffhunk://#diff-68e0bc9c71e86736439c4423d8a513b834a0b58ac39cb4ba28ceb76f9151d36eR269-R272) [[8]](diffhunk://#diff-68e0bc9c71e86736439c4423d8a513b834a0b58ac39cb4ba28ceb76f9151d36eR283-R284)

**General codebase cleanup**

* Removed direct usage of `libc::mprotect` from `src/policy/copyspace.rs`, replacing it with platform-abstracted memory protection functions. [[1]](diffhunk://#diff-1b7c1c2299844a62873b737ac0064caeb57b258cc68f455d1a9c0df5559cfcf6L16) [[2]](diffhunk://#diff-1b7c1c2299844a62873b737ac0064caeb57b258cc68f455d1a9c0df5559cfcf6L296-R296) [[3]](diffhunk://#diff-1b7c1c2299844a62873b737ac0064caeb57b258cc68f455d1a9c0df5559cfcf6L311-R315)

These changes lay the foundation for building, testing, and running the project on Windows, with further improvements expected for full platform parity.

### DIFF
--- a/src/util/malloc/malloc_ms_util.rs
+++ b/src/util/malloc/malloc_ms_util.rs
@@ -4,12 +4,6 @@ use crate::util::Address;
 use crate::vm::VMBinding;
 
 /// Allocate with alignment. This also guarantees the memory is zero initialized.
-/// This uses posix_memalign, which is not available on Windows.
-/// This would somehow affect `MallocMarkSweep` performance on Windows.
-#[cfg(all(
-    not(target_os = "windows"),
-    not(any(feature = "malloc_jemalloc", feature = "malloc_mimalloc"))
-))]
 pub fn align_alloc(size: usize, align: usize) -> Address {
     let mut ptr = std::ptr::null_mut::<libc::c_void>();
     let ptr_ptr = std::ptr::addr_of_mut!(ptr);
@@ -75,24 +69,16 @@ pub fn get_malloc_usable_size(address: Address, is_offset_malloc: bool) -> usize
 pub fn alloc<VM: VMBinding>(size: usize, align: usize, offset: usize) -> (Address, bool) {
     let address: Address;
     let mut is_offset_malloc = false;
-    // malloc returns 16 bytes aligned address.
-    // So if the alignment is smaller than 16 bytes, we do not need to align.
 
     match (align, offset) {
+        // malloc returns 16 bytes aligned address.
+        // So if the alignment is smaller than 16 bytes, we do not need to align.
         (a, 0) if a <= 16 => {
             let raw = unsafe { calloc(1, size) };
             address = Address::from_mut_ptr(raw);
             debug_assert!(address.is_aligned_to(align));
         }
-        #[cfg(all(
-            not(target_os = "windows"),
-            not(any(feature = "malloc_jemalloc", feature = "malloc_mimalloc"))
-        ))]
-        // On non-Windows platforms with posix_memalign, we can use align_alloc for alignments > 16
-        // However, on Windows, there is no equivalent function.
-        // The memory alloc by `align_alloc` may not be freed correctly by `free`.
-        // So we use offset allocation for all alignments > 16 on Windows.
-        (a, 0) if a > 16 => {
+        (a, 0) if a > 16 && SUPPORT_ALIGNED_MALLOC => {
             address = align_alloc(size, align);
             debug_assert!(
                 address.is_aligned_to(align),
--- a/src/util/os/imp/windows.rs
+++ b/src/util/os/imp/windows.rs
@@ -139,8 +139,14 @@ impl OSMemory for Windows {
     fn set_memory_access(start: Address, size: usize, prot: MmapProtection) -> Result<()> {
         use windows_sys::Win32::System::Memory::*;
         let mut old_protect = 0;
-        let res =
-            unsafe { VirtualProtect(start.to_mut_ptr(), size, prot.get_native_flags(), &mut old_protect) };
+        let res = unsafe {
+            VirtualProtect(
+                start.to_mut_ptr(),
+                size,
+                prot.get_native_flags(),
+                &mut old_protect,
+            )
+        };
         if res == 0 {
             Err(std::io::Error::last_os_error())
         } else {
--- a/src/util/malloc/library.rs
+++ b/src/util/malloc/library.rs
@@ -1,19 +1,19 @@
 // Export one of the malloc libraries.
 
-#[cfg(feature = "malloc_jemalloc")]
-pub use self::jemalloc::*;
-#[cfg(all(
-    not(target_os = "windows"),
-    not(any(feature = "malloc_jemalloc", feature = "malloc_mimalloc"))
-))]
-pub use self::libc_malloc::*;
-#[cfg(feature = "malloc_mimalloc")]
-pub use self::mimalloc::*;
-#[cfg(all(
-    target_os = "windows",
-    not(any(feature = "malloc_jemalloc", feature = "malloc_mimalloc"))
-))]
-pub use self::win_malloc::*;
+cfg_if::cfg_if! {
+    // These two libraries support all the platforms.
+    if #[cfg(feature = "malloc_jemalloc")] {
+        pub use self::jemalloc::*;
+    } else if #[cfg(feature = "malloc_mimalloc")] {
+        pub use self::mimalloc::*;
+    } else if #[cfg(target_os = "windows")] {
+        // Use our own Windows malloc implementation on Windows.
+        pub use self::win_malloc::*;
+    } else {
+        // Otherwise use libc's implementation.
+        pub use self::libc_malloc::*;
+    }
+}
 
 /// When we count page usage of library malloc, we assume they allocate in pages. For some malloc implementations,
 /// they may use a larger page (e.g. mimalloc's 64K page). For libraries that we are not sure, we assume they use
@@ -28,6 +28,12 @@ pub const BYTES_IN_MALLOC_PAGE: usize = 1 << LOG_BYTES_IN_MALLOC_PAGE;
 mod jemalloc {
     // Normal 4K page
     pub const LOG_BYTES_IN_MALLOC_PAGE: u8 = crate::util::constants::LOG_BYTES_IN_PAGE;
+    // Whether this allocator supports arbitrary-alignment allocation.
+    // If true, the allocator must provide a function equivalent to
+    // `posix_memalign`, and the returned pointer must be
+    // fully compatible with the rest of the allocator API (i.e., it can be
+    // passed to `free`, `realloc`, and `malloc_usable_size`).
+    pub const SUPPORT_ALIGNED_MALLOC: bool = true;
     // ANSI C
     pub use jemalloc_sys::{calloc, free, malloc, realloc};
     // Posix
@@ -40,12 +46,17 @@ mod jemalloc {
 mod mimalloc {
     // Normal 4K page accounting
     pub const LOG_BYTES_IN_MALLOC_PAGE: u8 = crate::util::constants::LOG_BYTES_IN_PAGE;
+    // Whether this allocator supports arbitrary-alignment allocation.
+    // If true, the allocator must provide a function equivalent to
+    // `posix_memalign`, and the returned pointer must be
+    // fully compatible with the rest of the allocator API (i.e., it can be
+    // passed to `free`, `realloc`, and `malloc_usable_size`).
+    pub const SUPPORT_ALIGNED_MALLOC: bool = true;
     // ANSI C
     pub use mimalloc_sys::{
         mi_calloc as calloc, mi_free as free, mi_malloc as malloc, mi_realloc as realloc,
     };
     // Posix
-    #[cfg_attr(target_os = "windows", allow(unused_imports))]
     pub use mimalloc_sys::mi_posix_memalign as posix_memalign;
     // GNU
     pub use mimalloc_sys::mi_malloc_usable_size as malloc_usable_size;
@@ -59,6 +70,12 @@ mod mimalloc {
 mod libc_malloc {
     // Normal 4K page
     pub const LOG_BYTES_IN_MALLOC_PAGE: u8 = crate::util::constants::LOG_BYTES_IN_PAGE;
+    // Whether this allocator supports arbitrary-alignment allocation.
+    // If true, the allocator must provide a function equivalent to
+    // `posix_memalign`, and the returned pointer must be
+    // fully compatible with the rest of the allocator API (i.e., it can be
+    // passed to `free`, `realloc`, and `malloc_usable_size`).
+    pub const SUPPORT_ALIGNED_MALLOC: bool = true;
     // ANSI C
     pub use libc::{calloc, free, malloc, realloc};
     // Posix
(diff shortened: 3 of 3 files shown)


.github/workflows/minimal-tests-core.yml
```diff
@@ -69,6 +74,7 @@ jobs:
 
       # Setup Environments
       - name: Setup Environments
+        if: runner.os != 'Windows'
```
qinsoon: We could possibly have an empty setup script for Windows, rather than skipping this step.

src/util/malloc/malloc_ms_util.rs
```diff
@@ -71,28 +75,37 @@ pub fn alloc<VM: VMBinding>(size: usize, align: usize, offset: usize) -> (Addres
...
-            offset
-        );
+
+    match (align, offset) {
+        (a, 0) if a <= 16 => {
+            let raw = unsafe { calloc(1, size) };
+            address = Address::from_mut_ptr(raw);
+            debug_assert!(address.is_aligned_to(align));
+        }
+        #[cfg(all(
```
qinsoon: This is a bit confusing. This skips the case for `align > 16 and offset == 0`, and instead uses the more general (and more expensive) case that allows both align and offset. 

If I understand right, the issue is that the skipped case uses `align_alloc` which uses `posix_memalign`.  Windows does not provide something like that `posix_memalign` and allows the memory to be freed using a normal 'free'. It is possible to implement something like `posix_memalign` on Windows, but it needs to be freed with a matching free method.

This sounds okay to me, but this should be documented in the code explicitly.

qinsoon: In the future, we could also simply remove `align_alloc` and the use of `posix_memalign`, and use the align/offset method instead. It will be slower, but it is only used in `MallocMarkSweep` which we don't care much about performance. This is just a note for future work, no changes is needed to remove `align_alloc` in this PR.

src/util/memory.rs
```diff
@@ -157,10 +174,19 @@ impl std::fmt::Display for MmapAnnotation<'_> {
...
     match result {
         Ok(_) => false,
         Err(err) => err.raw_os_error().unwrap() == libc::EEXIST,
     }
+    #[cfg(target_os = "windows")]
+    match result {
+        Ok(_) => false,
+        Err(err) => {
+            // ERROR_INVALID_ADDRESS may be returned if the address is already mapped or invalid
+            err.raw_os_error().unwrap() == 487 // ERROR_INVALID_ADDRESS
```
qinsoon: Use the constant: https://docs.rs/windows-sys/0.61.2/windows_sys/Win32/Foundation/constant.ERROR_INVALID_ADDRESS.html

src/util/memory.rs
```diff
@@ -242,64 +280,148 @@ fn mmap_fixed(
...
+        let ptr = start.to_mut_ptr();
+        let prot = strategy.prot.into_native_flags();
+        let commit =
+            (flags & MAP_NORESERVE) == 0 && !matches!(strategy.prot, MmapProtection::NoAccess);
+
+        // Query the state of the memory region
+        let mut mbi: MEMORY_BASIC_INFORMATION = unsafe { std::mem::zeroed() };
+        let query_res = unsafe {
+            VirtualQuery(
+                ptr,
```
qinsoon: This returns the memory state for a contiguous region that starts at `ptr` and has the same memory state (free/reserve/commit/etc) (doc: https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualquery#remarks). In other words, this does not guarantee that all the address range in `[ptr, ptr + size)` are in the same state. In the branches of the following `match` where `VirtualAlloc` is not used, `size` is simply ignored. We would have to read the size from mbi to check if it covers all the requested memory. If not, we need to keep querying all the address range and consolidate the states.

`VirtualAlloc` also may overwrite existing mappings. But this is a minor issue. For macOS where there is no equivalent to `MAP_FIXED_NOREPLACE`, we also overwrite existing mappings.

src/util/memory.rs
```diff
@@ -329,18 +451,37 @@ pub fn handle_mmap_error<VM: VMBinding>(
...
                 // If it is OOM, we invoke out_of_memory() through the VM interface.
+                #[cfg(not(target_os = "windows"))]
                 if os_errno == libc::ENOMEM {
                     // Signal `MmapOutOfMemory`. Expect the VM to abort immediately.
                     trace!("Signal MmapOutOfMemory!");
                     VM::VMCollection::out_of_memory(tls, AllocationError::MmapOutOfMemory);
                     unreachable!()
                 }
+                #[cfg(target_os = "windows")]
+                if os_errno == 8 {
```
qinsoon: https://docs.rs/windows-sys/0.61.2/windows_sys/Win32/Foundation/constant.ERROR_NOT_ENOUGH_MEMORY.html

qinsoon (commented):
The implementation looks reasonable to me. The major issue is the implementation `mmap_fixed` (see details below).

sepcnt:
> @sepcnt Thanks for opening this PR. Just to better understand the context: do you plan to use MMTk on Windows for a specific project or runtime, or is this PR mainly to add general Windows support to mmtk-core?
> 
> Either way is fine -- I’d just like to understand your use case so we can make better long-term decisions around Windows support.

While I am planning to build a runtime on the top of mmtk, this PR is mainly focus on making a 1:1 port.

src/util/memory.rs
sepcnt: We can actually mock the behavior of `MAP_FIXED_NOREPLACE` if we make a full scan. However, the impact needs further investigation.

wks:
> > @sepcnt Thank you for your PR. While it is nice to support a popular platform, I think this change should be made after we refactor the mmtk-core code base and isolate OS-specific parts into dedicated modules, as suggested in #1420.
> 
> I feel it is a good idea to merge this PR before the refactoring. Supporting for an OS that is significantly different from Unix-like systems could actually help guide the refactoring process for an OS interface for us.

A porting effort like this can definitely guide our progress of porting to another OS.  But given that none of our core members can test this PR locally, and this PR is making a **massive** amount of changes to the code base using AI-assisted tools, I consider it a risk to merge it into the master branch directly.  We may keep this pull request, and do our refactoring with this PR as a reference.

sepcnt:
> > > @sepcnt Thank you for your PR. While it is nice to support a popular platform, I think this change should be made after we refactor the mmtk-core code base and isolate OS-specific parts into dedicated modules, as suggested in #1420.
> > 
> > 
> > I feel it is a good idea to merge this PR before the refactoring. Supporting for an OS that is significantly different from Unix-like systems could actually help guide the refactoring process for an OS interface for us.
> 
> A porting effort like this can definitely guide our progress of porting to another OS. But given that none of our core members can test this PR locally, and this PR is making a **massive** amount of changes to the code base using AI-assisted tools, I consider it a risk to merge it into the master branch directly. We may keep this pull request, and do our refactoring with this PR as a reference.

Most of the changes are gated with `cfg!`. While this PR targets at Windows, I also tested on Ubuntu with below configurations:

```bash
sudo dpkg --add-architecture i386
sudo apt-get update && sudo apt-get install -y mingw-w64 wine64 wine32 xvfb
rustup target add x86_64-pc-windows-gnu
```

And in `.cargo/config.toml`:

```toml
[target.x86_64-pc-windows-gnu]
runner = "xvfb-run -a wine"
linker = "x86_64-w64-mingw32-gcc"
```

wks:
> Honestly, that does not seem to be a prerequisite for a PR to be merged. To me, for a PR to be merged, it needs to:
> 
>     1. pass all the tests/checks, and better have the changes tested.
> 
>     2. pass code review.
> 
>     3. align with the project's goal and design.
> 

If it's not yet a prerequisite, it should be.

Code reviewing and testing are ways to know if *this version* works.  But if we accept this pull request, not only does it mean that we are accepting this change, but it also means we will have the responsibility to maintain it.  It means for every change we make, we need to ensure we don't break Windows.  And if some changes accidentally breaks Windows, we need to debug it.  Given the result of this PR, i.e. `#[cfg]` scattered all over the place, it will be a nightmare for maintenance.  So unless someone volunteers to do the architectural refactoring immediately, this is not going to be friendly to developers.  However, all of our core members are currently quite occupied.  

> We have code reviews and CI. Those are intended to mitigate the risk of merging to master. You can do code review and list the risky parts or code that may cause issues, and expect them to be addressed properly before merging.
> 
> As for AI/LLM coding, there is a lot of discussion recently. I don't want to go deep on this. To my own experience, human can write bad code, and AI (at this stage) under human supervision can write good code. So my general attitude is neutral -- I judge a PR by its content, rather than who wrote it (AI or human).

There are also legal concerns about AI-generated code.  AI-trained coding agents may be trained from and/or directly produce snippets that are incompatible with our license (Apache2 + MIT).  Particularly, we cannot accept snippets that are GPL-licensed.

src/util/os/imp/windows.rs
```diff
@@ -0,0 +1,231 @@
...
+                    size,
+                    allocation_type,
+                    strategy.prot.get_native_flags(),
+                );
+                if res.is_null() {
+                    return Err(io::Error::last_os_error());
+                }
+
+                Ok(start)
+            } else {
```
qinsoon, line 99: @wks commented in https://github.com/mmtk/mmtk-core/pull/1439:

> Despite all the VirtualQuery checks we do, the subsequent VirtualAlloc may still fail in multi-threaded programs due to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use). And it takes time if the given region contains multiple mmap entries. But it is useful for sanity check.
> 
> I suggest we guard all the checks with a feature or debug_assertion so that we don't do them in production. (It's even better if we can extract all the sanity check parts to a separate function so that we can reuse it.) In production, we only call VirtualAlloc. And we can port the sanity check to Unix-like systems by parsing /proc/self/maps (or using a third-party crate for parsing it), but that still has the TICTOU problem and can only serve debug purposes.
> 
> Ideally, we should reserve the region of memory for metadata and the heap so that we won't need such checks.

src/util/os/imp/windows.rs
```diff
@@ -0,0 +1,231 @@
...
+        let res = unsafe { VirtualFree(start.to_mut_ptr(), size, MEM_DECOMMIT) };
+        if res == 0 {
+            // If decommit fails, we try to release the memory. This might happen if the memory was
+            // only reserved.
+            let res_release = unsafe { VirtualFree(start.to_mut_ptr(), 0, MEM_RELEASE) };
+            if res_release == 0 {
+                Err(std::io::Error::last_os_error())
+            } else {
+                Ok(())
+            }
```
qinsoon, line 133: @wks commented in https://github.com/mmtk/mmtk-core/pull/1439#discussion_r2667297644:

> We shouldn't use MEM_RELEASE.
> 
> 1. MEM_RELEASE is supposed to be paired with VirualAlloc and be used as if they were malloc-free pairs. The [doc](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualfree) says "If you specify this value (MEM_RELEASE), dwSize must be 0 (zero), and lpAddress must point to the base address returned by the [VirtualAlloc](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc) function when the region is reserved. The function fails if either of these conditions is not met." Obviously this is completely different from how we use munmap.
> 2. It will remove the reserved state. We don't do it, just like we don't "un-quarantine" memory in Unix-like systems.
> 
> Instead if MEM_DECOMMIT fails, we should return failure.

qinsoon:
> Maybe we can find a way to let Git-Bash use CR LF line ending in the `read` command. If that's impossible, the easiest solution is setting `core.autocrlf` to false in Git.

We can do something like this: https://github.com/rust-lang/rust/blob/859951e3c7c9d0322c39bad49221937455bdffcd/.gitattributes#L3

PR #1419, opened 2025-11-26
Let CI use both OpenJDK 11 and OpenJDK 21

wks:
When running tests using OpenJDK, including binding tests and performance tests, we use both the `jdk-11` branch and the `jdk-21` branch of the repo https://github.com/mmtk/mmtk-openjdk

Specifically, for binding tests, including both minimal tests and extended tests, we split OpenJDK 11 and OpenJDK 21 so that they are treated as two different bindings, and the CI will run tests for OpenJDK 11 and OpenJDK 21 in parallel.

For performance tests, we simply use the `jdk-11` branch instead of the `master` branch.  Since it points to the same branch as the current `master`, its behavior should not change.

### DIFF
--- a/src/policy/immortalspace.rs
+++ b/src/policy/immortalspace.rs
@@ -85,6 +85,10 @@ impl<VM: VMBinding> SFT for ImmortalSpace<VM> {
     ) -> ObjectReference {
         self.trace_object(queue, object)
     }
+    fn debug_print_object_info(&self, object: ObjectReference) {
+        println!("marked = {}", self.mark_state.is_marked::<VM>(object));
+        self.common.debug_print_object_global_info(object);
+    }
 }
 
 impl<VM: VMBinding> Space<VM> for ImmortalSpace<VM> {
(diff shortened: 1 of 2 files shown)


qinsoon:
Do you plan to directly switch to 21 in this PR? If so, testing for `jdk-11` needs to be included, and we need to start a new epoch. I generally suggest doing it in two steps: 1. support correctness testing with both jdk-11 and jdk-21, and perf testing with 11, 2. switch perf testing to 21 and start a new epoch.

wks:
Both minimal tests and extended tests for OpenJDK 11 and OpenJDK 21 finished independently.  OpenJDK 21 took longer to finish.  Individual benchmarks run slightly slower, but since it is fastdebug, this is OK.  Another factor that has greater impact on CI time is compiling OpenJDK itself.  OpenJDK 21 took longer to compile than OpenJDK 11.

I'll need to change the ready-to-merge criteria so that it refers to "extended-tests-openjdk11" and "extended-tests-openjdk21", separately.

Issue #1426, opened 2025-12-02
The `get-merged-pr` workflow sometimes gets empty output


qinsoon:
`get-merged-pr` is triggered when a PR is merged to get its pull request number, and set the pull request number as the output for `binding-refs` to read and check what binding PRs are associated with it.
https://github.com/mmtk/mmtk-core/blob/76ef95a3da8b1f2f1420bfb71122ac37c77761b5/.github/workflows/auto-merge.yml#L12-L25

However, we saw this workflow failed to get the pull request number in https://github.com/mmtk/mmtk-core/actions/runs/19812966874/job/56860107289. I attempted a few reruns and it still failed. 

I also saw this happened before once, but it worked after a rerun.

Will need to investigate the issue.

qinsoon:
The workflow also worked after another rerun https://github.com/mmtk/mmtk-core/actions/runs/19812966874/job/56860107289 (need to rerun all the jobs, rather than failed jobs). I can't reproduce the issue locally. The `gh pr list` worked fine locally.

Possibly there was a delay for the `gh pr list` query. We could add retry for this workflow if the PR number is empty.

wks:
Is the commit hash stable w.r.t. squash merging and multiple jobs in the merge queue?

qinsoon:
> Is the commit hash stable w.r.t. squash merging and multiple jobs in the merge queue?

I think so.
1. We only use squash merging in `mmtk-core` so it works fine for squash merging. 
2. The workflow is triggered when a commit is pushed to `master`. So it doesn't matter how many jobs there are in the merge queue, this workflow is triggered for each commit in `master`.

PR #1427, opened 2025-12-02
Allow retry in get-merged-pr

qinsoon:
This should fix https://github.com/mmtk/mmtk-core/issues/1426. I didn't test it, but the retry pattern is the same as in https://github.com/mmtk/mmtk-core/blob/76ef95a3da8b1f2f1420bfb71122ac37c77761b5/.github/workflows/auto-merge-inner.yml#L98-L103

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -16,19 +16,18 @@ jobs:
       pr: ${{ steps.get-pr.outputs.pr }}
     steps:
       - id: get-commit-hash
-        run: echo "commit=${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
-      - name: Check commit hash and PR
         run: |
+          echo "commit=${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
           echo "Commit hash: ${GITHUB_SHA}"
-          echo "Pull request for the commit is below:"
-          gh pr list --search "${GITHUB_SHA}" --state merged --repo mmtk/mmtk-core
-        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-merge.yml
```diff
@@ -26,9 +26,23 @@ jobs:
...
       - id: get-pr
         run: |
-          PR_NUMBER=$(gh pr list --search "${GITHUB_SHA}" --state merged --repo mmtk/mmtk-core --json number --jq '.[0].number')
-          echo "pr=$PR_NUMBER" >> "$GITHUB_OUTPUT"
-          echo "Output pr=$PR_NUMBER"
+          for n in {1..5}; do
+              echo "Wait 30s then try get the pull request number"
+              sleep 30
+
+              PR_NUMBER=$(gh pr list --search "${GITHUB_SHA}" --state merged --repo mmtk/mmtk-core --json number --jq '.[0].number')
```
wks, line 32: If we need to retry, it means something has gone wrong.  For the ease of debugging, I suggest we print out the human-readable output before each try, and print out the value of `$PR_NUMBER` regardless whether it was successful.

```suggestion
              echo "Pull request for the commit is below:"
              gh pr list --search "${GITHUB_SHA}" --state merged --repo mmtk/mmtk-core
              PR_NUMBER=$(gh pr list --search "${GITHUB_SHA}" --state merged --repo mmtk/mmtk-core --json number --jq '.[0].number')
              echo "Output pr=$PR_NUMBER"              
```

qinsoon: Done.

Issue #1430, opened 2025-12-05
Platform tiers


wks:
This issue is about establishing our list of platform tiers.

# Tiers

Similar to the Rust project which [categorizes platforms (targets) into multiple tiers](https://doc.rust-lang.org/nightly/rustc/platform-support.html), we can define our own platform tiers.

According to Rust's tiers,

-   **Tier 1** targets are guaranteed to work.
-   **Tier 2** targets are guaranteed to build.
-   **Tier 3** targets have support in the Rust code base.

We can extend this definition to MMTk based on the amount of test we do.

-   **Tier 1**: We test it in minimal-tests-core, and test it in minimal-tests-bindings or extended-tests-bindings or both.
-   **Tier 2**: We test it in minimal-tests-core.
-   **Tier 3**: We don't test it on CI, but we support it.  Users (from the core team or from outside) can test and use it if needed.

# Status quo

Here is a list of platforms which MMTk *currently* supports or we plan to support.

| Platform | Rust tier | MMTk tier | Notes |
|---|---|---|---|
| x86_64-unknown-linux-gnu | 1 | 1 | This is our most actively supported platform |
| i686-unknown-linux-gnu | 1 | 1 | We test it with JikesRVM |
| x86_64-apple-darwin | 2 | 2 | We test it with the core, but Rust has [demoted][x86appledem] it recently. |
| aarch64-unknown-linux-gnu | 1 | 3 | We don't test it in CI for now, but it is known to work with OpenJDK 21. |
| riscv64gc-unknown-linux-gnu | 2 | 3 | We don't test it in CI for now, but it is known to work with OpenJDK 21. |
| aarch64-apple-darwin | 1 | x | [Bug about page size][pagesizebug] prevents MMTk to work on this platform |
| x86_64-pc-windows-gnu | 1 | x | [Contribution][winpr] exists, but not merged. |

[x86appledem]: https://github.com/mmtk/mmtk-core/issues/1365
[pagesizebug]: https://github.com/mmtk/mmtk-core/issues/1370
[winpr]: https://github.com/mmtk/mmtk-core/pull/1418

# What should we do?

We may document it somewhere in the user guide or README file in mmtk-core.

We may also list on the website of the MMTk project.

# Tier changes

*We should discuss this at some point.*

Here is a list of known decisions:

-   **aarch64-apple-darwin**: This should be Tier 2 at least, and replace x86_64-apple-darwin.
-   **x86_64-apple-darwin**: We should drop it.
[...]

PR #1431, opened 2025-12-08
Document supported platforms

wks:
We add a section in README.md and mention the level of support we provide for each platform (target triple) in a similar way the Rust project defines "tiers".

Closes: https://github.com/mmtk/mmtk-core/issues/1430

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,30 +136,27 @@ a stable state.
 
 ## Supported platforms
 
-[Like Rust][rustps], supports for different platforms (targets) are organized into three tiers.
+Like [Rust platform support tiers][rustps], supports for MMTk core on different platforms (targets) are organized into three tiers.
 
--   Tier 1 platforms are guaranteed to work.  We do CI tests with the MMTk core and at least one binding.
--   Tier 2 platforms are guaranteed to build.  We do CI tests with the MMTk core.
+-   Tier 1 platforms are guaranteed to work.
+-   Tier 2 platforms are guaranteed to build.
 -   Tier 3 platforms have support in the mmtk-core code base, but we make no guarantee whether they will build or work.
 
 Here is a list of supported platforms.
 
 | Platform                    | Rust tier | MMTk tier | Notes |
 |-----------------------------|-----------|-----------|-------|
-| x86_64-unknown-linux-gnu    | 1         | 1         | (1)   |
-| i686-unknown-linux-gnu      | 1         | 1         | (2)   |
-| x86_64-apple-darwin         | 2         | 2         | (3)   |
-| aarch64-unknown-linux-gnu   | 1         | 3         | (4)   |
-| riscv64gc-unknown-linux-gnu | 2         | 3         | (4)   |
-| aarch64-linux-android       | 2         | 3         | (5)   |
+| x86_64-unknown-linux-gnu    | 1         | 1         |       |
+| i686-unknown-linux-gnu      | 1         | 1         |       |
+| x86_64-apple-darwin         | 2         | 2         | (1)   |
+| aarch64-unknown-linux-gnu   | 1         | 3         |       |
+| riscv64gc-unknown-linux-gnu | 2         | 3         |       |
+| aarch64-linux-android       | 2         | 3         |       |
+| x86_64-linux-android        | 2         | 3         |       |
 
 Notes:
 
-1.  x86_64-unknown-linux-gnu is the most actively supported platform.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -134,6 +134,37 @@ latest Rust toolchain rather than relying on an outdated version of Rust.
 Note, however, that we may switch to a more conservative MSRV policy in the future when MMTk reaches
 a stable state.
 
+## Supported platforms
+
+[Like Rust][rustps], supports for different platforms (targets) are organized into three tiers.
```
qinsoon:
```suggestion
Like [Rust platform support tiers][rustps], supports for MMTk core on different platforms (targets) are organized into three tiers.
```

README.md
```diff
@@ -134,6 +134,37 @@ latest Rust toolchain rather than relying on an outdated version of Rust.
...
+
+-   Tier 1 platforms are guaranteed to work.  We do CI tests with the MMTk core and at least one binding.
+-   Tier 2 platforms are guaranteed to build.  We do CI tests with the MMTk core.
+-   Tier 3 platforms have support in the mmtk-core code base, but we make no guarantee whether they will build or work.
```
qinsoon, line 143: After more thoughts on this, I think we could just adopt Rust's tiers: T1 = 'guarantee to work' (not requiring at least one binding), T2 = 'guarantee to build' (not requiring pass CI tests), T3 = 'we have support, but no guarantee'.

Currently there is inconsistency in the current tier 2 definition: 'guarantee to build', and 'pass CI tests'. Those are two different things. Rust's T2 allows tests to fail. In the future, we will expect mmtk-core's unit tests to be more complex, as seen in https://github.com/mmtk/mmtk-core/pull/1408. I would expect 'passing tests' to be closer to 'guarantee to work' (instead of 'guarantee to build'). 

We may change the tier for a platform in the future, but we don't generally want to change the definition of tiers. Having 'at least one binding' is important for now, but it may become less important in the future when we have more unit tests for MMTk core, and what one binding may test eventually may become a strict subset of what our unit tests cover in MMTk core. Consider two cases. For OpenJDK, we already knew that it only tests the default space and the large object space. Eventually when we have more complex unit tests, we would expect passing CI tests to be more credible than having one binding. For JikesRVM, it only includes a subset of MMTk GC plans. I don't really feel it gives us more confidence in the platform than what the current unit testing delivers. Plus, porting MMTk core to a new platform (especially to a different architecture) should be easier than porting a binding to a new platform. We may have a platform that we could pass all the tests with good code coverage, but we don't have a real binding that targets it. Do we still put the platform as T2 because we don't have a binding to work on it (even if the unit tests may already have more coverage)?

If we really want to show the importance of having a real binding for the platform, we could add a T0 on top of T1/2/3 (which are the same as Rust's tiers): T0 requires at least one binding. In this case, T0 highlights the existence of a real binding being tested with the platform, but T1 also means we expect MMTk to work on the platform.

README.md
```diff
@@ -134,6 +134,37 @@ latest Rust toolchain rather than relying on an outdated version of Rust.
...
+| Platform                    | Rust tier | MMTk tier | Notes |
+|-----------------------------|-----------|-----------|-------|
+| x86_64-unknown-linux-gnu    | 1         | 1         | (1)   |
+| i686-unknown-linux-gnu      | 1         | 1         | (2)   |
+| x86_64-apple-darwin         | 2         | 2         | (3)   |
+| aarch64-unknown-linux-gnu   | 1         | 3         | (4)   |
+| riscv64gc-unknown-linux-gnu | 2         | 3         | (4)   |
+| aarch64-linux-android       | 2         | 3         | (5)   |
+
+Notes:
```
qinsoon, line 157: I think we can remove the notes from `README.md` to keep README succinct. The notes are basically the reasons why we put them in this tier, and we only need to let users know the tiers (rather than the reasons). We can create another doc `docs/team/tiers.md` for these notes with a link in `README.md`

README.md
wks: Done.

README.md
wks: I removed CI and bindings from the definition of tiers.  I think that should be enough for the user, and we leave some room for us to change our CI policy as we need.

README.md
wks: I removed all notes except the one for x86_64-apple-darwin.  It tells the user that x86_64-apple-darwin may be replaced in the future.

qinsoon (approved):

PR #1435, opened 2025-12-12
Fix Clippy warnings for Rust 1.92

wks:
Clippy warned "this assertion has a constant value" in a `debug_assert`.  We replaced the `cfg!` expression with the `#[cfg]` attribute.

Clippy treated a tuple constructor as double parenthesis by mistake.  We extracted the tuple constructor into an explicit let-binding.


udesou (approved):
LGTM.

PR #1438, opened 2025-12-22
Fetch before trying to atomically mark in Compressor

no-defun-allowed:
Tracing is exceedingly slow on biojava if we don't test-and-test-and-set. My guess is that we get cache line ping-pong when many workers try to mark the same popular objects; but every policy I checked has some sort of test-and-test-and-set pattern, and this seems like a common sense thing to do.

[Plotty](http://squirrel.anu.edu.au/plotty/hayleyp/plots/p/GkUHfS) for a detailed before-and-after, although both configurations have some other changes to Compressor which I will upstream later. But this change still speeds up upstream - here is before:

```
============================ MMTk Statistics Totals ============================
GC	time.other	time.stw	total-work.time.min	total-work.time.total	total-work.time.max	total-work.count	
14	8575.41	14920.23	inf	0.000	-inf	0	
Total time: 23495.63 ms
------------------------------ End MMTk Statistics -----------------------------
```

And after:

```
============================ MMTk Statistics Totals ============================
GC	time.other	time.stw	total-work.time.total	total-work.time.max	total-work.time.min	total-work.count	
14	8826.60	9548.96	0.000	-inf	inf	0	
Total time: 18375.56 ms
------------------------------ End MMTk Statistics -----------------------------
```

### DIFF
--- a/src/policy/compressor/compressorspace.rs
+++ b/src/policy/compressor/compressorspace.rs
@@ -272,8 +272,8 @@ impl<VM: VMBinding> CompressorSpace<VM> {
         forwarding::MARK_SPEC
             .fetch_update_atomic::<u8, _>(
                 object.to_raw_address(),
-                Ordering::Relaxed,
                 Ordering::SeqCst,
+                Ordering::Relaxed,
                 |v| {
                     if v == 0 {
                         Some(1)


src/policy/compressor/compressorspace.rs
```diff
@@ -271,19 +269,26 @@ impl<VM: VMBinding> CompressorSpace<VM> {
...
-            object.to_raw_address(),
-            GC_MARK_BIT_MASK,
-            Ordering::SeqCst,
-        );
-        (old & GC_MARK_BIT_MASK) == 0
+        forwarding::MARK_SPEC
+            .fetch_update_atomic::<u8, _>(
+                object.to_raw_address(),
+                Ordering::Relaxed,
+                Ordering::SeqCst,
```
wks: The two memory orders are `set_order` and `fetch_order`, respectively.  I think you meant that when it is successfully updated, the order shall be `SeqCst`; when not updated, the order shall be `Relaxed`.  But these two lines of code are flipped.

But even if we let the "fetch_order" be `Relaxed`, it is inconsistent with `is_marked` because it has `SeqCst` order.  I think we can do one of the following.

-   Change all of the `set_order`, the `fetch_order`, and the memory order of the load operation in `is_marked` below to `Relaxed`.  Currently we only use the mark bit to arbitrate which thread should work on the object if two threads try to mark it. We are not using the mark bit for lock-free synchronization of other data structures (e.g. we don't enforce things like "if a thread sees the object marked, it must also see the object in a queue" or "if a thread sees the object marked, it must also see its last word marked").  So `Relaxed` is OK for the mark bit.
-   Change all of the three memory orders to `SeqCst`.  This is the safest, although it may have some performance impact.

wks (changes requested):
There is a typo in the memory orders, and the values are inconsistent.

src/policy/compressor/compressorspace.rs
no-defun-allowed: Yeah, they should be flipped. It's okay for the fetch order to be relaxed, as we just want a cheap filter against pointless SeqCst operations and the accompanying fences, pulling cache lines exclusively, and such.

no-defun-allowed: I mean, we *could* make both relaxed, as duplicate edges entirely inside the Compressor space are fine (only not-Compressor -> Compressor edges need to be unique with the remset). But relaxed-fetch seqcst-set is what I intended to write.

wks (approved):
LGTM.  I tested locally and I can still reproduce the performance improvement after changing the memory order of successful `fetch_update` to `SeqCst`.

Issue #1420, opened 2025-11-27
OS interface


qinsoon:
We currently only have `util::memory` that is designed to be a module that abstracts out the memory-related system calls. Based on the observation of the changes in https://github.com/mmtk/mmtk-core/pull/1418, clearly we have other parts that is OS dependent. We probably want a more proper OS interface.

We could use `traits` to require each OS to implement certain functions and constants, similar to `VMBinding`. However, we don't need it as a type argument. Instead, we can just statically include one of such implementations based on the host. Elsewhere In the code base, we use the specific implementation and traits.

For example
```rust
trait OperatingSystem {
  type Memory;
  type Process;
}

#[cfg(target_os = "linux")]
mod linux {
  struct Linux;
  impl OperatingSystem for Linux {
    ...
  }
}
#[cfg(target_os = "linux")]
pub use linux::Linux as OS;
```

Elsewhere in the code
```rust
// Hopefully Rust allows us to do the following rather than `<OS::Memory as Memory>::mmap_fixed()`
OS::Memory::mmap_fixed(...);
OS::Process::get_total_num_cpus();
```

qinsoon:
@wks Hopefully we won't need too much CPU-specific implementation in mmtk-core. I mean we probably won't need ISA-specific things (x86, ARM, etc.). Things like endianness or pointer width is necessary.

wks:
> [@wks](https://github.com/wks) Hopefully we won't need too much CPU-specific implementation in mmtk-core. I mean we probably won't need ISA-specific things (x86, ARM, etc.). Things like endianness or pointer width is necessary.

Probably.  Most of the things can be implemented in pure Rust.   If we need to optimize for a certain CPU, for example, scanning bit fields, using CPU-specific instructions, and we want to implement it by ourselves instead of using third-party crates, we may specialize those algorithms in the CPU module, and provide a generic implementation for others in Rust.

And there are CPU-specific properties that are important to us (page sizes, for example), but can be queried dynamically at run time from the OS ([`getpagesize()`](https://linux.die.net/man/2/getpagesize)).  We can do that in OS-specific modules.

PR #1439, opened 2026-01-07
OS interface

qinsoon:
This PR addresses https://github.com/mmtk/mmtk-core/issues/1420.

This PR is based on top of https://github.com/mmtk/mmtk-core/pull/1418, and includes all the changes for Windows (for testing purpose). It is likely that Windows support will be removed from this PR, and will be merged separately.

This PR does not try to refactor our malloc interface -- I am not sure if malloc should be included in the OS interface or not.

This PR consolidates the current multiple mmap functions (such as dzmmap, mmap_fixed, mmap_noreplace, mmap_noreserve, etc), and use `MmapStrategy` to specify the expected mmap behavior.

### DIFF
--- a/src/util/malloc/malloc_ms_util.rs
+++ b/src/util/malloc/malloc_ms_util.rs
@@ -3,6 +3,7 @@ use crate::util::malloc::library::*;
 use crate::util::Address;
 use crate::vm::VMBinding;
 
+/// Allocate with alignment. This also guarantees the memory is zero initialized.
 pub fn align_alloc(size: usize, align: usize) -> Address {
     let mut ptr = std::ptr::null_mut::<libc::c_void>();
     let ptr_ptr = std::ptr::addr_of_mut!(ptr);


src/util/os/imp/windows.rs
```diff
@@ -0,0 +1,237 @@
...
+                    size,
+                    allocation_type,
+                    strategy.prot.get_native_flags(),
+                );
+                if res.is_null() {
+                    return Err(io::Error::last_os_error());
+                }
+
+                Ok(start)
+            } else {
```
wks: Despite all the `VirtualQuery` checks we do, the subsequent `VirtualAlloc` may still fail in multi-threaded programs due to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use).  And it takes time if the given region contains multiple mmap entries.  But it is useful for sanity check.

I suggest we guard all the checks with a feature or `debug_assertion` so that we don't do them in production.  (It's even better if we can extract all the sanity check parts to a separate function so that we can reuse it.)  In production, we only call `VirtualAlloc`.  And we can port the sanity check to Unix-like systems by parsing `/proc/self/maps` (or using a third-party crate for parsing it), but that still has the TICTOU problem and can only serve debug purposes.

Ideally, we should reserve the region of memory for metadata and the heap so that we won't need such checks.

src/util/os/imp/mod.rs
```diff
@@ -0,0 +1,25 @@
+#[cfg(target_os = "windows")]
+pub(crate) mod windows;
+
+#[cfg(target_os = "windows")]
+pub use windows::WindowsMemoryImpl as OSMemory;
+#[cfg(target_os = "windows")]
+pub use windows::WindowsProcessImpl as OSProcess;
```
wks: The more traits we have, the more types we will need to `use` here.  And it is proportional to the number of platforms, too.

I suggest we make a super trait

```rust
trait OS : OSProcess + OSMemory {
}
```

so that each OS provides one type that implements all the required traits.

```rust
struct WindowsOS;
impl OS for WindowsOS {}
impl OSProcess for WindowsOS {
    ...
}
impl OSMemory for WindowsOS {
    ...
}
```

And we can simply import one type here

```suggestion
#[cfg(target_os = "windows")]
pub use windows::WindowsOS as OS;
```

This would probably make the code more concise.

src/util/os/imp/windows.rs
```diff
@@ -0,0 +1,237 @@
...
+        let res = unsafe { VirtualFree(start.to_mut_ptr(), size, MEM_DECOMMIT) };
+        if res == 0 {
+            // If decommit fails, we try to release the memory. This might happen if the memory was
+            // only reserved.
+            let res_release = unsafe { VirtualFree(start.to_mut_ptr(), 0, MEM_RELEASE) };
+            if res_release == 0 {
+                Err(std::io::Error::last_os_error())
+            } else {
+                Ok(())
+            }
```
wks: We shouldn't use `MEM_RELEASE`.

1.  `MEM_RELEASE` is supposed to be paired with `VirualAlloc` and be used as if they were malloc-free pairs.  The [doc](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualfree) says "If you specify this value (`MEM_RELEASE`), dwSize must be 0 (zero), and lpAddress must point to the base address returned by the [VirtualAlloc](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc) function when the region is reserved. The function fails if either of these conditions is not met."  Obviously this is completely different from how we use `munmap`.
2.  It will remove the reserved state.  We don't do it, just like we don't "un-quarantine" memory in Unix-like systems.

Instead if `MEM_DECOMMIT` fails, we should return failure.

```suggestion
            Err(std::io::Error::last_os_error())
```

wks (commented):
Some High-level comments:

# `windows` crate or `windows-sys` crate

There is a `windows` crate and it claims to be more idiomatic for Rust than `windows-sys` (for example, `windows` uses `Option<*mut XxxxxXxxType>` where `windows-sys` uses raw `*mut XxxxxXxxType` if it is nullable).  Personally I think we only call a limited number of Windows API functions, so it doesn't really matter.

# Granularity or reserving and committing

According to the [documentation `VirtualAlloc`](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc),

1.  The granularity of reserving and the granularity of committing may be different, and
2.  They are not constants, but can be queried using the [`GetSystemInfo()`](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsysteminfo) function.

We don't have to fix it in this PR because `Memory::dzmmap` maps the exactly memory range the user requested, which is usually a chunk.  But this is a reminder that we should make a related PR to remove the hard-coded 4K page size because MacOS and Android use different page sizes, too.

qinsoon:
I don't intend to include Windows-related changes in this PR for merging. See the PR description. This PR is only about OS interface refactoring. 

I will keep Windows-related comments unresolved, and will copy them to the Windows support PR.

src/util/os/imp/mod.rs
qinsoon: Sounds good. Done in https://github.com/mmtk/mmtk-core/pull/1439/commits/72d98ea0af174d7aee832e659332bf0857a070e5.

.github/scripts/ci-common.sh
```diff
@@ -12,6 +12,19 @@ dummyvm_toml=$project_root/docs/dummyvm/Cargo.toml
...
+  printf '%s' "${s%$'\r'}"
+}
+
+# Trim all whitespace (for feature keys)
+trim_ws() {
+  local s=$1
+  # remove all whitespace characters
+  printf '%s' "${s//[[:space:]]/}"
+}
+
```
wks: We shouldn't have any CR in any text files in mmtk-core, and the last line of any text files should end with an LF character.  We have a CI style check in mmtk-openjdk that enforces that.  See: https://github.com/mmtk/mmtk-openjdk/blob/jdk-11/.github/scripts/ci-check-lineends.sh  That was introduced specifically for avoiding complex line-end handling in scripts.  If we need it, we can copy it to mmtk-core.

.github/scripts/ci-common.sh
```diff
@@ -46,6 +59,8 @@ init_non_exclusive_features() {
     i=0
 
     while IFS= read -r line; do
+        line=$(strip_cr "$line")
```
wks: Similarly, if we don't have CR in the first place, this line and other changes in this file will be unnecessary.

wks (commented):
We should ensure that all text files in mmtk-core use LF (instead of CR LF or CR-only) line ends and ensure that the last line always ends with an LF.  By doing so, we can avoid the complexity of handling line ending in CI scripts.

.github/workflows/minimal-tests-core.yml
```diff
@@ -42,6 +42,10 @@ jobs:
     name: minimal-tests-core/${{ matrix.target.triple }}/${{ matrix.rust }}
     runs-on: ${{ matrix.target.os }}
 
+    defaults:
```
wks: I think this is here to support windows.  If we are not adding Windows support in this PR, we can omit this and the other occurrence in this file.

I think there should be a per-file or global setting that allow us to set the shell type in one place.  On Windows, we may need to specify whether we want the Bash from Cygwin or MinGW or whatever Unix-compatible layer.  We can leave it to the windows-support PR.

src/policy/copyspace.rs
```diff
@@ -293,8 +293,8 @@ impl<VM: VMBinding> CopySpace<VM> {
         }
         let start = self.common().start;
         let extent = self.common().extent;
-        unsafe {
-            mprotect(start.to_mut_ptr(), extent, PROT_NONE);
+        if let Err(e) = crate::util::os::OS::mprotect(start, extent) {
```
wks: We have added `use crate::util::os::*`.  We may omit the `crate::util::os::` prefix here, and the other invocation in `fn unprotect()` below.

src/util/os/memory.rs
```diff
@@ -0,0 +1,321 @@
...
+    fn memzero(start: Address, len: usize) {
+        Self::memset(start, 0, len);
+    }
+
+    /// Set a memory region to a specific value.
+    fn memset(start: Address, val: u8, len: usize) {
+        unsafe {
+            std::ptr::write_bytes::<u8>(start.to_mut_ptr(), val, len);
+        }
+    }
```
wks: The word "region" has a specific meaning in MMTk, that is, a memory range that is power of two in size, and aligned to its size.  We can replace the word "region" with "range" in the comment.

`std::ptr::write_bytes` is probably an intrinsic function and the compile will be able to optimize it with SIMD instructions.  But this is not really OS-specific.  We may leave `memzero` and `memset` in the OS-neutral `crate::util::memory` module.

What could be OS-specific is bulk-zeroing memory by copying from `/dev/zero`, in which case the OS may be able to help using mmap tricks internal to the kernel.  We need to investigate further if we need such optimization.  But if we make an abstraction of that, the API may need to be different.  For example, the memory range may need to be page-aligned.

src/policy/lockfreeimmortalspace.rs
```diff
@@ -255,11 +254,12 @@ impl<VM: VMBinding> LockFreeImmortalSpace<VM> {
         };
 
         // Eagerly memory map the entire heap (also zero all the memory)
-        let strategy = MmapStrategy::new(
-            *args.options.transparent_hugepages,
-            crate::util::memory::MmapProtection::ReadWrite,
-        );
-        crate::util::memory::dzmmap_noreplace(
+        let strategy = MmapStrategy::default()
```
wks, line 257: The builder pattern doesn't look necessary in the Rust programming language.  The struct construction syntax with `..defaultvalue` looks pretty readable.

```rust
let strategy = MmapStrategy {
  huge_page_option: HugePageSupport::from(*args.options.transparent_hugepages),
  prot: MmapProtection::ReadWrite,
  replace: false,
  reserve: true,
  ..Default::default()
};
```

We may separately decide whether we want to let `HugePageSupport` implement `From<bool>` so that we can just call `args.options.transparent_hugepages.into()`.  `MmapStrategy::transparent_hugepages` also assumes a boolean can be converted to a `HugePageSupport`, but since it is an `enum`, we may add other values in the future.  For now, we can keep using the if statement here, or make a method such as `HugePageSupport::with_transparent_hugepage(enabled: bool) -> bool` as a shorthand which does the same thing as `MmapStrategy::transparent_hugepages`.

.github/scripts/ci-common.sh
qinsoon: I think by default configuration, git will convert LF to CRLF if the repo is checked out on Windows. So if we run this script on Windows, we need to handle the character.

I can remove this change and later commit it with the windows changes if requested.

wks: Indeed.  `core.autocrlf` is `true` on Windows.  But if text files use CR LF on Windows, scripts should be able to interpret CR LF as line separator automatically and not require manual line-end conversion inside scripts.  I guess for some reason the GitHub CI was executing programs (such as the `bash` in this case) that assumes LF line ending.  This could happen if the Bash (provided by MinGW, CygWin, etc.) program happens to be configured to follow the Unix convention.  We could either let Git do the conversion on checkout and let Bash assume Windows convention, or disable line-end conversion on Git and let Bash assume Unix convention.  We can do it when we add Windows support.

src/util/os/memory.rs
qinsoon: Done. I added back the `crate::util::memory` module, and moved `set/zero` back to the module.

.github/scripts/ci-common.sh
qinsoon: I have removed those changes. Those will be included in the Windows PR.

src/policy/lockfreeimmortalspace.rs
qinsoon: We use both the builder style (e.g. `MMTkBuilder`) and the constructor (e.g. `PlanConstraint`) in the code base. I don't see one is better than the other in this case. I feel we can keep it as it is.

.github/scripts/ci-common.sh
```diff
@@ -66,7 +66,8 @@ init_non_exclusive_features() {
             IFS='='; feature=($line); unset IFS;
             if [[ ! -z "$feature" ]]; then
                 # Trim whitespaces
-                features[i]=$(echo $feature)
+                feature=$(echo $feature)
+                features[i]=$feature
```
wks: I don't understand why [the original PR](https://github.com/mmtk/mmtk-core/pull/1418) introduced one intermediate step.  This needs some explanation.

src/policy/lockfreeimmortalspace.rs
wks: Sorry.  What I really meant was the ["fluent" interface](https://en.wikipedia.org/wiki/Fluent_interface) which sets attributes via methods.  I initially thought it was unnecessary given that Rust has the struct initialization syntax.  But after a second thought, having fluent interface is OK because Rust also has a `std::thread::Builder` which also uses the fluent interface.

src/util/os/memory.rs
```diff
@@ -0,0 +1,302 @@
...
+use crate::{
+    util::{address::Address, VMThread},
+    vm::VMBinding,
+};
+
+/// Abstraction for OS memory operations.
+pub trait OSMemory {
+    /// Perform a demand-zero mmap.
+    ///
+    /// Falback: `annotation` is only used for debugging. For platforms that do not support mmap annotations, this parameter can be ignored.
```
wks:
```suggestion
    /// Fallback: `annotation` is only used for debugging. For platforms that do not support mmap annotations, this parameter can be ignored.
```

src/util/os/memory.rs
```diff
@@ -0,0 +1,302 @@
...
+        panic!("Unexpected mmap failure: {:?}", error)
+    }
+
+    /// Check whether the given OS error number indicates an out-of-memory condition.
+    fn is_mmap_oom(os_errno: i32) -> bool;
+
+    /// Unmap a memory region.
+    fn munmap(start: Address, size: usize) -> Result<()>;
+
+    /// Change the protection of a memory region to no access to forbit any access to the memory region.
```
wks: Typo, but it'll be irrelevant if we merge `mprotect` and `munprotect` as I suggested in another comment.

```suggestion
    /// Change the protection of a memory region to no access to forbid any access to the memory region.
```

src/util/os/memory.rs
```diff
@@ -0,0 +1,302 @@
...
+    fn mprotect(start: Address, size: usize) -> Result<()>;
+
+    /// Change the protection of a memory region to the specified protection.
+    fn munprotect(start: Address, size: usize, prot: MmapProtection) -> Result<()>;
```
wks: I suggest we combine `mprotect` and `munprotect` into one function `set_memory_access` or `set_memory_protection` or just `mprotect`.

1.  The underlying system call of both `mprotect` and `munprotect` is "mprotect".  This can confuse users who are familiar with Unix programming when they found "munprotect" actually calls "mprotect".
2.  `mprotect(start, size)` should be an alias of `munprotect(start, size, MmapProtection::NoAccess)`, making it redundant.
3.  `munprotect` can actualy modify the access right of the memory range.  But "unprotect" means reverting the "protecting" operation.

So we can just have one function that does what the current `munprotect` does, but name it differently.

Given that the master branch had `mprotect` and `munprotect` in the `memory` module, and this PR is about refactoring the OS interface, we may choose to make this change separately if we feel that this PR is not the right time to do it.

src/util/os/process.rs
```diff
@@ -0,0 +1,30 @@
...
+
+/// Abstraction for OS process operations.
+pub trait OSProcess {
+    /// Get the memory maps for the process. The returned string is a multi-line string.
+    /// Fallback: This is only used for debugging. For unimplemented cases, this function can return a placeholder Ok value.
+    fn get_process_memory_maps() -> Result<String>;
+
+    /// Get the process ID as a string.
+    /// Fallback: This is only used for debugging. For unimplemented cases, this function can return a placeholder Ok value.
+    fn get_process_id() -> Result<String>;
```
wks: Returning the process ID as string is not as useful as returning the raw value.  If some other parts of MMTk-core is OS-specific, they may use the raw integer value.

We can let each OS define its PID type because they may be different.

```rust
pub trait OSProcess {
    type ProcessIDType: Display;
    type ThreadIDType: Display;
    fn get_process_id() -> Result<Self::ProcessIDType>;
    fn get_thread_id() -> Result<Self::ThreadIDType>;
}
```

On Linux, both PID and TID have the `pid_t` type whch is an alias of i32.

```rust
impl OSProcess for Linux {
    type ProcessIDType = libc::pid_t;
    type ThreadIDType = libc::pid_t;

    fn get_process_id() -> Result<Self::ProcessIDType> {
        Ok(unsafe { libc::getpid() })
    }

    fn get_thread_id() -> Result<Self::ThreadIDType> {
        Ok(unsafe { libc::gettid() })
    }
}
```

On Windows, both PID and TID are `DWORD` (`u32`).  MacOS may need to use `pthread_threadid_np` for TID. (see https://stackoverflow.com/questions/19247228/what-is-thread-id-of-posix-on-mac-osx)

Because both the PID type and the TID type are required to implement `Display`, `debug_process_thread_id` can still directly use them in formatting.

src/util/rust_util/mod.rs
```diff
@@ -108,19 +108,12 @@ unsafe impl<T> Sync for InitializeOnce<T> {}
...
-        // TODO: When we support other platforms, use platform-specific methods to get thread
-        // identifiers.
-        format!("PID: {}", pid)
-    }
+    use crate::util::os::*;
+    format!(
+        "PID: {}, TID: {}",
+        OS::get_process_id().unwrap_or_default(),
+        OS::get_thread_id().unwrap_or_default()
+    )
```
wks, line 116: If we keep PID and TID as strings, the default value will be an empty string.  If we refactor the interface so that PID and TID types are defined by each OS, then they are required to implement `Display` but not `Default`.

In either case, I don't think the default value of the type makes sense because it is confusing to see an empty string where a numerical ID should appear, and 0 basically means an invalid PID/TID, which can be encoded as `Err`.  We should display something like "(Failed get PID)" if PID cannot be obtained via `OS::get_process_id()`.

```suggestion
    let pid_str: Cow<'_, str> = OS::get_process_id()
        .map(|x| Cow::from(x.to_string()))
        .unwrap_or_else(|_| Cow::from("(Failed to get PID)"));
    let tid_str = OS::get_thread_id()
        .map(|x| Cow::from(x.to_string()))
        .unwrap_or_else(|_| Cow::from("(Failed to get TID)"));
    format!("PID: {pid_str}, TID: {tid_str}",)
```

.github/scripts/ci-common.sh
qinsoon: I reverted the changes.

src/util/os/memory.rs
qinsoon: The two functions are changed to `set_memory_access`.

src/util/os/process.rs
qinsoon: I made the change. I think just requiring `Display` for pid/tid won't sufficient. I also added `Eq` and `Copy` so we can do equity compare and pass by value. We may need to expand this requirement in the future.

> If some other parts of MMTk-core is OS-specific, they may use the raw integer value.

I don't think OS-specific code really should use the OS interface, as the OS interface is for non OS specific code. OS specific code can always directly call getpid etc.

src/util/os/imp/unix_like/unix_common.rs
```diff
@@ -0,0 +1,60 @@
...
+
+pub type ProcessIDType = libc::pid_t;
+pub type ThreadIDType = libc::pthread_t;
+
+pub fn get_process_id() -> Result<ProcessIDType> {
+    Ok(unsafe { libc::getpid() })
+}
+
+pub fn get_thread_id() -> Result<ThreadIDType> {
+    Ok(unsafe { libc::pthread_self() })
```
wks, line 50: Linux has the `gettid()` system call, available in the `libc` crate in Rust, although it doesn't have a C language wrapper in libc.  I think it is better than `pthread_self` on Linux because that is what Linux users call "TID", and it can be shown by the command `ps -efL` or in the `top` utility by pressing the 'H' key.  It is confusing to call the return value of `pthread_self` "TID" on Linux.

Android should be able to use `gettid()`, too.

src/util/os/imp/unix_like/linux_like/linux.rs
```diff
@@ -0,0 +1,90 @@
...
+pub struct Linux;
+
+impl OSMemory for Linux {
+    fn dzmmap(
+        start: Address,
+        size: usize,
+        strategy: MmapStrategy,
+        annotation: &MmapAnnotation<'_>,
+    ) -> Result<Address> {
+        let addr = unix_common::mmap(start, size, strategy)?;
```
wks: This implementation is identical to Android.  We may extract it to `linux_common.rs`

src/util/os/imp/unix_like/linux_like/linux.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn set_memory_access(start: Address, size: usize, prot: MmapProtection) -> Result<()> {
+        unix_common::mprotect(start, size, prot)
+    }
+
+    fn is_mmap_oom(os_errno: i32) -> bool {
+        unix_common::is_mmap_oom(os_errno)
+    }
+
+    fn panic_if_unmapped(start: Address, size: usize) {
+        let strategy = MmapStrategy {
```
wks: This can be extracted to `linux_common.rs`, too.

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -93,10 +93,10 @@ pub(super) fn align_metadata_address(
 /// Unmaps the specified metadata range, or panics.
 #[cfg(test)]
 pub(crate) fn ensure_munmap_metadata(start: Address, size: usize) {
-    use crate::util::memory;
+    use crate::util::os::*;
```
wks: We already added a `use` in the same file at line 9.

src/util/os/imp/unix_like/linux_like/linux_common.rs
```diff
@@ -0,0 +1,102 @@
...
+    f.read_to_string(&mut data)?;
+    Ok(data)
+}
+
+pub fn get_total_num_cpus() -> CoreNum {
+    use std::mem::MaybeUninit;
+    unsafe {
+        let mut cs = MaybeUninit::zeroed().assume_init();
+        CPU_ZERO(&mut cs);
+        sched_getaffinity(0, std::mem::size_of::<cpu_set_t>(), &mut cs);
```
wks, line 77: This doesn't get the total number of CPUs.  It actually gets the number of CPUs in the current affinity mask, although its initial value should be the number of CPUs if the process has not set any affinity mask, yet.  I think it is a bug, and we should fix it separately.

k-sareen: It's not a bug. It's the correct way to figure out how many cores a process has been assigned. `taskset` changes the affinity mask before we spawn GC threads. So we can incorrectly spawn more GC threads than cores assigned to the process if we don't actually check the mask properly. This is also [what HotSpot does](https://github.com/openjdk/jdk/blob/6c8d5daad1f388dc9fc8af6c9b3674846050dc7e/src/hotspot/os/linux/os_linux.cpp#L4711).

src/util/os/imp/unix_like/unix_common.rs
qinsoon: The thread ID here is for MMTk. It doesn't matter if it is the pthread or the tid, as long as it can uniquely identify a thread for MMTk.

MacOS does not have `gettid`, while all those systems have pthread pointers. That's why we used `pthread_self`.

wks: Well, one important use case is GDB and rr.  But GDB shows both the "Thread ID" (`pthread_self()`) and the LWP (`gettid()`).  So it should be OK to use `pthread_self` as the "thread ID".

src/util/malloc/malloc_ms_util.rs
```diff
@@ -3,7 +3,6 @@ use crate::util::malloc::library::*;
 use crate::util::Address;
 use crate::vm::VMBinding;
 
-/// Allocate with alignment. This also guarantees the memory is zero initialized.
```
wks, line 6: Why is this comment removed?  The function still guarantees the memory is zero initialized.

qinsoon: That was unintended. I removed some windows related `cfg`s there. I added it back.

PR #1441, opened 2026-01-08
Use carryless multiply for Compressor forwarding metadata

no-defun-allowed:
This PR introduces an algorithm for computing the offset vector in the Compressor which uses the carryless multiply instruction, based on the branch-free and bit-parallel algorithm in https://branchfree.org/2019/03/06/code-fragment-finding-quote-pairs-with-carry-less-multiply-pclmulqdq/

### DIFF
--- a/src/policy/compressor/forwarding.rs
+++ b/src/policy/compressor/forwarding.rs
@@ -273,7 +273,7 @@ impl<VM: VMBinding> ForwardingMetadata<VM> {
 
     #[cfg(target_arch = "x86_64")]
     #[target_feature(enable = "pclmulqdq,popcnt")]
-    fn forward_clmul(&self, address: Address) -> Address {
+    unsafe fn forward_clmul(&self, address: Address) -> Address {
         debug_assert!(processor_can_clmul());
         let block = Block::from_unaligned_address(address);
         let (mut to, mut carry) = {
@@ -336,7 +336,7 @@ cfg_if::cfg_if! {
         }
 
         #[target_feature(enable = "pclmulqdq,popcnt")]
-        fn clmul_step(to: &mut Address, carry: &mut i64, word: usize) {
+        unsafe fn clmul_step(to: &mut Address, carry: &mut i64, word: usize) {
             use std::arch::x86_64;
             // Compute the prefix sum of this word of mark bitmap.
             let ones = x86_64::_mm_set1_epi8(0xFFu8 as i8);


no-defun-allowed:
Some issues come to mind.

- I've now introduced CPU-specific code to MMTk, I don't know if we have a plan for how to put that in the codebase. My current approach of dumping `#[cfg(target_arch = "x86_64")]` in the Compressor directory is rather undisciplined.
- `ranges::break_byte_range` is patterned after `ranges::break_bit_range`; I don't know if that's a good approach, and it's currently undocumented and untested.
- I have a portable prefix sum in `forwarding::prefix_sum` — although according to [the Wikipedia article](https://en.wikipedia.org/wiki/Prefix_sum#Scan_higher_order_function) it's a xor-scan, although xor is addition over ℤ₂ if you're so inclined — which appears about as fast as the original branchy algorithm. Is it worthwhile to keep the original branchy code, or should the fallback use that prefix sum algorithm?
- I'm still waiting for properly-measured results, but so far the carryless multiply algorithm is either within the noise or up to 5% faster in STW time than the branchy algorithm. This might not be a large enough speedup to deserve the complexity.

qinsoon:
> [Plotty](http://squirrel.anu.edu.au/plotty/hayleyp/plots/p/ZjSS35) and here are the geomean results for clmul-enabled relative to clmul-disabled:
> Heap factor 	Total time 	STW time
> 1000 	0.982 	0.970
> 1416 	0.997 	0.984
> 1892 	0.989 	0.970
> 2428 	0.997 	0.983
> 3023 	0.996 	0.980
> 3678 	0.996 	0.977
> 4392 	0.994 	0.970
> 5166 	0.995 	0.981
> 6000 	1.001 	0.965

To clarify, do both builds include the change for computing multiple regions in one `CalculateOffsetVector` package? Just asking, as the PR has two changes (`clmul` and the work package change), and the performance impact of the work package change was not mentioned.

no-defun-allowed:
> I also neglected to mention that I shrunk regions from 1MiB to 256kiB, to improve work balancing.

I tested this PR against upstream; [Plotty](http://squirrel.anu.edu.au/plotty/hayleyp/plots/p/8W9rQR) and a tally of changes to stop-the-world times:

|Speedup|Count|
|-:|-:|
|+20–25.1%|3|
|+10–20%|2|
|+5–10%|2|
|+2–5%|1|
|Too small to say|3|
|Too kafka to say|1|
|-2–5%|5|
|-5–5.4%|2|

Some benchmarks see large improvements due to the better work balancing, some see small regressions due to worse locality with smaller region sizes. (I only found the *mutator* time to be consistently worse on tradesoap, which also has the worst STW regression at 5.4% slower.)

no-defun-allowed:
The big wins are actually in being able to clmul-ise `.forward()`. [Plotty](http://squirrel.anu.edu.au/plotty/hayleyp/plots/p/FjSEHj) and another table of geomeans versus upstream:

|Heap factor|Total time|Non-zero STW times|
|-:|-:|-:|
|1000|0.760|0.531|
|1416|0.832|0.538|
|1892|0.869|0.541|
|2428|0.889|0.544|
|3023|0.915|0.570|
|3678|0.923|0.559|
|4392|0.937|0.567|
|5166|0.943|0.577|
|6000|0.951|0.574|

No benchmarks regress either, which is neat.

PR #1444, opened 2026-01-15
Use ci-perf-kit 0.8.6 (new epoch for CI runner package update on 5th Jan)

qinsoon:
We observed a performance slowdown in CI results recently. The slowdown first appears in the 6th Jan CI run (before the Rust version upgrade which was on 12th Jan): the canary showed a ~3% regression on luindex, and openjdk-stickyimmix showed a similar slowdown. (Other benchmarks also show regressions; luindex is used here as the performance difference for luindex is most obvious)

I also tested 3 builds ([`20251215`](https://github.com/mmtk/mmtk-core/commit/2ad6d36ab3d82be46135a7d76d4fe59cc84620ed), [`20260106`](https://github.com/mmtk/mmtk-core/commit/473fd586a9f532fa85b934fd5955c7e2db8281ff), and [`20260112`](https://github.com/mmtk/mmtk-core/commit/e30d0a0866fe66d352eb09eb7f9e9466dc649172)) with luindex (on a Zen 3 machine, rather than CI runenr's Zen 2), there was no measurable perf difference between those 3 builds. [link](https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|bear-2026-01-15-Thu-042440&benchmark^build^invocation^iteration&bmtime&|10&iteration^1^1|20&1^invocation|30&1&benchmark&build;jdk-mmtk-251215|41&Histogram%20(with%20CI)^build^benchmark&)

ScreenShot_2026-01-15_181157_830

@no-defun-allowed confirmed that there was a big update of all the packages for CI runners on 5th Jan. We started to see the difference on the commit in 6th Jan. So the performance difference is unrelated with MMTk moving to Rust 1.92 (which happened on 12th Jan).

This PR updates `ci-perf-kit` to 0.8.6 which will plot 5th Jan as a new epoch.

### DIFF
--- a/.github/workflows/micro-bm.yml
+++ b/.github/workflows/micro-bm.yml
@@ -69,7 +69,7 @@ jobs:
         with:
           repository: mmtk/ci-perf-kit
           token: ${{ secrets.GITHUB_TOKEN }}
-          ref: "0.8.5"
+          ref: "0.8.6"
           path: ci-perf-kit
           submodules: true
       # Use rust-toolchain in the trunk (it doesnt matter much - if the toolchains defined in the trunk and the branch are different, we cant run anyway)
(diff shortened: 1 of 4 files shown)


wks (approved):
LGTM

Issue #1351, opened 2025-07-24
Handle mmap failures and memory occupied by other programs


qinsoon:
When MMTk uses a predefined heap range, and fails to mmap the required memory, MMTk should be able to step through these 'holes' in its heap, and continue to work.

This only applies to the heap memory, and does not apply to the side metadata memory. We have strict mapping between data range to side metadata range. MMTk should quarantine side metadata memory to avoid such situations.

qinsoon:
We have been talking about this for a while, but I didn't seem to find an issue for it. Let me know if there is a duplicate issue.

PR #1449, opened 2026-02-19
Dynamic side metadata base address

qinsoon:
This addresses part of the issues with https://github.com/mmtk/mmtk-core/issues/1351. This PR allows side metadata to be mmapped dynamically, and also allows side metadata to use a fixed address specified in the option `side_metadata_base_address`.

### DIFF
--- a/benches/regular_bench/bulk_meta/access.rs
+++ b/benches/regular_bench/bulk_meta/access.rs
@@ -7,7 +7,7 @@ use mmtk::util::{
     conversions::raw_align_up,
     metadata::side_metadata::SideMetadataSpec,
     os::{MmapAnnotation, MmapProtection, MmapStrategy, OSMemory, OS},
-    test_private::side_metadata_address_to_meta_address,
+    test_private::address_to_meta_address,
     Address,
 };
 use std::hint::black_box;
@@ -37,7 +37,7 @@ pub fn bench(c: &mut Criterion) {
         b.iter(|| {
             let mut checksum = 0usize;
             for data_addr in &addrs {
-                let meta_addr = side_metadata_address_to_meta_address(&BENCH_SPEC, *data_addr);
+                let meta_addr = address_to_meta_address(&BENCH_SPEC, *data_addr);
                 checksum ^= meta_addr.as_usize();
             }
             black_box(checksum);
@@ -51,7 +51,7 @@ pub fn bench(c: &mut Criterion) {
         let mut meta_min = Address::MAX;
         let mut meta_max = Address::ZERO;
         for data_addr in &addrs {
-            let meta_addr = side_metadata_address_to_meta_address(&BENCH_SPEC, *data_addr);
+            let meta_addr = address_to_meta_address(&BENCH_SPEC, *data_addr);
             if meta_addr < meta_min {
                 meta_min = meta_addr;
             }
--- a/src/util/os/imp/unix_like/macos.rs
+++ b/src/util/os/imp/unix_like/macos.rs
@@ -31,11 +31,7 @@ impl OSMemory for MacOS {
         strategy: MmapStrategy,
         _annotation: &MmapAnnotation<'_>,
     ) -> Result<Address> {
-        unix_common::mmap_anywhere(
-            size,
-            align,
-            strategy.prot(MmapProtection::NoAccess).reserve(false),
-        )
+        unix_common::mmap_anywhere(size, align, strategy)
     }
 
     fn munmap(start: Address, size: usize) -> Result<()> {
--- a/src/util/test_private/mod.rs
+++ b/src/util/test_private/mod.rs
@@ -38,10 +38,7 @@ pub fn set_meta_bits(
 
 /// Expose `address_to_meta_address` when running `cargo bench`.
 #[inline(always)]
-pub fn side_metadata_address_to_meta_address(
-    metadata_spec: &SideMetadataSpec,
-    data_addr: Address,
-) -> Address {
+pub fn address_to_meta_address(metadata_spec: &SideMetadataSpec, data_addr: Address) -> Address {
     crate::util::metadata::side_metadata::address_to_meta_address(metadata_spec, data_addr)
 }
 


qinsoon:
I will further look into the micro benchmarks and the Immix performance.

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -161,6 +160,7 @@ pub(super) fn try_mmap_contiguous_metadata_space(
 }
 
 /// Performs the translation of data address (`data_addr`) to metadata address for the specified metadata (`metadata_spec`).
+#[inline(always)]
```
qinsoon: This `#[inline(always)]` directive is necessary. With this inlined, the microbenchmark shows no slowdown on loading side metadata.
```
side_metadata_load      time:   [2.6933 µs 2.6940 µs 2.6948 µs]
                        change: [−68.453% −68.423% −68.397%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low severe
  1 (1.00%) high mild
```

I am running Immix again, and see if the inline fixes the slowdown for Immix. 

Note that I use PGO when building OpenJDK. If the manual inline directive fixes the slowdown for Immix, that means somehow PGO does not inline the key function for side metadata address calculation.

wks: > ... With this inlined, the microbenchmark shows no slowdown...

That's right.  The compiler can usually figure out what to inline in real-world VM bindings, but for microbenchmarks (`cargo bench`), manual inlining matters.  I have observed this before, and added some annotations in `src/util/test_private/mod.rs`.  I think you can try adding wrappers in the `test_private` mod that have `#[inline(always)]` and keep `side_metadata/helpers.rs` free of inlining annotations.  If manual inlining is still necessary for the OpenJDK binding, we can keep `#[inline(always)]` in `helpers.rs`.

qinsoon: The manual inline does not fix the performance for Immix on dacapo: https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|shrew-2026-02-24-Tue-031617&benchmark^build^invocation^iteration&time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark&build;jdk-21-constant-side-metadata|41&Histogram%20(with%20CI)^build^benchmark&

src/util/metadata/side_metadata/layout.rs
```diff
@@ -0,0 +1,215 @@
...
+use crate::util::{constants::LOG_BYTES_IN_PAGE, conversions::raw_align_up};
+use crate::MMAPPER;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Once;
+
+// The compile-time base offset for global side metadata layout. We treat offsets as relative
+// (starting from zero) and add the runtime base address when computing actual addresses.
+pub(crate) const GLOBAL_SIDE_METADATA_BASE_OFFSET: usize = 0;
+
+static mut SIDE_METADATA_BASE_ADDRESS: Address = Address::ZERO;
```
qinsoon: I replaced this `static mut` with a `OnceLock`, and got the following results on `biojava` which showed no slowdown for this PR. 
https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|shrew-2026-02-26-Thu-071955&benchmark^build^invocation^iteration^stickyix&time^time.other^time.stw&|10&iteration^1^4&stickyix^1^null|20&1^invocation|30&1&benchmark&build;jdk-21-constant-side-metadata

If we use `static mut` for the base address, Rust needs to repetitively load from the variable every time we access side metadata. I inspected the assembly code of `ImmixSpace::trace_object`, and in around 100 instructions, there are 3 extra loads for this PR to get the side metadata address.

Intuitively `OnceLock` should be slower, as `get()` always checks if the value is initialized or not before loading it. However, if we use `oncelock.get().unwrap_unchecked()`, it seems the compiler can assume that there is no need to check or branch, and the value will not be changed so there is no need to repetitively load from it.

Using `OnceLock` seems to fix the slowdown on `biojava`, and I am running the experiments again for all the benchmarks.

qinsoon: This is the results for all the benchmarks. With OnceLock, the slowdown for STW time is 3.5% (compared to 4.3% with `static mut`).
https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|shrew-2026-02-26-Thu-225410&benchmark^build^invocation^iteration^stickyix&time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark&build;jdk-21-constant-side-metadata|41&Histogram%20(with%20CI)^build^benchmark&

Some benchmarks showed large error bars. The following is a rerun (which hasn't finished yet):
https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|shrew-2026-03-01-Sun-070849&benchmark^build^invocation^iteration^stickyix&time^time.other^time.stw&|10&iteration^1^4|20&1^invocation|30&1&benchmark&build;jdk-21-constant-side-metadata|41&Histogram%20(with%20CI)^build^benchmark&

Both seem suggest with OnceLock, we see slowdown for `batik`, but for other benchmarks, OnceLock is a performance improvement.

src/mmtk.rs
```diff
@@ -207,6 +213,12 @@ impl<VM: VMBinding> MMTK<VM> {
             },
         );
 
+        // The order here is important:
+        // Initialize side metadat sanity first
+        plan.verify_side_metadata_sanity();
+        // Then intiialize SFT because it may use side metadata
+        plan.initialize_sft();
```
qinsoon, line 218: These two lines use side metadata, so they have to happen after side metadata is initialized. They used to happen in `Plan::new()`. I extracted them, as at some point, I called `initialize_side_metadata()` after `Plan::new()`. So I had to move those two lines after `Plan::new()` and `initialize_side_metadata()`.

Now `initialize_side_metadata()` is called before `Plan::new()`. It is not necessary to have these two lines here. But I think this is still clearer.

src/util/metadata/side_metadata/layout.rs
qinsoon: I will still need to investigate the slowdown.

src/util/os/imp/unix_like/linux_like/linux_common.rs
```diff
@@ -120,6 +122,23 @@ pub fn dzmmap(
...
+pub fn dzmmap_anywhere(
+    size: usize,
+    align: usize,
+    strategy: MmapStrategy,
+    annotation: &MmapAnnotation<'_>,
+) -> Result<Address> {
+    let addr = unix_common::mmap_anywhere(
+        size,
+        align,
+        strategy.prot(MmapProtection::NoAccess).reserve(false),
```
wks: I think `.prot(MmapProtection::NoAccess).reserve(false)` should be orthogonal to `dzmamap_anywhere`.  it is `ChunkStateMmapper::quarantine_address_range_anywhere` that requires the map to be "no access" and "no reserve" because it is "quarantine".

src/util/metadata/side_metadata/layout.rs
```diff
@@ -0,0 +1,218 @@
...
+        };
+        info!(
+            "Quarantine side metadata range: total_bytes=0x{:x}, pages=0x{:x}, granularity=0x{:x}",
+            total_bytes,
+            pages,
+            MMAPPER.granularity()
+        );
+        let base = if specified_base.is_zero() {
+            MMAPPER
+                .quarantine_address_range_anywhere(pages, MmapStrategy::SIDE_METADATA, &anno)
```
wks: We probably don't need to pass `MmapStrategy::SIDE_METADATA` to `quarantine_address_range_anywhere` because when quarantining, it must have `MmapProtection::NoAccess` and `reserve: true`.

src/util/heap/layout/mmapper/csm/mod.rs
```diff
@@ -182,6 +182,37 @@ impl Mmapper for ChunkStateMmapper {
             })
     }
 
+    fn quarantine_address_range_anywhere(
+        &self,
+        pages: usize,
+        strategy: MmapStrategy,
```
wks: We don't need the strategy to be passed in.  Like the following excerpt from `quarantine_address_range`, we can construct the strategy locally:

```rust
                        let mmap_strategy = MmapStrategy::default()
                            .huge_page(huge_page_option)
                            .prot(MmapProtection::NoAccess)
                            .reserve(false)
                            .replace(false);
                        OS::dzmmap(group_start, group_bytes, mmap_strategy, anno)?;
```

We can probably add a constant in `impl MmapStrategy`:

```rust
                        pub const QUARANTINE = MmapStrategy::default()
                            .prot(MmapProtection::NoAccess)
                            .reserve(false)
                            .replace(false);
                        OS::dzmmap(group_start, group_bytes, mmap_strategy, anno)?;
```

so that we can directly use `MmapStrategy::QUARANTINE.huge_page(huge_page_option)`.

src/util/os/imp/unix_like/unix_common.rs
```diff
@@ -30,6 +31,23 @@ pub fn mmap(
...
+    let prot = strategy.prot.get_native_flags();
+    let flags = strategy.get_posix_mmap_flags(false);
+
+    let ptr = unsafe { libc::mmap(std::ptr::null_mut(), alloc_size, prot, flags, -1, 0) };
+    if ptr == libc::MAP_FAILED {
+        return Err(std::io::Error::last_os_error());
+    }
+
+    let start = Address::from_mut_ptr(ptr);
+    Ok(start.align_up(align))
```
wks: If we align up the start, the returned value is no longer the starting address of the mmap, and subsequent `munmap` may have problems.  For example, `ChunkStateMmapper::quarantine_address_range_anywhere` will try to call `munmap` with the returned address from `mmap`, and leave memory still mapped at the beginning and the end.

If the mmap is successful, we should either

-   immediately `munmap` the extraneous memory range in the beginning and in the end to make the resulting mmap exactly `start...(start+aligned_size)`, or
-   return the actual starting address and the `alloc_size`, and let the caller do the `munmap`.

src/util/metadata/side_metadata/layout.rs
qinsoon: I can't reproduce the same slowdown for `luindex` after turn on `work_packet_stats` and `perf_counter`: http://squirrel.anu.edu.au/plotty/yilin/mmtk/p/zPrqcq

I don't plan to spend more time on it.

k-sareen: `work_packet_stats` slows down GC by about 40% from what I recalled. So it's not surprising that the slowdown gets masked.

qinsoon: `work_packet_stats` alone should only slow down GC by a few percentage (~5% from my memory). 40% or so is only when you read perf counter per packet -- I was not doing that. I just enabled `perf_counter`, but didn't set any perf event to read.

qinsoon: The previous run that showed slowdown for `luindex` got `luindex`'s stw as
| | constant side metadata | dynamic side metadata (static mut) | dynamic side metadata (oncelock) |
|--|--|--|--|
|stw time (ms) | 62.259 | 63.421 | 66.629 |

My new run (and new builds) with `work_packet_stats` and `perf_counter` got stw as
| | constant side metadata | dynamic side metadata (static mut) | dynamic side metadata (oncelock) |
|--|--|--|--|
|stw time (ms) | 67.688 | 67.154 | 68.134 |

So everything got slower (which is expected), but the build for dynamic side metadata with oncelock is less affected. It seems to suggest the issue with oncelock is concurrency. `work_packet_stats` requires every packets to acquire locks to update the stats counters.

src/util/os/imp/unix_like/linux_like/linux_common.rs
qinsoon: Done.

src/util/heap/layout/mmapper/csm/mod.rs
qinsoon: I have changed `quarantine_address_range_anywhere` to have similar args as `quarantine_address_range` (i.e. taking an arg for `huge_page_option`, and no strategy arg).

src/util/os/imp/unix_like/unix_common.rs
qinsoon: Now `munmap` for those unaligned leading/trailing memory is unmapped in the function.

qinsoon:
> A high-level comment: `OnceLock` has a method `OnceLock::get_unchecked()` that can bypass the intermediate `Option<&T>` from `OnceLock::get()`. I haven't tried whether it is faster than `l.get().unwrap_unchecked()`, but it is worth trying.

This is a good point. And according to https://github.com/rust-lang/libs-team/issues/654, `get().unwrap_unchecked()` does not completely eliminate the atomic load in `get()`. However, we need to wait until `OnceLock::get_unchecked()` is stabilized. I added TODO in the code.

benches/regular_bench/bulk_meta/access.rs
```diff
@@ -0,0 +1,90 @@
...
+const BENCH_SPEC: SideMetadataSpec = SideMetadataSpec {
+    name: "bench",
+    is_global: true,
+    offset: 0,
+    log_num_of_bits: 0,
+    log_bytes_in_region: LOG_BYTES_IN_WORD as usize,
+};
+
+const NUM_ADDRS: usize = 4096;
+const DATA_STRIDE: usize = 64;
```
wks, line 25: To test random access, you can increase the stride size,  do something like `addr = base + (i * VERY_LARGE_STRIDE_SIZE) % VERY_LARGE_RANGE_SIZE`, or use a pseudo random number generator with a known seed.

src/util/metadata/side_metadata/global.rs
```diff
@@ -50,63 +50,58 @@ impl SideMetadataSpec {
     }
 
     /// Get the absolute offset for the spec.
-    pub const fn get_absolute_offset(&self) -> Address {
+    pub fn get_absolute_offset(&self) -> Address {
```
wks: I think the term "absolute offset" is confusing.  If it is absolute, it should be just an address, not offset.  In fact this function returns an `Address`.

I think the name came from the fact that the `SideMetadataSpec::offset` field was previously a union type, so we used the term "absolute offset" and "relative offset" to distinguish between them.  Now that the `SideMetadataSpec::offset` is always an offset from a certain base address, we should just abandon the name "absolute offset" and call it "starting address".


```suggestion
    pub fn starting_address(&self) -> Address {
```

src/util/metadata/side_metadata/global.rs
```diff
@@ -50,63 +50,58 @@ impl SideMetadataSpec {
...
         debug_assert!(self.is_absolute_offset());
-        unsafe { self.offset.addr }
+        let base =
+            crate::util::metadata::side_metadata::layout::global_side_metadata_base_address();
+        base + self.offset
     }
 
     /// Get the relative offset for the spec.
     pub const fn get_rel_offset(&self) -> usize {
```
wks: Similarly, we can just call it "offset".

```suggestion
    pub const fn offset(&self) -> usize {
```

src/util/heap/layout/mmapper/mod.rs
```diff
@@ -77,6 +77,21 @@ pub trait Mmapper: Sync {
         anno: &MmapAnnotation,
     ) -> MmapResult<()>;
 
+    /// Quarantine/reserve address range at any available address and return the base address.
+    /// The returned address is aligned to the mmapper's granularity.
+    ///
+    /// Arguments:
+    /// * `pages`: Number of pages to quarantine
+    /// * `strategy`: The mmap strategy.  The `prot` field is ignored because we always use
+    ///   `PROT_NONE`.
```
wks: This line in the doc is outdated.

src/util/metadata/side_metadata/global.rs
```diff
@@ -50,63 +50,58 @@ impl SideMetadataSpec {
...
     /// Return the upperbound offset for the side metadata. The next side metadata should be laid out at this offset.
     #[cfg(target_pointer_width = "64")]
-    pub const fn upper_bound_offset(&self) -> SideMetadataOffset {
+    pub const fn upper_bound_offset(&self) -> usize {
         debug_assert!(self.is_absolute_offset());
-        SideMetadataOffset {
-            addr: unsafe { self.offset.addr }
-                .add(crate::util::metadata::side_metadata::metadata_address_range_size(self)),
-        }
+        self.offset + crate::util::metadata::side_metadata::metadata_address_range_size(self)
```
wks: In this file, we have quite a few lines with fully qualified identifiers.  We may import them instead to improve readablilty.

src/util/metadata/side_metadata/global.rs
```diff
@@ -50,63 +50,58 @@ impl SideMetadataSpec {
...
 
     /// The upper bound address for metadata address computed for this global spec. The computed metadata address
     /// should never be larger than this address. Otherwise, we are accessing the metadata that is laid out
     /// after this spec. This spec must be a contiguous side metadata spec (which uses address
     /// as offset).
-    pub const fn upper_bound_address_for_contiguous(&self) -> Address {
+    pub fn upper_bound_address_for_contiguous(&self) -> Address {
         debug_assert!(self.is_absolute_offset());
-        unsafe { self.upper_bound_offset().addr }
+        self.get_absolute_offset()
```
wks: We could just use the `+` operator which is a synonym of `Address::add`.  There are a few other use cases in the same file, too.

src/util/metadata/side_metadata/global.rs
```diff
@@ -50,63 +50,58 @@ impl SideMetadataSpec {
     }
```
wks: These two functions are pointless now because each `SideMetadataSpec` always has an `offset` which is always relative to a certain base address (side meta base, or chunk start).  We can just use `uses_contiguous_side_metadata`.  In fact, in `src/util/metadata/side_metadata/sanity.rs`, we have the following lines:

```rs
    let metadata_addr_bound = if spec.is_absolute_offset() {
        spec.upper_bound_address_for_contiguous()
```

It clearly means "if spec is contiguous, call the function with `_contiguous`".

src/util/metadata/side_metadata/helpers.rs
```diff
@@ -403,7 +403,21 @@ mod tests {
...
     use crate::util::metadata::side_metadata::*;
 
+    fn should_skip_spec_on_this_target(spec: &SideMetadataSpec) -> bool {
+        use layout::LOG_GLOBAL_SIDE_METADATA_WORST_CASE_RATIO;
+        log_data_meta_ratio(spec) < LOG_GLOBAL_SIDE_METADATA_WORST_CASE_RATIO
+    }
+
     fn test_round_trip_conversion(spec: &SideMetadataSpec, test_data: &[Address]) {
+        core_test_initialize_side_metadata();
+        if should_skip_spec_on_this_target(spec) {
```
wks: The `should_skip_spec_on_this_target` is side-effect-free.  Should we move the `if` statement before `core_test_initialize_side_metadata();` so that we won't do the initialization if `should_skip_spec_on_this_target` returns true?

src/util/metadata/side_metadata/layout.rs
```diff
@@ -0,0 +1,225 @@
...
+
+// The following steps are needed before using side metadata.
+// The functions are intended as 'pub(super)'. We expect most people to use [`crate::util::metadata::side_metadata::initialize_side_metadata()`] instead, which calls these functions internally.
+
+// Step 1: Call `set_vm-side_metadata_specs()` to register VM side metadata layout. This is needed for the startup reservation to cover VM side metadata.
+
+/// Record VM side metadata layout so startup reservation can cover VM specs.
+/// This must be called before `initialize_side_metadata_base()`.
+pub(super) fn set_vm_side_metadata_specs(specs: &[SideMetadataSpec]) {
+    let _ = VM_SIDE_METADATA_UPPER_BOUND_OFFSET.get_or_init(|| {
```
wks: I think `VM_SIDE_METADATA_UPPER_BOUND_OFFSET` must be initialized exactly once.  `get_or_init` may fail silently if it is initialized twice.  We can use `OnceLock::set` and panic if the result is `Err`.  In this way, the error won't go undetected.

src/util/metadata/side_metadata/layout.rs
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// Step 2: Call `initialize_side_metadata_base()` to reserve address space for side metadata.
+
+/// Initialize the side metadata base address by reserving address space with quarantine mmap.
+pub(super) fn initialize_side_metadata_base(
+    specified_base: Address,
+    huge_page_support: HugePageSupport,
+) {
+    SIDE_METADATA_BASE_ADDRESS.get_or_init(|| {
```
wks: Similarly, we can use `OnceLock::set` here, too.

src/util/metadata/side_metadata/layout.rs
```diff
@@ -0,0 +1,225 @@
...
+                core_end,
+                unsafe { Address::from_usize(vm_end) }
+            );
+            if VM_SIDE_METADATA_UPPER_BOUND_OFFSET.get().is_none() {
+                warn!(
+                    "Initializing side metadata base before VM side metadata layout was registered"
+                );
+                let bt = std::backtrace::Backtrace::capture();
+                debug!("backtrace for early side metadata base initialization:\n{bt}");
+            }
```
wks: We called `VM_SIDE_METADATA_UPPER_BOUND_OFFSET.get()` several times.  But we called `unwrap` the first time, but attempted to detect if it is uninitialized later.  If it were uninitialized, it would have already panicked when calling `unwrap()`.


```suggestion
            let core_end = super::spec_defs::LAST_LOCAL_SIDE_METADATA_SPEC.upper_bound_offset();
            let Some(vm_end) = VM_SIDE_METADATA_UPPER_BOUND_OFFSET.get().copied() else {
                panic!(
                    "VM_SIDE_METADATA_UPPER_BOUND_OFFSET is not initialized.  \
                     We are trying to initialize side metadata base \
                     before VM side metadata layout was registered."
                );
            };
            info!(
                "Initializing side metadata base: core_end={} vm_end={}",
                core_end,
                unsafe { Address::from_usize(vm_end) }
            );
```

src/util/metadata/side_metadata/layout.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+        let total_bytes = side_metadata_reserved_bytes();
+        let pages = total_bytes >> LOG_BYTES_IN_PAGE;
+        let anno = MmapAnnotation::SideMeta {
+            space: "side-metadata",
+            meta: "all",
```
wks: This will be displayed in `/proc/<pid>/maps` as `"mmtk:sidemeta:side-metadata:all"`.  I think it looks better if it is `"mmtk:sidemeta:quarantined:quarantined"`, or `"mmtk:misc:quarantined"`.

Alternatively we can invent a new `MmapAnnotation` variant like this:

```rust
MmapAnnotation::Quarantined {
    kind: "metadata",
}
```

and make it display as `"mmtk:quarantined:metadata"`.

src/util/metadata/side_metadata/layout.rs
```diff
@@ -0,0 +1,225 @@
...
+            );
+        }
+
+        assert!(
+            SIDE_METADATA_BASE_ADDRESS.get().is_some(),
+            "global_side_metadata_base_address() called before side metadata base was initialized"
+        );
+    }
+
+    // TODO: use `OncLock::get_unchecked()` once it is stabilized.
```
wks:
```suggestion
    // TODO: use `OnceLock::get_unchecked()` once it is stabilized.
```

Also note that the third-party crate `oncelock` also has [the `OnceLock::get_unchecked()` method](https://docs.rs/oncelock/latest/oncelock/sync/struct.OnceLock.html#method.get_unchecked), and it is stable.  See:

benches/regular_bench/bulk_meta/access.rs
qinsoon: This cargo bench was just created to help early development. I don't intend to use it or tune it more for this PR. However, I kept it in the repo as it may be useful in the future, no matter for someone to directly benchmark it, or use it as a reference to write further benches. If requested, I can remove it from the repo.

src/util/metadata/side_metadata/global.rs
qinsoon: Done. I changed it to `get_starting_address`, and changed `get_rel_offset` to `get_offset_for_chunked`. `is_absolute_offset` and `is_rel_offset` are removed as well -- we directly use `uses_contiguous_side_metadata` and `uses_chunked_side_metadata`.

src/util/metadata/side_metadata/layout.rs
qinsoon: Sounds good. The function was using `get_or_init`, as for tests, the function may get called multiple times. Now I added a check for tests to skip calling these functions if side metadata is initialized by other tests.

src/util/metadata/side_metadata/layout.rs
qinsoon: I tried the suggestion. However, I don't feel it is clearer.
```rust
MmapAnnotation::Quarantined {
    kind: "metadata",
}
```

We already have `MmapAnnotation::Space` and `SideMetadata`, and `Quarantined` usually are generated from `Space`/`SideMetadata`. It would be the best if we do something like below, but it seems too complicated.
```rust
MmapAnnotation::Quarantined(Box<MmapAnnotation>)
```

I ended up with
```
MmapAnnotation::SideMeta {
            space: "all",
            meta: "all-quarantined"
}
```
It should be clear enough.

src/util/metadata/side_metadata/layout.rs
qinsoon: Typo fixed.

We may just wait for Rust to stabilize the API. There was only 1 version (`0.1) published for the third party crate. It does not feel like a serious project that someone would maintain overtime.

benches/regular_bench/bulk_meta/access.rs
wks: We can keep it, just in case someone wants to run it.

src/util/os/imp/unix_like/macos.rs
```diff
@@ -25,6 +25,19 @@ impl OSMemory for MacOS {
...
+    fn dzmmap_anywhere(
+        size: usize,
+        align: usize,
+        strategy: MmapStrategy,
+        _annotation: &MmapAnnotation<'_>,
+    ) -> Result<Address> {
+        unix_common::mmap_anywhere(
+            size,
+            align,
+            strategy.prot(MmapProtection::NoAccess).reserve(false),
```
wks: Why do we need to explicitly add `NoAccess` and "no reserve" on top of the `strategy`?  I think the caller knows whether it should have access or not.  The only call site of `dzmap_anywhere` is `quarantine_address_range_anywhere`, and it uses `MmapStrategy::QUARANTINE` which is already set to`NoAccess` and `reserve: false`.

I guess the code here is the stale code before a certain refactoring.

src/util/test_private/mod.rs
```diff
@@ -35,5 +36,20 @@ pub fn set_meta_bits(
     SideMetadataSpec::set_meta_bits(meta_start_addr, meta_start_bit, meta_end_addr, meta_end_bit)
 }
 
+/// Expose `address_to_meta_address` when running `cargo bench`.
+#[inline(always)]
+pub fn side_metadata_address_to_meta_address(
```
wks: It converts from data address to meta address, not from "side metadata address".  It's OK to just call it `address_to_meta_address` and the caller can distinguish this from the private `crate::util::metadata::side_metadata::address_to_meta_address` function using module names.

```suggestion
pub fn address_to_meta_address(
```

qinsoon: Done.

Issue #1450, opened 2026-02-20
Confusing mmap error


qinsoon:
We do mmap for both the space and its side metadata in
https://github.com/mmtk/mmtk-core/blob/d9b40daff7ab9bf25334e11c866814ddbee837f5/src/policy/space.rs#L185-L208

However, when we see an error, we call `handle_mmap_error` with the space address
https://github.com/mmtk/mmtk-core/blob/d9b40daff7ab9bf25334e11c866814ddbee837f5/src/policy/space.rs#L210

`handle_mmap_error` will then print `Failed to mmap {}, size {}` for the space address even if the failure is from mmapping the side metadata. This is very confusing.

The current `std::io::Error` does not include the error address. We should consider creating our own `Error` type and let `Error` to carry the address info.
1. Add `struct MmapError { error_address: Address, bytes: usize, ... }`
2. Change the return type of `dzmmap()` to `Result<Address, MmapError>`
3. `handle_mmap_error` no longer needs the `addr` and `bytes` arguments. It can get those from the error.

PR #1451, opened 2026-02-23
Introduce MmapResult

qinsoon:
This PR fixes https://github.com/mmtk/mmtk-core/issues/1450.

### DIFF
--- a/src/util/os/imp/unix_like/macos.rs
+++ b/src/util/os/imp/unix_like/macos.rs
@@ -11,9 +11,9 @@ impl OSMemory for MacOS {
         start: Address,
         size: usize,
         strategy: MmapStrategy,
-        _annotation: &MmapAnnotation<'_>,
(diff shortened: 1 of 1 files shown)


src/util/os/imp/unix_like/macos.rs
```diff
@@ -12,8 +12,8 @@ impl OSMemory for MacOS {
         size: usize,
         strategy: MmapStrategy,
         _annotation: &MmapAnnotation<'_>,
-    ) -> Result<Address> {
-        let addr = unix_common::mmap(start, size, strategy)?;
+    ) -> MmapResult<Address> {
+        let addr = unix_common::mmap(start, size, strategy, _annotation)?;
```
wks: We can remove the `_` name prefix because the `_annotation` parameter is no longer unused.

wks (approved):
LGTM, except for one very minor style problem.

PR #1452, opened 2026-02-25
Add CI script to check line ends

wks:
We add CI scripts to ensure that all text files

-   use UNIX line ends (LF only), and
-   have a newline character at the end of the file.

This PR is inspired by the scripts introduced in the OpenJDK binding in https://github.com/mmtk/mmtk-openjdk/pull/105, but also offers the ability to automatically fix line ends for the user:

```shell
./.github/scripts/ci-check-lineends.sh -f
```

This PR also fixes existing text files that have wrong line ends.

### DIFF
--- a/.github/scripts/ci-style.sh
+++ b/.github/scripts/ci-style.sh
@@ -4,7 +4,7 @@ export RUSTFLAGS="-D warnings -A unknown-lints"
 
 # --- Check line ends of text files ---
 
-if ! $project_root/.github/scripts/ci-check-lineends.sh -v; then
+if ! $project_root/.github/scripts/ci-check-lineends.sh; then
     echo "ERROR: Line ends check failed."
     exit 1
 fi
--- a/.github/scripts/ci-check-lineends.sh
+++ b/.github/scripts/ci-check-lineends.sh
@@ -12,6 +12,10 @@
 #
 #   ./.github/scripts/ci-check-lineends.sh -f
 #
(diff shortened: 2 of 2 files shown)


wks:
The `dos2unix --info=e` option is only available since version 7.5.2.

The `dos2unix --add-eol` option is only available since version 7.5.0.

But Ubuntu 22.04 (on which we run mmtk-core CI tests) only provides `dos2unix` 7.4.2, and Ubuntu 24.04 only provides 7.5.1.  This means we won't be able to implement line-end checking/fixing using dos2unix until we switch to the not-yet-released Ubuntu 26.04 image, or install dos2unix manually.

.github/scripts/ci-style.sh
```diff
@@ -2,6 +2,13 @@
 
 export RUSTFLAGS="-D warnings -A unknown-lints"
 
+# --- Check line ends of text files ---
+
+if ! $project_root/.github/scripts/ci-check-lineends.sh -v; then
```
qinsoon: I don't think we want `-v`. From the log in https://github.com/mmtk/mmtk-core/actions/runs/22472311227/job/65091763326?pr=1453, without `-v`, we will skip printing "Processing file" lines, and the output will only show the files that violate the check, which is much cleaner and useful.

wks: Yes.  That makes sense.  I just tried to mimic the style of our existing bash scripts which use `-x`.  I think it is cleaner to just print files with errors.

PR #1454, opened 2026-03-03
Support pinning/tpinning roots for concurrent Immix

qinsoon:
`PinningProcessEdges` for `ConcurrentImmixGCWorkContext` is set to `UnsupportedProcessEdges`. But concurrent marking does not move objects, so in theory there is no issue to support pinning process edges. This PR allows using pinning process edges for concurrent Immix.

### DIFF
--- a/src/scheduler/gc_work.rs
+++ b/src/scheduler/gc_work.rs
@@ -634,6 +634,12 @@ pub trait ProcessEdgesWork:
     ///
     /// `roots` indicates if we are creating a packet for root scanning.  It is only true when this
     /// method is called to handle `RootsWorkFactory::create_process_pinning_roots_work`.
+    ///
+    /// It normally returns `Some(work_packet)` and the `work_packet` should be added to the same
+    /// work bucket as `self`.  In some special cases, such as ConcurrentImmix, this function may
+    /// return `None`, which means the function has enqueued plan-specific object scanning work
+    /// packets that defer from `Self::ScanObjectsWorkType`.  In that case, there is no work packets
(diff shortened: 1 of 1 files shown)


src/scheduler/gc_work.rs
```diff
@@ -633,14 +634,16 @@ pub trait ProcessEdgesWork:
     ///
     /// `roots` indicates if we are creating a packet for root scanning.  It is only true when this
     /// method is called to handle `RootsWorkFactory::create_process_pinning_roots_work`.
-    fn create_scan_work(&self, nodes: Vec<ObjectReference>) -> Self::ScanObjectsWorkType;
+    fn create_scan_work(&self, nodes: Vec<ObjectReference>) -> Option<Self::ScanObjectsWorkType>;
```
wks, line 643: This is a significant change to the interface.  Although it is internal, I think we should take a note why we make the change here.

```suggestion
    ///
    /// It normally returns `Some(work_packet)` and the `work_packet` should be added to the same
    /// work bucket as `self`.  In some special cases, such as ConcurrentImmix, this function may
    /// return `None`, which means the function has enqueued plan-specific object scanning work
    /// packets that defer from `Self::ScanObjectsWorkType`.  In that case, there is no work packets
    /// for the caller to add.
    fn create_scan_work(&self, nodes: Vec<ObjectReference>) -> Option<Self::ScanObjectsWorkType>;
```

wks (approved):
LGTM, but we can add some comments so that we know what was going on when we look at the code later.

Issue #922, opened 2023-08-25
Remove Java-specific constants


wks:
TL;DR: `src/util/constants.rs` defines many Java-specific constants, such as `BYTES_IN_INT` and `BYTES_IN_LONG`.  Many other places in mmtk-core use those constants.  We should replace them with Rust counterparts, or more meaningful constants.

# Problem

MMTk was ported from JikesRVM written in Java, and Java integer types have precisely defined sizes.  The Java MMTk had constants that represent the sizes of those built-in integer types.  They are: `{,LOG_}{BITS,BYTES}_IN_{CHAR,SHORT,INT,LONG}` and `{MAX,MIN}_INT`.

When porting to Rust, we copied those constants over.  But those constants are foreign for Rust users because there is no `short`, `int` or `long` in Rust, but there are `i16`, `i32` and `i64` instead.  Moreover, as a language-agnostic GC framework, MMTk should not be Java-specific.  So it doesn't make sense to use those Java-specific constants (i.e. to depend on Java's semantics of `short`, `int` and `long`, etc.).

# Which parts are using those constants

We can let the compiler tell us which parts of mmtk-core are using those constants by adding `#[deprecated]` to those constants.

## src/plan/generational/barrier.rs

In `memory_region_copy_slow`, there is an assertion

```rs
            debug_assert_eq!(
                dst.bytes() & (BYTES_IN_INT - 1),
                0,
                "bytes should be a multiple of 32-bit words"
            );
```

It was `BYTES_IN_ADDRESS`, but was changed to `BYTES_IN_INT` in https://github.com/mmtk/mmtk-core/pull/899. It may be related to CompressedOops.

## src/plan/plan_constraints.rs

In `PlanConstraints::default()`, `max_non_los_default_alloc_bytes` and `max_non_los_copy_bytes` are set to `MAX_INT`.  They are both thresholds for deciding whether objects should go to LOS.  There are separate values for "alloc" and "copy" because when copying, object size may grow (for example, for implementing address-based hashing).

I think it is a mistake.  `MAX_INT` is not a meaningful value for anything

-   If it is intended to express "no limit" in the sense that the default space can hold objects of any sizes, it should have been `BYTES_IN_ADDRESS`.
-   If it is giving a reasonable default for plans, I don't think `MAX_INT` is a good default value because a 2GB object is too big for almost anything other than LOS.  A reasonable value should be something like "4 * block_size".

## src/policy/marksweepspace/malloc_ms/global.rs

[...]

PR #1456, opened 2026-03-06
Remove Java-specific constants.

wks:
The Rust MMTk was ported from JikesRVM which is a JVM.  Since the Rust MMTk is VM-agnostic, it should not depend on Java-specific constants.

Fixes: https://github.com/mmtk/mmtk-core/issues/922

### DIFF
--- a/src/plan/generational/barrier.rs
+++ b/src/plan/generational/barrier.rs
@@ -90,11 +90,6 @@ impl<VM: VMBinding, P: GenerationalPlanExt<VM> + PlanTraceObject<VM>> BarrierSem
         // Only enqueue array slices in mature spaces
         if !dst_in_nursery {
             // enqueue
-            debug_assert_eq!(
-                dst.bytes() & (std::mem::size_of::<i32>() - 1),
-                0,
-                "bytes should be a multiple of 32-bit words"
-            );
             self.region_modbuf.push(dst);
             self.region_modbuf
                 .is_full()
--- a/src/plan/plan_constraints.rs
+++ b/src/plan/plan_constraints.rs
@@ -55,10 +55,10 @@ impl PlanConstraints {
         PlanConstraints {
             collects_garbage: true,
             moves_objects: false,
-            // This size is copied from JikesRVM.
-            max_non_los_default_alloc_bytes: i32::MAX as _,
-            // This size is copied from JikesRVM.
-            max_non_los_copy_bytes: i32::MAX as _,
+            // No limit by default.  All objects can go to non-LOS space.
+            max_non_los_default_alloc_bytes: usize::MAX,
+            // No limit by default.  All objects can go to non-LOS space.
+            max_non_los_copy_bytes: usize::MAX,
             // As `LAZY_SWEEP` is true, needs_linear_scan is true for all the plans. This is strange.
             // https://github.com/mmtk/mmtk-core/issues/1027 tracks the issue.
             needs_linear_scan: crate::util::constants::SUPPORT_CARD_SCANNING
--- a/src/policy/marksweepspace/malloc_ms/global.rs
+++ b/src/policy/marksweepspace/malloc_ms/global.rs
@@ -272,7 +272,7 @@ impl<VM: VMBinding> crate::policy::gc_work::PolicyTraceObject<VM> for MallocSpac
 
 // Actually no max object size.
 #[allow(dead_code)]
-pub const MAX_OBJECT_SIZE: usize = i32::MAX as _;
+pub const MAX_OBJECT_SIZE: usize = usize::MAX;
 
 impl<VM: VMBinding> MallocSpace<VM> {
     pub fn extend_global_side_metadata_specs(specs: &mut Vec<SideMetadataSpec>) {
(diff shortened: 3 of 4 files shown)


src/plan/generational/barrier.rs
```diff
@@ -92,7 +91,7 @@ impl<VM: VMBinding, P: GenerationalPlanExt<VM> + PlanTraceObject<VM>> BarrierSem
         if !dst_in_nursery {
             // enqueue
             debug_assert_eq!(
-                dst.bytes() & (BYTES_IN_INT - 1),
+                dst.bytes() & (std::mem::size_of::<i32>() - 1),
```
wks: This place used to assert that the `dst.bytes()` is a multiple of `BYTES_IN_ADDRESS`.  PR https://github.com/mmtk/mmtk-core/pull/899 added support for compressed oops for OpenJDK.  It changed the alignment assertion to 32 bits because the array in a Java `Object[]` must be 32-bit aligned if compressed oops is enabled.  Otherwise it must be 64-bit aligned.

This assertion is OpenJDK-specific.  In general there is no requirement for the `src` or `dst` memory slices because the VMs can layout its reference arrays (or tagged reference arrays or struct arrays) arbitrarily.  We can safely remove this assertion in mmtk-core, and the mmtk-openjdk binding can do its own assertion in the wrapper code that calls into mmtk-core.

src/plan/plan_constraints.rs
```diff
@@ -55,8 +55,10 @@ impl PlanConstraints {
         PlanConstraints {
             collects_garbage: true,
             moves_objects: false,
-            max_non_los_default_alloc_bytes: MAX_INT,
-            max_non_los_copy_bytes: MAX_INT,
+            // This size is copied from JikesRVM.
+            max_non_los_default_alloc_bytes: i32::MAX as _,
+            // This size is copied from JikesRVM.
+            max_non_los_copy_bytes: i32::MAX as _,
```
wks: In JikesRVM, we used `MAX_INT` probably because we want a number that is so large that if there is no size limit on the non-LOS space, all objects will be allocated in the regular non-LOS space.  In 32-bit address spaces, `MAX_INT` is the largest object size we can allocate, and any size will be smaller than that.

If that's the case, we could have just used `usize::MAX` in the Rust MMTk.

src/policy/marksweepspace/malloc_ms/global.rs
```diff
@@ -272,7 +272,7 @@ impl<VM: VMBinding> crate::policy::gc_work::PolicyTraceObject<VM> for MallocSpac
 
 // Actually no max object size.
 #[allow(dead_code)]
-pub const MAX_OBJECT_SIZE: usize = crate::util::constants::MAX_INT;
+pub const MAX_OBJECT_SIZE: usize = i32::MAX as _;
```
wks: This is not really dead code.  The `plan::marksweep::global` module uses the `MAX_OBJECT_SIZE` constant, to set `PlanConstrants::max_non_los_default_alloc_bytes`.  Since the malloc MS space has no object size limit, it should be set to `usize::MAX`, too.

src/util/alloc/allocator.rs
```diff
@@ -158,7 +157,7 @@ pub fn align_allocation_inner<VM: VMBinding>(
     // Make sure MIN_ALIGNMENT is reasonable.
     #[allow(clippy::assertions_on_constants)]
     {
-        debug_assert!(VM::MIN_ALIGNMENT >= BYTES_IN_INT);
+        debug_assert!(VM::MIN_ALIGNMENT >= std::mem::size_of::<i32>());
```
wks, line 166: This line asserts the min alignment must be at least 4 bytes.

This line was introduced at the early stage of porting: https://github.com/mmtk/mmtk-core/commit/568a0d979163b39763ec1a539d35521b2f477c2b  The reason was unclear, but it is possible that some parts of the algorithm below that line assumes the min alignment is at least 4.

We need some explicit rules about the min and max alignments, and enforce them statically if possible.  For example, we may require `4 <= VM::MIN_ALIGNMENT` and `VM::MIN_ALIGNMENT <= VM::MAX_ALIGNMENT`.

src/util/raw_memory_freelist.rs
```diff
@@ -5,7 +5,7 @@ use crate::util::conversions;
 use crate::util::os::*;
 
 /** log2 of the number of bits used by a free list entry (two entries per unit) */
-const LOG_ENTRY_BITS: usize = LOG_BITS_IN_INT as _;
+const LOG_ENTRY_BITS: usize = i32::BITS.ilog2() as _;
```
wks, line 8: In the free list, each entry is a 32-bit int.  So `i32::BITS.ilog2()` is correct.

qinsoon (approved):
LGTM

src/plan/plan_constraints.rs
```diff
@@ -55,8 +55,10 @@ impl PlanConstraints {
         PlanConstraints {
             collects_garbage: true,
             moves_objects: false,
-            max_non_los_default_alloc_bytes: MAX_INT,
-            max_non_los_copy_bytes: MAX_INT,
+            // No limit by default.  All objects can go to non-LOS space.
+            max_non_los_default_alloc_bytes: usize::MAX,
```
qinsoon, line 59: We probably cast this value to wrong types in OpenJDK, which caused regression. E.g. here: https://github.com/mmtk/mmtk-openjdk/blob/b692c5a2f14883401519410a7acc4c64713af23e/openjdk/share/mmtkBarrierSetC2.cpp#L105 There might be more cases. @wks

PR #1457, opened 2026-03-11
Reintroduce GC disable/enable API

qinsoon:
This PR removes `Collection::is_collection_enabled`, and adds `memory_manager::enable_collection`, `memory_manager::disable_collection` and `memory_manager::is_collection_enabled`.

This PR also removes the return type `PauseRequestOutcome` of `try_request_pause`, and use `Result<(), GcStatus>` instead. This simplifies the code.

### DIFF
--- a/src/global_state.rs
+++ b/src/global_state.rs
@@ -404,15 +404,16 @@ impl GcStatusWord {
             .is_ok()
     }
 
-    /// `Disabled(depth)` -> `Disabled(depth - 1)`, or `Disabled(1)` -> `NotInGC`. Panics if
-    /// collection is not currently disabled. Returns `true` if this call actually re-enabled
-    /// collection (i.e. it was the outermost `Disabled(1)` -> `NotInGC` transition), `false` if
-    /// it only decremented the nesting depth.
+    /// `Disabled(depth)` -> `Disabled(depth - 1)`, or `Disabled(1)` -> `NotInGC`. If collection is
+    /// not currently disabled, this is a no-op (the status is left unchanged). Returns `true` if
+    /// this call actually re-enabled collection (i.e. it was the outermost `Disabled(1)` ->
+    /// `NotInGC` transition), `false` if it only decremented the nesting depth, or if collection
+    /// was already enabled.
     pub(crate) fn set_enabled(&self) -> bool {
         let old = self.transition(|status| match status {
             GcStatus::Disabled(1) => GcStatus::NotInGC,
             GcStatus::Disabled(depth) => GcStatus::Disabled(depth - 1),
-            _ => panic!("Trying to enable GC when it is not disabled"),
+            other => other,
         });
         old == GcStatus::Disabled(1)
     }
@@ -652,9 +653,18 @@ mod gc_status_tests {
     }
 
     #[test]
-    #[should_panic(expected = "not disabled")]
-    fn set_enabled_panics_if_not_disabled() {
-        GcStatusWord::new(GcStatus::NotInGC).set_enabled();
+    fn set_enabled_is_noop_if_not_disabled() {
+        for status in [
+            GcStatus::Uninitialized,
+            GcStatus::NotInGC,
+            GcStatus::InConcurrentGC,
+            GcStatus::InPause,
+            GcStatus::PauseRequested,
+        ] {
+            let word = GcStatusWord::new(status);
+            assert!(!word.set_enabled());
+            assert_eq!(word.load(), status);
+        }
     }
 
     #[test]
--- a/src/global_state.rs
+++ b/src/global_state.rs
@@ -254,6 +254,9 @@ impl GcStatus {
 pub(crate) enum PauseRequestOutcome {
     /// Collection is currently disabled; no pause was (or could be) requested.
     Disabled,
+    /// MMTk has not been initialized yet (`initialize_collection()` has not been called), so
+    /// there are no GC worker threads to run a collection; no pause was (or could be) requested.
+    Uninitialized,
     /// A pause was already requested (by this call or a racing one); the caller does not need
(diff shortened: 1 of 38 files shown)


src/global_state.rs
```diff
@@ -387,17 +387,58 @@ impl GcStatusWord {
...
+                let next = match Self::decode(bits) {
+                    GcStatus::Disabled(depth) => GcStatus::Disabled(depth + 1),
+                    GcStatus::NotInGC => GcStatus::Disabled(1),
+                    _ => return None,
+                };
+                Some(Self::encode(next))
+            })
+            .map(|old_bits| Self::decode(old_bits) == GcStatus::NotInGC)
+            .map_err(Self::decode)
+    }
```
wks: This code pattern is the same as `try_request_pause`.  We can extract the pattern into `try_transition` which, like `transition`, will handle the encoding and decoding of the bits.  It can greatly simplify the implementation of `set_disable` and `try_request_pause`.

I made the change on my branch. You can selectively merge it into this PR: https://github.com/mmtk/mmtk-core/commit/a0731f4de741e7f41441b47afc2d1e808f52ccfd

src/vm/tests/mock_tests/mock_test_wait_for_no_collection_in_progress.rs
```diff
@@ -0,0 +1,128 @@
...
+            // Thread B: as `disable_collection()`'s documentation instructs, retry (yielding
+            // between attempts) until it succeeds.
+            let thread_to_disable_gc = std::thread::spawn(move || {
+                while mmtk.disable_collection().is_err() {
+                    std::thread::yield_now();
+                }
+            });
+
+            // Give thread B a chance to run, then confirm it is still retrying.
+            std::thread::sleep(Duration::from_millis(200));
```
wks: This greatly depends on the OS scheduler, and the scheduling of GitHub runners is very unstable due to shared hosting.  We should use condition variables to wait for a synchronized flag instead.

I have made changes to this test case in my branch, and you can selective pull it into this PR: https://github.com/mmtk/mmtk-core/commit/1ba186813bcf157e72e4af83300f75059b89a278

wks (commented):
The algorithm is sound, but we can make the code of `set_disabled` and `try_request_pause` far more concise than the current state by extracting a `try_transition` function similar to the style of `fetch_update` and `transition`.  There is also a test case `disable_collection_fails_while_gc_in_progress` which relied on timers and yield-waiting, which is unreliable.  It can be made reliable by waiting for variable changes using Condvar.

I have made changes in my branch.  You can selectively adopt them into this PR.

branch: https://github.com/mmtk/mmtk-core/compare/master...wks:mmtk-core:review/reintroduce-gc-disable-api
-   `try_transition`: https://github.com/mmtk/mmtk-core/commit/a0731f4de741e7f41441b47afc2d1e808f52ccfd
-   Use condvar: https://github.com/mmtk/mmtk-core/commit/1ba186813bcf157e72e4af83300f75059b89a278

qinsoon:
I cherry-picked both commits into this PR.

src/util/heap/gc_trigger.rs
```diff
@@ -75,29 +74,61 @@ impl<VM: VMBinding> GCTrigger<VM> {
...
+    /// [`GCTrigger::enable_collection`].
+    pub fn disable_collection(&self) -> Result<bool, GcStatus> {
+        self.state.gc_status.set_disabled()
+    }
+
+    /// Re-enable collection. If collection is not currently disabled (e.g. there was no prior
+    /// matching call to [`GCTrigger::disable_collection`]), this is a no-op.
+    /// Returns `true` if this call actually re-enabled collection (i.e. it was the outermost
+    /// matching call), `false` if it only decremented the nesting depth, or if collection was
+    /// already enabled.
```
wks, line 122: The return value of calling this function without a pairing `disable_collection` is unspecified.  It neither re-enabled collection (i.e. previous state was `Disabled`) nor decremented the nest level (was already `NotInGC`).  The actual return value is false because `old == GcStatus::Disabled(1)` is false.

If `disable_collection` and `enable_collection` is used correctly, they should always be paired, so unpaired `enable_collection` is, in theory, an error.  I think we can `Err(old_state)` to indicate the error (i.e. if the old state is anything other than `Disabled(x)` where `x >= 1`).  `panic!()` is also OK.

I am uncertain whether `Result` or `panic!` is better.  Correct use of `enable_collection` should always be successful, whereas correct use of `disable_collection` still has to deal with races.  But `Result` can give the VM binding a chance to recover from error, and since `disable_collection` returns `Result`, `Err` makes the API symmetric.

qinsoon: The current behavior is allowing enable GC even if GC is not disabled (no previous disable_collection call). I think it is more flexible than requiring strict paring of enable/disable (as we are not sure what other runtimes would do). I can document the current behavior.

If you have other suggestions, let me know.

qinsoon: Actually the return value of enabling collection while collection is not disabled is already documented.
> `false` if it only decremented the nesting depth, or if collection was already enabled.

Initially, my implementation was requiring strict pairing of disable/enable, and there has to be a disable call before enable calls. But I saw panics from this in both Julia and Ruby. So I changed it to allow calling enable without previous disable calls.

About the return value, `bool` is not as expressive as `Result<bool, GcStatus>`. But it seems working fine. The runtime mostly cares whether the state has been changed or not. If we see `bool` is insufficient, we can change it later.

wks (approved):
LGTM

qinsoon:
@wks I removed the `binding-refs` comment so the auto merge will not be triggered. The binding PRs need to be merged manually.

PR #1460, opened 2026-03-16
Add an option to disable concurrent marking for concurrent immix

qinsoon:


src/plan/concurrent/immix/global.rs
```diff
@@ -86,18 +86,25 @@ impl<VM: VMBinding> Plan for ConcurrentImmix<VM> {
...
-        let allocated = used_pages_now.saturating_sub(used_pages_after_last_gc);
-        if !concurrent_marking_in_progress && allocated > threshold {
-            info!("Allocated {allocated} pages since last GC ({used_pages_now} - {used_pages_after_last_gc} > {threshold}): Do concurrent marking");
-            debug_assert!(
-                self.common.base.scheduler.work_buckets[WorkBucketStage::Concurrent].is_empty()
-            );
-            debug_assert!(!self.concurrent_marking_in_progress());
-            debug_assert_ne!(self.previous_pause(), Some(Pause::InitialMark));
-            return true;
+        if !*self
```
wks: I personally think short-circuiting early return is cleaner in terms of control flow.  But the current code is good enough.

```rust
if concurrent_marking_is_disabled() {
    return false;
}

// continue here
```

wks (approved):
LGTM.  I have a minor style suggestion, but the current code is good enough.

PR #1473, opened 2026-04-08
Fix infinite loop if we return from `Collection::out_of_memory`

k-sareen:
This commit also fixes a minor issue wherein we were using the unaligned object size when checking for OOM in the large object allocator.

### DIFF
--- a/src/util/alloc/allocator.rs
+++ b/src/util/alloc/allocator.rs
@@ -472,11 +472,6 @@ pub trait Allocator<VM: VMBinding>: Downcast {
             if !is_mutator {
                 debug_assert!(!result.is_zero());
                 debug_assert!(!self.get_context().thrown_oom.load(Ordering::Relaxed));
-                debug_assert!(!self
-                    .get_context()
-                    .state
-                    .allocation_success
-                    .load(Ordering::Relaxed));
                 return result;
             }
 
--- a/src/util/alloc/allocator.rs
+++ b/src/util/alloc/allocator.rs
@@ -248,7 +248,8 @@ pub(crate) fn assert_allocation_args<VM: VMBinding>(size: usize, align: usize, o
 /// used as `Arc<AllocatorContext>` inside all allocator implementations since
 /// we need the entire struct to be `Send`.
 ///
-/// See here for more information: https://github.com/mmtk/mmtk-core/issues/1474
+/// See doc comment on `impl Sync` for `AllocationOptionsHolder` above.
+/// See here for more information: <https://github.com/mmtk/mmtk-core/issues/1474>
 pub struct AllocatorContext<VM: VMBinding> {
     alloc_options: AllocationOptionsHolder,
     pub state: Arc<GlobalState>,
--- a/src/util/alloc/allocator.rs
+++ b/src/util/alloc/allocator.rs
@@ -287,6 +287,16 @@ impl<VM: VMBinding> AllocatorContext<VM> {
     }
 }
 
+macro_rules! reset_allocation_state {
+    ($receiver:expr) => {
+        {
+            let context = $receiver.get_context();
+            // Relaxed store is fine since this is a thread-local boolean.
+            context.thrown_oom.store(false, Ordering::Relaxed);
+        }
+    };
+}
+
 /// A trait which implements allocation routines. Every allocator needs to implements this trait.
 pub trait Allocator<VM: VMBinding>: Downcast {
     /// Return the [`VMThread`] associated with this allocator instance.
@@ -330,6 +340,12 @@ pub trait Allocator<VM: VMBinding>: Downcast {
             return true;
         }
         false
+    }
+
+    fn out_of_memory(&self, tls: VMThread) {
+        VM::VMCollection::out_of_memory(tls, AllocationError::HeapOutOfMemory);
+        // Relaxed store is fine since this is a thread-local boolean.
+        self.get_context().thrown_oom.store(true, Ordering::Relaxed);
     }
 
     /// An allocation attempt. The implementation of this function depends on the allocator used.
@@ -430,6 +446,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
         let tls = self.get_tls();
         let is_mutator = VM::VMActivePlan::is_mutator(tls);
         let stress_test = self.get_context().options.is_stress_test_gc_enabled();
+        assert!(!self.get_context().thrown_oom.load(Ordering::Relaxed), "We should not enter alloc_slow_inline if we have already thrown OOM for this allocation request.");
 
         // Information about the previous collection.
         let mut emergency_collection = false;
@@ -457,16 +474,14 @@ pub trait Allocator<VM: VMBinding>: Downcast {
 
             if !is_mutator {
                 debug_assert!(!result.is_zero());
+                debug_assert!(!self.get_context().thrown_oom.load(Ordering::Relaxed));
+                debug_assert!(!self.get_context().state.allocation_success.load(Ordering::Relaxed));
                 return result;
             }
 
             if !result.is_zero() {
                 // Report allocation success to assist OutOfMemory handling.
-                // Relaxed storing is fine since this is a thread-local boolean.
-                self.get_context()
-                    .thrown_oom
-                    .store(false, Ordering::Relaxed);
-                if !self
+				if !self
                     .get_context()
                     .state
                     .allocation_success
@@ -477,6 +492,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
                         .allocation_success
                         .store(true, Ordering::SeqCst);
                 }
+                reset_allocation_state!(self);
 
                 // Only update the allocation bytes if we haven't failed a previous allocation in this loop
                 if stress_test && self.get_context().state.is_initialized() && !previous_result_zero
@@ -526,6 +542,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
                 // the code beyond this point tests OOM conditions and, if not OOM, try to allocate
                 // again.  Since we didn't block for GC, the allocation will fail again if we try
                 // again. So we return null immediately.
+                reset_allocation_state!(self);
                 return Address::ZERO;
             }
 
@@ -534,9 +551,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
             if self.get_context().thrown_oom.load(Ordering::Relaxed) {
                 // Need to reset the thrown_oom state since we're giving up on this allocation,
                 // that is to say, the thrown_oom state is *per* allocation request
-                self.get_context()
-                    .thrown_oom
-                    .store(false, Ordering::Relaxed);
+                reset_allocation_state!(self);
                 return Address::ZERO;
             }
 
@@ -560,9 +575,8 @@ pub trait Allocator<VM: VMBinding>: Downcast {
                 if fail_with_oom {
                     // Note that we throw a `HeapOutOfMemory` error here and return a null ptr back to the VM
                     trace!("Throw HeapOutOfMemory!");
-                    VM::VMCollection::out_of_memory(tls, AllocationError::HeapOutOfMemory);
-                    // Relaxed store is fine since this is a thread-local boolean.
-                    self.get_context().thrown_oom.store(true, Ordering::Relaxed);
+                    self.out_of_memory(tls);
+                    reset_allocation_state!(self);
                     self.get_context()
                         .state
                         .allocation_success
--- a/src/util/alloc/allocator.rs
+++ b/src/util/alloc/allocator.rs
@@ -288,13 +288,11 @@ impl<VM: VMBinding> AllocatorContext<VM> {
 }
 
 macro_rules! reset_allocation_state {
-    ($receiver:expr) => {
-        {
-            let context = $receiver.get_context();
-            // Relaxed store is fine since this is a thread-local boolean.
-            context.thrown_oom.store(false, Ordering::Relaxed);
-        }
-    };
+    ($receiver:expr) => {{
+        let context = $receiver.get_context();
+        // Relaxed store is fine since this is a thread-local boolean.
+        context.thrown_oom.store(false, Ordering::Relaxed);
+    }};
 }
 
 /// A trait which implements allocation routines. Every allocator needs to implements this trait.
@@ -475,13 +473,17 @@ pub trait Allocator<VM: VMBinding>: Downcast {
             if !is_mutator {
                 debug_assert!(!result.is_zero());
                 debug_assert!(!self.get_context().thrown_oom.load(Ordering::Relaxed));
-                debug_assert!(!self.get_context().state.allocation_success.load(Ordering::Relaxed));
+                debug_assert!(!self
+                    .get_context()
+                    .state
+                    .allocation_success
+                    .load(Ordering::Relaxed));
                 return result;
             }
 
             if !result.is_zero() {
                 // Report allocation success to assist OutOfMemory handling.
-				if !self
+                if !self
                     .get_context()
                     .state
                     .allocation_success
--- a/src/util/alloc/allocator.rs
+++ b/src/util/alloc/allocator.rs
@@ -7,7 +7,7 @@ use crate::util::options::Options;
 use crate::MMTK;
 
 use std::cell::RefCell;
-use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use crate::policy::space::Space;
@@ -246,6 +246,7 @@ pub(crate) fn assert_allocation_args<VM: VMBinding>(size: usize, align: usize, o
 pub struct AllocatorContext<VM: VMBinding> {
     alloc_options: AllocationOptionsHolder,
     pub state: Arc<GlobalState>,
+    pub thrown_oom: AtomicBool,
     pub options: Arc<Options>,
     pub gc_trigger: Arc<GCTrigger<VM>>,
     #[cfg(feature = "analysis")]
@@ -257,6 +258,7 @@ impl<VM: VMBinding> AllocatorContext<VM> {
         Self {
             alloc_options: AllocationOptionsHolder::new(AllocationOptions::default()),
             state: mmtk.state.clone(),
+            thrown_oom: AtomicBool::new(false),
             options: mmtk.options.clone(),
             gc_trigger: mmtk.gc_trigger.clone(),
             #[cfg(feature = "analysis")]
@@ -452,6 +454,10 @@ pub trait Allocator<VM: VMBinding>: Downcast {
 
             if !result.is_zero() {
                 // Report allocation success to assist OutOfMemory handling.
+                // Relaxed storing is fine since this is a thread-local boolean.
+                self.get_context()
+                    .thrown_oom
+                    .store(false, Ordering::Relaxed);
                 if !self
                     .get_context()
                     .state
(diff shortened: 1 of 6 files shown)


src/util/alloc/allocator.rs
```diff
@@ -246,6 +246,7 @@ pub(crate) fn assert_allocation_args<VM: VMBinding>(size: usize, align: usize, o
 pub struct AllocatorContext<VM: VMBinding> {
     alloc_options: AllocationOptionsHolder,
     pub state: Arc<GlobalState>,
+    pub thrown_oom: AtomicBool,
```
qinsoon, line 257: Is it easy to use `bool` instead, given this flag is not shared across threads? For the functions that access the flag, they have `&mut` to the allocator.

qinsoon (commented):
Maybe add at least one test to test your case that the allocation size is large and `out_of_memory` is not throwing/panicking.

src/util/alloc/allocator.rs
k-sareen: Ah fair enough. I just auto-defaulted to `AtomicBool`. Will update the PR

k-sareen: That will require adding a `get_context_mut` function to every allocator. I can do it if you think it's cleaner.

qinsoon: Yeah. That sounds reasonable to me.

k-sareen: For some reason, `context` is an `Arc<AllocatorContext<VM>>` inside each allocator. So we can't get a mutable reference to it. Why is it even an `Arc`? This looks like a thread-local struct to me.

k-sareen: Perhaps it's not actually thread-local? Hence the `Arc`?

k-sareen: It's actually cloned! https://github.com/mmtk/mmtk-core/blob/06e55309fcb95fd9fb5ecd6856aa1ef2c621e3e5/src/util/alloc/allocators.rs#L123

That's very different from what I was expecting. The name `context` seemed to imply it's a thread-local struct.

qinsoon: I feel it is thread-local, and we should use `Rc` (instead of `Arc`).

k-sareen: Hmm... I don't have a good location to smuggle the boolean then. The mutator context is not accessible inside an allocator.

qinsoon: Yeah. Just use `AtomicBool`.

k-sareen: But if it's not actually thread-local, then using `AtomicBool` is also incorrect.

qinsoon: It is thread-local. 

Allocators exist in mutators and GC workers. We transfer `GCWorker` among threads (when we create a new worker thread, or when a worker thread surrenders). So `GCWorker` needs to be `Send`. `Rc` is not thread safe, so it is `!Send`. This is the reason we use `Arc`.

k-sareen: Okay. I think I will just copy-paste your comment because I think it's relevant documentation. Thank you for looking into it.

src/util/alloc/allocator.rs
```diff
@@ -495,6 +508,17 @@ pub trait Allocator<VM: VMBinding>: Downcast {
...
 
+            // If we have already thrown an OOM for this allocation then return a zero.
+            // Relaxed load and store is fine given this is a thread-local boolean.
+            if self.get_context().thrown_oom.load(Ordering::Relaxed) {
+                // Need to reset the thrown_oom state since we're giving up on this allocation,
+                // that is to say, the thrown_oom state is *per* allocation request
+                self.get_context()
+                    .thrown_oom
+                    .store(false, Ordering::Relaxed);
+                return Address::ZERO;
```
wks, line 549: This means `memory_manager::alloc` can now return `Address::ZERO` even if `AllocationOptions::at_safepoint == true`.  This is a change to the API between mmtk-core and the VM binding.

wks (commented):
If we urgently need to fix the bug, this PR does that.

But if we are not in a hurry, I would prefer a more thorough refactoring of both (a) the allocation API and (b) the internal contract between allocator components, including `alloc`, `alloc_slow`, `alloc_slow_once`, `alloc_slow_inline` and `Space::acquire`.

-   Currently the VM binding never expects `memory_manager::alloc` to return `Address::ZERO`.  After this PR, it is now possible to return `Address::ZERO`.  We also didn't expect `Collection::out_of_memory` to return, and this PR changes that, too.
-   `alloc_slow_once` can fail with more than one reasons, including page exhaustion and unsatisfiable request.  They need to be handled differently.

# alloc returning `Address::ZERO`

Now that `memory_manager::alloc` can return `Address::ZERO`, the VM binding needs to be aware of this and raise OOM error.  For example, OpenJDK should let the JIT-compiled code transfer the control flow and continue with an exception.

Our current code was ported from JikesRVM, and probably has assumed `Collection::out_of_memory` never returns.  This is also part of the API change.  I'll elaborate this in a new Issue, but returning `Address::ZERO` seems to be on the right track.  It needs to be done anyway to let the VM binding properly raise an exception to the application code.

# Internal contracts

When `alloc_slow_once` returns unsuccessfully, it may either mean

1.  We have gone all the way to the `Space::acquire` and it reported that it ran out of pages.  In this case, we need to perform a GC so that we have more spare pages to allocate into.  Or
2.  The request is obviously unreasonable and cannot be satisfied, so we should raise OOM error immediately and never try again.

Obviously, returning `Address::NULL` is not enough to distinguish between the two cases.  Therefore, this PR introduced `thrown_oom: AtomicBool` to distinguish between the two cases.

But there is one obvious idiomatic Rust solution: returning `Result<Address, AllocationError>` where `AllocationError` can be `PageExhausted` or `BadRequest`.  `alloc_slow_inline` can then match against the result (`Ok(addr)`, `Err(PageExhaust)` and `Err(BadRequest)` to decide what to do next. 

https://github.com/mmtk/mmtk-core/issues/1223 mentioned that some internal allocation calls should have returned `Option<Address>`.  We may extend it to a `Result` type.  I don't think the performance concern is relevant here because this happens in the slow path.

qinsoon:
I do not mind having this PR as an incremental step before a more complete overhaul. But as I mentioned in https://github.com/mmtk/mmtk-core/pull/1473#pullrequestreview-4072785721, we should add tests in this PR to verify the current behavior when returning from `out_of_memory`.

k-sareen:
> Currently the VM binding never expects memory_manager::alloc to return Address::ZERO. After this PR, it is now possible to return Address::ZERO. We also didn't expect Collection::out_of_memory to return, and this PR changes that, too.

But this is not true. `Collection:out_of_memory` *can* return and also we've been returning `Address::ZERO` for a while. I did this refactoring to support throwing OOM exceptions correctly in OpenJDK. Note the doc comment of `Collection::out_of_memory` in [mmtk-core](https://github.com/mmtk/mmtk-core/blob/06e55309fcb95fd9fb5ecd6856aa1ef2c621e3e5/src/vm/collection.rs#L66-L67) and [mmtk-openjdk](https://github.com/mmtk/mmtk-openjdk/blob/4037ffaa03eddf8d43be9f402f87c7d0a5bd13e0/openjdk/share/mmtkUpcalls.cpp#L150). The [documentation in `mmtk_alloc`](https://github.com/mmtk/mmtk-core/blob/06e55309fcb95fd9fb5ecd6856aa1ef2c621e3e5/src/memory_manager.rs#L165) also mentions that it can return an `Address::ZERO` if it's out of memory.

I think this is a straightforward bug given the short-circuiting and infinite loop, not a breaking change to the API. 

Original PR introducing `Collection::out_of_memory`: https://github.com/mmtk/mmtk-core/pull/535

We also discussed doing `Result<Address, Err>` in the PR there and I had a branch which did it at some point. I think the issue was that `Result` is not FFI compatible, so we would have to use `Option<Address>` instead.

src/util/alloc/allocator.rs
k-sareen: See my comment below.

k-sareen:
@qinsoon The `handle_obvious_oom_case` test is failing because I've moved the size alignment before the OOM check in the LOS allocator. This results in a different assertion being triggered. I think this is actually the correct thing to do because we need to check the actual size we will allocate, not the given size from the runtime. But I know that the Haskell folks depend on this current behavior. What do you think we should do in this case?

qinsoon:
> @qinsoon The `handle_obvious_oom_case` test is failing because I've moved the size alignment before the OOM check in the LOS allocator. This results in a different assertion being triggered. I think this is actually the correct thing to do because we need to check the actual size we will allocate, not the given size from the runtime. But I know that the Haskell folks depend on this current behavior. What do you think we should do in this case?

I think it is an issue with the tests -- the tests try to allocate objects of size `usize::MAX`, and MMTk expects the allocation size is aligned to `VM::MIN_ALIGNMENT`. You can just change the tests to align down `usize::MAX` to `VM::MIN_ALIGNMENT`, and call `alloc` with the aligned size.

k-sareen:
>  You can just change the tests to align down `usize::MAX` to `VM::MIN_ALIGNMENT`

Needs to be `usize::MAX - BYTES_IN_PAGE` to not overflow when doing addition. I've changed the test case to be the same.

```java
import java.util.ArrayList;
import java.util.Random;

class Test {
        Random r = new Random();

        int r1 = r.nextInt();
        ArrayList<Integer> arr = new ArrayList<>(Integer.MAX_VALUE - 2);
        arr.set(r1, r.nextInt());
        System.out.println(arr.get(r1));
    }
}
```

A simple Java test case like the above will get stuck in an infinite loop on master but correctly throw OOM in the PR. Since we panic in `Collection::out_of_memory` on master, we can never get into the infinite loop situation. How would you like me to create a test case for this @qinsoon?

qinsoon:
> Since we panic in `Collection::out_of_memory` on master, we can never get into the infinite loop situation. How would you like me to create a test case for this @qinsoon?

To create a test case that reproduces your original error:
1. `out_of_memory` does not throw. You can do it like this: https://github.com/mmtk/mmtk-core/blob/06e55309fcb95fd9fb5ecd6856aa1ef2c621e3e5/src/vm/tests/mock_tests/mock_test_allocate_no_gc_oom_on_acquire_allow_oom_call.rs#L14
2. allocate memory that is large enough to directly fail in the check `will_oom_on_acquire`.

We should expect 1. `out_of_memory` is called, 2. the test can finish (no infinite loop) and 3. the return value from alloc is 0.

k-sareen:
Ahh thank you. I never understood how to use the `MockVM`. ~~Unfortunately, the test case I've made using it does not have an infinite loop on master. But clearly there is an infinite loop given the Java test case above.~~ Nvm. I didn't add it to the `mod.rs` so it never actually ran. I've added a failing test now.

src/util/alloc/allocator.rs
```diff
@@ -495,6 +509,17 @@ pub trait Allocator<VM: VMBinding>: Downcast {
                 return Address::ZERO;
```
qinsoon, line 540: We need to clear `thrown_oom` here, right?

src/util/alloc/allocator.rs
qinsoon, line 578: `thrown_oom` is set, but not cleared before we return.

qinsoon (commented):
There seems to be a few cases where `thrown_oom` is cleared before returning.

I think we can set and clear the flags more systematically. See https://github.com/qinsoon/mmtk-core/commit/edcfb266440e918bd336e75c6874ddca269ac43f. The diff also includes other minor refactoring to move `handle_obvious_oom_request` from `Space` to `GCTrigger` -- I can open a PR for that separately.

src/util/alloc/allocator.rs
k-sareen: Ah right. That's true. Thank you. This approach seems a bit brittle since we have many return paths.

src/util/alloc/allocator.rs
k-sareen: Hm. This is actually a bit confusing. Do we want to set `thrown_oom` to `false` immediately after setting it to `true` above? I'm not sure how to deal with this.

k-sareen:
Your refactoring is good, maybe my only suggestion is we can convert the function into a macro since I don't really think we need a function there.

qinsoon:
> Your refactoring is good, maybe my only suggestion is we can convert the function into a macro since I don't really think we need a function there.

Yeah. Do whatever you think is necessary.

k-sareen:
~~Do we even need `allocation_success` anymore? I think if `previous_result_zero` is `true` then `allocation_success` is also `true`. So we can just do `fail_with_oom = previous_result_zero;` and the remove `allocation_success` entirely. It's a bit strange to have a global constant imo.~~ Ah this is used for determining whether we need to do an emergency collection or not.

k-sareen:
The tests are failing because I've added an assertion checking that `allocation_success` is always `true` if it's the GC worker performing an allocation.

qinsoon:
> The tests are failing because I've added an assertion checking that `allocation_success` is always `true` if it's the GC worker performing an allocation.

`allocation_success` is a global state. It is not bound to one mutator or one worker.

src/util/alloc/allocator.rs
```diff
@@ -277,6 +287,14 @@ impl<VM: VMBinding> AllocatorContext<VM> {
     }
 }
 
+macro_rules! reset_allocation_state {
```
wks: We can use a function instead of macro. We can also be a method of the `Allocator` trait if we don't mind making it publicly visible.

```diff
diff --git a/src/util/alloc/allocator.rs b/src/util/alloc/allocator.rs
index 21d09db269..e0ff26a6c5 100644
--- a/src/util/alloc/allocator.rs
+++ b/src/util/alloc/allocator.rs
@@ -287,12 +287,10 @@ impl<VM: VMBinding> AllocatorContext<VM> {
     }
 }
 
-macro_rules! reset_allocation_state {
-    ($receiver:expr) => {{
-        let context = $receiver.get_context();
-        // Relaxed store is fine since this is a thread-local boolean.
-        context.thrown_oom.store(false, Ordering::Relaxed);
-    }};
+fn reset_allocation_state<VM: VMBinding, A: Allocator<VM> + ?Sized>(allocator: &A) {
+    let context = allocator.get_context();
+    // Relaxed store is fine since this is a thread-local boolean.
+    context.thrown_oom.store(false, Ordering::Relaxed);
 }
 
 /// A trait which implements allocation routines. Every allocator needs to implements this trait.
@@ -488,7 +486,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
                         .allocation_success
                         .store(true, Ordering::SeqCst);
                 }
-                reset_allocation_state!(self);
+                reset_allocation_state(self);
 
                 // Only update the allocation bytes if we haven't failed a previous allocation in this loop
                 if stress_test && self.get_context().state.is_initialized() && !previous_result_zero
@@ -538,7 +536,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
                 // the code beyond this point tests OOM conditions and, if not OOM, try to allocate
                 // again.  Since we didn't block for GC, the allocation will fail again if we try
                 // again. So we return null immediately.
-                reset_allocation_state!(self);
+                reset_allocation_state(self);
                 return Address::ZERO;
             }
 
@@ -547,7 +545,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
             if self.get_context().thrown_oom.load(Ordering::Relaxed) {
                 // Need to reset the thrown_oom state since we're giving up on this allocation,
                 // that is to say, the thrown_oom state is *per* allocation request
-                reset_allocation_state!(self);
+                reset_allocation_state(self);
                 return Address::ZERO;
             }
 
@@ -572,7 +570,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
                     // Note that we throw a `HeapOutOfMemory` error here and return a null ptr back to the VM
                     trace!("Throw HeapOutOfMemory!");
                     self.out_of_memory(tls);
-                    reset_allocation_state!(self);
+                    reset_allocation_state(self);
                     self.get_context()
                         .state
                         .allocation_success
```

k-sareen: Done

src/vm/tests/mock_tests/mock_test_issue867_allocate_unrealistically_large_object.rs
```diff
@@ -37,7 +37,13 @@ pub fn allocate_max_size_object() {
     with_mockvm(
         default_setup,
         || {
-            let (size, align) = (usize::MAX, 8);
+            use crate::util::constants::BYTES_IN_PAGE;
+
+            let align = 8;
+            // Maximum size we allow for allocation is usize::MAX - BYTES_IN_PAGE aligned down to
+            // page size. This will panic with an 'attempt to add with overflow' otherwise.
```
wks, line 44: It probably panics with "attempt to add with overflow" because MarkCompact space will add per-object overhead before the object payload.

```rust
impl<VM: VMBinding> Allocator<VM> for MarkCompactAllocator<VM> {
    fn alloc(&mut self, size: usize, align: usize, offset: usize) -> Address {
        let rtn = self
            .bump_allocator
            .alloc(size + Self::HEADER_RESERVED_IN_BYTES, align, offset);
        // ...
    }
}
```

We may consider it a separate problem because `MarkCompactgAllocator` should really call `Collection::out_of_memory` and return `Address::ZERO` if `size + Self::HEADER_RESERVED_IN_BYTES` caused overflow.  This also means the handling of "obvious OOM requests` can be made earlier.  We can leave it to another day if it is not easy to fix here.

k-sareen: Yes let's do it in a separate PR

src/util/alloc/allocator.rs
```diff
@@ -463,6 +486,7 @@ pub trait Allocator<VM: VMBinding>: Downcast {
                         .allocation_success
                         .store(true, Ordering::SeqCst);
                 }
+                reset_allocation_state(self);
```
wks: Instead of resetting allocation state, we should assert that `thrown_oom` is false because the allocation is successful.  It can't both throw OOM and be successful.

k-sareen: ~~It needs to be reset in case the OOM is caught and handled by the runtime.~~ Oh sorry I mistook the context. Right. I personally don't think this is a big issue since it's a thread-local boolean without any atomic synchronization. Would you rather I just make it an assertion?

wks: Yes.  I think an assertion is appropriate.  As mentioned in https://github.com/mmtk/mmtk-core/pull/1473#pullrequestreview-4073622166, the `thrown_oom` is just a way to distinguish between OOM and retrying after GC.  As I mentioned in the "proposed refactoring" section in https://github.com/mmtk/mmtk-core/issues/1475, all code paths that return `InternalAllocationFailure::BadRequest` from `alloc_slow_once` should result in the `alloc_slow_inline` function returning immediately.  The line at this comment is on a "good path", i.e. allocation is successful, so `thrown_oom` must not have been set.

k-sareen: Done.

src/util/alloc/allocator.rs
wks: Yes.  It is sound to add `reset_allocation_state` here.  No change is needed because it is sound, but you can do the following if you think appropriate.

If we swapped the order of the next `if` block

```rs
if self.get_context().thrown_oom.load(Ordering::Relaxed) {
    // ...
}
```

with this one

```rs
if !self.get_context().get_alloc_options().at_safepoint {
    // ...
}
```

we would not have needed to add `reset_allocaiton_state` here.  As we discussed in https://github.com/mmtk/mmtk-core/issues/1475#proposed-refactoring, `alloc_slow_once` can return four mutually exclusive results:

-   Success. (`!result.is_zero()` here, or `Ok(result)` as I proposed)
-   Obvious OOM requests. (`thrown_oom == true` here, or `Err(BadRequest)` as I proposed)
-   Need GC but not at safepoint. (`at_safepoint` here, or `Err(WouldBlock)` as I proposed)
-   Should retry. (the rest of the code here, or `Err(Retry)` as I proposed)

So `thrown_oom` should only be set in the second case, but should be `false` for all the other cases.

k-sareen: Right. I think let's try to get this merged as-is if there are no issues with it.

wks (approved):
LGTM.

PR #1476, opened 2026-04-09
Move a few misplaced functions about obvious OOM handling

qinsoon:
These functions are clearly misplaced:
* `will_oom_on_acquire`: this function only needs `GCTrigger`. This PR moves it from `Space` to `GCTrigger`.
* `handle_obvious_oom_request`: this function is only used by `Allocator` and also takes `AllocationOptions` as an argument. This PR moves it from `Space` to `Allocator`.


wks (approved):
LGTM.

CI indicates API changes, but that is because we exposed the `Allocator` type to the VM binding for embedding it into Mutator.  I don't think it is really an API-breaking change.

PR #1480, opened 2026-04-17
Add a warning about unwinding.

wks:
In the doc comment of `Collection::out_of_memory`, we added a warning about implementing out-of-memory exceptions by unwinding the stack from within `Collection::out_of_memory`, and recommended returning from it instead.

### DIFF
--- a/src/vm/collection.rs
+++ b/src/vm/collection.rs
@@ -71,6 +71,47 @@ pub trait Collection<VM: VMBinding> {
     /// Arguments:
     /// * `tls`: The thread pointer for the mutator which failed the allocation and triggered the OOM.
     /// * `err_kind`: The type of OOM error that was encountered.
+    ///
+    /// # Warnings about stack unwinding
+    ///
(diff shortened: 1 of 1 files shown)


src/vm/collection.rs
```diff
@@ -71,6 +71,47 @@ pub trait Collection<VM: VMBinding> {
     /// Arguments:
     /// * `tls`: The thread pointer for the mutator which failed the allocation and triggered the OOM.
     /// * `err_kind`: The type of OOM error that was encountered.
+    ///
+    /// # Warnings about stack unwinding
```
qinsoon, line 75: All of these are for 'Heap OOM'. Make sure this is clear. For critical OOM (mmap OOM), we expect the function to abort the VM and not return.

qinsoon (approved):
LGTM

src/vm/collection.rs
wks: Yes.  I mentioned "... when the heap is out of memory" in the first sentence.

PR #1484, opened 2026-04-21
Create a slice from raw backing memory for RawMemoryFreeList

caizixian:
Reduce the unsafe to one place that is easy to verify

### DIFF
--- a/src/util/raw_memory_freelist.rs
+++ b/src/util/raw_memory_freelist.rs
@@ -25,6 +25,7 @@ pub struct RawMemoryFreeList {
     current_units: i32,
     pages_per_block: i32,
     strategy: MmapStrategy,
+    slice: &'static mut [i32],
 }
 
 impl FreeList for RawMemoryFreeList {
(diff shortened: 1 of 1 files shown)


k-sareen (approved):
LGTM. Thanks!

PS: Do we need a `SAFETY` comment on L118?

src/util/raw_memory_freelist.rs
```diff
@@ -111,6 +115,9 @@ impl RawMemoryFreeList {
             current_units: 0,
             pages_per_block,
             strategy,
+            slice: unsafe {
+                std::slice::from_raw_parts_mut(std::ptr::NonNull::<i32>::dangling().as_ptr(), 0)
```
wks: We don't really need `dangling()`.  Technically, when a RawMemoryFreelist is created, its address range is `base..base`, a zero-sized slice starting at `base`.

caizixian:
@k-sareen @wks please take another look and merge if you are happy with the change.

k-sareen (approved):
LGTM. Thanks!

src/util/raw_memory_freelist.rs
```diff
@@ -35,21 +36,24 @@ impl FreeList for RawMemoryFreeList {
...
-        let offset = (index << LOG_BYTES_IN_ENTRY) as usize;
-        debug_assert!(self.base + offset >= self.base && self.base + offset < self.high_water);
-        unsafe { (self.base + offset).load() }
+        let len = (self.high_water - self.base) >> LOG_BYTES_IN_ENTRY;
+        assert!(
+            (index as usize) < len,
+            "index out of bounds: the len is {} but the index is {}",
+            len,
+            index
+        );
```
wks: If we use slice, the length is just `self.slice.len()` because we set the length to that value in `grow_freelist`.  And the expression `self.slice[index as usize]` already includes a bounds check.  Never the less, it is still OK to do an assertion to see if the `high_water`, the `base` and the `slice.len()` are consistent.


```suggestion
        #[cfg(debug_assertions)]
        {
            let len = (self.high_water - self.base) >> LOG_BYTES_IN_ENTRY;
            debug_assert_eq!(
                len,
                self.slice.len(),
                "Length does not match.  \
                high_water: {h}, base: {b}, computed len: {len}, \
                slice len: {sl}",
                h = self.high_water,
                b = self.base,
                sl = self.slice.len()
            );
        }
```

src/util/raw_memory_freelist.rs
```diff
@@ -35,21 +36,24 @@ impl FreeList for RawMemoryFreeList {
...
-            offset,
-            self.base + offset,
-            self.high_water
+        let len = (self.high_water - self.base) >> LOG_BYTES_IN_ENTRY;
+        assert!(
+            (index as usize) < len,
+            "index out of bounds: the len is {} but the index is {}",
+            len,
+            index
         );
```
wks: This is similar

PR #1483, opened 2026-04-21
Replace MaybeUninit in allocator arrays with Option

caizixian:
I don't think the changes here are performance sensitive but I can run some benchmarks if needed

Issue #1318, opened 2025-05-11
Remove repr(C) from Mutator


wks:
The reason why the `Mutator` struct was made `#[repr(C)]` was to make it possible for bindings to embed `Mutator` into their own C/C++ structs in thread-local storage.  This is supposed to improve performance by allowing inlined fast-path code to access important fields (such as `BumpPointer`) quickly from the TLS.  However, embedding the entire `Mutaotr` is neither practical nor necessary.

It is impractical because

-   Because the `Mutator` struct is so big, it is beyond the addressable range of immediate addressing in some architectures, such as RISC-V.  There is already [a proposal to make `Mutator` smaller](https://github.com/mmtk/mmtk-core/issues/1001).
-   The C/C++ part of the VM binding has to precisely replicate the `Mutator` struct.  This is tedious and error-prone, but both the [OpenJDK](https://github.com/mmtk/mmtk-openjdk/blob/056d331be285d994881b848e2e14cb2cc6af2ea1/openjdk/mmtkMutator.hpp) and the [Julia](https://github.com/mmtk/mmtk-julia/blob/30ac651f284bbc80d404968672b7f7d6aa278551/mmtk/api/mmtkMutator.h) binding have done that.  And the presence of Rust-specific data types, such as `&dyn T`, makes it hard to get the object layout right.
-   To make it `repr(C)`, the `Allocators` sub-struct has to rely on fixed-size arrays `[T; N]`.  By doing this, the number of allocators of each kind are limited and unknown at run time.
    -   Changing the maximum number of each allocator will alter the layout of `Mutator`, resulting in a series of cascading changes in VM bindings that depend on the precise layout, such as [this PR](https://github.com/mmtk/mmtk-core/pull/1308) and its associated changes in the [OpenJDK](https://github.com/mmtk/mmtk-openjdk/pull/308) and [Julia](https://github.com/mmtk/mmtk-julia/pull/238) bindings.
    -   Because the *actual* number of each kind of allocator is determined by the concrete Plan, the `Allocators` struct has to use `MaybeUninit` (like `[MaybeUninit<BumpAllocator<VM>>; MAX_BUMP_ALLOCATORS]`) so that array elements can remain uninitialized if the concrete plan doesn't need that many allocators of that kind.  This is also the reason why the `Allocators::get_allocator` and `Allocators::get_allocator_mut` have to be made `unsafe`.

[...]

PR #1486, opened 2026-04-22
Remove unsafety for uninitialized allocatr. Initialize with unusable space instead.

qinsoon:
Inspired by https://github.com/mmtk/mmtk-core/pull/1483.
* Initialize all the allocators with unusable space first.
* Then initialize the allocators that a plan may use with proper spaces.
* Refactor allocators so all allocators reference a `&dyn Space`. This requires changing bindings, as the layout of allocators are changed.

### DIFF
--- a/src/plan/compressor/mutator.rs
+++ b/src/plan/compressor/mutator.rs
@@ -61,13 +61,11 @@ pub fn create_compressor_mutator<VM: VMBinding>(
 
 pub fn compressor_mutator_release<VM: VMBinding>(mutator: &mut Mutator<VM>, tls: VMWorkerThread) {
     // reset the thread-local allocation bump pointer
-    let bump_allocator = unsafe {
-        mutator
-            .allocators
-            .get_allocator_mut(mutator.config.allocator_mapping[AllocationSemantics::Default])
-    }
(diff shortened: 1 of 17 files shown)


qinsoon:
I will make binding side changes if we think this is a good idea.

k-sareen:
I think this solution is preferred over #1483 since we need to ensure that `MutatorContext` is FFI-safe in case bindings want to embed it entirely.

wks:
> I think this solution is preferred over #1483 since we need to ensure that `MutatorContext` is FFI-safe in case bindings want to embed it entirely.

Alternatively, we can first investigate whether disallowing embedding `Mutator` in favor for embedding only `BumpPointer` can keep the performance good in OpenJDK.  If that's possible, we can refactor the `Mutator` even more aggressively.

See:

-   https://github.com/mmtk/mmtk-core/issues/1318
-   This comment mentioned OpenJDK performance. https://github.com/mmtk/mmtk-core/issues/1318#issuecomment-2870738079

qinsoon:
https://github.com/mmtk/mmtk-core/pull/1483 seems good to me. If it is merged, we can close this PR.

PR #1487, opened 2026-04-22
Add bytes parameter to initialize_object_metadata

wenyuzhao:
Pass the object allocation size to `initialize_object_metadata` so that spaces can initialize metadata for the object's entire address range. For example, LXR uses it to initialize per-page side metadata for LOS objects.


qinsoon (approved):
LGTM

wks (approved):
LGTM

Issue #1346, opened 2025-07-21
Quarantine the entire heap memory range


wks:
Currently, in both Map32 and Map64, we assume MMTk is in control of the entire memory range from `heap_start` to `heap_end`.  But if the VM uses third-party libraries, they may call `mmap` and the OS will return a random memory range, which could possibly be within the heap range.  This actually happened in some VMs, such as ART.

Although we currently have `Mmapper::quarantine_address_range`, we only use it for side metadata, but not the memory for objects themselves.  On ART, we quarantine the object spaces, too, not just metadata. (@k-sareen, correct me if I am wrong.)  We have not seen problems on OpenJDK, but in theory it could happen, too.  In JikesRVM, the ["malloc returned something that is in RVM address space"](https://github.com/mmtk/mmtk-jikesrvm/issues/158) error could possibly be caused by malloc getting mmap memory within RVM address space, although we haven't confirmed it, yet.

I suggest mmtk-core should quarantine the entire MMTk heap address range at start-up.  For Map64, it is 0x200_0000_0000 to 0x2200_0000_0000, or whatever the VM binding configures at set-up time.  It will help ART, and other VMs can avoid potential problems not yet observed.

Because we quarantine the entire heap, it is no longer needed to have the `MapState::Quarantined` state because it is always quarantined as long as it is within the heap range.

qinsoon:
> Because we quarantine the entire heap, it is no longer needed to have the MapState::Quarantined state because it is always quarantined as long as it is within the heap range.

Better still keep it. My gut feeling is that there will be cases where we don't want to/can't quarantine addresses.

k-sareen:
Why do you need to quarantine the entire address range? You only need to quarantine the address range you intend on using. Like Yi said, that's bounded by the max heap size. That's what I'm doing in ART. It doesn't make sense to me to quarantine everything.

Issue #1482, opened 2026-04-21
Dynamic side metadata range may overlap with heap range.


wks:
If the side metadata range is picked at random, it may overlap with the heap range.

Currently, `MMTK::new` eagerly maps the side metadata memory.

https://github.com/mmtk/mmtk-core/blob/cafcece9f3753c4ad3c13077c2eb12b7493d470b/src/mmtk.rs#L167

By default, the range of the side metadata is picked at random.  Currently, we don't quarantine the heap memory range (0x0000_0200_0000_0000usize to 0x0000_2200_0000_0000usize).  Therefore, there is a chance that the randomly picked side metadata range will overlap with the heap memory range.

The heap memory is mapped on demand in `Space::get_new_pages_and_initialize`.

https://github.com/mmtk/mmtk-core/blob/cafcece9f3753c4ad3c13077c2eb12b7493d470b/src/policy/space.rs#L150-L163

If the memory is already mapped, it will result in mmap error, and will call `OS::handle_mmap_error`.  It will call `Collection::out_of_memory` and it is supposed to terminate the VM immediately.

I tried running OpenJDK with `-XX:-UseCompressedOops` (otherwise the heap will be in the lowest 35-bit address range and will be very unlikely to have conflict) with GenCopy (which has the most number of spaces), and here are the address ranges from the log:

```
[2026-04-21T05:56:19Z INFO  mmtk::util::metadata::side_metadata::layout] Side metadata base initialized at 0x7456b4c00000 (range: 0x7456b4c00000 - 0x7f0338c00000)
```

```
[2026-04-21T05:57:09Z INFO  mmtk::util::metadata::side_metadata::layout] Side metadata base initialized at 0x750278c00000 (range: 0x750278c00000 - 0x7faefcc00000)
```

```
[2026-04-21T06:02:20Z INFO  mmtk::util::metadata::side_metadata::layout] Side metadata base initialized at 0x7468b4c00000 (range: 0x7468b4c00000 - 0x7f1538c00000)
```

And the spaces are always in the same ranges

```
[2026-04-21T06:02:20Z DEBUG mmtk::policy::space] Created space copyspace0 [0x20000000000, 0x40000000000) for 2199023255552 bytes
[2026-04-21T06:02:20Z DEBUG mmtk::policy::space] Created space copyspace1 [0x40000000000, 0x60000000000) for 2199023255552 bytes
[...]

Issue #1407, opened 2025-10-23
Optimize bulk metadata operations for Region


wks:
Currently, we have `SideMetadataSpec::bset_metadata` and `SideMetadataSpec::bzero_metadata`.  They are general.  They take care of corner cases such as the starting or ending bits are in the middle of a metadata byte.  This complicated the issue and we developed complex algorithms (e.g. https://github.com/mmtk/mmtk-core/pull/1181) to deal with the general cases.

However, we often deal with side metadata for `Region`.  A `Region` is power of two in size, and is aligned to its size.  A corollary of this is that **any side metadata of a `Region` must be aligned to its own size (in bits), provided that the granularity of the side metadata is finer than the `Region`**.

For example, an Immix line is a 256-byte region, and the VO bit side metadata is one bit per 8 bytes.  The VO bits for an Immix line is therefore 32 bits in size.  We can prove that the side metadata range (in terms of the bits offset from the side metadata base address) must also be a multiple of 32 bits.  As long as the side metadata base address is 4-byte aligned, the metadata for a `Region` must also be an aligned `u32`.

This gives us many interesting properties.  For example, bulk-clearing the VO bit side metadata for an Immix line should be, in theory, just one 32-bit store instruction.  There should be no loops or boundary checks involved.

# Implementation

The key is to persuade Rust to consider both the side metadata spec and the Region size as constant.  Things like `VO_BITS.bzero_metadata::<Line>(line)` should be expanded to one 32-bit assignment with some not-too-complicated address computation.

To avoid unnecessary address computation, if we are, for example, modifying multiple Immix lines in one Immix block, we know their VO bits metadata must be adjacent, too.  We may iterate through all the 32-bit VO bit metadata words for an Immix block while we sweep all lines in a block.  We only need to add 8 to the metadata address when we move to the next line, and there is no need to handle discontiguous metadata in this case because a `Block` must be within a `Chunk`.

Issue #1370, opened 2025-08-14
[meta] Refactoring related to Mmapper and page sizes


wks:
We recently discussed some changes related to page sizes and mmapper.  This issue keeps a check list for those issues and their dependencies.

-   [ ] Refactor `Mmapper` to combine `ByteMapMmapper` and `TwoLevelMmapper`.
    -   https://github.com/mmtk/mmtk-core/pull/1369
    -   This is a software-enginnering change and does not change the semantics of `Mmapper`.
    -   This tidies up the code and should make it easier to make other changes.
-   [ ] Remove the `Protected` state and dead methods related to it.
    -   https://github.com/mmtk/mmtk-core/issues/1348
    -   It is better to tidy up `Mmapper` before doing this because this changes the semantics of `Mmapper`.
-   [ ] Reduce the amount of code dependent on the page size.
    -   Particularly, refactor methods of `Mmapper` so that they take the number of bytes as arguments.  Concrete `Mmapper` implementations may work on a coarser granularity anyway.
-   [ ] Introduce an `os` module to hold OS-dependent parts of mmtk-core.
    -   This should include the page size which varies between OSes and architectures.

wks:
PR https://github.com/mmtk/mmtk-core/pull/1599 adds support for aarch64 macos.  It exposes the page size as a per-OS constant: `OSMemory::LOG_BYTES_IN_PAGE`.  It may have different values depending on the target architecture (aarch64 or x86_64).

While this approach works, there are corner cases where an operating system on an architecture can support multiple page sizes.  One example is Linux on aarch64.  The [Linux Kernel Documentation](https://docs.kernel.org/arch/arm64/memory.html) mentioned 4KB and 64KB page size, while Android is advocating 16KB page size.  So the safest way to know the page size is asking the operating system, using the command line `getconf PAGE_SIZE` or the C function `sysconf`.

In this way, the page size is no longer a constant.  I am still optimistic about the performance implication because we don't use the page size in any fast paths.  The page size is used in the low-level OS-facing memory allocation.

PR #1491, opened 2026-04-30
Quarantine address ranges for spaces

qinsoon:
This PR ensures that contiguous spaces are always created with its address range quarantined. There are two cases:
1. A contiguous space is created with `VMRequest::Fixed`. We quarantine the specified address range.
2. A contiguous space is created with `VMRequest::Extent/AlignedExtent` (for 64 bits targets, this is the default for all spaces). We quarantine an address range with a fixed address. If the fixed address is not available, we will see a panic.

This resolves most issues for https://github.com/mmtk/mmtk-core/issues/1346. 

Minor changes:
* ci-check-lineends.sh ignores the `book` directory which may contain unexpected line end patterns.
* `OSMemory::dzmmap_anywhere` now also returns `MmapResult`. This is consistent with other dzmmap methods.
* Allow replace for mmaping raw memory free list (it mmaps into the space address range which is now quarantined)
* Add `OSMemory::dzmmap_preferred` and `Mmapper::quarantine_address_range_preferred`. However, those are unused for now.


qinsoon:
The return type of `OSMemory::dzmmap_anywhere` is changed in this PR, which made the CI API checks fail. I don't think we need a migration entry for this, as `dzmmap_anywhere` was recently introduced, and is not in any release version yet.

wks:
I tried this on the mmtk-ruby binding.  In the debug mode, the start-up time changed from 130ms to 300ms, and in debug mode from 14-19ms to 20-30ms.

While this looks small, I worry that it may affect short-running scripting languages (such as Ruby).  I suggest we need to optimize the `MapState` data structure so that it doesn't record the `Quarantined` state.  Instead we can just record quarantined memory ranges in a `Vec<(Address, usize)>`.  Because we quarantine memory much less frequently than mmapping (which is even more true after we pre-mmap the entire side metadata memory), the Vec should contain only a few entries and should be very fast for linear searching.  But if we guarantee the memory of all spaces are quarantined, we don't even need to check the `MapState` of each chunk to decide whether we set `MmapStrategy::replace`.  It should always set `replace` to true.  We can just assert that every time we call `ensure_mapped`, the memory range is within one of the quarantined ranges.

qinsoon:
> I tried this on the mmtk-ruby binding. In the debug mode, the start-up time changed from 130ms to 300ms, and in debug mode from 14-19ms to 20-30ms.

Do you mean 14-19ms to 20-30ms for a release build?

How many spaces do a Ruby binding use? I think one thing we can do is to allow bindings to specify what space/allocation semantics they would use, and we can skip the initialization for spaces that are unused for a binding.

src/util/heap/heap_meta.rs
```diff
@@ -14,25 +16,52 @@ impl HeapMeta {
...
+            self.heap_cursor
         };
 
+        let actual = mmapper.quarantine_address_range_preferred(
+            preferred,
+            crate::util::conversions::bytes_to_pages_up(extent),
+            align,
+            huge_page_option,
+            anno,
+        )?;
```
wks: This is not preferred/actual.  The memory needs to be mapped at the exact address.  On Map64, the address range of a space is determined by the space index.  That means if a space has index 2, its address must be 0x400_0000_0000 to 0x600_0000_000.  Many parts, specifically the SFT and the SFTSpaceMap, depends on this deterministic address layout.

I think the right solution is even simpler.  We do the quarantining after all spaces are instantiated and their space indices are known.  Then

-   For Map64, the address range of each space just the exact 0x200_0000_0000 byte range determined by the space index of each space.
-   For Map32, currently we only need to quarantine the unique "discontiguous address range".  We currently don't have any "contiguous spaces" in Map32 (the nursery in GenCopy and GenImmix used to be contiguous, but we changed that), so we can just call `unimplemented!()` if we see anything other than `VMRequest::Discontiguous` in map32. (lockfreeimmortalspace may be an exception.)

To properly address the address layout problem with Map32, we should probably introduce the concept of "partition" (https://github.com/mmtk/mmtk-core/issues/932).  Currently the address range calculation of spaces is a mess, and the issue https://github.com/mmtk/mmtk-core/issues/984 attempts to address it by changing the way spaces are instantiated (but the code examples in that issue may not be optimal).

wks:
> > I tried this on the mmtk-ruby binding. In the debug mode, the start-up time changed from 130ms to 300ms, and in debug mode from 14-19ms to 20-30ms.
> 
> Do you mean 14-19ms to 20-30ms for a release build?

Yes.  On master, running `ruby --mmtk --version` with the release build takes between 14 to 19 ms.  With this PR, running the same command with release build takes 20-30 ms.

> 
> How many spaces do a Ruby binding use? I think one thing we can do is to allow bindings to specify what space/allocation semantics they would use, and we can skip the initialization for spaces that are unused for a binding.

CRuby by default uses StickyImmix.  That has four spaces: "immix", "immortal", "los" and "nonmoving".  CRuby currently doesn't use the immortal space or the non-moving space.  But I think it complicates things if we selectively skip quarantining spaces.

src/util/heap/heap_meta.rs
qinsoon: > On Map64, the address range of a space is determined by the space index. That means if a space has index 2, its address must be 0x400_0000_0000 to 0x600_0000_000. Many parts, specifically the SFT and the SFTSpaceMap, depends on this deterministic address layout.

I agree with this. I realize there might be some assumptions with SFT, SFTSpaceMap, or SpaceDescriptor. I will write some tests to check this, and how we may fix it.

> I think the right solution is even simpler. We do the quarantining after all spaces are instantiated and their space indices are known.

The issue is that we cannot guarantee we will always successfully quarantine the address ranges that we want.

qinsoon:
> I think it complicates things if we selectively skip quarantining spaces.

We can selectively not initialize allocators/spaces that are not in use, as they are not used anyway.

wks:
> > I think it complicates things if we selectively skip quarantining spaces.
> 
> We can selectively not initialize allocators/spaces that are not in use, as they are not used anyway.

That would complicate things even more.  Then the plan needs to be aware that some spaces are not initialized.  That may affect, for example, the dispatching of `trace_object`, the `Space::in_space` call, etc.

qinsoon:
> > > I think it complicates things if we selectively skip quarantining spaces.
> > 
> > 
> > We can selectively not initialize allocators/spaces that are not in use, as they are not used anyway.
> 
> That would complicate things even more. Then the plan needs to be aware that some spaces are not initialized. That may affect, for example, the dispatching of `trace_object`, the `Space::in_space` call, etc.

A binding is aware of what allocation semantics they use. Disable the unused ones sounds reasonable to me. We could even use that info to simplify the if-else for dispatching trace_object, or for other purposes.

I don't want to separate the quarantined state from the normal mmaper unless we have to. This, to me, is actually complicating things. Besides, having the mmapper to know all the virtual memory states seems beneficial. At this point, I don't see we have to go down that track.

src/util/heap/heap_meta.rs
wks: > The issue is that we cannot guarantee we will always successfully quarantine the address ranges that we want.

We probably can't guarantee this unless we dynamically select the memory range of spaces, too (https://github.com/mmtk/mmtk-core/issues/1347), because the loader or *any* `mmap` invocation could (in theory) occupy that range.  But according to the behavior of the Linux kernel and the loader (see https://github.com/mmtk/mmtk-core/issues/1482#issuecomment-4384918418), it is very likely to succeed if we do the quarantining early enough.  The best we can do for now is to panic if we fail to quarantine the address range.

wks:
In today's meeting, we agreed that

1.  We can leave the Mmapper and the ChunkStateMmapper unchanged in this PR.  I'll later try to write another Mmapper implementation that maintains the Quarantined state with a different data structure, and make another pull request.
2.  We may add compile-time options (cargo features) that allow VM bindings to disable spaces that they don't need, such as the immortal space and the non-moving space.

qinsoon:
@wks This PR is ready for review again.

src/policy/lockfreeimmortalspace.rs
```diff
@@ -234,10 +234,36 @@ impl<VM: VMBinding> LockFreeImmortalSpace<VM> {
...
-            unreachable!()
+        let (extent, align, top) = match vmrequest {
+            VMRequest::Extent { extent, top } => (extent, None, top),
+            VMRequest::AlignedExtent { extent, align, top } => (extent, Some(align), top),
+            _ => unreachable!(),
         };
-        let start = args.heap.reserve(extent, top);
+        let anno = MmapAnnotation::Space { name: args.name };
+        let huge_page_option: HugePageSupport =
+            args.options.transparent_hugepages_as_huge_page_support();
```
wks: Rust can infer its type.  You may have double-clicked the inferred type name in VSCode.

```suggestion
        let huge_page_option = args.options.transparent_hugepages_as_huge_page_support();
```

src/util/heap/heap_meta.rs
```diff
@@ -14,25 +16,62 @@ impl HeapMeta {
...
+
         assert!(
-            self.heap_cursor <= self.heap_limit,
-            "Out of virtual address space at {} ({} > {})",
-            self.heap_cursor - extent,
+            actual >= self.heap_cursor && actual + extent <= self.heap_limit,
```
wks: We assert the actual range is within the `heap_cursor` and `heap_limit`, but the critical invariants of SFTSpaceMap, i.e. each space occupies a non-overlapping 2TB range, is still not enforced.

wks (commented):
The problem is still SFT.  With Map64, we use SFTSpaceMap, and the map only has one entry for each space.  We currently update the SFT when we grow spaces.

If the address range of a space is not fixed, but "preferred", and `dzmmap_preferred` is allowed to return a different address, we need to enforce an important invariant: each 2TB region is used by at most one space.  If `dzmmap_preferred` happens to return two address ranges for two spaces, but they reside in the same 2TB region, it will be a conflict in SFT.  This PR currently does not check against that, but only checks if the returned address range is within the heap range.

Currently each space has a `SpaceDescriptor`.  In both Map64 and the discontiguous partition in Map32, the `SpaceDescriptor` will encode a "space index".  for Map64, the "space index" is calculated the same way as SFTSpaceMap, and the "space index" is also used by ChunkMap as the identifier of the space.  We should check we don't have two spaces sharing the same "space index".

I think we should add such checks, and introduce some test cases to ensure the behavior when `actual != preferred` is acceptable.  If the return value from the OS does not meet the SFTSpaceMap's requirement, we can just use plain `dzmmap` and panic if it fails.

qinsoon:
> we need to enforce an important invariant: each 2TB region is used by at most one space.

Right. I got what you meant. This particular invariant above is actually enforced, as the space start is aligned to 2TB (even for the preferred case). The problem is that `SFTSpaceMap` assumes the space address range is a specific range (e.g. any address that is larger than the given range is considered to be in the last space). There might be similar assumptions in `SpaceDescriptor`.

I will write some tests to verify this and fix if possible.

qinsoon:
> The problem is that SFTSpaceMap assumes the space address range is a specific range (e.g. any address that is larger than the given range is considered to be in the last space). There might be similar assumptions in SpaceDescriptor.

There are a few issues, but can be fixed -- I have a test that quarantines the entire current heap range and forces all the spaces to be mmapped to a different address range, and it seems working fine. There are a few things that are fixed to make it work:
1. `MAX_SPACES`: currently we set `MAX_SPACES` to 16. Types such as `SFTSpaceMap`, `SpaceDescriptor` and `Map64` use `MAX_SPACES` to initialize their internal arrays, and thus only supports the 16 spaces and the address ranges for those 16 spaces. As we use 2TB virtual address space, and 47 bits for address space (`LOG_ARCH_ADDRESS_SPACE`), at max we may have 64 slots for 2TB spaces. If we set `MAX_SPACES` to 64 on 64bits, then most issues for those types are gone.
2. `VMLayout` allows users to set `log_space_extent` and `force_use_contiguous_spaces`. When `force_use_contiguous_spaces` is set, the default value for `log_space_extent` is 41 (2TB). But it can be set to other values. However, I believe only 2TB works for `SFTSpaceMap`/`SpaceDescriptor`/`Map64`. We probably should make this explicit -- when we set `force_use_contiguous_spaces`, the space extent is 2TB.

I turned this PR back to draft, as I will need to do some cleanup before it is ready for review again.

src/util/heap/heap_meta.rs
```diff
@@ -14,25 +16,53 @@ impl HeapMeta {
...
+            let raw_start = self.heap_cursor;
+            if let Some(align) = align {
+                raw_start.align_up(align)
+            } else {
+                raw_start
+            }
         };
 
+        // TODO: The following call do an fixed mmap. We should try to allow the OS to choose the address if the fixed mmap fails.
+        mmapper.quarantine_address_range(
```
qinsoon, line 45: This is reverted to use `quarantine_address_range` instead of `quarantine_address_range_preferred`. So this PR only quarantine the current address range used by spaces. It no longer tries to get an arbitrary address from OS mmap.

src/policy/lockfreeimmortalspace.rs
wks: This hasn't been fixed.

src/util/metadata/side_metadata/layout.rs
```diff
@@ -80,7 +80,7 @@ pub(super) fn initialize_side_metadata_base(
     );
     let base = if specified_base.is_zero() {
         MMAPPER
-            .quarantine_address_range_anywhere(pages, huge_page_support, &anno)
+            .quarantine_address_range_anywhere(pages, None, huge_page_support, &anno)
```
wks, line 83: It's OK to use `None` here because the OS usually guarantees the mmap returns page-aligned addresses.

Actually it may be helpful if we guarantee side metadata are also aligned to some value.  For example, if a an Immix line is 256-byte aligned, and the VO bit is 1 bit per 8 bytes on 64-bit machines, then the VO bits of all 64-bit words in an Immix line will fit into a 32-bit word.  If the VO bit side metadata is at least 32-bit aligned, we can guarantee that the VO bits in an Immix line is in an *aligned* 32-bit word. (See https://github.com/mmtk/mmtk-core/issues/1407)  Currently the page-alignment from the OS is so strong that we don't need to specify this 32-bit alignment requirement for VO bits.  But I feel we may want to expose the alignment guarantee provided by the `mmap` system call in the `os` module.  We can deal with it later.  (See https://github.com/mmtk/mmtk-core/issues/1370)

wks (changes requested):
Mostly OK.  Only a small problem about a type annotation that can be auto-inferred.

wks (approved):
LGTM

qinsoon:
@wks The Ruby tests failed. Is it related with this PR?

wks:
> @wks The Ruby tests failed. Is it related with this PR?

No.  I have seen this error before.  It is triggered in rare cases.  ~~I'll restart the Ruby tests.~~  Let's restart the Ruby tests after the OpenJDK and Julia tests finish.

PR #1493, opened 2026-05-13
Concurrent Immix should do a full collection on user requests

qinsoon:
On user collection requests, concurrent Immix should make some efforts on reclaiming memory (either doing a final marking pause, or a full STW pause). Though how we handle user collection requests is totally up to the plan, triggering an initial marking pause on user requested collection does not reclaim anything and is almost unexpected.

### DIFF
--- a/src/plan/concurrent/immix/global.rs
+++ b/src/plan/concurrent/immix/global.rs
@@ -146,7 +146,11 @@ impl<VM: VMBinding> Plan for ConcurrentImmix<VM> {
             // It is related to defragmentation.  See https://github.com/mmtk/mmtk-core/issues/1357 for more details.
             // We currently force `FinalMark` to happen if the last pause is `InitialMark`.
             Pause::FinalMark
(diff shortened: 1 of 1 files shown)


wks (approved):
LGTM

wks:
By the way, In `GCTrigger::handle_user_collection_request`, we have the following snippet:

```rust
            if exhaustive {
                if let Some(gen) = self.plan().generational() {
                    gen.force_full_heap_collection();
                }
            }
```

Can we do something similar?  If a plan implements ConcurrentPlan, we can add a method, e.g. `ConcurrentPlan::force_concurrent_gc` to let it do full GC in the coming GC.  For ConcurrentImmix, it shall set `ConcurrentImmix::should_do_full_gc`.  But I haven't figured out its interaction with an in-progress concurrent marking.  What will happen if the user calls `handle_user_collection_request` when concurrent marking is in progress?  

The current PR seems to be the cleanest solution at this moment.

oscardssmith:
yeah. We likely want the ability to request both concurrent and pausing GCs.

Issue #840, opened 2023-05-30
Replace compare_exchange with fetch_update


wks:
# Task list

-   [ ] Add method `fetch_update` to `Address` (if you feel necessary)
-   [ ] Replace existing uses of `compare_exchange` (for atomic types, Address and header/side metadata, etc.) with `fetch_update` (or `fetch_add`, `fetch_sub`, `fetch_and`, `fetch_or`, `fetch_xor`, `swap`, etc. if appropriate).

# Rationale

TL;DR: `fetch_update` is definitely more concise than `compare_exchange`, and may be faster.  Let's use it.

[Compare exchange (a.k.a. compare-and-swap, or CAS)](https://en.wikipedia.org/wiki/Compare-and-swap) is one way to implement atomic read-modify-write operations.  In mmtk-core, we widely use the compare exchange operation, mainly for atomic metadata accesses.

In Rust, compare exchange is usually used in a loop.  For example,

```rust
let a: &AtomicUsize = ...;

loop {
    let old_value = a.load(...);

    if !need_change(old_value) {
        break;
    }

    let new_value = compute(old_value);
    let result = a.compare_exchange_weak(old_value, new_value, ...);
    if result.is_ok() {
        break;
    }
}
```

This approach has two problems.

1.  It is verbose.
2.  Compare exchange is not a primitive instruction for some architectures.

It is verbose compared to the `fetch_update` method of built-in atomic types.  The same idea in the code above can be expressed with `fetch_update` more concisely, without an explicit loop.

```rust
let a: &AtomicUsize = ...;

a.fetch_update(/* insert memory orders here */, |old_value| {
    if need_change(old_value) {
        let new_value = compute(old_value);
        Some(new_value)
    } else {
        None
    }
});
```

Not all architectures have CMPXCHG or CAS instructions.  A popular alternative is [load-link/store-conditional (LL/SC)](https://en.wikipedia.org/wiki/Load-link/store-conditional).  Using LL/SC, the same idea as the code above can be written as the following pseudo-code:

```rust
let a: &AtomicUsize = ...;

loop {
    let old_value = a.load_link();  // pseudocode. Load from `a` and link `a` with the current processor.  Break previous links of `a`.

    if !need_change(old_value) {
        break;
    }

    let new_value = compute(old_value);
    let successful = a.store_conditional(new_value);  // pseudocode.  Store into `a` only if `a` is still linked to the current processor.
    if successful {
        break;
    }
}
```

[...]

PR #1497, opened 2026-05-19
`ObjectBarrier::object_is_unlogged` should be atomic

ThePuzzlemaker:
This change fixes a probable soundness bug in this code, and updates the doc comment which appears to have been copied from `log_object` and was incorrect for this given function.

I believe this is a soundness bug because this function is called in the fastpath of ObjectBarrier, which amounts to (when expanded):
```rs
if self.object_is_unlogged(src) {
  if self.log_object(src) {
    ...snip...
  }
}
```

`log_object` is atomic, but `object_is_unlogged` is **not**. Interleaving atomic and non-atomic operations like this, as well as using non-atomic accesses concurrently when this metadata bit may be modified concurrently, [is unsound][1].

This bug does not affect SATBBarrier, as it already uses proper atomic operations in both the `object_is_unlogged` and `log_object` cases.

[1]: https://doc.rust-lang.org/stable/std/sync/atomic/index.html#memory-model-for-atomic-accesses

I have used `SeqCst` here even though I believe another ordering may be more apt, as I think those changes should be in a different PR as part of #840, and I am not entirely sure of this due to the prevalent usage of `SeqCst` within the MMTk codebase already. You may refer to [my thoughts on this that I posted on the Zulip][2].

[2]: https://mmtk.zulipchat.com/#narrow/channel/262673-mmtk-core/topic/Extending.20atomics.20audit.20.28.23840.29.20to.20usage.20of.20SeqCst/near/596237532

### DIFF
--- a/src/plan/barriers.rs
+++ b/src/plan/barriers.rs
@@ -192,10 +192,9 @@ impl<S: BarrierSemantics> ObjectBarrier<S> {
         Self { semantics }
     }
 
-    /// Attempt to atomically log an object.
-    /// Returns true if the object is not logged previously.
+    /// Returns true if the object is not logged.
     fn object_is_unlogged(&self, object: ObjectReference) -> bool {
-        unsafe { S::UNLOG_BIT_SPEC.load::<S::VM, u8>(object, None) != 0 }
+        S::UNLOG_BIT_SPEC.load_atomic::<S::VM, u8>(object, None, Ordering::SeqCst) != 0
     }
 
     /// Attempt to atomically log an object.


qinsoon (approved):
LGTM. Thanks.

The memory ordering may not need to be as strong as `SeqCst`. But this is consistent with what we do for `SATBBarrier`. Further changes on the memory ordering can be left to a different PR.

wks:
Actually `Relaxed` is sufficient for ObjectBarrier.  It only needs to ensure all objects that have been written to between two GCs are identified by the second GC.  `Relaxed` is sufficient because when GC is triggered, mutator threads will call `Collection::block_for_gc`, and it will form a happens-before relationship between mutators that are stopped and GC workers that are waken up after all mutators are stopped.

PR #1498, opened 2026-05-24
Make the visualize script compatible with Python 3.10

qinsoon:
This is the only issue that I encountered. But it might not be the only issue with Python 3.10.


wks (approved):
LGTM

PR #1501, opened 2026-05-26
Log pause time with fractional milliseconds

qinsoon:
`elapsed.as_millis()` truncates the value down to the nearest integral value (milliseconds). When the pause is short enough, the fractional part becomes significant. This PR switches the log to use floating-point millisecond values so short pauses can be displayed with sub-millisecond precision.


wks (approved):
LGTM

oscardssmith:
merge?

PR #1502, opened 2026-05-26
Add support to find next non-zero bit in side-metadata

k-sareen:


qinsoon:
Can you add some tests to make sure the implementation is correct?

Optionally, it would be good if you can extract common code between search backwards and search forwards, and reduce code duplication where possible. I feel only a few lines are actually different while the rest are identical. Using a constant generic parameter to identify search direction should work.

k-sareen:
There's a fair amount of code-duplication between the `find_{prev, next}_non_zero_value` functions. I was not comfortable refactoring it, so I've implemented a mirror basically.

k-sareen:
Sure, I'll add some tests. I'll create analogues for the tests for the `find_prev` one. As a single datapoint, the function works correctly in my fork since I use it, but I'm only using it at exactly one location.

k-sareen:
~~Regarding the fix to the edge case above -- the option was to either do `while cursor <= meta_end {}` in `find_first_non_zero_bit_in_metadata_bytes` or to align up the `end_addr` in the `find_next_non_zero_value_fast`. Which one do you prefer @wks? I think perhaps the `while cursor <= meta_end` is more accurately saying what we want, i.e. the metadata byte can be exactly at the end address in which case we still need to load it to make sure we're correctly finding the next set bit.~~

EDIT: Actually the change to the while loop bounds is not working correctly for all cases. So perhaps the current fix is the best one?

wks:
Is there an immediate need to use forward searching in a binding?

If not, I'd suggest we do some refactoring on the algorithm.  The current implementation involves too many things:

1.  Searching for the first/last non-zero bit in a byte/word.
2.  Searching for the next/previous non-zero bit in an address range.
3.  Searching for the next/previous non-zero byte/word in an address range.
4.  Dealing with contiguous/discontiguous metadata.
5.  Checking if a chunk is mapped.
6.  Selecting searching direction (forward/backward).

Even a low-level helper function `helper::find_last_non_zero_bit_in_metadata_bytes` (and the new `helper::find_first_non_zero_bit_in_metadata_bytes` introduced in this PR) is dealing with several things (1, 2, 3, 5) at once (although it calls `helper::find_last_non_zero_bit_in_metadata_bits` for (1), but `helper::find_last_non_zero_bit_in_metadata_bits` deals with (5), too, which I think is a break of abstraction).

I introduced the `ranges::BitByteRange` type to solve an Rust value ownership problem, but it may not be most elegant solution.

I think we could extract a very simple primitive (maybe someone 3rd party library already implemented that) that simply **searches for the first/last non-zero byte in an address range**.  We just make a composite algorithm that:

1.  We split the metadata address range into chunks and ensure each chunk is mapped and contiguous.
2.  We call that function to search for the first/last non-zero byte in the chunk.
3.  We search for the first/last non-zero bit in the returned byte.

The extracted function that searches for non-zero byte should be able to use SIMD instructions on architectures that support it (and third-party libraries may have already implemented it that way).

k-sareen:
There's no immediate need. I needed it in my fork so I thought I can quickly upstream it. Having it was better than not having it, I thought.

src/util/metadata/side_metadata/global.rs
```diff
@@ -1123,6 +1123,167 @@ impl SideMetadataSpec {
...
+    ) -> Option<Address> {
+        let region_bytes = 1 << self.log_bytes_in_region;
+        // Figure out the range that we need to search.
+        let start_addr = data_addr.align_down(region_bytes);
+        let end_addr = data_addr + search_limit_bytes;
+
+        let mut cursor = start_addr;
+        while cursor < end_addr {
+            // We encounter an unmapped address. Just return None.
+            if !cursor.is_mapped() {
```
wks: The `MMAPPER` has a granularity.  If the current `cursor` is mapped, we can skip the check until the next boundary of `MMAPPER.granularity()`.  See `find_last_non_zero_bit_in_metadata_bytes`.

wks (approved):
I think the code quality is as good as the existing one.  There is only one minor issue in `find_next_non_zero_value_simple`.  But if no performance critical code is depending on it, we can merge this PR as is.

src/util/metadata/side_metadata/global.rs
k-sareen: Right. I was mirroring the implementation in the `find_prev_non_zero_value_simple`. Should I change the implementation of that too?

wks: > Right. I was mirroring the implementation in the `find_prev_non_zero_value_simple`. Should I change the implementation of that too?

Yes.  It should be easy enough to fix both.

wks (approved):
LGTM

Issue #1137, opened 2024-05-20
Custom tracing unit/packet


wks:
This is one way to implement https://github.com/mmtk/mmtk-core/issues/710

# What is a custom tracing unit?

It is a unit of work that processes multiple object graph edges.  Examples include

-   Process one slot, tracing the reference in it and updating the slot.
-   Process a contiguous range of slots (i.e. `MemorySlice`), tracing and updating each slot.
-   Process a native object (a struct held in a malloc cell) that is associated to an in-heap object and contains object references.  For example, some Ruby objects that have off-heap buffers.
-   Process a native object (a struct held anywhere outside the GC heap) that is not part of any heap object.  Examples include the indirect reference table used by JNI, the weak reference table, finalization table, string table, and other strong or weak tables.
-   Process a stack, tracing and updating each root slot, except we cannot use the `trait Edge` (or `trait Slot` after [this](https://github.com/mmtk/mmtk-core/pull/1134)) for various reasons.

The common part is that all of them need to call `trace_object` to trace the edges, and create `ScanObjects` work packets for newly visited children.

What's not common is that the "unit" can be small or large.  It can be one single slot, and it can be multiple objects to be scanned, and it can be a whole stack.

## In theory, Edge (Slot) and MemorySlice are custom tracing units

Yes.   But we don't want to replace them yet.  They work pretty well in MMTk for now.

# Representation of a custom tracing unit

## As a closure

The simplest way to represent such a thing is a `FnOnce(impl ObjectTracer)`, or a trait like this:

```rust
trait CustomScannableUnit {
    fn run(self, object_tracer: &mut impl ObjectTracer);
}
```

That is, it is a runnable thing that can contain arbitrary data as context, and it uses a `ObjectTracer` (which provides `trace_object` when running.

But the key point is that it is not given a reference to `ObjectTracer` until it is executed.  This is important because we can only call `trace_object` at certain times, such as `TPinningClosure`, `PinningRootsTrace`, `Closure`, and `*RefClosure`.  Importantly, we cannot call `trace_object` in `Prepare`.

## As a work packet

Because it is a unit of work, we can wrap it in a work packet.

But because a "custom tracing unit" can be small, we can pack multiple such units into one packet.

[...]

PR #1503, opened 2026-05-28
Reorganize scheduler/gc_work.rs

### LINKED ISSUE #1279: Reorganize src/scheduler/gc_work.rs
The file `src/scheduler/gc_work.rs` contains about 1200 lines of code, with more than 20 structs and more than 50 `impl` items.

-   There are simply too many items, and
-   they are not ordered or grouped in the way they are used. Not by use cases, and not in the temporal order in a GC.

This made it difficult for developers to find the right type or function when they want to make modification to MMTk.  But one design goal of MMTk is to make this easy.

# Reorganizing this file into multiple files

We can categorize the work packets and traits by their use cases.

Unless specified, items listed below are `struct` items.

-   Generic GC scheduling work packets.
    -   `ScheduleCollection`
    -   `StopMutators`
    -   `Prepare`
    -   `PrepareMutator`
    -   `PrepareCollector`
    -   `Release`
    -   `ReleaseMutator`
    -   `ReleaseCollector`
    -   `VMPostForwarding` (running concurrently with `Release`)
-   Root scanning
    -   `ScanMutatorRoots`
    -   `ScanVMSpecificRoots`
    -   `ProcessEdgesWorkRootsWorkFactory` (implements `RootsWorkFactory` defined in `src/vm/scanning.rs`)
    -   enum `RootsKind`
-   Tracing (transitive closure)
    -   Slots/edges
        -   `ProcessEdgesBase` (This is all about processing slots instead of edges).
        -   trait `ProcessEdgesWork` (This part really needs to be refactored.  `ProcessEdgesWork::trace_object` shouldn't be specific to slot-enqueued tracing.)
        -   `SFTProcessEdges`
        -   `PlanProcessEdges`
        -   `ProcessEdgesWorkTracer` (implements `ObjectTracer` defined in `src/vm/scanning.rs`)
        -   `UnsupportedProcessEdges`
    -   Nodes
        -   trait `ScanObjectsWork`
        -   `ScanObjects`
        -   `PlanScanObjects`
        -   `ProcessRootNode` (for pinning and transitive-pinning roots)
-   Finalizer and weak references
    -   `ProcessEdgesWorkTracerContext` (implements `TracerContext` defined in `src/vm/scanning.rs`)
    -   `VMProcessWeakRefs`
    -   `VMForwardWeakRefs`

# Refactoring slots-processing work packets

In the early stage of the Rust port, MMTk was very focused on slot-enqueuing tracing because that's what OpenJDK and JikesRVM do.  One design decision which I think is bad is that it mixed up "edge processing" and "slots processing".

-   Slot processing: For each slot in a list of slots, load from it, call `trace_object`, and store back forwarded reference.
-   Edge visiting: The `trace_object(obj) -> new_obj` method.  This effectively visits an object graph edge that points to `obj`, regardless whether `obj` is loaded from a given `Slot` or obtained in VM-specific ways in `Scanning::scan_object_and_trace_edges`.  In the end, if a VM binding only supports node-enqueuing tracing (such as Ruby), we have to wrap `ProcessEdgesWork` behind the `ObjectTracer` trait which only provides the `trace_object` method (that is, `ProcessEdgesWorkTracer`).

I had a branch that refactors the `trace_object` into a dedicated type and leaves the `ProcessEdgesWork` dedicated to slot processing.  The code is here: https://github.com/mmtk/mmtk-core/pull/1278, but it is very old.

And https://github.com/mmtk/mmtk-core/issues/599 contains more information about my intended refactoring.

wks:
The file `scheduler/gc_work.rs` contains too many lines of code.  We reorganize the work packets in that file into multiple modules.

`scheduler::gc_work` still contains essential work packets for scheduling, stopping mutators, preparing, releasing, and root scanning (not including making work packets from roots, which is done by implementations of `RootsWorkFacgory`) during a GC.  The following work packets remain in this module:

-   `ScheduleCollection`
-   `Prepare`
-   `PrepareMutator`
-   `PrepareCollector`
-   `Release`
-   `ReleaseMutator`
-   `ReleaseCollector`
-   `StopMutators`
-   `ScanMutatorRoots`
-   `ScanVMSpecificRoots`

The remaining work packets are all relevant to tracing GC, and are moved to `plan::tracing::gc_work`.  They are further split into three sub-modules.

`plan::tracing::gc_work::root` contains work packets and types for root processing

-   `DefaultRootsWorkFactory`
-   `ProcessPinningRoots`

`plan::tracing::gc_work::closure` contains work packets for computing the transitive closure

-   `ProcessSlots`
-   `ProcessNodes`

`plan::tracing::gc_work::weakref` contains work packets for the VM-neutral weakref/finalization API.

-   `VMProcessWeakRefs`
-   `VMForwardWeakRefs`

These work packets and the `ObjectTracerContext` trait allow the VM binding to retain not-yet-reached objects and expanding the transitive closure, therefore they are placed under the `tracing` module.  Reference counting collectors may need a different set of primitives for weakref/finalization.

*(Note that the work packets for the built-in reference processor and finalizable processor, namely `{Soft,Weak,Phantom}RefProcesing`, `RefForwarding`, `RefEnqueue`, `Finalization` and `ForwardFinalization`, remain in the `util::reference_processor` and `util::finalizable_processor` modules, respectively.*)

The `DefaultObjectTracerContext` and the `DefaultObjectTracer` are put under `plan::tracing::gc_work`.  The `ObjectTracerContext` trait was designed for binding-side weak reference processing so that the VM binding can selectively retain objects.  But (1) currently `ProcessNodes` reuses the `DefaultObjectTracerContext` to do node-enqueuing tracing, and (2) in the future we may let the VM binding implement some root-scanning, object-scanning or non-object node scanning (see https://github.com/mmtk/mmtk-core/issues/1137), and we may use the `ObjectTracerContext` interface for that.  We leave them in `plan::tracing::gc_work` instead of `plan::tracing::gc_work::weakref`.

This is a structural refactoring, and should not have any changes to the performance.

Fixes: https://github.com/mmtk/mmtk-core/issues/1279


wks:
I also considered moving `GCWorkScheduler::schedule_common_work` into `plan::tracing::gc_work`.  See: https://github.com/wks/mmtk-core/pull/new/fix/move-work-packets-around2

But I am not sure how specific the current `GCWorkScheduler::schedule_common_work`.  It currently adds `StopMutators`, `Prepare` and `Release`, and weak reference processing, which are not specific to tracing GC.  Reference counting GC needs them, too, although it may not be profitable to do weak reference processing in every RC pause.

Currently, plans deviate from `schedule_common_work` for two reasons

1.  mark-compact, including `MarkCompact` and `Compressor`.  Such collectors need a second transitive closure for compaction.
2.  concurrent collectors, including `ConcurrentImmix`.  When scheduling concurrent collections, it only schedules `StopMutators` and `Prepare` in the initial pause, and other work packets in the final pause.

We may refactor it in the future we we have more insights about different kinds of collectors.

src/plan/tracing/gc_work/closure.rs
```diff
@@ -0,0 +1,236 @@
...
+        let queue = self.process_slots(worker, trace);
+
+        self.flush(worker, queue);
+    }
+}
+
+/// A work packet for scanning objects and optionally do node-enqueuing tracing during a
+/// stop-the-world tracing GC and the final mark pause of a concurrent GC.
+///
+/// It will scan each objects.  For objects that supports slot enqueuing, it will collect their
```
k-sareen:
```suggestion
/// It will scan each object.  For objects that supports slot enqueuing, it will collect their
```

src/plan/tracing/gc_work/root.rs
```diff
@@ -0,0 +1,202 @@
...
+        },
+        VectorObjectQueue,
+    },
+    scheduler::{GCWork, GCWorker, WorkBucketStage},
+    util::ObjectReference,
+    vm::{RootsKind, RootsWorkFactory, VMBinding},
+    MMTK,
+};
+
+/// An implementation of [`RootsWorkFactory`] for stop-the-world tracing GC, i.e. finding the
```
k-sareen: `RootsWorkFactory` doesn't find the transitive closure though. It's just stashing the roots to be processed in the `Closure` phase. I think I know what the comment is trying to say, but it can be a bit confusing if you just read this as you can come to the conclusion that `RootsWorkFactory` is doing the transitive closure.

Perhaps:
```
An implementation of [`RootsWorkFactory`] for stop-the-world tracing GC wherein we cache root slots to start the transitive closure from.
```

src/plan/tracing/gc_work/root.rs
```diff
@@ -0,0 +1,202 @@
...
+            mmtk,
+            phantom: PhantomData,
+        }
+    }
+}
+
+/// This work packet processes pinning roots during stop-the-world tracing GC.
+///
+/// Note that by definition, a "root" is an *edge* from outside the object graph to an object.  This
+/// work packet represents each edge as the `ObjectReference` of the object the edge points to.
```
k-sareen:
```suggestion
/// work packet represents each edge as the `ObjectReference` of the object the edge points to (i.e. the referent).
```

src/plan/tracing/gc_work/closure.rs
```diff
@@ -0,0 +1,236 @@
...
+impl<T: Trace> ProcessNodes<T> {
+    pub fn new(objects: Vec<ObjectReference>, bucket: WorkBucketStage) -> Self {
+        Self {
+            objects,
+            bucket,
+            phantom_data: PhantomData,
+        }
+    }
+
+    fn increase_live_bytes(
```
k-sareen: We have to be careful. If we skip the `ProcessNodes` work-packet, for example, if we're only doing edge-enqueuing, then we may never increase the live bytes. Perhaps we should add a TODO or comment here for the future?

src/plan/tracing/gc_work/closure.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+}
+
+/// A work packet for scanning objects and optionally do node-enqueuing tracing during a
+/// stop-the-world tracing GC and the final mark pause of a concurrent GC.
+///
+/// It will scan each objects.  For objects that supports slot enqueuing, it will collect their
+/// slots and spawn [`ProcessSlots`] work packets to trace them.  For objects that don't
+/// support slot enqueuing, it will immediately trace their slots and spawn other
+/// [`ProcessNodes`] work packets to process their newly traced children.
```
k-sareen:
```suggestion
/// [`ProcessNodes`] work packets to process their newly traced children. It is the VM's responsibility to update the references to point to the new addresses in such a case.
```

k-sareen (approved):
LGTM. Thanks! Just some very minor things.

src/plan/tracing/gc_work/closure.rs
wks: Fixed.

src/plan/tracing/gc_work/root.rs
wks: I slightly rephrased it to avoid this confusion.

src/plan/tracing/gc_work/root.rs
wks: Fixed.

src/plan/tracing/gc_work/closure.rs
wks: I moved this function here to fix a compilation error because `GCWorkerShared` is private.  From the version history, when the live bytes calculation feature was added, the `worker.shared.live_bytes_per_space` field was put in `GCWorkerShared`.  Therefore, the `increase_live_bytes` method was added to `GCWorkerShared`, too.

I reverted this change and moved this method back to `GCWorkerShared`.  I made `GCWorkerShared` `pub(crate)`.  The fact that `ProcessNodes::try_enqueue_slots` is accessing `worker.shared.live_bytes_per_space.borrow_mut()` means `GCWorkerShared` should have been `pub(crate)`.

src/plan/tracing/gc_work/closure.rs
wks: I added this sentence.  I further added that the VM binding updates references by implementing `Scanning::scan_object_and_trace_edges`.

k-sareen:
Thanks for the changes!

PR #1504, opened 2026-05-29
Concurrent worker threads

qinsoon:
This PR allows the scheduler to use a subset of workers as 'active workers'. This mechanism is used to support running concurrent marking with a specified number of concurrent workers. The PR is mostly done by Codex.


src/scheduler/worker_monitor.rs
```diff
@@ -109,9 +153,17 @@ impl WorkerMonitor {
...
+                condvar.notify_one();
+            }
         } else {
-            self.workers_have_anything_to_do.notify_one();
+            let sync = self.sync.lock().unwrap();
+            let active_workers = self.active_workers();
+            if let Some(ordinal) = sync.parker.first_parked_active_worker(active_workers) {
+                debug!("Notifying parked worker {} for available work.", ordinal);
+                self.workers_have_anything_to_do[ordinal].notify_one();
+            }
```
wks: There is a race between the thread that notifies and the thread that waits.

1.  Assume all threads are busy now.
2.  Thread 1 tries to find some work to do, but couldn't find any.
3.  Thread 2 publishes one piece of work.  It tries to notify one work.  But it didn't find any active worker.  It ended up not notifying any condvar.
4.  Thread 1 sets itself as parked (by entering `WorkerMonitor::park_and_wait`).  It tries to wait on the Condvar `workers_have_anything_to_do[1]`.  Since the Condvar is not notified, Thread 1 successfully blocked itself on the condvar.

One important property of Condvar is that if a Condvar is notified before another thread waits on it, the second thread is guaranteed not to be able to block.

wks (commented):
I think this PR tries to limit the number of GC workers that wake up by notifying only some of the workers when some work is available.  But the algorithm in this PR introduces a race that prevent it from working.

I think we can still use one `workers_have_anything_to_do: Condvar` to notify all active workers.  When deactivating workers, we can let the deactivated workers instead wait for another condvar, e.g. `active_worker_number_changed: Condvar`.  The inactive workers will wait until the number or active workers change.  Here is a pseudocode.

```rust
fn run_worker(ordinal) {
    loop {
        // Now it is inactive.  Wait until activated.
        active_worker_number_changed.wait_until(|| { ordinal < active_worker_count });

        loop {
            // Now it is active.  Find something to do.
            while let Some work = find_some_work_to_do() {
                work.do_work();

                // Optional.  If we allow changing active worker count during a GC, we can add this `if`.
                // If we only allow changing active worker count when all workers are parked, we can remove this `if`.
                if ordinal >= active_worker_count {
                    break;
                }
            }

            // Nothing to do.  Park.
            park();
            workers_have_anything_to_do.wait();
            unpark();

            // Check if the number of active workers changed.
            if ordinal >= active_worker_count {
                break;
            }
        }

        // Now it is inactive again.
    }
}
```

When changing the number of active workers, we notify both Condvar instances: the `active_worker_number_changed` and the `workers_have_anything_to_do`.  Then all workers will wake up and check, but the de-activated workers (ordinal too big) will wait on `active_worker_number_changed` instead.  Other workers will check if there is anything to do, and then go back to sleep again.

src/util/options.rs
```diff
@@ -878,6 +902,9 @@ options! {
     plan:                   PlanSelector            [always_valid] = PlanSelector::GenImmix,
     /// Number of GC worker threads.
     threads:                usize                   [|v: &usize| *v > 0] = num_cpus::get(),
+    /// Maximum number of GC worker threads that may run concurrent GC work.
+    /// If this exceeds the total number of GC worker threads, all workers may participate.
+    concurrent_threads:     usize                   [|v: &usize| *v > 0] = Options::compute_default_concurrent_threads(num_cpus::get()),
```
qinsoon, line 907: An alternative is to merge this option with `threads`. Users can set something like `threads=8` (8 stw GC threads, the default number of concurrent GC threads), or `threads=8,2` (8 STW GC threads, 2 concurrent GC threads). Let me know if you think that's better.

src/scheduler/worker_monitor.rs
```diff
@@ -79,37 +89,82 @@ impl WorkerParker {
...
     pub fn make_request(&self, goal: WorkerGoal) {
-        let mut guard = self.sync.lock().unwrap();
-        let newly_requested = guard.goals.set_request(goal);
+        // A request always needs all workers, including currently inactive ones, to make
+        // progress.
+        self.set_active_workers(self.worker_count);
+        let newly_requested = {
+            let mut guard = self.sync.lock().unwrap();
+            guard.goals.set_request(goal)
+        };
```
wks: With the scope of the mutex reduced, this block will acquire the `sync` lock and release immediately.  Then when we call `self.notify_work_available(true);` below, it will acquire `sync` again.  We can merge the two mutex scopes by calling `notify_work_available_while_locked` instead.

src/scheduler/worker_monitor.rs
```diff
@@ -157,14 +212,17 @@ impl WorkerMonitor {
...
+                    self.notify_work_available_while_locked(true);
                 }
             }
         } else {
             should_wait = true;
         }
 
-        if should_wait {
+        while should_wait || !self.is_worker_active(ordinal) {
+            should_wait = false;
```
wks: This is correct, but is a bit counterintuitive.  We should add some comments to explain what it does.  For example, explain that `should_wait` only forces the worker to enter the loop the first time, but if the worker is deactivated during waiting, it will wait until it is re-activated.

Alternatively, we can rewrite the loop like this to avoid modifying `should_wait`.  I think this is clearer.

```rust
if should_wait {
    assert!(worker is active);
    wait on workers_have_anything_to_do
}

/// The current worker may become inactive while waiting
while worker is inactive {
    wait on active_worker_number_changed
}

assert!(worker is active);
```

qinsoon: Done. Both reviews are addressed in https://github.com/mmtk/mmtk-core/pull/1504/commits/399d30c5eeee95afab56be6b71c48f85b5fcc9be

wks (approved):
LGTM

PR #1507, opened 2026-06-04
Remove double enqueuing

wks:
This PR refactors `ProcessSlots` and `ProcessNodes` into simple textbook-style two-level loops, moving away from double-enqueuing, i.e. letting `ProcessSlots` and `ProcessNodes` work packets generate each other.

**DRAFT: This is still a draft.**

-   [X] Fix `ProcessSlots` so that it can compute the transitive closure alone, without `ProcessNodes`.
-   [ ] Fix `ProcessNodes` so that it can compute the transitive closure alone, without `ProcessSlots`.
-   [ ] Add an option for the VM binding to select between `ProcessNodes` or `ProcessSlots`.  CRuby must use `ProcessNodes` because it only implements `Scanning::scan_object_and_trace_edges`.
-   [ ] How to properly "fork" the work packets?  Or how should we do the work stealing properly?  We should discuss this.

### DIFF
--- a/src/plan/tracing/gc_work/closure.rs
+++ b/src/plan/tracing/gc_work/closure.rs
@@ -1,11 +1,11 @@
-use std::marker::PhantomData;
+use std::{collections::VecDeque, marker::PhantomData};
 
 use crate::{
     plan::{
         tracing::{gc_work::DefaultObjectTracerContext, SlotOfTrace, Trace},
-        VectorObjectQueue, VectorQueue,
+        VectorQueue,
     },
-    scheduler::{GCWork, GCWorker, GCWorkerShared, WorkBucketStage},
+    scheduler::{GCWork, GCWorker, GCWorkerShared, WorkBucketStage, EDGES_WORK_BUFFER_SIZE},
     util::{ObjectReference, VMWorkerThread},
     vm::{slot::Slot, ObjectTracerContext, Scanning, VMBinding},
     MMTK,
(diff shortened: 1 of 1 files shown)


wks:
Preliminary test with lusearch from DaCapo Chopin, StickyImmix plan, with 50M heap size shows that it almost cut the time to compute the transitive closure in half.  Sadly, the bottleneck of STW time of a nursery GC of StickyImmix in lusearch is currently weak reference and finalization.  Overall, the reduction of STW time is modest.

| | GC | time | time.other | time.stw |
|---|---|---|---|---|
| master | 768 | 3031 | 2525 | 506 |
| pr | 786 | 2786 | 2319 | 468 |

master:
image

pr:
image

k-sareen:
FYI this is the commit I have that implements slot enqueuing for `ProcessEdgesWork`: https://github.com/k-sareen/mmtk-core/commit/6a3da1088726034d156836c77a61d5203ba7ea70. It's based off of LXR, including the work packet splitting heuristic for better load balancing.

wks:
Tried again with SemiSpace (still lusearch, 50M), but this time this PR is slower.  The timeline shows there is obvious load-balancing problem, and the workers are 50% idle during transitive closure.
| | 	GC | 	time |	time.other | 	time.stw |
|---|---|---|---|---|
| master | 	1339 | 	3988 | 2438 | 1549 |
| pr | 	1339 | 	3984 | 	2312 | 	1671 |

master:
image

pr:
image

wks:
> FYI this is the commit I have that implements slot enqueuing for `ProcessEdgesWork`: [k-sareen@6a3da10](https://github.com/k-sareen/mmtk-core/commit/6a3da1088726034d156836c77a61d5203ba7ea70). It's based off of LXR, including the work packet splitting heuristic for better load balancing.

Thank you.  I tried your load-balancing technique, i.e. splitting the work packet for every 4096 pushes since the last split.  The load-balancing is slightly better.

image

I also noticed that the optimal work packet size may no longer be 4096 (which we chose arbitrarily anyway).  As shown above, a work packet that started with 241 slots can end up visiting 13018 slots during its execution, and it is split only three times.

wks:
I shrunk the constant `EDGES_WORK_BUFFER_SIZE` to 512 so that the `ProcessNodes work packets are at most 512 elements initially.  There is no noticeable improvement.

image

The strange thing is the work packet I highlighted.  A zoomed-in version is below:

image

The work packet started with only 109 slots, but ended up visiting 3010 slots.  It split itself 14 times.  But those split work packets are all concentrated in the two workers.  Why didn't other GC workers help consuming those work packets?

wks:
I think the culprit is `GCWorker::local_work_buffer`.  When calling `GCWorker::add_work`, it will try to add work packet to the worker's local buffer before adding to the global work buckets.  By default, `GCWorker::add_work` will add to the local buffer if it has fewer than 16 work packets.  But a worker can move work packets from work buckets to the local buffer in batch using `Injector::steal_batch_and_pop(crossbeam_deque::Worker)`.

```rust
    pub fn add_work(&mut self, bucket: WorkBucketStage, work: impl GCWork<VM>) {
        if !self.scheduler().work_buckets[bucket].is_open()
            || self.local_work_buffer.len() >= Self::LOCALLY_CACHED_WORK_PACKETS  // KUNSHAN: It's 16 by default
        {
            self.scheduler.work_buckets[bucket].add(work);
            return;
        }
        self.local_work_buffer.push(Box::new(work));
        probe!(mmtk, local_cache, self.local_work_buffer.len());  // KUNSHAN: I added this
    }
```

I added eBPF probes to visualize the number of work packets in the local cache of each worker.  And the result is just as I expected.

image

As the counters show, thread 123463 was holding 16 work packets locally, and only started to "share" its work packets with other threads when the local cache overflows the 16-packet capacity.

The other thread 123461 was holding up to 3 work packets at the point of the red arrow.  Because the local buffer never overflowed, the 3 work packets were only executed by thread 123461.

I think the problem here is that when adding work packets into the local queue, the current worker doesn't notify other GC workers of the opportunity to steal.

wks:
I added the following line at the end of `GCWorker::add_work`, which is right after the worker adds a work packet to its local buffer.

```rust
        self.scheduler.worker_monitor.notify_work_available(false);
```

Screenshot_20260605_233550

In the GC shown above, thread 138437 split 5 times, and other threads were able to steal them.

But in this version, the work packet still splits after visiting 512 slots or overflowing its 512-element capacity.  I don't think adding a `notify_work_available` is the proper solution.  To make the load balanced, either

-   the active work packet should split more frequently when more workers are idle, or
-   the active work packet should make its **work items** steal-able, i.e. work stealing at sub-work-packet level.

The latter is addressed by the paper *The paper Work Packets: A New Abstraction for GC Software Engineering, Optimization, and Innovation* by Zhao et al.

k-sareen:
I don't think this PR should address load balancing. That can be a separate PR when Wenyu (or whomever) has the time to upstream the changes to the work packet system.

wks:
I just used a simple heuristics that a work packet should split itself into more parts if there are GC workers parked.  It only requires a small change to the `WorkerMonitor`, but it looks effective.  When there are workers idle, the work packet splits more eagerly.

The two timelines below both use lusearch, Semispace and 50M heap size.

In the following timeline, packet size is 512 slots.  When full or after 512 pushes, split the work packet into n+2 parts if there are n workers parked.  This works pretty well.

image

In the following timeline, packet size is 4096 slots.  When full or after 4096 pushes, split the work packet into n+2 parts if there are n workers parked.  4096 is just too big.  Each work packet visits way more than 4096 slots.

image

Here are other plans.

StickyImmix, 512 slots.

image

Immix, 512 slots.  Note that it even managed to split the work packets for the closure of soft reference processing.

image


> I don't think this PR should address load balancing. That can be a separate PR when Wenyu (or whomever) has the time to upstream the changes to the work packet system.

That's right.  Wenyu showed me the code for sub-work-packet work stealing.  It requires some fundamental changes to the scheduler, the GC workers, etc.  I don't plan to do it in this PR.  I just made minimal effort to make the load balance problem not that serious.

qinsoon:
> I just used a simple heuristics that a work packet should split itself into more parts if there are GC workers parked.

This kind of optimization might not always be effective. If a work packet does not produce much further work, the cost of waking workers and fetching packets may outweigh the benefits of splitting. In other words, it could cut worst case scenarios, but may by average cause slowdown. I think we generally want to look at average numbers across benchmarks to confirm an optimization is effective.

I would suggest that 
1. doing a faithful refactoring (no optimization, just remove the double enqueueing), and 
3. doing a full performance evaluation (running all the benchmarks).

Then we could check if there is any slowdown, or even double check if the results match Angus's thesis. If we see issues, then optimize.

PR #1508, opened 2026-06-05
Simplified LXR GC

wenyuzhao:
This pull request introduces a simplified version of the LXR collector.  There are many features not included in this version of LXR, such as reference processing, and they will be ported to the master branch in the future.

In addition to the LXR plan, this PR also makes the following changes to mmtk-core

-   **Field unlog bits** metadata and **field-logging barrier**.  They can be used by generational and concurrent Immix plans to implement remembered set and SATB.
-   **`Pause::RefCount`**: a new pause kind for reference counting.
-   **`RootsWorkFactory::create_process_roots_work_with_root_kind`**: an experimental API with `RootKind`.
-   **Renamed pause-start and pause-end hooks** to merge the two mechanisms from the master branch and the LXR branch to make the names consistent.
-   **`UnstraddlableRegion`** to prevent the invocation of `region.contains` on regions which objects can straddle, specifically `Line`.
-   **Many changes to the `ImmixSpace` and `LargeObjectSpace`** to support reference counting-based Immix plans.
-   **Additional object metadata** to support LXR.
-   **Additional scheduler stages** to support LXR.
-   Merged the implementations of `ObjectReference::iterate_fields`.
-   The definition of "chunk index" is changed.
-   Map32: 0 is now a valid entry of the linked list, and invalid entries are represented as -1.

Related PRs:

-   mmtk-openjdk JDK-21: https://github.com/mmtk/mmtk-openjdk/pull/358
-   mmtk-jikesrvm: https://github.com/mmtk/mmtk-jikesrvm/pull/195
-   mmtk-ruby: https://github.com/mmtk/mmtk-ruby/pull/161


src/plan/lxr/gc_work/mature_evac.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    fn process_slot(&mut self, s: VM::VMSlot, reuse: u8, lxr: &LXR<VM>) -> bool {
+        // Skip slots that does not contain a real oop
+        if !self.address_is_valid_oop_slot(s, reuse, lxr) {
+            return false;
+        }
+        // Skip objects that are dead or out of the collection set.
+        let v = unsafe { s.to_address().load::<u32>() };
+        if v & 0b111 != 0 {
```
qinsoon: This assertion seems specific to bindings. It is a normal load of u32 from an address (not load on Slot which returns an object reference).

src/mmtk.rs
```diff
@@ -343,6 +349,10 @@ impl<VM: VMBinding> MMTK<VM> {
     pub fn harness_begin(&self, tls: VMMutatorThread) {
         probe!(mmtk, harness_begin);
         self.handle_user_collection_request(tls, true, true);
+        if tls.0 .0.is_null() {
+            use crate::vm::Collection;
+            VM::VMCollection::block_for_gc(tls);
+        }
```
wks, line 356: Even in `master`, mmtk-openjdk has always been passing NULL `tls` when calling `mmtk.harness_begin`, and MMTk passes the NULL `tls` back to mmtk-openjdk via `VM::VMCollection::block_for_gc`, and mmtk-openjdk simply ignores the `tls` parameter in `Collection::block_for_gc`.

So why do we skip the `block_for_gc` here when `tls` is null (which is always true for OpenJDK)?  There may be some LXR-specific considerations.

qinsoon: Similarly in `handle_user_collection_request`, `block_for_gc` is only called if tls is not null. This causes one test to hang (as `tls` in the test is null and `block_for_gc` is never called).
```rust
if !tls.0 .0.is_null() {
    VM::VMCollection::block_for_gc(tls);
}
```

qinsoon:
Two issues need to be addressed:
1. This PR assumes we only have the default immix space and LOS.
2. There are some raw conversion between address and object reference.

src/mmtk.rs
qinsoon: I changed the test tls to 1 to work around the issue. There is a FIXME to remind us to remove the workaround before merging the PR.

wenyuzhao: The LXR openjdk binding enables [vm thread triggered GC](https://github.com/mmtk/mmtk-openjdk/pull/358/changes#diff-42fd4b78fac9296caa55ac659d82b4c541173d46956a32dc5249f7d093aa51c8R374). This is mainly to support metaspace GC.

Such GC requests are triggered by VM threads, and handle_user_collection_request should not block so that VM thread can continue execution. For these GC requests there is no TLS available.

The one in harness_begin is to just force block current thread again, since harness_begin has TLS == NULL as well.

qinsoon: > The LXR openjdk binding enables [vm thread triggered GC](https://github.com/mmtk/mmtk-openjdk/pull/358/changes#diff-42fd4b78fac9296caa55ac659d82b4c541173d46956a32dc5249f7d093aa51c8R374). This is mainly to support metaspace GC.

We can add an argument for `handle_user_collection_request` to indicate non blocking. Or, we could have a different API, as requests from vm thread are not 'user requests' anyway.

src/policy/immix/block.rs
```diff
@@ -269,17 +500,15 @@ impl Block {
...
                 // There are some marked lines. Keep the block live.
-                let is_reusable = marked_lines != Block::LINES;
-                if is_reusable {
+                if marked_lines != Block::LINES {
                     // There are holes. Mark the block as reusable.
                     self.set_state(BlockState::Reusable {
-                        unavailable_lines: marked_lines as _,
+                        unavailable_lines: usize::min(marked_lines, u8::MAX as usize) as _,
                     });
-                    space.reusable_blocks.push(*self)
```
wks, line 282: Why is this `space.reusable_blocks.push` removed?  If a block contains lines that are not marked, it should be reusable.  We already executed `self.set_state(BlockState::Reusable { ...`.  Why don't we add it to `space.reusable_blocks`?  After this PR, the only two places we call `space.reusable_blocks.push` are in `rc_sweep_nursery` and `rc_sweep_mature`, respectively.

src/plan/barriers.rs
```diff
@@ -271,6 +276,84 @@ impl<S: BarrierSemantics> Barrier<S::VM> for ObjectBarrier<S> {
...
+        unimplemented!()
+    }
+}
+
+/// The value stored in the log bit/byte indicating that the object or field is unlogged, i.e. it
+/// has not yet been recorded in the remembered set and the write barrier should still take its slow path.
+pub const UNLOGGED_VALUE: u8 = 0b1;
+/// The value stored in the log bit/byte indicating that the object or field is logged, i.e. it
+/// has already been recorded in the remembered set and the write barrier can skip its slow path.
+pub const LOGGED_VALUE: u8 = 0b0;
```
wks: These should be moved to `log_bit.rs` and shared between the object unlog bits and field unlog bits.

src/plan/tracing/mod.rs
```diff
@@ -360,11 +365,21 @@ impl ObjectQueue for VectorQueue<ObjectReference> {
...
 // And it currently only works if the object supports slot enqueuing (i.e. `Scanning::scan_object` is implemented).
 // We may refactor the interface according to <https://github.com/mmtk/mmtk-core/issues/1375>
-pub(crate) struct SlotIterator<VM: VMBinding> {
+pub struct SlotIterator<VM: VMBinding> {
     _p: PhantomData<VM>,
 }
 
 impl<VM: VMBinding> SlotIterator<VM> {
+    /// Iterate over the slots of an object by applying a function to each slot.
+    pub fn iterate(o: ObjectReference, f: impl FnMut(VM::VMSlot)) {
```
wks: The `iterate_fields` below is ported from `lxr` when we ported ConcurrentImmix last time, and it takes `support_slot_enqueuing` into consideration.  We should just use that.

src/plan/lxr/global.rs
```diff
@@ -0,0 +1,836 @@
...
+        super::NO_EVAC.store(false, Ordering::SeqCst);
+        let pause = self.current_pause().unwrap();
+
+        super::SURVIVAL_RATIO_PREDICTOR
+            .set_alloc_size(self.block_allocation.total_young_allocation_in_bytes());
+
+        if pause == Pause::Full || pause == Pause::InitialMark {
+            // Reset block mark and object mark table.
+            let work_packets = self.generate_full_trace_prepare_tasks();
+            self.immix_space.scheduler().work_buckets[WorkBucketStage::RCProcessIncs]
```
wks, line 292: `self.immix_space.scheduler()`: Can just use the `_scheduler` parameter.

src/plan/mutator_context.rs
```diff
@@ -25,7 +25,7 @@ pub(crate) fn unreachable_prepare_func<VM: VMBinding>(
     _mutator: &mut Mutator<VM>,
     _tls: VMWorkerThread,
 ) {
-    unreachable!("`MutatorConfig::prepare_func` must not be called for the current plan.")
+    // unreachable!("`MutatorConfig::prepare_func` must not be called for the current plan.")
```
wks: May be a leftover from an ancient workaround.  We should un-comment this and test it again.

src/plan/mutator_context.rs
```diff
@@ -468,6 +468,11 @@ pub trait MutatorContext<VM: VMBinding>: Send + 'static {
     fn flush(&mut self) {
         self.flush_remembered_sets();
     }
+    /// Assert that this mutator's barrier has been flushed (i.e. it has no pending remembered-set
+    /// entries buffered). Used for sanity checking.
+    fn assert_is_flushed(&mut self) {
+        // self.barrier().assert_is_flushed();
```
wks: This, too.  If it makes sense, we should enable the assertion and test it again.

src/policy/immix/block.rs
```diff
@@ -94,6 +108,10 @@ impl BlockMayHaveObjects for Block {
 }
 
 impl Block {
+    /// Log bytes in block
+    pub const LOG_BYTES: usize = <Self as Region>::LOG_BYTES;
+    /// Bytes in block
+    pub const BYTES: usize = 1 << Self::LOG_BYTES;
```
wks: `Block` already implements `Region` and already provides the byte and log-byte size.  These can be omitted.

src/policy/immix/block.rs
```diff
@@ -110,6 +128,59 @@ impl Block {
...
+
+    /// Get block start address
+    pub const fn start(&self) -> Address {
+        self.0
+    }
+
+    /// Get block end address
+    pub const fn end(&self) -> Address {
+        self.0.add(Self::BYTES)
+    }
```
wks: These methods are all available from the `Region` trait.

src/policy/immix/block.rs
```diff
@@ -195,16 +323,119 @@ impl Block {
...
+
+    pub(crate) fn initialize_mark_table_as_marked<VM: VMBinding>(&self) {
+        let meta = VM::VMObjectModel::LOCAL_MARK_BIT_SPEC.extract_side_spec();
+        let start: *mut u8 = address_to_meta_address(meta, self.start()).to_mut_ptr();
+        let limit: *mut u8 = address_to_meta_address(meta, self.end()).to_mut_ptr();
+        unsafe {
+            let bytes = limit.offset_from(start) as usize;
+            std::ptr::write_bytes(start, 0xffu8, bytes);
+        }
```
wks, line 316: If we can, we should just use `bset_metadata`.  If performance is a concern, we may extract an optimized form for `Region`.  See https://github.com/mmtk/mmtk-core/issues/1407

src/policy/immix/block.rs
```diff
@@ -195,16 +323,119 @@ impl Block {
...
+            if old_value == 1 {
+                return false;
+            }
+            if Self::LOG_TABLE
+                .compare_exchange_atomic(self.start(), 0u8, 1u8, Ordering::SeqCst, Ordering::SeqCst)
+                .is_ok()
+            {
+                return true;
+            }
+        }
```
wks, line 331: Consider using `fetch_update`

src/policy/immix/block.rs
```diff
@@ -195,16 +323,119 @@ impl Block {
...
+    pub fn initialize_field_unlog_table_as_unlogged<VM: VMBinding>(&self) {
+        let meta = *VM::VMObjectModel::GLOBAL_FIELD_UNLOG_BIT_SPEC
+            .as_spec()
+            .extract_side_spec();
+        let start: *mut u8 = address_to_meta_address(&meta, self.start()).to_mut_ptr();
+        let limit: *mut u8 = address_to_meta_address(&meta, self.end()).to_mut_ptr();
+        unsafe {
+            let bytes = limit.offset_from(start) as usize;
+            std::ptr::write_bytes(start, 0xffu8, bytes);
+        }
```
wks, line 369: This should use `bset_metadata` or some specialized method for `Region`, too.  Strangely `clear_field_unlog_table` uses `bzero_metadata`, but this method doesn't.  Maybe this method is hotter than the `clear_field_unlog_table` so the performance is more important.

src/policy/immix/block.rs
```diff
@@ -195,16 +323,119 @@ impl Block {
...
+        let start =
+            address_to_meta_address(&crate::util::rc::RC_TABLE, self.start()).to_ptr::<UInt>();
+        let limit =
+            address_to_meta_address(&crate::util::rc::RC_TABLE, self.end()).to_ptr::<UInt>();
+        let rc_table = unsafe { std::slice::from_raw_parts(start, limit.offset_from(start) as _) };
+        for x in rc_table {
+            if *x != 0 {
+                return false;
+            }
+        }
```
wks, line 391: This is a linear searching algorithm, but at byte granularity.  If performance is important, it deserves a specialized method in side metadata.

src/policy/immix/block.rs
```diff
@@ -195,16 +323,119 @@ impl Block {
...
+
     /// Sweep this block.
+    /// Return true if the block is swept.
     pub fn sweep<VM: VMBinding>(
         &self,
         space: &ImmixSpace<VM>,
         mark_histogram: &mut Histogram,
         line_mark_state: Option<u8>,
-    ) -> BlockSweepResult {
+    ) -> bool {
```
wks: I think the `BlockSweepResult` is accidentally rolled back during rebasing.  It is easy to restore the `BlockSweepResult` return value so that we can see block-reuse statistics in the eBPF timeline.

But since a call of `space.reusable_blocks.push` in this method is removed, we need to confirm if any behavior related to block reusing is changed in this PR.

wks (commented):
I haven't gone through the code.  I just listed some obvious problems so that we can address them one by one.

src/util/metadata/side_metadata/spec_defs.rs
```diff
@@ -59,6 +59,10 @@ define_side_metadata_specs!(
     VO_BIT       = (global: true, log_num_of_bits: 0, log_bytes_in_region: LOG_MIN_OBJECT_SIZE as usize),
     // Track the index in SFT map for a chunk (only used for SFT sparse chunk map)
     SFT_DENSE_CHUNK_MAP_INDEX   = (global: true, log_num_of_bits: 3, log_bytes_in_region: LOG_BYTES_IN_CHUNK),
+    // Reference counts
+    RC_TABLE = (global: true, log_num_of_bits: crate::util::rc::LOG_REF_COUNT_BITS, log_bytes_in_region: crate::util::rc::LOG_MIN_OBJECT_SIZE),
+    // Record defrag state for immix blocks
+    IX_BLOCK_DEFRAG = (global: true, log_num_of_bits: 3, log_bytes_in_region: crate::policy::immix::block::Block::LOG_BYTES),
```
qinsoon: What is the reason that this metadata is moved to global?

src/util/alloc/immix_allocator.rs
```diff
@@ -308,14 +302,6 @@ impl<VM: VMBinding> ImmixAllocator<VM> {
                     block.start(),
                     block.end()
                 );
-                // Bulk clear stale line mark state
-                Line::MARK_TABLE
-                    .bzero_metadata(block.start(), crate::policy::immix::block::Block::BYTES);
-                // mark objects if concurrent marking is active
-                if self.immix_space().should_allocate_as_live() {
```
qinsoon, line 315: I am not sure why it is removed. I didn't find an alternative approach in the PR to mark new allocated objects as live.

src/util/metadata/side_metadata/global.rs
```diff
@@ -1463,6 +1507,12 @@ impl SideMetadataContext {
         // and both policies use the chunk map, we just add the chunk map table globally.
         ret.push(crate::util::heap::chunk_map::ChunkMap::ALLOC_TABLE);
 
+        // `Block::init()` unconditionally records per-block defrag state (e.g. whether a block
+        // is a defrag source) for every Immix-family plan, not just plans with reference
+        // counting enabled. So this table must always be mapped, regardless of which plan
+        // requests it.
+        ret.push(crate::policy::immix::block::Block::DEFRAG_STATE_TABLE);
```
qinsoon: This is added by Claude during fixing mmtk-core test/build. It is clearly wrong. We need to revert this.

src/scheduler/scheduler.rs
```diff
@@ -673,16 +716,33 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
...
+    pub(super) fn schedule_concurrent_packets(&self, mmtk: &MMTK<VM>) -> bool {
+        // Only LXR defers work into the inactive queue (via `add_deferred`/`bulk_add_deferred`),
+        // so only LXR needs the `Concurrent` bucket's queue flipped here. For every other plan
+        // (e.g. the generic `ConcurrentImmix`), flipping would silently orphan any packets left
+        // in the currently-active queue when a STW pause interrupts an in-progress concurrent
+        // phase: `is_empty()` would check the other (empty) queue and the bucket would be closed
+        // as if drained, even though unprocessed concurrent-marking work is still sitting in the
+        // now-inactive queue. Keep this the same enable/disable-only mechanism as master unless
+        // the plan is LXR.
+        let is_lxr = mmtk.get_plan().downcast_ref::<LXR<VM>>().is_some();
```
qinsoon: This is a workaround for now. Should fix it before merging the PR.

src/policy/immix/immixspace.rs
```diff
@@ -383,18 +468,26 @@ impl<VM: VMBinding> ImmixSpace<VM> {
...
+            rc: RefCountHelper::NEW,
         }
     }
 
     /// Flush the thread-local queues in BlockPageResource
     pub fn flush_page_resource(&self) {
-        self.reusable_blocks.flush_all();
+        // FIXME: Do we need this for LXR? We observed this to cause fails on conix.
+        if !self.rc_enabled {
+            self.reusable_blocks.flush_all();
```
qinsoon, line 502: This line was removed from this PR. Now it has been added back for non-LXR plans.

src/util/alloc/immix_allocator.rs
qinsoon: This is added back for non-LXR plans.

qinsoon:
binding-refs
OPENJDK21_BINDING_REPO=wenyuzhao/mmtk-openjdk
OPENJDK21_BINDING_REF=lxr-x/jdk21-simplified
OPENJDK11_BINDING_REPO=qinsoon/mmtk-openjdk
OPENJDK11_BINDING_REF=adapt-lxr-related-api-changes
JULIA_REPO=qinsoon/julia
JULIA_REF=update-mmtk-lxr-api-changes
JIKESRVM_BINDING_REF=update-mmtk-lxr-api-changes
RUBY_BINDING_REF=update-mmtk-lxr-api-changes

src/policy/immix/block.rs
wks: It was an error during rebasing, we restored this line.

src/plan/tracing/mod.rs
wks: We merged the two implementations to into just `ObjectReference::iterate_fields`

src/plan/mutator_context.rs
wks: We restored it.  Currently, only NoGC uses `unreachable_prepare_func`.

src/plan/mutator_context.rs
wks: `Barrier::assert_is_flushed` was removed in https://github.com/mmtk/mmtk-core/pull/650  I removed `Mutator::assert_is_flushed` here.

src/plan/lxr/global.rs
wks: The whole file `src/plan/lxr/global.rs` mentions `self.immix_space.scheduler()` many times in multiple function.  This is an instance of [Feature Envy](https://refactoring.guru/smells/feature-envy) and is an indication that the LXR plan (or all plans in general) has the need to interact with the `Scheduler` to add work packets.  It shows a structural problem, but doesn't prevent this PR from being merged.  We may choose to fix it now, or fix it in the scope of the whole MMTk for all plans in the future.

src/policy/immix/block.rs
wks: I restored the `BlockSweepResult`.

src/plan/barriers.rs
wks: I moved them to `log_bit.rs`

src/policy/largeobjectspace.rs
```diff
@@ -31,15 +36,37 @@ pub struct LargeObjectSpace<VM: VMBinding> {
...
         self.get_name()
     }
     fn is_live(&self, object: ObjectReference) -> bool {
+        if self.rc_enabled {
+            if self.is_end_of_satb_or_full_gc {
+                return self.is_marked(object) && self.rc.count(object) > 0;
+            }
+            return self.rc.count(object) > 0;
+        }
+        if self.trace_in_progress {
```
qinsoon: This causes correctness issues. `trace_in_progress` is set in the space's `prepare/release`. So LOS objects are considered as 'live' between prepare and release. If a weak ref points to an LOS object, the LOS object is falsely considered as 'live' and the weak ref won't be cleared. But the actual LOS object is reclaimed -- creating a dangling pointer.

src/util/metadata/side_metadata/spec_defs.rs
```diff
@@ -77,6 +79,14 @@ define_side_metadata_specs!(
     IX_BLOCK_DEFRAG = (global: false, log_num_of_bits: 3, log_bytes_in_region: crate::policy::immix::block::Block::LOG_BYTES),
     // Mark blocks by immix
     IX_BLOCK_MARK   = (global: false, log_num_of_bits: 3, log_bytes_in_region: crate::policy::immix::block::Block::LOG_BYTES),
+    // Straddle line marks
+    RC_STRADDLE_LINES = (global: false, log_num_of_bits: 3, log_bytes_in_region: crate::policy::immix::line::Line::LOG_BYTES),
```
qinsoon, line 83: This is 1 byte per line, but is only used to store 1s or 0s -- is this intended for performance?

Also this side metadata does not appear to be related with RC, even though it is in the `RefCountHelper` and have `RC_` in its name.

src/scheduler/scheduler.rs
```diff
@@ -239,16 +259,16 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         // `VMProcessWeakRefs` packet can be an ordinary packet (doesn't have to be a sentinel)
         // because there are no other packets in the bucket.  We set it as sentinel for
         // consistency.
-        self.work_buckets[WorkBucketStage::VMRefClosure]
-            .set_sentinel(Box::new(VMProcessWeakRefs::<C::DefaultTrace>::new()));
+        // self.work_buckets[WorkBucketStage::VMRefClosure]
```
qinsoon: `VMProcessWeakRefs` and `VMPostForwarding` are skipped. This would cause issues for bindings that use those two APIs.

src/policy/immix/immixspace.rs
```diff
@@ -606,52 +770,104 @@ impl<VM: VMBinding> ImmixSpace<VM> {
...
         #[cfg(feature = "vo_bit")]
         vo_bit::helper::on_trace_object::<VM>(object);
 
-        if self.attempt_mark(object, self.mark_state) {
-            // Mark block and lines
-            if !super::BLOCK_ONLY {
-                if !super::MARK_LINE_AT_SCAN_TIME {
-                    self.mark_lines(object);
+        if self.attempt_mark(object) {
+            if self.rc_enabled {
```
qinsoon, line 850: This conflicts with the previous assertion that this function should not be called for LXR.

src/policy/sft_map.rs
```diff
@@ -471,6 +471,9 @@ mod sparse_chunk_map {
 
     impl SFTMap for SFTSparseChunkMap {
         fn has_sft_entry(&self, addr: Address) -> bool {
+            if addr < vm_layout().heap_start || addr >= vm_layout().heap_end {
```
qinsoon: This causes issues with VM space in JikesRVM/Julia (and potentially other issues). SFT map covers all the addressable memory, rather than the heap range.

qinsoon: A related issue is from `conversions::address_to_chunk_index/chunk_index_to_address`, which computes chunk index relative to heap start. After removing the above line in the sft map (https://github.com/mmtk/mmtk-core/pull/1508/commits/6d7566fc56ad2a98a6760adaa4915f39a5e0c580), we started to see failures in our 32 bits tests.

qinsoon: It seems none of these changes are needed. I reverted them.

src/scheduler/scheduler.rs
```diff
@@ -79,6 +82,25 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         })
     }
 
+    pub fn process_concurrent_packets_in_pause(&self) {
+        let mut packets = vec![];
+        // Buggy
```
wks, line 87: What does this "Buggy" comment refer to?  Is it still considered buggy now?

qinsoon:
I started a performance run to compare simplified LXR and concurrent Immix from this PR, and full featured LXR from https://github.com/wenyuzhao/mmtk-core/tree/lxr. Weak reference and class unloading is disabled. All tests use 16 total GC threads (and 16 concurrent GC threads).

~The experiment is still running. We may not see all the benchmarks until tomorrow.~ I found that I was not using a right build for full featured LXR (I used the jdk11 binding with jdk21, and surprisingly it ran fine without weak ref and class unloading).
~https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|boar-2026-08-18-Tue-020623&benchmark^buildstring^invocation^iteration&time^time.other^time.stw&|10&iteration^1^2|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-lxr-full.lxr.common.threads-16.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&~

The following is the new result. I included 4 configs:
* Full featured LXR
* Full featured LXR without weak ref and class unloading (however this didn't run, so we can't have a 'fair' comparison)
* Simplified LXR without weak ref and class unloading
* Concurrent Immix without weak ref and class unloading

Time
simplified-vs-full-time

Latency
simplified-vs-full-latency

wks:
binding-refs
OPENJDK21_BINDING_REPO=wenyuzhao/mmtk-openjdk
OPENJDK21_BINDING_REF=lxr-x/jdk21-simplified
JIKESRVM_BINDING_REF=update-mmtk-lxr-api-changes
RUBY_BINDING_REF=update-mmtk-lxr-api-changes

src/policy/immix/block.rs
wks: I removed those duplicated objects.  I also introduced `UnstraddlableRegion` which has the `containing` method, and `Line` does NOT implement it.  It forces the programmer to specify whether the line should contain the starting address or the object reference's raw address.

src/util/metadata/side_metadata/spec_defs.rs
wks: `IX_BLOCK_DEFRAG` is now local.  `RC_TABLE` remains global because in theory all objects can have ref count.

src/policy/immix/immixspace.rs
wks: This function can still be called during RC if `LXR_MATURE_EVACUATION` is false.  I changed the assertion and added `if !self.rc_enabled` around code related to the VO bit.

src/scheduler/scheduler.rs
wks: They are now reverted.

wks:
LXR crashes when using MemBalancer (OpenJDK without `-Xms` or `-Xmx`).  The following log is from running GCBench, but it is also reproducible when running lusearch, too.

```
wks@luna ~/c/j/gcbench> MMTK_NO_REFERENCE_TYPES=true MMTK_PLAN=LXR taskset -c 0-15 /home/wks/projects/mmtk-github/parallels/lxr/lxr-x-simplified/openjdk/build/linux-x86_64-server-slowdebug/jdk/bin/java -XX:MetaspaceSize=500M -XX:+UseThirdPartyHeap -cp . GCBench
[2026-08-19T02:42:06Z INFO  mmtk::util::metadata::side_metadata::layout] Initializing side metadata base: vm_specs_registered=true core_end=15069496737792 vm_end=0xfb6a4000000
[2026-08-19T02:42:06Z INFO  mmtk::util::metadata::side_metadata::layout] Quarantine side metadata range: total_bytes=0xfb6a4000000, pages=0xfb6a4000, granularity=0x400000
[2026-08-19T02:42:07Z INFO  mmtk::util::metadata::side_metadata::layout] Side metadata base initialized at 0x6f46d8c00000 (range: 0x6f46d8c00000 - 0x7efd7cc00000)
[2026-08-19T02:42:07Z INFO  mmtk::memory_manager] Initialized MMTk with LXR (DynamicHeapSize(6815744, 8327790592))
Running in a single thread...
Garbage Collector Test
 Stretching memory with a binary tree of depth 20
 Total memory available=6815744 bytes  Free memory=2494464 bytes
[2026-08-19T02:42:07Z INFO  mmtk::util::heap::gc_trigger] [POLL] immix: Triggering collection (1672/1664 pages)
[2026-08-19T02:42:07Z INFO  mmtk::plan::lxr::global] Scheduling RC collection...
[2026-08-19T02:42:07Z INFO  mmtk::scheduler::scheduler] End of GC (1626/1727 pages, took 24.25 ms)
[2026-08-19T02:42:07Z INFO  mmtk::util::heap::gc_trigger] [POLL] immix: Triggering collection (1729/1727 pages)
[2026-08-19T02:42:07Z INFO  mmtk::plan::lxr::global] Scheduling emergency full-heap collection...
[2026-08-19T02:42:07Z INFO  mmtk::scheduler::scheduler] End of GC (823/1664 pages, took 69.32 ms)
[2026-08-19T02:42:07Z INFO  mmtk::util::heap::gc_trigger] [POLL] immix: Triggering collection (1666/1664 pages)
[2026-08-19T02:42:07Z INFO  mmtk::plan::lxr::global] Scheduling RC collection...
[2026-08-19T02:42:07Z INFO  mmtk::scheduler::scheduler] End of GC (1789/2074 pages, took 50.53 ms)
[2026-08-19T02:42:07Z INFO  mmtk::util::heap::gc_trigger] [POLL] immix: Triggering collection (2076/2074 pages)
[2026-08-19T02:42:07Z INFO  mmtk::plan::lxr::global] Scheduling emergency full-heap collection...
[2026-08-19T02:42:08Z INFO  mmtk::scheduler::scheduler] End of GC (2067/2607 pages, took 111.79 ms)
[2026-08-19T02:42:08Z INFO  mmtk::util::heap::gc_trigger] [POLL] immix: Triggering collection (2613/2607 pages)
[2026-08-19T02:42:08Z INFO  mmtk::plan::lxr::global] Scheduling RC collection...
[2026-08-19T02:42:08Z INFO  mmtk::scheduler::scheduler] End of GC (2732/3258 pages, took 40.93 ms)
[2026-08-19T02:42:08Z INFO  mmtk::util::heap::gc_trigger] [POLL] immix: Triggering collection (3263/3258 pages)
[2026-08-19T02:42:08Z INFO  mmtk::plan::lxr::global] Scheduling concurrent marking initial pause...
[2026-08-19T02:42:08Z INFO  mmtk::scheduler::scheduler] End of GC (3485/3258 pages, took 39.41 ms)
[2026-08-19T02:42:08Z INFO  mmtk::util::heap::gc_trigger] [POLL] immix: Triggering collection (3497/3258 pages)
[2026-08-19T02:42:08Z INFO  mmtk::plan::lxr::global] Scheduling concurrent marking final pause...

thread '<unnamed>' (29882) panicked at /home/wks/projects/mmtk-github/parallels/lxr/lxr-x-simplified/mmtk-core/src/plan/lxr/gc_work/mature_evac.rs:135:13:
Invalid slot: OpenJDKSlot { addr: 0x400e9378 } -> 0x827f004

thread '<unnamed>' (29886) panicked at /home/wks/projects/mmtk-github/parallels/lxr/lxr-x-simplified/mmtk-core/src/plan/lxr/gc_work/mature_evac.rs:135:13:
Invalid slot: OpenJDKSlot { addr: 0x40a54794 } -> 0x805000c

thread '<unnamed>' (29880) panicked at /home/wks/projects/mmtk-github/parallels/lxr/lxr-x-simplified/mmtk-core/src/plan/lxr/gc_work/mature_evac.rs:135:13:
Invalid slot: OpenJDKSlot { addr: 0x409f04cc } -> 0x8011002

thread '<unnamed>' (29883) panicked at /home/wks/projects/mmtk-github/parallels/lxr/lxr-x-simplified/mmtk-core/src/plan/lxr/gc_work/mature_evac.rs:135:13:
Invalid slot: OpenJDKSlot { addr: 0x40d31ff8 } -> 0x81e9e94
fish: Job 1, 'MMTK_NO_REFERENCE_TYPES=true MM…' terminated by signal SIGABRT (Abort)
```

The panic happens in the `v & 0b111` branch:

```rust
        let v: u32 = unsafe { s.to_address().load::<u32>() };
        if v & 0b111 != 0 {
            panic!("Invalid slot: {s:?} -> {v:#x}");
        }
```

qinsoon (approved):
LGTM.

There are many issues related to the LXR impl and the other changes introduced in this PR that still need to be addressed properly. Besides, this PR is very large and includes many changes that should have been split into smaller PRs.

Normally, we would not consider a PR in this state ready to merge. However, we have decided to make an exception in this case and address the remaining issues after the merge:
* This PR contains many years' worth of optimization work -- we don't want to accidentally revert important optimizations. Merging it and letting CI track the performance can reduce the risk of causing any regressions.
* This PR introduces changes that may need proper redesign. Merging it and having future refactoring/redesign go through code review is also helpful.
* We are currently under time pressure for LXR-related binding work. Merging the PR first also helps us parallelize future work on LXR.

Thus, I think this is okay to merge this PR right now, as we have done our best to make sure no regression for bindings. This is an exception to our usual review and merge criteria and should not be taken as a precedent for future PRs.

wks:
OpenJDK 11 and 21 tests timed out while running `sudo apt-get update -y`.  It is possible that the network was under load at that time.

qinsoon:
> OpenJDK 11 and 21 tests timed out while running `sudo apt-get update -y`. It is possible that the network was under load at that time.

Yeah. I noticed some tests failed at `apt-get` these few days.

qinsoon:
Updated results with full featured LXR

simplified-vs-full-time
simplified-vs-full-latency
 logs: https://squirrel.anu.edu.au/plotty/yilin/mmtk/#0|boar-2026-08-19-Wed-025947&benchmark^buildstring^invocation^iteration&time^time.other^time.stw&|10&iteration^1^2|20&1^invocation|30&1&benchmark&buildstring;jdk-mmtk-lxr-full.lxr.common.threads-16.dacapochopin|41&Histogram%20(with%20CI)^buildstring^benchmark&

PR #1510, opened 2026-06-08
Remove ptr/ref to Plan from work packets

wks:
`GCWork::do_work` has an `mmtk` parameter, and the `MMTK::get_plan()` and `MMTK::get_plan_mut()` methods provide access to the plan instance. It is therefore pointless for a work packet to capture a pointer or reference inside the struct itself.

This PR does not fix the fact that `Plan::prepare` and `Plan::release` have `&mut self` arguments.  Therefore, getting `&mut Plan` from an `MMTK` instance is still unsafe.  See:
https://github.com/mmtk/mmtk-core/issues/852

### DIFF
--- a/src/plan/compressor/global.rs
+++ b/src/plan/compressor/global.rs
@@ -91,7 +91,7 @@ impl<VM: VMBinding> Plan for Compressor<VM> {
 
         // Prepare global/collectors/mutators
         scheduler.work_buckets[WorkBucketStage::Prepare]
-            .add(Prepare::<CompressorWorkContext<VM>>::new(self));
+            .add(Prepare::<CompressorWorkContext<VM>>::new());
 
         scheduler.work_buckets[WorkBucketStage::CalculateForwarding].add(GenerateWork::new(
             &self.compressor_space,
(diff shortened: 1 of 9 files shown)


wks:
I am not sure if it is worth making this change.  It is always safer to use fewer raw pointers and weird casting statements like `unsafe { &mut *(self.plan as *const _ as *mut _) }`.  But this PR doesn't really make `mmtk.get_plan()` or `mmtk.get_plan_mut()` safe because `MMTK::plan` is still an `UnsafeCell<Box<dyn Plan<VM = VM>>>`.

qinsoon:
> I am not sure if it is worth making this change. It is always safer to use fewer raw pointers and weird casting statements like `unsafe { &mut *(self.plan as *const _ as *mut _) }`. But this PR doesn't really make `mmtk.get_plan()` or `mmtk.get_plan_mut()` safe because `MMTK::plan` is still an `UnsafeCell<Box<dyn Plan<VM = VM>>>`.

I think it is cleaner. With the refactoring, all we need to address is the `unsafe` on `MMTK::get_plan/get_plan_mut`.

PR #1513, opened 2026-06-22
Add mmtk_shutdown API

qinsoon:
This PR adds `mmtk_shutdown`. It requests all GC workers to stop (similar to `prepare_to_fork`).

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -86,6 +86,13 @@ pub fn mmtk_init<VM: VMBinding>(builder: &MMTKBuilder) -> Box<MMTK<VM>> {
     Box::new(mmtk)
 }
 
+/// Shut down an MMTk instance.
+/// This would asycnronously request GC workers to stop, but it will not be blocked until
+/// the GC workers actually quit. A binding needs to check if all GC workers have quit.
+pub fn mmtk_shutdown<VM: VMBinding>(mmtk: &'static MMTK<VM>) {
+    mmtk.shutdown();
+}
+
 /// Add an externally mmapped region to the VM space. A VM space can be set through MMTk options (`vm_space_start` and `vm_space_size`),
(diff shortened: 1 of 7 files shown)


src/memory_manager.rs
```diff
@@ -86,6 +86,13 @@ pub fn mmtk_init<VM: VMBinding>(builder: &MMTKBuilder) -> Box<MMTK<VM>> {
     Box::new(mmtk)
 }
 
+/// Shut down an MMTk instance.
+/// This would asycnronously request GC workers to stop, but it will not be blocked until
+/// the GC workers actually quit. A binding needs to check if all GC workers have quit.
```
wks: > asycnronously

Should be "asynchronously"

> ... it will not be blocked until the GC workers actually quit.

"will not... until..." sounds as if the current thread will be blocked when GC workers actually quit.  We can change the wording a bit to eliminate the ambiguity.

```suggestion
/// This would asynchronously request GC workers to stop and return immediately.
/// Bindings need to check if all GC workers have quit in binding-specific ways.
```

src/scheduler/scheduler.rs
```diff
@@ -100,6 +100,14 @@ impl<VM: VMBinding> GCWorkScheduler<VM> {
         self.worker_monitor.make_request(WorkerGoal::StopForFork);
     }
 
+    /// Ask all GC workers to exit permanently.
+    pub fn shutdown_gc_threads(self: &Arc<Self>) {
+        self.worker_group.prepare_surrender_buffer();
+
+        debug!("A mutator is requesting GC threads to shut down...");
```
wks: We may change level to `info!` or higher.  The log for forking is `debug!` because an application may fork many times in the normal case.  But shutting down MMTk should be important enough for the `info!` log level.

wks (commented):
Just some minor comments.

wks (approved):
LGTM

PR #1515, opened 2026-06-24
Mmap side metadata after space creation

qinsoon:
This PR changes the initialization order: we create the plan and the spaces first, then do initialization for side metadata. This means within plan/space constructor, we should not use side metadata. Any side metadata initialization is done in `Plan/Space::initialize_side_metadata` and is called after side metadata is initialized.

With this PR and https://github.com/mmtk/mmtk-core/pull/1491, we should not see side metadata using any space address range.


wks (approved):
LGTM

PR #1517, opened 2026-06-26
fix(linux): support musl builds

playX18:
On musl ThreadIDType is not Display and is just `*mut c_void`, I updated the helper to return libc::pid_t directly.

### DIFF
--- a/src/util/os/imp/unix_like/linux_like/linux.rs
+++ b/src/util/os/imp/unix_like/linux_like/linux.rs
@@ -56,7 +56,7 @@ impl OSMemory for Linux {
 
 impl OSProcess for Linux {
     type ProcessIDType = unix_common::ProcessIDType;
-    type ThreadIDType = unix_common::ThreadIDType;
+    type ThreadIDType = libc::pid_t;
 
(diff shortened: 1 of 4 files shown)


src/util/os/imp/unix_like/linux_like/linux.rs
```diff
@@ -56,7 +56,7 @@ impl OSMemory for Linux {
 
 impl OSProcess for Linux {
     type ProcessIDType = unix_common::ProcessIDType;
-    type ThreadIDType = unix_common::ThreadIDType;
+    type ThreadIDType = libc::pid_t;
```
qinsoon: Maybe an easier change is to require `ThreadIDType`  to be `Debug` rather than `Display`
```rust
    /// The process ID type for the OS.
    type ProcessIDType: Display + Eq + Copy;
    /// The thread ID type for the OS.
    type ThreadIDType: Display + Eq + Copy;
```

playX18: Hmm that may be better. I'll update it.

qinsoon (approved):
LGTM

PR #1521, opened 2026-07-08
Add a workflow to run cargo fix/fmt etc

qinsoon:
Add a workflow to run `cargo fix`, `cargo clippy --fix` and `cargo fmt`, attempting to resolve some simple CI check failures. The workflow is triggered on a pull request when a label "PR-cargo-auto-fix" is attached. It will commit any change from the above commands back to the PR, and then remove the label.

Done with claude.


wks:
I think it is an overkill to run `cargo fmt --fix` in GitHub Actions,  We can instead add a Git commit hook that prints warnings after `git commit` so that developers will immediately notice there is `cargo fmt` errors before running `git push`.

qinsoon:
> I think it is an overkill to run `cargo fmt --fix` in GitHub Actions, We can instead add a Git commit hook that prints warnings after `git commit` so that developers will immediately notice there is `cargo fmt` errors before running `git push`.

Git commit hooks are local. Some people may not have it set up. A github action can run on any PR with one click to fix those simple problems. It is a QoL improvement for maintainers and contributors.

qinsoon:
To test this, I will open a PR from a branch within the mmtk main repo instead.

PR #1522, opened 2026-07-08
Allow 403 in check-broken-links-in-docs

qinsoon:


wks:
If a website replied 403 because of robots, I think we should stop crawling the site for the sake of respecting their robot policies.

<https://dl.acm.org/robots.txt> contains `Crawl-delay: 1` which asks bots to craw slower, but it doesn't forbid unknown user agents (`*`) from crawling https://dl.acm.org/doi/xx.xxxx/xxxxxxx (where xxx are numbers).  But when I tried running `lychee` locally, or running `curl` to access one of its URLs, it returns 403 which includes a JavaScript to redirect to a captcha page.  So it means it still disallows robot access despite `robots.txt` allows it.  For this reason, I think we should still exclude dl.acm.org because any attempt to visit its pages would be futile.  Its a wast of our CI time, and a waste of ACM's bandwidth.

Steve's homepage is now <https://www.steveblackburn.org/>, and <https://www.steveblackburn.org/robots.txt> doesn't contain any machine-readable directives.  We should be able to craw Steve's homepage.  When I tried running `lychee` locally, it followed 301 link and then replied 200.

```
   [200] https://users.cecs.anu.edu.au/~steveb/pubs/papers/vmmagic-vee-2009.pdf (at 98:688) | 200 OK | Followed 1 redirect. Redirects: https://users.cecs.anu.edu.au/~steveb/pubs/papers/vmmagic-vee-2009.pdf --[301]--> https://www.steveblackburn.org/pubs/papers/vmmagic-vee-2009.pdf
```

I think we should be able to remove Steve's home page from the exclusion list.  But if it still gets 403, we should update the URLs to point to https://www.steveblackburn.org/ instead.

qinsoon:
The motivation is that we see this error recently:
> Errors in docs/userguide/src/tutorial/further_reading.md
>    [403] https://learning.oreilly.com/library/view/the-garbage-collection/9781315388007 | Rejected status code: 403 Forbidden (configurable with "accept" option)

We could specifically add this link to the exclude list. However, I think it might be better to just accept 403, and avoid adding more entries to the exclude list.

wks:
> The motivation is that we see this error recently:
> 
> > Errors in docs/userguide/src/tutorial/further_reading.md
> > [403] https://learning.oreilly.com/library/view/the-garbage-collection/9781315388007 | Rejected status code: 403 Forbidden (configurable with "accept" option)
> 
> We could specifically add this link to the exclude list. However, I think it might be better to just accept 403, and avoid adding more entries to the exclude list.

The Garbage Collection Handbook has its own homepage <https://gchandbook.org/> and it is crawlable.  We should change the URL to that.

I have some mixed feeling about 403.  If a web site replies 403, it means it does not want us to crawl it, and we should respect that.  But our docs may may contain links to web pages that don't allow crawling.  I think maintaining our blacklist is still better.  We can make a contributor policy that if our docs link to a website that is not friendly to crawlers, blacklist it.

qinsoon:
OK. I don't have a strong opinion on this. 

The purpose of our CI check is to make sure as much as we can, that the links are valid for the readers of our docs. If we got 403 on a link, it doesn't mean the link is invalid to readers -- it just means our CI work cannot access it. So treating 403 as acceptable sounds okay to me.

> The Garbage Collection Handbook has its own homepage https://gchandbook.org/

We may not always be able to find an alternative link.

But like I said, I don't have a strong opinion on this, as long as we fix the CI job. Feel free to submit a PR. @wks

qinsoon:
Closed in favor of https://github.com/mmtk/mmtk-core/pull/1527.

PR #1524, opened 2026-07-08
Add a workflow to run cargo fmt

qinsoon:
Add a workflow to run `cargo fmt`, attempting to resolve some simple CI style check failures. The workflow is triggered on a pull request when a label "PR-cargo-auto-fix" is attached. It will commit any change from the above commands back to the PR, and then remove the label.

Done with claude.


.github/workflows/cargo-auto-fix.yml
```diff
@@ -0,0 +1,62 @@
...
+          # Pin to the exact commit rather than the mutable branch name.
+          ref: ${{ github.event.pull_request.head.sha }}
+          # Use a token with push access so we can push the fix commit back to the PR branch.
+          token: ${{ secrets.CI_ACCESS_TOKEN }}
+
+      - name: Install Rust toolchain
+        run: rustup component add rustfmt clippy
+
+      - name: Run cargo fix, clippy --fix, and fmt
+        run: ./.github/scripts/ci-cargo-fix.sh
```
wks, line 41: I think the errors reportd by clippy and fmt belong to two categories.  Formatting errors are almost always acceptible, while clippy warnings usually requires the developer's personal judgement.  It will be better if we can report clippy and fmt errors separately.

.github/workflows/cargo-auto-fix.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Commit and push fixes
+        run: |
+          git config user.name "mmtkgc-bot"
+          git config user.email "mmtkgc.bot@gmail.com"
+          git add .
+          if git diff --cached --quiet; then
+            echo "No changes to commit."
+          else
+            git commit -m "Auto-fix cargo fmt/clippy issues"
+            git push origin HEAD:${{ github.event.pull_request.head.ref }}
```
wks, line 52: This is risky because it is pushing to the developers branch, usually residing in the developer's personal repository.  I think such a direct and automated push is reaching too far.  It is not very polite even if the merge request has granted us the privilege to push into the developers repos.

I think it is still better to use comments to notify humans to make changes.  We may automate it a bit by adding an "accept changes" button after humans review the changes.

wks (commented):
-   `cargo fmt` should usually be OK.  It only involves format changes.
-   `cargo clippy`, on the other hand, judges the correctness of programs using built-in rules, and can often deviate from the developer's true intent.

But regardless whether the suggested changes are good, I think it is not a good idea to directly commit and push to the developer's branch.  From one hand, the changes (specifically the change from `cargo clippy --fix`) may be undesirable.  The developers often want to suppress the warning, or rewrite the program in a way not suggested by `cargo clippy`.  On the other hand, automated commit and push can change the commit tree of the developer's branch, causing their next push to have conflict.

I suggest we do something like the comment from the "github-advanced-security" bot (See https://github.com/ruby/ruby/pull/17190#discussion_r3345594734).  It replies to the pull request, alert the developers to make changes, and auto-closes the warning when the problem goes away.  It also has a button "dismiss alert".

We can slightly add some automation over it by adding an "accept changes" button (if buttons are not possible, we can drive it with a reply message), and the bot will make the push after humans reviewed the changes.

wks:
There is also a risk that in rare cases, it may introduce an infinite stream of commits.

Committing and pushing will trigger the CI.  It will terminate the current work flows, and start a new one.  In rare case, if there is a bug in `cargo clippy` or `cargo fmt`, one change may not fully solve the problem, and cause another warning to be raised by them after the commit and push.  Then it will trigger CI again and the buggy tools will make another push.  This can keep going until it exhausts our CI quota.

I think human supervision must be in the loop.  The best approach, I think, is just alert the developer and let them run `cargo clippy` and `cargo fix` offline, and make a push personally.  If we want automation, we need human to push the "push" button.

.github/workflows/cargo-auto-fix.yml
qinsoon: > It is not very polite even if the merge request has granted us the privilege to push into the developers repos.

Maintainers are only allowed to push to a person's branch if they tick 'allow maintainers to edit my pull request'. By doing that, they give the consent to maintainers to modify their branch.

So if a person does not want maintainers to modify their branch, they should not tick that option -- and this action will fail as it does not permission to push.

We force standard Rust format and lints in our CI. A PR cannot be merged unless they comply and pass the CI checks. This action is simply doing everyone a favor: when you have style issues, you have to fix it -- this action simply automates that.

.github/workflows/cargo-auto-fix.yml
qinsoon: I think a safer approach is to only automate `cargo fmt` in this action.

wks:
Here is a concrete example.  Clippy fixes warnings conservatively, and leaves curly braces to preserve the scope semantics:

```rust
warning: this block may be rewritten with the `?` operator
   --> src/util/rust_util/rev_group.rs:114:13
    |
114 | /             if let Some(item) = self.iter.next() {
115 | |                 // The next group has at least one item.
116 | |                 // This is the key of the group.
117 | |                 let key = (self.get_key)(&item);
...   |
120 | |                 return None;
121 | |             }
    | |_____________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#question_mark
help: replace it with
    |
114 ~             {
115 +                 let item = self.iter.next()?;
116 +                 // The next group has at least one item.
117 +                 // This is the key of the group.
118 +                 let key = (self.get_key)(&item);
119 +                 (item, key)
120 +             }
    |
```

When humans use Clippy, they will try to remove the curly braces.  See:

https://github.com/mmtk/mmtk-core/pull/1529/changes#diff-a6fe2cf1d2ee33eadf7bd233e2b58707e18d79df0c0ed456e808ae4cfef391a5R114

qinsoon:
> Here is a concrete example. Clippy fixes warnings conservatively, and leaves curly braces to preserve the scope semantics:
> 
> ```rust
> warning: this block may be rewritten with the `?` operator
>    --> src/util/rust_util/rev_group.rs:114:13
>     |
> 114 | /             if let Some(item) = self.iter.next() {
> 115 | |                 // The next group has at least one item.
> 116 | |                 // This is the key of the group.
> 117 | |                 let key = (self.get_key)(&item);
> ...   |
> 120 | |                 return None;
> 121 | |             }
>     | |_____________^
>     |
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#question_mark
> help: replace it with
>     |
> 114 ~             {
> 115 +                 let item = self.iter.next()?;
> 116 +                 // The next group has at least one item.
> 117 +                 // This is the key of the group.
> 118 +                 let key = (self.get_key)(&item);
> 119 +                 (item, key)
> 120 +             }
>     |
> ```
> 
> When humans use Clippy, they will try to remove the curly braces. See:
> 
> https://github.com/mmtk/mmtk-core/pull/1529/changes#diff-a6fe2cf1d2ee33eadf7bd233e2b58707e18d79df0c0ed456e808ae4cfef391a5R114

Either is good code. We don't forbid people from submitting code of different styles -- as long as the style matches what our CI expects, and the style does not degrade readability.

qinsoon:
> There is also a risk that in rare cases, it may introduce an infinite stream of commits.
> 
> Committing and pushing will trigger the CI. It will terminate the current work flows, and start a new one. In rare case, if there is a bug in `cargo clippy` or `cargo fmt`, one change may not fully solve the problem, and cause another warning to be raised by them after the commit and push. Then it will trigger CI again and the buggy tools will make another push. This can keep going until it exhausts our CI quota.
> 
> I think human supervision must be in the loop. The best approach, I think, is just alert the developer and let them run `cargo clippy` and `cargo fix` offline, and make a push personally. If we want automation, we need human to push the "push" button.

I don't see how an infinite loop would happen. This action strictly runs once when someone labels 'PR-cargo-auto-fix'.
1. This action will not be cancelled because of new commits (no concurrency control for this action)
2. This action will always remove the label. 
3. This action is not triggered for new commits (even if the label is present).

If you see a reasonable scenario that this action may introduce 'an infinite stream of commits', let me know. I can try to fix the problem.

.github/workflows/cargo-auto-fix.yml
qinsoon: Now the script `ci-cargo-fix.sh` only runs `cargo fmt`.

wks (approved):
LGTM

PR #1527, opened 2026-07-08
Fix links that disallow robot access

wks:
Some web sites do not allow robot access.  We modified links and the Lychee checking options to respect the policies of the web sites.

The GC Handbook now links to https://gchandbook.org/

We no longer exclude the host `users.cecs.anu.edu.au`.  Although the server no longer blocks our Lychee checker now, we modified one affected URL to directly link to https://www.steveblackburn.org/ because it will be redirected anyway.

The link-check workflow now notifies the user about 403 and suggests excluding the affected URLs.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -39,8 +39,7 @@ jobs:
       - name: On link check failure
         if: failure()
         run: |
-          echo "Summary:"
-          cat lychee/out.md
+          echo "Link check failed.  Please read the summary in the last step."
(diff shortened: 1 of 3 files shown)


qinsoon (approved):
LGTM

wks:
FYI, https://github.com/mmtk/mmtk-core/actions/runs/28922504104/job/85802623014#step:5:61 provides an example of CI output when some URLs result in 403.  I removed the `cat lychee/out.md` command from the "On link check failure" step because it is already printed in the previous step.

Issue #1247, opened 2024-12-04
Refactoring the write barrier API


wks:
TL;DR: This issue addresses some recent discussion about write barriers and omitting the `slot` and `target` parameters of the write barrier API functions.

We have some previous discussion about *generalizing* the *subsuming barrier* API for [tagged references](https://github.com/mmtk/mmtk-core/issues/1034) and [atomic RMW (including CAS)](https://github.com/mmtk/mmtk-core/issues/1038).  This issue does not discuss subsuming barrier in depth, except acknowledging that the most general form of subsuming barrier sucks.

# The most general write barrier API sucks!

The write barrier that is ultimately general, w.r.t. the field representation (pointer, compressed pointer, offsetted pointer, tagged pointer, handle, etc.), the operation (store, compare-and-swap, atomic exchange, etc.), whether the object is multi-copy (like Sapphire where write barriers write to both the old copy and the new copy), whether non-reference fields need barriers (like Sapphire), and the kind of barrier (object-logging barrier, field-logging barrier, SATB barrier, XOR zone barrier, generational barrier, etc.), is a **subsuming barrier** that lets the VM binding implement the actual write operation, has multiple `object` fields, an optional `slot` field, an optional  and it can be very complicated.

```rust
fn object_reference_write(mutator: Mutator,
    /// The object
    object: ObjectReference,
    /// For GC like Sapphire, the "new copy" or "old copy" of the current object
    mirrored_object: Option<ObjectReference>,
    /// Only used by field-logging barriers (LXR).  Other barriers just let `operation` do the actual write.
    slot_addr: Option<Address>,
    /// The old target in the slot, or None if the slot was holding NULL, None, nil, false, true, nothing, missing, undef, small integer, etc.
    old_target: Option<ObjectReference>,
    /// The target object, or None if storing NULL, None, nil, false, true, nothing, missing, undef, small integer, etc.
    new_target: Option<ObjectReference>,
    /// A routine provided by VM binding to do the actual write/swap/CAS.
    /// Return the actual old target if different from the `old_target` argument.
    /// This can happen in CAS.
    operation: FnOnce() -> Option<ObjectReference>);
```

[...]

Issue #1038, opened 2023-12-05
Substituting write barriers for atomic read-modify-write operations


wks:
Currently, the `Barrier` trait provides the substituting barrier, `object_reference_write`, which includes both the write operation to the slot, and the write barrier semantics (such as checking the unlogged bits for the object-remembering barrier). However, this is not enough.

# Memory orders

The read and write operations may have acquire and/or release semantics.  For example, in Java,

-   [AtomicReference.set()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/atomic/AtomicReference.html#set(V)) has the strongest memory order, similar to `seq_cst` in C++14.
-   [AtomicReference.setAcquire()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/atomic/AtomicReference.html#setRelease(V)) has the memory order equivalent to `release` in C++14.
-   [AtomicReference.setOpaque()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/atomic/AtomicReference.html#setOpaque(V)) and [AtomicReference.setPlain()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/atomic/AtomicReference.html#setPlain(V)) have weaker memory order.

Implementation-wise, those memory orders translate to different machine-level instructions.

# Atomic read-modify-write operations

Atomic read-modify-write operations do both a read and (optionally) a write operation.  For example, in Java,

-   [AtomicReference.getAndSet(newValue)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/atomic/AtomicReference.html#getAndSet(V)) exchanges the reference in the slot with the reference in the argument.
-   [AtomicReference.compareAndExchange{,Acquire,Release}(expectedValue, newValue)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/atomic/AtomicReference.html#compareAndExchange(V,V)) does strong CAS operation.  It has several variants, with seq_cst, acquire and release semantics, respectively.  It also has variants that do weak CAS operations.
[...]

PR #1531, opened 2026-07-13
Require `Slot` to implement `Sync`.

wks:
`Slot` has pointer semantics and is immutable, so there should be no problem if two threads reference the same `Slot` value at the same time.

In the doc comment, we also mention the possibility that two GC workers may update the same slot concurrently during tracing, and require the VM binding to implement `Slot::store` in a way that is benign to such a race.

### DIFF
--- a/src/vm/tests/mock_tests/mock_test_slots.rs
+++ b/src/vm/tests/mock_tests/mock_test_slots.rs
@@ -68,6 +68,7 @@ mod compressed_oop {
     }
 
     unsafe impl Send for CompressedOopSlot {}
+    unsafe impl Sync for CompressedOopSlot {}
 
     impl CompressedOopSlot {
         pub fn from_address(address: Address) -> Self {
@@ -148,6 +149,7 @@ mod offset_slot {
     }
 
     unsafe impl Send for OffsetSlot {}
+    unsafe impl Sync for OffsetSlot {}
 
     impl OffsetSlot {
         pub fn new_no_offset(address: Address) -> Self {
@@ -183,6 +185,11 @@ mod offset_slot {
         fn store(&self, object: ObjectReference) {
             let begin = object.to_raw_address();
             let middle = begin + self.offset;
+
+            // Note on thread-safety: Because the `OffsetSlot` value itself is immutable,
+            // if multiple GC workers call `store` with the same `object` value,
+            // they always compute to the same `middle` value,
+            // and it is stored via a single atomic store.
             unsafe { (*self.slot_addr).store(middle, atomic::Ordering::Relaxed) }
         }
     }
@@ -242,6 +249,7 @@ mod tagged_slot {
     }
 
     unsafe impl Send for TaggedSlot {}
+    unsafe impl Sync for TaggedSlot {}
 
     impl TaggedSlot {
         // The DummyVM has OBJECT_REF_OFFSET = 4.
@@ -262,6 +270,10 @@ mod tagged_slot {
             ObjectReference::from_raw_address(unsafe { Address::from_usize(untagged) })
         }
 
+        // Note on thread-safety: `TaggedSlot::store` doesn't modify the tag bits.
+        // If multiple GC workers call `store` with the same `object` value,
+        // they will apply the same tag bits, resulting in the same `new_tagged` value.
+        // It is then stored via a single atomic store.
         fn store(&self, object: ObjectReference) {
             let old_tagged = unsafe { (*self.slot_addr).load(atomic::Ordering::Relaxed) };
             let new_untagged = object.to_raw_address().as_usize();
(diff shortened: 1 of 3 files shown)


src/vm/tests/mock_tests/mock_test_slots.rs
```diff
@@ -183,6 +185,11 @@ mod offset_slot {
         fn store(&self, object: ObjectReference) {
             let begin = object.to_raw_address();
             let middle = begin + self.offset;
+
+            // Note on thread-safety: Because the `OffsetSlot` value itself is immutable,
+            // if multiple GC workers call `store` with the same `object` value,
```
k-sareen, line 190: But `middle` is calculated based on what the value of the `object` is. So technically two GC workers, if they are trying to store different values to the same slot, can calculate different `middle` values. In general, we have the lost write problem if we are trying to write two things to a slot (imagine a subsuming barrier, for example). Or is that considered out of scope (and hence undefined behavior) for `Slot`?

wks: It may happen if two GC workers abuse the `Slot::store` method to store arbitrary `ObjectReference`.

But if the GC workers call `Slot::store` for the purpose of updating fields when the target object is moved during a moving GC (this is the intended use case of `Slot::store`), this will not happen.  During a GC, an object is moved to exactly one place.  So if two GC workers trace the same target object, they will receive the same "new `ObjectReference`" from `trace_object`.

```rust
fn process_slot(slot: &impl Slot) {
    if let Some(target) = slot.load() {
        let new_target = trace_object(target); // Always the same `new_target`.
        slot.store(new_target);
    }
}
```

In any trace, `trace_object` shall be idempotent.  Calling `trace_object(target)` will return the same `new_target`, and calling `trace_object(new_target)` will also return `new_target`.  The implementation of `Slot::store` needs to make use of this property to ensure both GC workers will write the same `new_target` value.  I have mentioned this in the doc comment of `Slot::store`.

In the case of `OffsetSlot`, the pointer in the slot is at an offset from the `ObjectReference` of the target object.  The target object may move, but the offset is determined by the source object (more precisely, determined by the specific slot).  Moving GC will not change the offset of the `OffsetSlot`.  Therefore, as long as two GC workers call `slot.store(new_target)` with the same `new_target` value, the computed `middle` variable is guaranteed to be the same.

k-sareen: Subsuming barriers also do `slot.store`, hence why I mentioned it above.

wks: In the doc comment of `Slot::store` I mentioned that it is not suitable for subsuming barriers.  The mere "write operation" sounds simple, but is very VM-specific and can have very rich semantics (See https://github.com/mmtk/mmtk-core/issues/1038 and https://github.com/mmtk/mmtk-core/issues/1247).

qinsoon: 1. It would be very confusing if you can't store into a slot by using `Slot::store`
2. Why can't we use `Slot::store` for write barrier? I looked at the referenced issues, and I only found the following statement in https://github.com/mmtk/mmtk-core/issues/1247, which is a conclusion (not a reason).
>  It needs the address of the field in order to access side metadata. Note that it is not the Slot type which is intended for updating an object graph edge using the Slot::store method, and a Slot may not necessarily be inside the MMTk heap (can be on the stack or in malloc memory).

wks: @qinsoon The main thing is that the `Slot` type is intended for slot updating, so it contains just enough information to update the reference in the slot, such as the offset.  The "write operation" has much richer semantics, as I mentioned in the issues linked.  Here are some examples:

-   type tags (When object is moved, the GC updates the slot and preserves the tag; but when mutator writes to a field, it can update the type tag (which means the slow now holds a reference of a different type), or even update field holding a reference to a non-reference value (Ruby and most dynamically typed languages allow this).  This is not easy to be captured by any VM-agnostic API such as `Slot::store(target: ObjectReference)`.
-   Atomic RMW operations and memory order.

I remember the last time we discussed with Steve, he seemed to agree with an API with a callback:

```rust
fn object_reference_write(object: ObjectReference, field: Slot, updating_function: FnOnce() -> ObjectReference) {
    object_reference_write_pre(object, field);
    updating_function();
    object_reference_write_post(object, field, new_value);
}
```

We can discuss this further.  But `Slot::store(object: ObjectReference)` is not rich enough for this task.

wks:
binding-refs
JULIA_REPO=wks/julia
JULIA_REF=fix/slot-sync

qinsoon (approved):
LGTM

PR #1532, opened 2026-07-15
Set allocation_success before oom call

qinsoon:
This PR fixes a possible infinite loop in allocation. `Collection::out_of_memory` may not return, thus setting `allocation_success` after `out_of_memory` may never be executed. This results in an inconsistent state: MMTk failed an allocation, but `allocation_success` is still recorded as `true`.

Minor edits:
* Rename `reset_allocation_state` to `reset_oom_state`. We can not reset allocation state after the `out_of_memory` call, as it may not return. But we can set oom state, as it is only set and reset after the oom call.

The unit test is written by Claude.


wks (approved):
LGTM

PR #1533, opened 2026-07-15
Add missing prefix pages in MMTk tutorial

qinsoon:
We have empty links in the MMTk tutorial, and the broken link check fails for empty links:
```console
Errors in docs/userguide/src/SUMMARY.md

    [ERROR] error: (at 10:7) | Empty URL found but a URL must not be empty
    [ERROR] error: (at 14:7) | Empty URL found but a URL must not be empty
    [ERROR] error: (at 17:7) | Empty URL found but a URL must not be empty

```

This PR also adds markdown pages for each section and links to the pages properly. `mdbook` seems to use empty links as 'draft chapter', which is not what we intended.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -35,6 +35,7 @@ jobs:
             --cache
             --max-cache-age 1d
             --exclude 'https://dl.acm.org/**'
+            --failIfEmpty false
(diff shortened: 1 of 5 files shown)


wks (approved):
LGTM

PR #1534, opened 2026-07-21
Refactor GcStatus

qinsoon:
This PR refactors the current `GcStatus`, and integrate several states into `GcStatus`, such as `GlobalState::initialized`, `GCTrigger::request_flag`.

### DIFF
--- a/src/global_state.rs
+++ b/src/global_state.rs
@@ -218,12 +218,27 @@ impl Default for GlobalState {
     }
 }
 
+/// The status of MMTk's GC subsystem. This doubles as the "is MMTk initialized" flag (via
+/// [`GcStatus::Uninitialized`]) and as the state that tracks whether a GC is running and, if so,
+/// what phase it is in. See [`GcStatusWord`] for how this is stored atomically and which
+/// transitions between variants are legal.
 #[derive(PartialEq, Copy, Clone, Debug)]
 pub enum GcStatus {
+    /// MMTk has not been initialized yet, i.e. `initialize_collection()` has not been called, so
+    /// there are no GC worker threads available to run a collection. This is the same condition
+    /// reported by [`PauseRequestOutcome::Uninitialized`]: [`GcStatusWord::try_request_pause`]
+    /// returns that variant exactly when the status is `GcStatus::Uninitialized`.
     Uninitialized,
+    /// MMTk is initialized, and no GC is running, pending, or requested.
     NotInGC,
+    /// A concurrent GC's background work (e.g. concurrent marking) is running while mutators
+    /// continue to run normally. See `ConcurrentPlan::concurrent_work_in_progress`.
     InConcurrentGC,
+    /// A stop-the-world pause is active: mutators are stopped and GC workers are doing pause
+    /// work (e.g. tracing).
     InPause,
+    /// A GC pause has been requested (by [`GcStatusWord::try_request_pause`]) but mutators have
+    /// not all stopped yet.
     PauseRequested,
 }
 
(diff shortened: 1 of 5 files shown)


src/global_state.rs
```diff
@@ -222,11 +218,335 @@ impl Default for GlobalState {
...
+/// instead of behind a `Mutex<GcStatus>`. The tag is kept to [`Self::TAG_BITS`] bits (rather than
+/// using the whole word) to leave room for a payload-carrying variant (e.g. a nesting depth) to
+/// be added later without needing to re-encode the rest.
+///
+/// `GcStatus` is a state machine: only a handful of transitions between its variants are legal.
+/// Every legal transition is exposed here as its own method, each performing its own
+/// compare-and-swap retry loop and asserting that the transition is legal for the status it
+/// finds. Do not add a generic "set the status to X" method: doing so would make it possible to
+/// bypass the state machine's invariants.
+pub(crate) struct GcStatusWord(AtomicUsize);
```
qinsoon, line 270: usize seems unnecessary for now, but https://github.com/mmtk/mmtk-core/pull/1457 will store a usize counter into this word.

src/global_state.rs
```diff
@@ -222,11 +218,335 @@ impl Default for GlobalState {
     }
 }
 
-#[derive(PartialEq)]
+#[derive(PartialEq, Copy, Clone, Debug)]
 pub enum GcStatus {
+    Uninitialized,
```
wks, line 231: I assume `GcStatus::Uninitialized` has the same meaning as `PauseRequestOutcome::Uninitialized`.  We should document it (and the `GcStatus` type and other enum variants, too).

src/global_state.rs
```diff
@@ -222,11 +218,335 @@ impl Default for GlobalState {
...
+            match self.0.compare_exchange_weak(
+                cur,
+                Self::encode(next),
+                Ordering::SeqCst,
+                Ordering::SeqCst,
+            ) {
+                Ok(_) => return next,
+                Err(actual) => cur = actual,
+            }
+        }
```
wks: We can use `AtomicUsize::update` because it is simpler and it already implements the "test and test-and-set" optimization.  See: https://github.com/mmtk/mmtk-core/issues/840

src/global_state.rs
```diff
@@ -222,11 +218,335 @@ impl Default for GlobalState {
...
+        let mut cur = self.0.load(Ordering::SeqCst);
+        loop {
+            let next = f(Self::decode(cur));
+            match self.0.compare_exchange_weak(
+                cur,
+                Self::encode(next),
+                Ordering::SeqCst,
+                Ordering::SeqCst,
+            ) {
+                Ok(_) => return next,
```
wks: I think it is better if `transition` returns the *old* `GcStatus` instead of the new value.  `AtomicUsize::update` returns the old value, for a good reason.  If multiple threads try to atomically transition a variable from A to B using `update`, only one thread will return A, and all others will return B.  The thread that returned A knows it "won" the race, while other threads "lost" the race.  If the `update` function returns the new value, and multiple threads want to transition from A to B, then all threads will return B, and it can't identify the winner.

If we can identify the winner, we can let the winner do an operation which should be done only once when transitioning from A to B.

Alternatively, we can let `transition` return a boolean where `true` means winning the race.  Alternatively, returning `Result<GcStatus, GcStatus>` also allows us to use `Ok(old)` and `Err(old)` to identify the winner.

src/global_state.rs
```diff
@@ -222,11 +218,335 @@ impl Default for GlobalState {
...
+            match self.0.compare_exchange_weak(
+                cur,
+                Self::encode(GcStatus::PauseRequested),
+                Ordering::SeqCst,
+                Ordering::SeqCst,
+            ) {
+                Ok(_) => return PauseRequestOutcome::Requested,
+                Err(actual) => cur = actual,
+            }
+        }
```
wks: We can use `AtomicUsize::try_update`.

wks (commented):
The logic is sound, but there are places we can improve documentation, and we can use standard library methods (`AtomicUsize::update` and `AtomicUsize::try_update`) to greatly simplify the implementation.

qinsoon:
`try_update` and `update` are only available in Rust 1.95. We are using Rust 1.92. `fetch_update` is available now, but will be deprecated in Rust 1.99.

Should I use `fetch_update`? Otherwise, we can just keep it as is, and we will get rid of those CAS later.

src/global_state.rs
qinsoon: Done. `GcStatus` is documented. I didn't realize that there was no document for `GcStatus`.

src/global_state.rs
qinsoon: `transition` now returns the old value.

wks:
> `try_update` and `update` are only available in Rust 1.95. We are using Rust 1.92. `fetch_update` is available now, but will be deprecated in Rust 1.99.
> 
> Should I use `fetch_update`? Otherwise, we can just keep it as is, and we will get rid of those CAS later.

Yes.  Our MSRV prevents us from using `update` and `try_update`.  We should use `fetch_update` for now, and make the switch when we bump MSRV.

qinsoon:
> > `try_update` and `update` are only available in Rust 1.95. We are using Rust 1.92. `fetch_update` is available now, but will be deprecated in Rust 1.99.
> > Should I use `fetch_update`? Otherwise, we can just keep it as is, and we will get rid of those CAS later.
> 
> Yes. Our MSRV prevents us from using `update` and `try_update`. We should use `fetch_update` for now, and make the switch when we bump MSRV.

Done. CAS in this PR is replaced with `fetch_update`.

wks (approved):
LGTM

PR #1538, opened 2026-07-30
Update doc comments in `AllocationSemantics`

k-sareen:


k-sareen:
Is there a particular reason why `AllocationSemantics` sits in `mmtk::plan` instead of `mmtk::util::alloc`?

wks (approved):
LGTM

wks:
> Is there a particular reason why `AllocationSemantics` sits in `mmtk::plan` instead of `mmtk::util::alloc`?

Back in JikesRVM, they were constants in the `Plan` abstract class.

https://github.com/JikesRVM/JikesRVM/blob/5072f19761115d987b6ee162f49a03522d36c697/MMTk/src/org/mmtk/plan/Plan.java#L76-L90

It was passed to `Mutator.alloc` which was a virtual function in JikesRVM.  Each plan implements a mutator class, such as `GenImmixMutator`.  What it does is basically dispatching the `Mutator.alloc` invocation to a particular space according to the `int allocator` parameter (equivalent to `AllocationSemantics`).  For example, in `GenImmixMutator.alloc`, if `allocator == GenImmix.ALLOC_MATURE`, it will be dispatched to `GenImmix.immixSpace`.  Otherwise it delegates to the superclass method `GenMutator.alloc`.  If `allocator == Gen.ALLOC_NURSERY`, then it will use a bump pointer to allocate into `Gen.nurserySpace`.

After porting to Rust, it becomes `enum Allocator` and then `enum AllocationSemantics`.  We now dispatch by matching against `AllocationSemantics`, which is equivalent to what JikesRVM was doing.  Since the dispatching is specific to each plan, I think it is legit to keep it in `src::plan`.

k-sareen:
Any particular reason why we should even keep these unused semantics around in the first place?

wks:
> Any particular reason why we should even keep these unused semantics around in the first place?

It was used in JikesRVM.  Currently we rely on the `exec_permission_on_all_spaces` feature to support JikesRVM, but having a dedicated code space should be better.  I think we still need it, but it does not have high priority now.

PR #1539, opened 2026-07-31
Rename on_gc_start/end to on_pause_start/end. Call on_pause_start after all threads stop

qinsoon:
This PR renames `on_gc_start/end` to `on_pause_start/end`.

`on_pause_start` now is called in the `StopMutators` packet after all the threads are stopped. Previously `on_gc_start` is called in `ScheduleCollection` when mutator threads are still running. In some trigger heuristics like mem balancer, it will record the used memory in `on_gc_start/end` to estimate how much memory is freed during the GC. If we call `on_gc_start` when mutator is still running, the used memory is miscounted, as there will be more allocation after the `on_gc_start` is called.

### DIFF
--- a/docs/userguide/src/migration/prefix.md
+++ b/docs/userguide/src/migration/prefix.md
@@ -32,6 +32,31 @@ Notes for the mmtk-core developers:
 
 ## 0.33.0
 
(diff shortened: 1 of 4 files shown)


qinsoon:
The Julia binding uses this API. I will fix it later.

wks (approved):
LGTM

PR #1542, opened 2026-07-31
GC status transition logging

wks:
We log the transition of GC status for debug purposes.


wks:
This log helped me identify the abnormal GC state transition in ConcurrentImmix when working on the lxr-x/simplified branch.

qinsoon (approved):
LGTM

PR #1544, opened 2026-08-04
Add `on_gc_start/end` to `GCTriggerPolicy`

qinsoon:
In addition to `on_pause_start/end`, this PR adds `on_gc_start/end` to `GCTriggerPolicy`. The implementation of `on_pause_start/end` of current trigger policy is now moved to `on_gc_start/end`.

`Plan::end_of_gc` is renamed to `Plan::end_of_pause`.

`on_gc_start/end` is called within the default implementation `Plan::notify_mutators_paused` and `Plan::end_of_pause` with an assertion that the plan is not concurrent. For any plan that overrides those above two `Plan` methods, they need to make sure they call `on_gc_start/end` properly.

### DIFF
--- a/docs/userguide/src/migration/prefix.md
+++ b/docs/userguide/src/migration/prefix.md
@@ -32,19 +32,24 @@ Notes for the mmtk-core developers:
 
 ## 0.33.0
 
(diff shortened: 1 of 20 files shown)


wks (approved):
LGTM

Issue #384, opened 2021-07-22
Remove redundant zeroing for OpenJDK and V8


wenyuzhao:
Currently, MMTk will always use zeroed memory for allocation. However, OpenJDK internally will do another round of zeroing, which is redundant. We need to either disable OpenJDK internal zeroing or make MMTk not doing zeroing at all.

V8 does not require zeroing. So we may want to disable zeroing for V8 as well.

qinsoon:
Bulk zeroing can only be done in mmtk-core. However, if a binding would like to zero each object themselves, we can support this (and avoid double zeroing) by providing an option to disable zeroing in mmtk-core.

qinsoon:
One other thing we may want to check is whether the OS gives zeroed memory with mmap. In Linux, the OS gives us zeroed memory, for which we do not need to zero it again. However, if a piece of memory is reused (not directly from mmap), we still need to zero.

PR #1545, opened 2026-08-04
Allow bindings to disable memory zeroing

qinsoon:
Some bindings/languages do not require MMTk to zero memory. We already knew that OpenJDK does not need zeroing. It seems Julia also does not need zeroing.

Related: https://github.com/mmtk/mmtk-core/issues/384


oscardssmith:
bump? What's needed here?

qinsoon:
> bump? What's needed here?

I would need to check if LXR has new allocation/zero paths, and test with OpenJDK and Julia.

PR #1548, opened 2026-08-10
Update ci-perf-kit to 0.9.1. Check cache before building OpenJDK

qinsoon:
This PR updates ci-perf-kit to https://github.com/mmtk/ci-perf-kit/releases/tag/0.9.1, which includes some style changes about plots, and the index page.

This PR also fixes a redundant build: we expect CI to build OpenJDK for every commit, but for weekly perf runs, we can get the build from the build cache without having to build it again.


k-sareen (approved):
LGTM. Thanks!

PR #1549, opened 2026-08-12
CI no longer uploads perf run logs

qinsoon:


k-sareen:
Why don't we want to upload logs? Isn't it helpful for debugging performance regressions? To figure out all the exact versions etc.

k-sareen (approved):
LGTM. Thanks!

PR #1550, opened 2026-08-13
GC poll from the last parked GC worker

### LINKED ISSUE #1523: Immediately trigger final mark pause when concurrent marking finishes
In the current concurrent Immix (or LXR) implementation, when concurrent marking is done, a final mark pause is only triggered when GC is polled (in the allocator slowpath, or `gc_poll`). Alternatively, when MMTk detects the concurrent work is done, it can immediately trigger a final mark pause. This triggering will bring all mutators to stop at the next safepoint.

We are not sure if this would have any measurable performance difference, as GC poll usually happens frequent enough. But allowing a mechanism like this could be useful.

qinsoon:
This PR allows the last parked GC worker to poll for a GC. When a concurrent GC has done its concurrent work, we can directly poll for a pause, instead of waiting for a failed allocation. Closes https://github.com/mmtk/mmtk-core/issues/1523.

This is not just an optimization. It is also needed for correctness. https://github.com/mmtk/mmtk-core/pull/1457 added a `gc_disable` API, which asks the binding to block if a concurrent GC is happening and MMTk cannot disable GC right now. If the only mutator thread calls `gc_disable` and gets blocked, no allocation would be made. Thus no one would poll for a pause when the concurrent marking is done. The mutator is blocked to wait for the concurrent GC to finish, but the concurent GC cannot finish as no one would poll for the final mark pause.


src/util/heap/gc_trigger.rs
```diff
@@ -162,6 +165,50 @@ impl<VM: VMBinding> GCTrigger<VM> {
...
+        if !concurrent_plan.concurrent_work_in_progress() {
+            return false;
+        }
+        if !self.scheduler.work_buckets[crate::scheduler::WorkBucketStage::Concurrent].is_drained()
+        {
+            return false;
+        }
+
+        let plan = self.plan();
+        if self.policy.is_gc_required(false, None, plan) {
```
qinsoon, line 189: We could additionally pass an argument to the plan to indicate that this is a poll from GC threads.

wks (approved):
LGTM

qinsoon:
I suggest we merge this PR after the LXR PR, in case this may cause any issue for LXR.

qinsoon:
https://github.com/mmtk/mmtk-openjdk/actions/runs/32793638361/job/97644261036 tested this PR. No regression (only pmd digest validation issue, and LXR's tomcat issue -- both are known issues)

Issue #1357, opened 2025-07-30
Cancellable (or upgradeable) concurrent GC


wks:
We have agreed that we should let https://github.com/mmtk/mmtk-core/pull/1355 get merged as long as it works, even if not perfectly.  But to properly support cancellable (or upgradable) concurrent GC, there are many things to be considered.

# Motivation

This is motivated by the Concurrent Immix plan.  A concurrent GC consists of an initial pause, a concurrent marking phase, and a final pause.  During the concurrent marking phase, mutators can be running concurrently with GC workers.  There is a possibility that mutators may run out of memory during concurrent marking.  At this time, we have multiple choices of what to do with it.

1.  Let the allocating mutator block until the current GC finishes (after the final pause).  Other mutators can continue, and may block, too, if they try to allocate and fail, too.
2.  Cancel the current concurrent marking, and restart with a stop-the-world GC.  (In other words, we "upgrade" the concurrent GC to a STW GC.)  Other mutators will stop, too, waiting for the next STW GC.

I (Kunshan) personally favors the first option, i.e. letting the concurrent GC finish.  Reasons are:

-    Cancelling the current GC will result in all the work done by the concurrent GC to be wasted.
-    The complexity of cancelling an already running GC is high, including resetting metadata, etc., and is prone to error.
-    Root-unreachable objects conservatively kept alive during this concurrent GC are guaranteed not to be reached during the next GC (even if it is concurrent).

On the other hands, the main reason for supporting cancellable concurrent GC is that

-   A concurrent GC works on the heap snapshot when the GC was triggered (i.e. the snapshot-at-the-beginning, SATB), and objects allocated during the concurrent GC are conservatively considered alive.  If the mutators manage to exhaust the memory, they must have been allocating so fast and generated many garbage objects.  Therefore the SATB-based concurrent GC may not be freeing up memory fast enough to keep up with the allocation speed of mutators.

This issue is mainly about what to do if we *do* want to implement cancellable (or upgradable) concurrent GC.

# Things to consider

In https://github.com/mmtk/mmtk-core/pull/1355, the InitialMark pause generates work packets in a hidden bucket, and moves the work packets into the Unconstrainted bucket after mutators are resumed.  Then GC workers start working on 

To cancel a concurrent GC, we need to

1.  Let all GC workers stop after finishing their current work packets, and park.
2.  The last parked worker removes all pending work packets from the Unconstrainted bucket.  (Other buckets should be empty.)  Then it closes all buckets.
[...]

PR #1554, opened 2026-08-17
Move unfinished concurrent marking work to final mark pause

qinsoon:
This is another attempt to address https://github.com/mmtk/mmtk-core/issues/1357. Unlike https://github.com/qinsoon/mmtk-core/commit/0e74020d969dd15d22f0b6427169569d4c3dfcb0, this PR will preserve finished concurrent marking progress, and keep doing the rest of marking work in the final mark pause, instead of discarding all the concurrent marking work and losing all the marking progress.


src/plan/concurrent/immix/global.rs
```diff
@@ -405,6 +409,18 @@ impl<VM: VMBinding> ConcurrentImmix<VM> {
         scheduler.work_buckets[WorkBucketStage::Unconstrained]
             .add(StopMutators::<ConcurrentImmixGCWorkContext<VM>>::new_no_scan_roots());
 
+        // If we have unfinished concurrent marking work, do it here.
+        if self.unfinished_concurrent_marking.load(Ordering::SeqCst) {
+            info!(
+                "Concurrent marking was interrupted. Moving remaining work to STW closure bucket."
+            );
+            let leftover_concurrent_work =
+                scheduler.work_buckets[WorkBucketStage::Concurrent].drain_all_packets();
```
wks: At this time, mutators have not stopped, yet, so some mutators can still add work packets to the `Concurrent` bucket.

src/plan/concurrent/immix/global.rs
```diff
@@ -494,4 +510,15 @@ impl<VM: VMBinding> ConcurrentPlan for ConcurrentImmix<VM> {
...
         self.concurrent_marking_in_progress()
     }
+
+    fn on_concurrent_work_interrupted(&self) {
+        assert!(!self.unfinished_concurrent_marking.load(Ordering::SeqCst));
+        // A pause is requested when we are doing concurrent marking.
+        // Set concurrent bucket as disabled now. Later (during collection scheduling),
+        // we will move all the remaining work to a STW bucket and continue.
+        // This preserves all marking progress already made; nothing is reset or re-traced.
+        self.common.base.scheduler.work_buckets[WorkBucketStage::Concurrent].set_enabled(false);
```
wks, line 522: This line disables the `Concurrent bucket`, but it still doesn't guarantee that all mutators will observe the bucket is disabled immediately.

src/plan/concurrent/barrier.rs
```diff
@@ -88,17 +83,22 @@ impl<VM: VMBinding, P: ConcurrentPlan<VM = VM> + PlanTraceObject<VM>, const KIND
...
         }
     }
 
+    /// Which bucket newly-flushed concurrent-marking work should be added to.
+    fn target_bucket(&self) -> WorkBucketStage {
+        if self.mmtk.scheduler.work_buckets[WorkBucketStage::Concurrent].is_enabled() {
+            WorkBucketStage::Concurrent
+        } else {
+            debug_assert_ne!(self.plan.current_pause(), Some(Pause::InitialMark));
+            WorkBucketStage::Closure
```
wks, line 93: The LXR PR introduced a new stage `WorkBucketStage::FinishConcurrentWork` which is the first STW stage, and is designed for this very purpose, i.e. joining outstanding concurrent work.

wks (changes requested):
My understanding is that when a concurrent GC is interrupted (i.e. a pause is triggered when concurrent marking is in progress), a GC worker will close the concurrent bucket and move all work packets to the `Closure` stage (can be the new `FinishConcurrentWork` stage introduced by the LXR PR which is for this very purpose).

But a problem with this is that for a mutator, the two actions: (1) checking whether the Concurrent bucket is enabled, and (2) adding work packet to the target bucket, are not atomic.  It is possible that

1.  A mutator observes the Concurrent bucket is enabled and decides that the target bucket is Concurrent, then
2.  a GC worker running `schedule_concurrent_marking_final_pause` disables the Concurrent bucket, and then
3.  the mutator will attempt to add the work packet to the target bucket, i.e. Concurrent, which is already closed.
4.  Because the Concurrent bucket is disabled, GC workers will not execute the work packets that slipped into the Concurrent bucket in step 3.

I think the easy fix is

1.  Don't disable the Concurrent bucket.
2.  Use a different flag, independent from the "enabled" flag of the Concurrent bucket, to determine the target bucket.
3.  After setting that flag, we still try to drain the Concurrent bucket in `schedule_concurrent_marking_final_pause`.
4.  ~~Add a work packet in the `FinishConcurrentWork` bucket to wait for all outstanding concurrent packets to finish.  This is for handling the corner case that the Concurrent bucket is empty in the last step when we try to drain the Concurrent bucket.  In that case, the work packet stage after Closure (i.e. `SoftRefClosure`) may be erroneously opened because the Concurrent bucket is not part of its opening condition, although `FinishConcurrentWork` (as the first STW stage) is part of the open condition.~~  Given that we can't open new work buckets until all GC workers have parked, it should be impossible.

It should be benign to let a few packets slip into the Concurrent bucket.  As long as the flag redirects the target bucket, and the Concurrent bucket is still open, GC workers will continue to execute work packets in the Concurrent bucket, and the Concurrent bucket will eventually be drained.

But there will be an interesting question: if ~~we have a mechanism to wait for all concurrent buckets to finish in `FinishConcurrentWork`~~ we can't open new buckets until all GC workers parked, do we still need to drain the Concurrent bucket?

qinsoon:
> Use a different flag, independent from the "enabled" flag of the Concurrent bucket, to determine the target bucket.

If the concurrent bucket is enabled and is open, a new pause can't start (e.g. the `ScheduleCollection`/`StopAllMutators` packet won't be scheduled -- no priority queue, and the packet queue is likely FIFO). What we would see is something like the perfetto trace in https://github.com/mmtk/mmtk-core/pull/1563 -- MMTk requested a pause, but MMTk did not start working on anything for the pause (including stop-the-world) until the concurrent work is drained.

> we can't open new buckets until all GC workers parked, do we still need to drain the Concurrent bucket?

The benefits of moving work to the normal STW closure bucket is that we will have all the STW workers to work in STW.

src/plan/concurrent/immix/global.rs
qinsoon: I guess we can do this after all the mutators are stopped.

wks:
> If the concurrent bucket is enabled and is open, a new pause can't start (e.g. the `ScheduleCollection`/`StopAllMutators` packet won't be scheduled -- no priority queue, and the packet queue is likely FIFO). What we would see is something like the perfetto trace in #1563 -- MMTk requested a pause, but MMTk did not start working on anything for the pause (including stop-the-world) until the concurrent work is drained.

What about this:

1.  We still disable the `Concurrent` bucket in `on_concurrent_work_interrupted`.  It will prevent workers from getting work packets out of it, but it won't prevent new packets from being added to it.
2.  We don't have to set "target" bucket.  It can just be the `Concurrent` bucket.  Because concurrent workers cannot get work packets out of the Concurrent buckets after it is disabled, all concurrent workers will eventually stop.
3.  When all mutators stopped, we move all work packets from the Concurrent bucket to `FinishConcurrentWork`

> The benefits of moving work to the normal STW closure bucket is that we will have all the STW workers to work in STW.

Yes.  That makes sense if the number of concurrent workers is different from the number of STW workers.

qinsoon:
> > If the concurrent bucket is enabled and is open, a new pause can't start (e.g. the `ScheduleCollection`/`StopAllMutators` packet won't be scheduled -- no priority queue, and the packet queue is likely FIFO). What we would see is something like the perfetto trace in #1563 -- MMTk requested a pause, but MMTk did not start working on anything for the pause (including stop-the-world) until the concurrent work is drained.
> 
> What about this:
> 
>     1. We still disable the `Concurrent` bucket in `on_concurrent_work_interrupted`.  It will prevent workers from getting work packets out of it, but it won't prevent new packets from being added to it.
> 
>     2. We don't have to set "target" bucket.  It can just be the `Concurrent` bucket.  Because concurrent workers cannot get work packets out of the Concurrent buckets after it is disabled, all concurrent workers will eventually stop.
> 
>     3. When all mutators stopped, we move all work packets from the Concurrent bucket to `FinishConcurrentWork`
> 
> 
> > The benefits of moving work to the normal STW closure bucket is that we will have all the STW workers to work in STW.
> 
> Yes. That makes sense if the number of concurrent workers is different from the number of STW workers.

I made those changes, as suggested.

For 2, we still need the 'target' bucket -- that code exists before this PR. If concurrent marking is in progress, add work to the concurrent bucket, otherwise add to the closure bucket.

One further change is that there is a check in the barriers to see if a bucket is disabled. If a bucket is disabled, we should use add_no_notify. If there is a race, that's fine -- we just wait up a worker with no new work they can work on.

wks (approved):
LGTM

qinsoon:
https://github.com/mmtk/mmtk-openjdk/actions/runs/32918285833 checked this PR on OpenJDK -- no regression (tomcat failed for LXR, and sunflow failed for concurrent Immix). Two known issues.

PR #1556, opened 2026-08-19
Add LXR to per-commit perf CI run

qinsoon:
This PR updates `ci-perf-kit` to 0.9.3 with LXR configs, and adds LXR to per-commit perf CI.

I didn't test this yet, but the configs for LXR is similar to other plans: same as concurrent Immix for ci-perf-kit (plot both time and latency), same as sticky Immix for per-commit runs.


wks (approved):
LGTM

Issue #1496, opened 2026-05-18
Remove "double enqueuing" (a.k.a. dual-queue)


wks:
TODO:

-   [ ] Step 1: Refactor both `ProcessSlots` and `ProcessNodes` into single-queue two-level-loop algorithms, with `ProcessSlots` doing edge-order slot enqueuing and `ProcessNodes` doing node-order ObjRef enqueuing (in Angus' terminology, see below).
-   [ ] Step 2: Let `ProcessSlots` try ObjRef-enqueuing (particularly for non-moving GC) and Tuple-enqueuing.  Also rename the work packets to reflect their true nature according to Angus' terminology.

Currently, tracing collectors in MMTk use the `ProcessEdgesWork` and `ScanObjectsWork` work packets.  They spawn each other.

-   `ProcessEdgesWork` traces a vector of slots and spawns `ScanObjectsWork` work packets to scan newly visited children.
-   `ScanObjectsWork` scans a vector of objects, collects their slots and spawns `ProcessEdgesWork` work packets to process the slots.

This is known as "double enqueuing".  This style of implementation spawns one work packet after following every level of reference, introducing overhead of creating work packets and scheduling work packets between workers.  It also effectively makes the traversal [breadth-first](https://en.wikipedia.org/wiki/Breadth-first_search), while the optimal traversing strategy for GC is [depth-first](https://en.wikipedia.org/wiki/Depth-first_search).  @angussidney called it "dual-queue design" in Section 5.2.1 of [his Thesis](https://www.steveblackburn.org/pubs/theses/atkinson-2023.pdf).  He mentioned that single-queue design would be more cache-friendly.

The pull request https://github.com/mmtk/mmtk-core/pull/1468 will replace the `ProcessEdgesWork` trait with the `Trace` trait.  With this change, the `trace_object` method will no longer be tied to "process edges", and it will be much easier to implement tracing with textbook-style two-level loops.  This gives a good opportunity to get rid of the dual-queue design.

# Angus' terminologies

[Angus' Thesis](https://www.steveblackburn.org/pubs/theses/atkinson-2023.pdf) made a distinction (in Section 5.2) between

-   The *timing* of when and object is marked.  It can be **Node**-ordered or **Edge**-ordered equeuing.
-   The *type* of item to put into the queue.  It can be **ObjRef**, **Slot** or a **Tuple** of both ObjRef and Slot.

[...]

PR #1559, opened 2026-08-20
Remove ProcessEdgesBase in LXR

wks:
DRAFT:

-   [ ] Give `LXRStopTheWorldProcessEdges` a better name, without "process edge".
-   [ ] ~~Refactor `LXRStopTheWorldProcessEdges`, inline `process_slots` and `flush` into `do_work` if possible because we don't have to follow the structure of the legacy `ProcessEdgesWork` trait.  Move fields to local variables if possible.  Replace `enqueue` with a local queue and a closure.~~

The `ProcessEdgesBase` was introduced to make the legacy code work.  We remove it.

### DIFF
--- a/src/plan/lxr/gc_work/mod.rs
+++ b/src/plan/lxr/gc_work/mod.rs
@@ -1,9 +1,10 @@
 use super::global::LXR;
+use crate::plan::lxr::gc_work::rc::{ProcessIncs, EDGE_KIND_ROOT};
 use crate::plan::tracing::UnsupportedTrace;
 use crate::scheduler::gc_work::RootKind;
 use crate::util::ObjectReference;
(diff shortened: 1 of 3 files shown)


wks:
`LXRStopTheWorldProcessEdges` is essentially the same as `plan::tracing::closure::ProcessSlots` (more precisely, edge-ordered slot enqueuing tracing), with the `Trace` being the implementation of `mature_evac_trace_object` and `full_gc_trace_object`.  It also has some special handling of root edges and remembered set recorded edges.

I think it would be better to refactor `LXRStopTheWorldProcessEdges` after we refactor `plan::tracing::closure::ProcessSlots` to [remove double enqueuing](https://github.com/mmtk/mmtk-core/issues/1496).  It may provide more insight.

PR #1562, opened 2026-08-24
fix: do not prepare/release NonMoving space in nursery GC

### LINKED ISSUE #1560: NonMoving space objects are swept away in Nursery GC
mmtk at commit: fb806c0a30a517a3236c55f1e0949571ac0d4970
OS: x86_64 Linux,
binding: [CapyScheme](https://codeberg.org/playxe/capy)

Hey all! I have been testing GenImmix/GenCopy plans across R7RS benchmarks and found out that NonMoving space is not traced appropriately during young GC. https://github.com/mmtk/mmtk-core/blob/fb806c0a30a517a3236c55f1e0949571ac0d4970/src/plan/generational/global.rs#L211 Only traces LOS or nursery objects and leaves NonMoving objects unmarked, which in turn leads them to being released, since https://github.com/mmtk/mmtk-core/blob/fb806c0a30a517a3236c55f1e0949571ac0d4970/src/plan/global.rs#L899 does always release NonMoving space objects.

playX18:
Resolves #1560

### DIFF
--- a/src/plan/generational/global.rs
+++ b/src/plan/generational/global.rs
@@ -74,6 +74,7 @@ impl<VM: VMBinding> CommonGenPlan<VM> {
     /// Release Gen. This should be called by a single thread in GC release work.
     pub fn release(&mut self, tls: VMWorkerThread) {
         let full_heap = !self.is_current_gc_nursery();
+
         self.common.release(tls, full_heap);
         self.nursery.release();
     }
--- a/src/plan/generational/global.rs
+++ b/src/plan/generational/global.rs
@@ -62,10 +62,16 @@ impl<VM: VMBinding> CommonGenPlan<VM> {
     /// Prepare Gen. This should be called by a single thread in GC prepare work.
     pub fn prepare(&mut self, tls: VMWorkerThread) {
         let full_heap = !self.is_current_gc_nursery();
+
+        // Only in case of full heap collection we prepare other spaces for collection that are not generational e.g NonMoving.
+        // LOS is generational space and is part of common plan so it is prepared in both cases.
         if full_heap {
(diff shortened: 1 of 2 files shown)


qinsoon (changes requested):
I think modifying `CommonGenPlan::prepare/release` would be cleaner.

Have something like 
```rust
    /// Prepare Gen. This should be called by a single thread in GC prepare work.
    pub fn prepare(&mut self, tls: VMWorkerThread) {
        let full_heap = !self.is_current_gc_nursery();
        if full_heap {
            self.full_heap_gc_count.lock().unwrap().inc();
            self.common.prepare(tls, full_heap);
        } else {
            self.los.prepare(full_heap);
        }
        self.nursery.prepare(true);
        self.nursery
            .set_copy_for_sft_trace(Some(CopySemantics::PromoteToMature));
    }

    /// Release Gen. This should be called by a single thread in GC release work.
    pub fn release(&mut self, tls: VMWorkerThread) {
        let full_heap = !self.is_current_gc_nursery();
        if full_heap {
            self.common.release(tls, full_heap);
         } else {
            self.common.los.release(tls, full_heap);
         }
        self.nursery.release();
    }
```

This would be doing the same thing as sticky Immix -- see how sticky Immix only prepares/releases LOS in a nursery GC and does not call `common.prepare/release`.
https://github.com/mmtk/mmtk-core/blob/fb806c0a30a517a3236c55f1e0949571ac0d4970/src/plan/sticky/immix/global.rs#L119-L156

qinsoon:
The style checks for stable Rust failed, but they are irrelevant for this PR. No need to fix them here.

qinsoon (approved):
LGTM. Thanks.

PR #1563, opened 2026-08-24
Show GC_REQUEST_WAIT time with visualization scripts

qinsoon:
In some bad scenarios where the concurrent marking threads cannot keep pace with allocation, the GC_REQUEST_WAIT (time-to-yield) may dominate. Plot them clearly.
ScreenShot_2026-08-24_200447_076


wks (approved):
LGTM

PR #1565, opened 2026-08-25
Measure time-to-yield and pause-time

qinsoon:
This adds a `LatencySampler` Counter for us to measure time-to-yield and pause-time. When we merge this, time-to-yield and pause-time data will be recorded with our CI. We just need to modify render scripts to render them for certain plans.

### DIFF
--- a/src/global_state.rs
+++ b/src/global_state.rs
@@ -15,8 +15,6 @@ use std::time::{Duration, Instant};
 pub struct GlobalState {
     /// The current GC status.
     pub(crate) gc_status: GcStatusWord,
-    /// When did the last GC start? Only accessed by the last parked worker.
-    pub(crate) gc_start_time: AtomicRefCell<Option<Instant>>,
     /// The time when a GC pause is requested. Used to calculate the time-to-yield metric.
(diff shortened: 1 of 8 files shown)


wks (approved):
LGTM

qinsoon:
91bca960a2df99530180cb9078cd952fe6fa3d64 allows a GC worker to poll a GC. This PR does not properly record the request time for that case, and we saw an assertion failure. 

```
[2026-08-25T05:16:13Z INFO  mmtk::util::heap::gc_trigger] [POLL] Requesting a concurrent GC's closing pause from the last parked GC worker

thread '<unnamed>' (53346) panicked at /home/runner/work/mmtk-core/mmtk-core/mmtk-openjdk/repos/mmtk-core/src/global_state.rs:220:14:
Pause requested time was not recorded
```

The issue should be fixed in 619282f2eb830b993f0cd2a47e82cf00d481950b.

qinsoon:
I further pushed 8f2edcee8e0251387e6289348e720c9a41015170 -- we also have a `gc_start_time` (which is actually pause start time). I think it is a duplicate, and removed it.

@wks If you would like to do another round of review, please go ahead. Otherwise, I will merge this later today.

PR #1567, opened 2026-08-26
Measure 50p instead avg in LatencySampler

qinsoon:
1. Change the avg reported by LatencySampler to 50p/median
2. If no sample at all, return 0 instead n/a (`ci-perf-kit` expects numbers: https://github.com/mmtk/mmtk-core/actions/runs/32908522388/job/98003634871)


wks (approved):
LGTM

PR #1568, opened 2026-08-27
Limited support for more spaces in LXR

qinsoon:
This PR adds limited support for more spaces in LXR. This PR is mostly tailored for the Julia case, which uses the immortal space and the VM space besides the normal LXR Immix space and LOS.

### DIFF
--- a/src/plan/global.rs
+++ b/src/plan/global.rs
@@ -877,10 +877,14 @@ impl<VM: VMBinding> CommonPlan<VM> {
     }
 
     #[allow(clippy::needless_return)]
-    fn prepare_nonmoving_space(&mut self, _full_heap: bool) {
-        // LXR does not support spaces with no LXR support
-        // FIXME: We want to properly deal with this. Essentially any space that is not supported by LXR should not be included when LXR is selected.
-        // This skip just works around for the nonmoving space (defaults to Immix space), for which we see a panic.
+    pub(crate) fn prepare_nonmoving_space(&mut self, _full_heap: bool) {
+        // FIXME: We need to handle nonmoving space properly.
+        // Nonmoving space is a bit special for LXR, as it could be a second ImmixSpace (as opposed to the default ImmixSpace).
+        // It is arguable whether we should use an LXR ImmixSpace here, or use a normal Immix space.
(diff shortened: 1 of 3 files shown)


wks:
This PR modified tracing code to call `self.lxr.common.trace_object`.  But `CommonPlan::release_nonmoving_space` still contains these lines:

```rust
    #[allow(clippy::needless_return)]
    fn release_nonmoving_space(&mut self, _full_heap: bool) {
        // LXR does not support spaces with no LXR support
        // FIXME: We want to properly deal with this. Essentially any space that is not supported by LXR should not be included when LXR is selected.
        // This skip just works around for the nonmoving space (defaults to Immix space), for which we see a panic.
        if *self.base.options.plan == PlanSelector::LXR {
            return;
        }
```

And `CommonPlan::end_of_gc_nonmoving_space` also excludes LXR in the same way.  Is this expected?

qinsoon:
> This PR modified tracing code to call `self.lxr.common.trace_object`. But `CommonPlan::release_nonmoving_space` still contains these lines:
> 
> ```rust
>     #[allow(clippy::needless_return)]
>     fn release_nonmoving_space(&mut self, _full_heap: bool) {
>         // LXR does not support spaces with no LXR support
>         // FIXME: We want to properly deal with this. Essentially any space that is not supported by LXR should not be included when LXR is selected.
>         // This skip just works around for the nonmoving space (defaults to Immix space), for which we see a panic.
>         if *self.base.options.plan == PlanSelector::LXR {
>             return;
>         }
> ```
> 
> And `CommonPlan::end_of_gc_nonmoving_space` also excludes LXR in the same way. Is this expected?

Thanks for the review. I made somechanges to prepare and release for LXR -- LXR only prepare/release those spaces for marking trace.

The non moving space is a bit special, as it is a second Immix Space (LXR is using 'ImmixSpace'). I still use an early return for it. See my updated comments for prepare/release/end_of_gc for the non moving space.

qinsoon:
The OpenJDK tests for this PR passed. This PR is ready for another review. @wks

wks (approved):
LGTM

PR #1569, opened 2026-08-27
Extra statistics on eBPF timeline

wks:
We show the reserved pages and total pages as two curves on the timeline.  This can show the heap pressure and the effectiveness of each GC.  Currently we only collect the data at two times: (1) when all mutator have stopped and (2) at the end of a pause.  We don't interpolate the data, but we may find convenient places to collect the data during mutator time without reducing performance.

In the `PrepareBlockState` work packet of Immix-based plans, we show the number of blocks prepared and the number of blocks prepared and the number of defrag source blocks selected.  This is useful for debugging defragmentation.

We aggregate the following data from individual work packets to the GC bar in the eBPF timeline:

-   The number and kind (normal, pinning or tpinning) of roots scanned.
-   The number of prepared blocks and defrag sources in Immix.
-   The number of blocks swept, reused and unreused during the release stage in Immix.

We still collect those data when the user specified `capture.py -e N` to skip some collections.  In skipped collections, the statistics on individual work packets will not be shown, but the aggregated numbers on the GC bar will still be shown.

### DIFF
--- a/src/policy/immix/immixspace.rs
+++ b/src/policy/immix/immixspace.rs
@@ -1402,6 +1402,9 @@ impl<VM: VMBinding> GCWork<VM> for PrepareBlockState<VM> {
     fn do_work(&mut self, _worker: &mut GCWorker<VM>, mmtk: &'static MMTK<VM>) {
         // Clear object mark table for this chunk
         self.reset_object_mark();
+        // The number of defrag source blocks.  For debugging.
+        let mut num_defrag_source_blocks = 0;
+        let mut num_blocks_prepared = 0;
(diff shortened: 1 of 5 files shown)


wks:
Here is a screenshot.  The perfetto UI is not interpolating the data points.  Since we currently only collect the stats at the beginning and the end of pauses, the graph looks like the data is jumping between values.

image

wks:
I decided not to worry about the reserved pages not being continuous.  It is the actual data we collected.  We may add more data-collection points in the future to make it smoother (and not lose accuracy, of course).

qinsoon (approved):
LGTM

PR #1572, opened 2026-08-31
Support pinning roots for LXR

qinsoon:
This PR implements `create_process_pinning_roots_work` for `LXRRootsWorkFactory`, and adds a bucket before the normal `ProcessIncs` to handle pinning roots. Changes are originated from https://github.com/mmtk/mmtk-core/pull/1564. Mostly done by Claude.

### DIFF
--- a/src/plan/lxr/gc_work/rc.rs
+++ b/src/plan/lxr/gc_work/rc.rs
@@ -4,6 +4,7 @@ use super::super::LXR;
 use super::super::{LAZY_DECREMENTS, MATURE_EVACUATION, NO_EVAC, NURSERY_EVACUATION};
 use super::tracing::LXRConcurrentTraceObjects;
 use super::tracing::LXRStopTheWorldProcessEdges;
+use super::tracing::ProcessModBufSATB;
 use super::ProcessEdgesBase;
 use crate::plan::VectorQueue;
 use crate::policy::immix::block::BlockState;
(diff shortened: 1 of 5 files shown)


src/plan/lxr/gc_work/rc.rs
```diff
@@ -81,6 +84,94 @@ impl<VM: VMBinding, const KIND: EdgeKind> ProcessIncs<VM, KIND> {
...
+                ),
+            ),
+            Pause::FinalMark => worker.add_work(
+                WorkBucketStage::PinningRootsTrace,
+                LXRStopTheWorldProcessNodes::<VM, false>::new(
+                    to_trace,
+                    mmtk,
+                    WorkBucketStage::Closure,
+                ),
+            ),
```
wks, line 159: It's strange that we still need to process roots in FinalMark in SATB, because SATB is a "black root" algorithm and doesn't need any root scanning in FinalMark.  But since `ProcessIncs::do_work` also schedules `LXRStopTheWorldProcessEdges::<_, false>` in FinalMark, we can leave it as is and resolve this issue later.

wks (approved):
LGTM.  The PR is obviously mimicking the style of `LXRStopTheWorldProcessEdges` to create `LXRStopTheWorldProcessNodes`.  We can leave the refactoring later, as I mentioned [here](https://github.com/mmtk/mmtk-core/pull/1559#issuecomment-5353722912)

PR #1573, opened 2026-08-31
Avoid snapshotting newly allocated objects

qinsoon:
Newly allocated objects don't need to be captured by SATB barriers: they are not in the snapshot, and they don't reference anything in the snapshot. We currently still trigger SATB barriers for newly allocated objects (log bits are bulk set to 1 in InitialMark), but it is benign -- the old references are zeros anyway if we do zeroing in MMTk. However, if we disable zeroing, the SATB barrier will read non-zero 'old values', treat them as references and fail.

### DIFF
--- a/src/util/metadata/log_bit.rs
+++ b/src/util/metadata/log_bit.rs
@@ -21,8 +21,8 @@ impl VMGlobalLogBitSpec {
         self.store_atomic::<VM, u8>(object, LOGGED_VALUE, None, order)
     }
 
-    /// Clear the unlog bits for a whole region
-    pub fn mark_region_as_logged(&self, start: Address, size: usize) {
+    /// Bulk clear the unlog bits for a memory range
+    pub fn bulk_mark_as_logged(&self, start: Address, size: usize) {
         if let MetadataSpec::OnSide(side) = self.as_spec() {
             side.bzero_metadata(start, size);
         }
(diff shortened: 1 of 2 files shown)


src/util/metadata/log_bit.rs
```diff
@@ -21,6 +21,13 @@ impl VMGlobalLogBitSpec {
         self.store_atomic::<VM, u8>(object, LOGGED_VALUE, None, order)
     }
 
+    /// Clear the unlog bits for a whole region
+    pub fn mark_region_as_logged(&self, start: Address, size: usize) {
```
wks: The word "region" has a specific meaning in MMTk core, i.e. power-of-two sized and aligned range of memory.  In this case, we are unlogging blocks and lines which are both regions.  But this function should be changed.  We should do one of the following:

-   Rename it to `mark_memory_range_as_logged` or `bulk_mark_as_logged` and keep the parameters unchanged.
-   Alternatively, keep the function name `mark_region_as_logged` but take a `region: impl Region` as parameter.  We can pass `Block` or `Line` to it.

wks (changes requested):
It is sound to skip actually logging the newly allocated objects.  But there is a minor problem with a function name.

src/util/metadata/log_bit.rs
qinsoon: I renamed the function to `bulk_mark_as_logged`.

wks (approved):
LGTM

k-sareen:
Perhaps we need to fix this properly once-and-for-all by refactoring how we initialize metadata (#1256). We had the same problem with the unlog bit for generational GCs (#1169).

PR #1576, opened 2026-09-03
LXR: for bindings with no unified object ref address, skip the last line in a hole

qinsoon:
For bindings that object start may be in the prior line of the object ref address, we conservatively keep the prior line alive by not using the last line in a hole.


wks (approved):
LGTM.  The code works, but we probably need to provide some abstractions over the mark/RC tables and the hole-finding alogrithm make the search for available lines more maintainable in the future.

PR #1577, opened 2026-09-03
LXR: introduce ObjectEnvelope and address non-unified obj ref

qinsoon:


src/util/rc.rs
```diff
@@ -33,6 +34,125 @@ pub const RC_STRADDLE_LINES: SideMetadataSpec =
...
+///                            |         latest possible object start = ref - LOWER
+///                            earliest possible object start = ref - UPPER = envelope start
+/// ```
+///
+/// All the RC metadata is stored at the envelope start.
+///
+/// To avoid the envelope start being outside of the block where the object is in, we reserve the leading 'slack'
+/// bytes of each block for the envelope reservation. See `ImmixAllocator::block_usable_start`.
+///
+/// TODO: Currently both LOS and Immix objects in LXR are using object envelope. But we don't reserve 'slack' for
```
qinsoon: This is no longer true. LOS also reserves some bytes.

src/vm/object_model.rs
```diff
@@ -471,6 +471,33 @@ pub trait ObjectModel<VM: VMBinding> {
...
+    ///
+    /// Defaults to [`Self::OBJECT_REF_OFFSET_LOWER_BOUND`], i.e. a fixed offset, which holds for
+    /// every binding that does not override this.
+    ///
+    /// We should have the invariant:
+    /// * object ref <= object_start + OBJECT_REF_OFFSET_UPPER_BOUND
+    // The doc comment above links to `crate::util::rc::ObjectEnvelope`. The `util::rc` module is
+    // private (not re-exported at the crate root), so that item is unreachable from outside the
+    // crate even though it is declared `pub`. `cargo doc --document-private-items` (used by GC
+    // implementers and by `ci-doc.sh`) documents it anyway, so the link does resolve there.
```
wks: I think we can move `ObjectEnvelop` to a dedicated `pub(crate)` or `pub` mod, such as `util::envelope`.  It is not only useful for RC, but also useful for tracing-based Immix.  In tracing-based Immix, it is still useful for deciding the first line to mark given an object reference. 

It could also possibly be used by other places to find the first region an object could possibly overlap with.

src/util/rc.rs
```diff
@@ -33,6 +34,125 @@ pub const RC_STRADDLE_LINES: SideMetadataSpec =
...
+/// large objects space.
+#[repr(transparent)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ObjectEnvelope(Address);
```
wks: It is a bit awkward if `ObjectEnvelope` only represents the starting address of the envelope.  Intuitively it should represent an address range, *conceptually* including both the start and the end.

I suggest we let `ObjectEnvelope` wrap the `ObjectReference` itself, while it still provides query functions to find the envelope start and end addresses.  In this case, `ObjectEnvelope::start()` will just return `objref - UPPER_BOUND`.

src/util/rc.rs
```diff
@@ -33,6 +34,125 @@ pub const RC_STRADDLE_LINES: SideMetadataSpec =
...
+
+    /// The constant distance between an object's reference address and its envelope start.
+    const fn offset<VM: VMBinding>() -> usize {
+        VM::VMObjectModel::OBJECT_REF_OFFSET_UPPER_BOUND as usize
+    }
+
+    /// The envelope's exclusive end for an object of `size` bytes: the latest address the object
+    /// could possibly end at, i.e. `ref - LOWER + size`.
+    pub fn end<VM: VMBinding>(self, size: usize) -> Address {
+        self.0 + size + Self::slack::<VM>()
```
wks: Intuitively, an envelope has an end, so `envelope.end()` should just return an address without the `size` argument.  We can know the size by calling `ObjectModel::get_current_size`.

If we have already computed the object size, we can introduce a new method `Envelope::end_from_known_size(self, size: usize) -> Address`.

src/util/rc.rs
```diff
@@ -33,6 +34,125 @@ pub const RC_STRADDLE_LINES: SideMetadataSpec =
...
+    /// The test is on the *envelope's* length, `size + slack`, not the object's: a range longer
+    /// than a line can touch three or more lines, and every line but the first and last needs a
+    /// mark. That is why the threshold is `Line::BYTES - slack` rather than `Line::BYTES` --
+    /// with a coarse bound this pulls a lot more objects into straddle marking, which is the
+    /// price of deriving the envelope from a single global offset.
+    ///
+    /// Every guard that decides whether to mark, unmark, or assert goes through here, so they
+    /// cannot disagree: a mark that unmark does not clear is a stray count that outlives its
+    /// object.
+    pub fn needs_straddle_marks<VM: VMBinding>(size: usize) -> bool {
```
wks, line 84: This method is highly specific to the Immix algorithm.  I think it is more appropriate to move this method to `Line` or a dedicated `LineMarkTable` type.  It can call `Envelope::slack::<VM>()` to get the slack value for the VM.

src/util/rc.rs
```diff
@@ -33,6 +34,125 @@ pub const RC_STRADDLE_LINES: SideMetadataSpec =
...
+    }
+
+    /// The envelope whose start is `a`. `a` must be an envelope start, i.e. an address that
+    /// [`Self::of`] produced; the reference-count table only ever holds counts at such addresses.
+    ///
+    /// # Safety
+    ///
+    /// The caller must guarantee that `a` is an envelope start, so that
+    /// [`Self::to_object_reference`] reproduces a real object reference.
+    pub unsafe fn from_start(a: Address) -> Self {
```
wks: Because the distance from the start to the objref is a per-VM constant (the upper bound), `from_start` can still be implemented by adding the upper bound.  It should still remain `unsafe` because it is the caller's responsibility to ensure the argument is the actual start.

src/util/rc.rs
```diff
@@ -33,6 +34,125 @@ pub const RC_STRADDLE_LINES: SideMetadataSpec =
...
+        unsafe { ObjectReference::from_raw_address_unchecked(self.0 + Self::offset::<VM>()) }
+    }
+
+    /// The envelope's start address, i.e. the address its reference count is stored at.
+    pub fn start(self) -> Address {
+        self.0
+    }
+
+    /// The line holding the envelope's start.
+    pub fn line(self) -> Line {
```
wks: This method is specific to Immix, and should be a method of `Line`.

And from the method name, it is not obvious the returned line contains the envelope start.  I suggest the following method:

```rust
impl Line {
    fn containing_envelope_start(envelope: Envelope) -> Line { ... }
}
```

or just `Line::from_unaligned_address(envelope.start())` which I think is clear enough.

Alternatively, we can parameterize the Line to make it general.

```rust
impl Envelope {
  fn first_overlapping_region::<VM: VMBinding, R: Region>(self) -> R { ... }  
  fn last_overlapping_region::<VM: VMBinding, R: Region>(self) -> R { ... }
}
```

src/util/rc.rs
```diff
@@ -205,41 +330,54 @@ impl<VM: VMBinding> RefCountHelper<VM> {
...
+        let line = ObjectEnvelope::of::<VM>(o).line();
         self.count(o) != 0 && unsafe { RC_STRADDLE_LINES.load::<u8>(line.start()) != 0 }
     }
 
+    /// The `[start, end)` range of lines that need a synthetic straddle mark for object `o`.
+    fn straddle_line_range(&self, o: ObjectReference, size: usize) -> (Line, Line) {
+        let envelope = ObjectEnvelope::of::<VM>(o);
+        let start_line = envelope.line().next();
+        let end_line = Line::from_unaligned_address(envelope.end::<VM>(size));
+        let block_end_line = Block::from_unaligned_address(o.to_raw_address()).end_line();
```
wks: We can use `Block::containing(o)` because Block is a `UnstraddlableRegion`.

src/util/rc.rs
```diff
@@ -205,41 +330,54 @@ impl<VM: VMBinding> RefCountHelper<VM> {
...
-        let line = Line::from_unaligned_address(o.to_raw_address());
+        let line = ObjectEnvelope::of::<VM>(o).line();
         self.count(o) != 0 && unsafe { RC_STRADDLE_LINES.load::<u8>(line.start()) != 0 }
     }
 
+    /// The `[start, end)` range of lines that need a synthetic straddle mark for object `o`.
+    fn straddle_line_range(&self, o: ObjectReference, size: usize) -> (Line, Line) {
+        let envelope = ObjectEnvelope::of::<VM>(o);
+        let start_line = envelope.line().next();
+        let end_line = Line::from_unaligned_address(envelope.end::<VM>(size));
```
wks: The last overlapping region of an envelope can be computed using `Line::from_unaligned_address(envelope.end::<VM>(size) - 1)` because `envelope.end::<VM>(size) - 1` must be a byte inside the envelope, and the line constructed from `Line::from_unaligned_address(...)` must contain that byte.  This will make the subsequent check `if end_line > block_end_line` redundant (but we can assert that just to be safe).

We can implement this as `Envelope::last_overlapping_region::<VM, R>()` for convenience.  I think this can be generalized to all non-empty address ranges, but we currently don't have a dedicated "address range" type.

src/util/alloc/immix_allocator.rs
```diff
@@ -242,7 +253,9 @@ impl<VM: VMBinding> ImmixAllocator<VM> {
                 self.immix_space().get_next_available_lines(self.copy, line)
             {
                 // Find recyclable lines. Update the bump allocation cursor and limit.
-                self.bump_pointer.cursor = start_line.start();
+                self.bump_pointer.cursor = start_line
+                    .start()
+                    .max(self.block_usable_start(line.block()));
```
wks, line 258: I suggested skipping the first `slack()` bytes of each block so that the envelope never crosses block boundaries.  But after a second thought, we may think of alternatives because

-   If the `slack()` is large, we may end up wasting too much memory.
-   The upper bound may be infinite, such as the SableVM layout.

In the latter case, we may give MMTk core an option to opt out the optimization of using constants to estimate the object start, but instead always use `ObjectModel::ref_to_object_start`.  One way to do it is to encapsulate related computation to a dedicated type, such as `LineMarkTable` or `LineRCTable` or others.  Operations include:

-   Find the first line to mark.
-   Find the address to store the RC.
-   How many bytes to skip in the beginning of each Block?
-   Should we skip the first line of each hole when allocating?
-   Should we skip the last line occupied by an object when marking lines?

MMTk core may provide different strategies depending on whether the `slack` is finite and whether the VM is able to identify small objects (`envelope < Line::BYTES`) at all.

We should discuss this in tomorrow's meeting and ask Steve for his opinion.

wks (commented):
I think it is good to use helper types to make the logic of RC-related address calculation clear.

I have some suggestions on the implementation of `ObjectEnvelope`.  I think it is easier to let `ObjectEnvelope` encapsulate `ObjectRefrerence` instead of the starting address, and it is clearer to let `ObjectEnvelope` represent a range instead of just the starting address.

And I think `ObjectEnvelope` deserves its own module because it is not only relevant to RC or LXR, but also tracing-based Immix, too.

I also have some thoughts of opting out the optimization of "using constants and envelope to estimate the object start" if the VM does not have an `UPPER_BOUND` (in other words, the `UPPER_BOUND` is infinite) in the case of, for example, the SableVM layout.  We should discuss it with Steve tomorrow.

qinsoon:
Wrapping around an `ObjectReference` sounds fine to me.

> And I think ObjectEnvelope deserves its own module because it is not only relevant to RC or LXR, but also tracing-based Immix, too.

On the contrary, I was considering naming `ObjectEnvelope` as `RCObjectEnvelope`. It is at the moment only used for RC related metadata (RC_TABLE and RC_STRADDLE_LINES). It is possible that it can be used for other cases -- by that time, this type can be moved to where appropriate.

wks:
> ... I was considering naming `ObjectEnvelope` as `RCObjectEnvelope`. It is at the moment only used for RC related metadata (RC_TABLE and RC_STRADDLE_LINES). It is possible that it can be used for other cases -- by that time, this type can be moved to where appropriate.

Well, that's still OK.  We can make it specific to LXR, and later generalize it if we find appropriate.

src/vm/object_model.rs
qinsoon: For now we just keep `ObjectEnvelope` with the `rc` module -- only `rc` is using it. If we plan to use `ObjectEnvelope` in other places, we can move it out of `rc`.

The document unnecessarily mentions `ObjectEnvelope` (Claude wrote it). What this constant defines is more general than `ObjectEnvelope` -- `ObjectEnvelope` is just the first consumer of this constant. I changed the doc here.

src/util/rc.rs
qinsoon: I changed to `ObjectEnvelope(ObjectReference)`. However, I still kept a few places where we take `size` as an argument (including `end()`). LXR tries to avoid redundantly reading object size, and passes the object size around (e.g. `promote` and `promote_with_size`). I kept the same convention for `ObjectEnvelope`.

src/util/rc.rs
qinsoon: See my comments in https://github.com/mmtk/mmtk-core/pull/1577#discussion_r3930133942

src/util/rc.rs
qinsoon: I moved it to `RefCountHelper` instead. I do want to constrain the changes of this PR to RC only at the moment -- we haven't tested if/how this works outside RC.

src/util/rc.rs
qinsoon: Done.

qinsoon:
https://github.com/mmtk/mmtk-openjdk/pull/378 is testing this PR.

qinsoon:
Using `ObjectEnvelope` like this PR means we mark and access metadata both at an offset address from object reference. Steve clarified that this is not necessary. We can still mark metadata at object reference, and only need to consider 'envelope' when we check RC table occupancy to find holes. I think for now, we just go for the solution in https://github.com/mmtk/mmtk-core/pull/1576. I turned this PR into a draft.

PR #1578, opened 2026-09-04
List LXR features in the section CI scripts parse

qinsoon:
Our CI scripts iterate all combinations of features and do checks. LXR features are not in the section that the scripts would parse. This PR moves LXR features below `# -- Non mutually exclusive features --`.


wks (approved):
LGTM

PR #1579, opened 2026-09-07
LXR: set straddle line mark at obj ref

qinsoon:
`RC_STRADDLE_LINE` is read from obj ref address.
https://github.com/mmtk/mmtk-core/blob/4011008057599f2e795562d4129061b7b91f1797/src/util/rc.rs#L220-L231

We should also set it at obj ref address. The object start line is handled by https://github.com/mmtk/mmtk-core/pull/1576.


wks (approved):
LGTM

PR #1580, opened 2026-09-09
LXR: use object log bit for probable write

qinsoon:
This PR adds object log bit to LXR as an optional feature. LXR uses field log bits. However it still implements `object_probable_write`, which does not specify the actual field to write -- so the barrier slowpath has to scan the object, push references to a buffer and log all the fields. For repetitive calls to `object_probable_write` to the same object, object scanning happens for every call. 

Julia happens to use `object_probable_write` often enough. The following is measured during the sysimg step during building Julia.

per GC | With object log bit | Without object log bit | ratio
-- | -- | -- | --
probable write slowpath | 8,162 | 2,098,992 | 257×
field visits | 233,040 | 1,510,149,064 | 6,480×
sysimg step | 62.79 s | 673.16 s | 10.7×
image output | 72.57 s | 1395.96 s | 19.2×
wall | 16:31 | 1:19:53 | 4.8×

This PR adds a feature that allows LXR to use object log bit to remember that an object is scanned in the probable write API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -210,6 +210,14 @@ lxr_stw = ["lxr_no_cm", "lxr_no_lazy"]
 lxr_no_nursery_evac = []
 lxr_no_mature_evac = []
 lxr_no_evac = ["lxr_no_nursery_evac", "lxr_no_mature_evac"]
+# Use the per-object log bit for LXR's barriers if no field is given.
+#
+# LXR uses a field write barrier. However, for APIs like `object_probable_write` where no
+# field is given, LXR's barrier would go through every field of the object every time the
+# barrier is called. In a runtime where this API is used often (such as Julia), we see
+# huge slowdown (10x) due to repetitively scan those objects in every barrier. In those cases,
+# this feature introduces per object log bit for LXR to address the pathological case.
+lxr-object-log = []
 
 # To collect statistics for each GC work packet. Enabling this may introduce a small overhead (several percentage slowdown on benchmark time).
 work_packet_stats = []
--- a/src/plan/lxr/barrier.rs
+++ b/src/plan/lxr/barrier.rs
@@ -24,6 +24,33 @@ use crate::vm::slot::Slot;
 use crate::vm::*;
 use crate::MMTK;
 
+/// Re-arm the per-object log bits that one mutator's barrier cleared, so the next epoch's first
+/// store to each of those objects reaches the barrier again.
+#[cfg(feature = "lxr-object-log")]
+pub struct RearmLoggedObjects<VM: VMBinding> {
+    objects: Vec<ObjectReference>,
+    _p: std::marker::PhantomData<VM>,
+}
+
+#[cfg(feature = "lxr-object-log")]
+impl<VM: VMBinding> RearmLoggedObjects<VM> {
+    pub fn new(objects: Vec<ObjectReference>) -> Self {
+        Self {
+            objects,
+            _p: std::marker::PhantomData,
+        }
+    }
+}
+
+#[cfg(feature = "lxr-object-log")]
+impl<VM: VMBinding> crate::scheduler::GCWork<VM> for RearmLoggedObjects<VM> {
+    fn do_work(&mut self, _worker: &mut crate::scheduler::GCWorker<VM>, _mmtk: &'static MMTK<VM>) {
+        for obj in &self.objects {
+            VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.mark_as_unlogged::<VM>(*obj, Ordering::SeqCst);
+        }
+    }
+}
+
 pub struct LXRFieldBarrierSemantics<VM: VMBinding> {
     mmtk: &'static MMTK<VM>,
     tls: VMMutatorThread,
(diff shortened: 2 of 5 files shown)


src/plan/lxr/barrier.rs
```diff
@@ -235,5 +255,23 @@ impl<VM: VMBinding> BarrierSemantics for LXRFieldBarrierSemantics<VM> {
...
+                None,
+                Ordering::SeqCst,
+            );
+            self.logged_objs.push(obj);
+            // Try keep the logged objects queue small.
+            // We can remove this check, and only clear logged objects during flush.
+            // This would cause logged object queue to grow unboundedly.
+            if self.logged_objs.is_full() {
+                self.clear_and_reset_logged_objects();
+            }
```
wks: I don't think we need to worry about "grow unboundedly".  The size of the `logged_objs` queue is strictly smaller than the number of slots pushed into `self.incs`.

src/plan/lxr/barrier.rs
```diff
@@ -185,6 +201,10 @@ impl<VM: VMBinding> BarrierSemantics for LXRFieldBarrierSemantics<VM> {
         self.flush_weak_refs();
         self.flush_incs();
         self.flush_decs_and_satb();
+        // Ends the coalescing epoch for the objects this mutator logged: each is armed
+        // again, so the next store to it is recorded.
+        #[cfg(feature = "lxr-object-log")]
+        self.clear_and_reset_logged_objects();
```
wks: `BarrierSemantics` will also be flushed when a mutator is flushed.  It may happen during mutator phase when a user thread finishes execution (`memory_manager::destroy_mutator`).  If we want to re-arm the object unlog bits at the next GC, we should flush the buffer into work packets.

src/plan/lxr/gc_work/rc.rs
```diff
@@ -270,6 +270,14 @@ impl<VM: VMBinding, const KIND: EdgeKind> ProcessIncs<VM, KIND> {
...
             }
         };
+        // Promotion above arms the per-field unlog bits, which is what LXR's own barrier
+        // consults. Bindings whose inlined write-barrier fast path cannot name the field
+        // instead gate on the per-object log bit, and nothing else sets it for an object
+        // promoted through the reference-counting path rather than through tracing. Leaving
+        // it clear makes such a binding skip the barrier for every mature object, losing the
+        // decrements and increments that keep reachable objects alive.
+        #[cfg(feature = "lxr-object-log")]
+        VM::VMObjectModel::GLOBAL_LOG_BIT_SPEC.mark_as_unlogged::<VM>(o, Ordering::SeqCst);
```
wks: If we reset object unlog bits in work packets created in `flush()`, this can be omitted.

wks (changes requested):
The idea of using object unlog bit to eliminate redundant object scanning is good, as the performance data has shown.

But the way this PR handles the `logged_objs` buffer is unsound.  If a mutator is terminated before the next GC, it will re-arm the obj unlog bits of all objects it has logged, forcing other threads to scan those objects again when accessed.

There is also an optimization in `object_probable_write` that tries to limit the size of the `logged_obj` buffer.  But I think it is unnecessary, because logged objects are strictly fewer (maybe order of magnitudes fewer) than logged fields.

I suggest we flush the `logged_objs` buffer into work packets like we do in `ObjectBarrier` and `SATBBarrier`, and re-arm the object unlog bits during the next GC.  It can be done by either traversing the flushed `logged_objs` buffers, or bulk-setting the object unlog bits per block.

src/plan/lxr/gc_work/rc.rs
qinsoon: I think we have to keep this. The object log bit starts as 0 for young objects, the probable write will skip the object (as expected). This promotion changes the object from young to old, and we set the log bit as 1. When the log bit is 1, the probable write barrier actually triggers for the object and sets the bit back to 0 (to avoid scanning it again). The flush makes sure if we sets the bit to 0, it will be flipped back to 1 again.

src/plan/lxr/barrier.rs
qinsoon: If we really want to spawn packets to do this reset, the above `logged_objs.is_full()` actually makes sense -- we would spawn packets with fixed sizes. Let me know what you think.

src/plan/lxr/barrier.rs
wks: I mean, when it is full, instead of clear and reset immediately, flush it so that we reset the unlog bits during pause.

```suggestion
            if self.logged_objs.is_full() {
                // flush the logged_objs queue, create work packets, postponing the resetting of object unlog bits to the next pause.
                self.flush_logged_objs();
            }
```

src/plan/lxr/barrier.rs
wks: Yes.  That's what I meant.  We still need to check `logged_objs.is_full()`, but we make work packet when it is full.

src/plan/lxr/gc_work/rc.rs
wks: OK.  That makes sense.  We should keep this `mark_as_unlogged` here.

qinsoon:
I have done the changes. Can you look at this PR again? @wks

wks (approved):
LGTM

oscardssmith:
IMO measurements against the current LXR julia PR aren't that useful for motivation here as the current field barrier isn't fully up to power, but if this works, that's good to me.

Cargo.toml
```diff
@@ -210,6 +210,14 @@ lxr_stw = ["lxr_no_cm", "lxr_no_lazy"]
...
 lxr_no_mature_evac = []
 lxr_no_evac = ["lxr_no_nursery_evac", "lxr_no_mature_evac"]
+# Use the per-object log bit for LXR's barriers if no field is given.
+#
+# LXR uses a field write barrier. However, for APIs like `object_probable_write` where no
+# field is given, LXR's barrier would go through every field of the object every time the
+# barrier is called. In a runtime where this API is used often (such as Julia), we see
+# huge slowdown (10x) due to repetitively scan those objects in every barrier. In those cases,
+# this feature introduces per object log bit for LXR to address the pathological case.
+lxr-object-log = []
```
k-sareen, line 220: Minor style thing: I think we normally use `_` instead of `-` for features. At least I don't think we have any features that use `-`. (Honestly I didn't even know that we could use `-` in feature names).

qinsoon: Fixed with https://github.com/mmtk/mmtk-core/pull/1583.

PR #1582, opened 2026-09-09
fix(scanning): trace child of the object, not object itself

playX18:
Edge tracing had a fatal bug where it would trace object instead of a child and thus panic on !MAY_MOVE_OBJECTS check, and in the end also produce segfault if not panic.


qinsoon (approved):
LGTM

PR #1583, opened 2026-09-11
Rename lxr-object-log to lxr_object_log

qinsoon:


wks (approved):
LGTM.  Actually, "_", "-" and "+" are all allowed, but it's good to be consistent within mmtk-core.

PR #1584, opened 2026-09-11
LXR: add tuning knobs, fix survival predictor for non moving build

qinsoon:
This PR exposes some tuning knobs for LXR as options, and maintains total promotion volume and ratio (besides the existing copy promotion volume and ratio). 

The behavior on OpenJDK using those options' default values should be identical to `master`.

### DIFF
--- a/src/plan/lxr/gc_work/rc.rs
+++ b/src/plan/lxr/gc_work/rc.rs
@@ -18,6 +18,7 @@ use crate::vm::slot::Slot;
 use crate::{
     plan::concurrent::global::ConcurrentPlan,
     plan::concurrent::Pause,
+    plan::global::Plan,
     policy::{immix::block::Block, space::Space},
     scheduler::{GCWork, GCWorker, WorkBucketStage},
     util::{metadata::side_metadata, object_forwarding, ObjectReference},
@@ -189,10 +190,8 @@ impl<VM: VMBinding, const KIND: EdgeKind> ProcessIncs<VM, KIND> {
                 block.set_as_in_place_promoted();
             }
             self.rc.promote_with_size(o, size);
-            if copied {
-                self.survival_ratio_predictor_local
-                    .record_copied_promotion(size);
-            }
+            self.survival_ratio_predictor_local
+                .record_promotion(size, copied);
         } else {
             // println!("promote los {:?} {}", o, self.immix().is_marked(o));
         }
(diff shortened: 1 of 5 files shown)


src/plan/lxr/gc_work/rc.rs
```diff
@@ -623,13 +622,17 @@ impl<VM: VMBinding, const KIND: EdgeKind> GCWork<VM> for ProcessIncs<VM, KIND> {
...
+        let split_size = *self.lxr.base().options.lxr_min_packet_split_size;
+        // Split depth defaults to 16 (split after 16 recursions).
+        let split_depth = *self.lxr.base().options.lxr_min_packet_split_depth;
         while !self.new_incs.is_empty() {
             self.new_incs_count = 0;
             depth += 1;
             incs.clear();
             self.new_incs.swap(&mut incs);
-            if ACTIVE_PACKET_SPLIT && depth >= 16 && incs.len() > 1 {
+            if depth as usize >= split_depth && incs.len() > split_size {
```
wks, line 635: I think the old code `incs.len() > 1` was a bug, and this fixes it.

wks:
`lxr_inc_buffer_limit` uses the value 0 for "unlimited", while `lxr_max_stw_sweep_nursery_blocks` and `lxr_min_packet_split_size` use `usize::MAX`.  Although it doesn't prevent the code work correctly, I think it is better to make them consistent.  I recommend using `usize::MAX` which is consistent with the `DEFAULT_STRESS_FACTOR` which is `usize::max`.

qinsoon:
> `lxr_inc_buffer_limit` uses the value 0 for "unlimited", while `lxr_max_stw_sweep_nursery_blocks` and `lxr_min_packet_split_size` use `usize::MAX`. Although it doesn't prevent the code work correctly, I think it is better to make them consistent. I recommend using `usize::MAX` which is consistent with the `DEFAULT_STRESS_FACTOR` which is `usize::max`.

Makes sense. I have made changes on that.

src/plan/lxr/global.rs
```diff
@@ -108,13 +108,23 @@ impl<VM: VMBinding> Plan for LXR<VM> {
...
-                << LOG_CONSERVATIVE_SURVIVAL_RATIO_MULTIPLER;
-        if predicted_survival_mb >= super::MAX_SURVIVAL_MB {
+        // Use copy promotion ratio, or just total promotion ratio.
+        let ratio = if LXR_CONSTRAINTS.moves_objects {
+            super::SURVIVAL_RATIO_PREDICTOR.copy_promote_ratio()
+        } else {
+            super::SURVIVAL_RATIO_PREDICTOR.promote_ratio()
+        };
+        let predicted_survival_mb: usize = ((total_young_alloc_pages as f64 * ratio) as usize)
+            << LOG_CONSERVATIVE_SURVIVAL_RATIO_MULTIPLER;
```
wks, line 126: This behaves identical to the old code, but I think it may have been a bug.  For copying LXR, it looks like in-place promoted objects didn't survive at all.  However, in-place promoted objects still occupy heap space, and they occupy even more space than evacuated objects due to lack of defragmentation.

I think it is OK for this PR, but we can discuss this in the group.

wks (approved):
LGTM.

I think there are bugs in the old code.  One is fixed by this PR, and the other can be discussed in group.  This PR is OK.

Issue #1585, opened 2026-09-14
Remove the hack about the block/non-block behavior for null tls in `harness_begin`/`handle_user_collection_request`


qinsoon:
In `handle_user_collection_request`, no `block_for_gc` if tls is null.
https://github.com/mmtk/mmtk-core/blob/8996f057c6e69387ad0819db1017894d9654a338/src/mmtk.rs#L353-L356

In `harness_begin`, `block_for_gc` if tls is null.
https://github.com/mmtk/mmtk-core/blob/8996f057c6e69387ad0819db1017894d9654a338/src/mmtk.rs#L469-L471

---

I think the original intention is to use `handle_user_collection_request` from the OpenJDK side to trigger a GC from a VM thread. Thus `block_for_gc` is not needed. The hack is to take null tls as a 'hint' (which is always true in OpenJDK) to skip the block.

To `compensate` this skip, in `harness_begin`, after calling `handle_user_collection_request`, a `block_for_gc` call is added.

I think we should do these:
1. See if we still need to request a GC from OpenJDK's VM threads. https://github.com/mmtk/mmtk-core/pull/1550 already allows polling for GC from an MMTk GC worker.
2. If we still need to request GC from a VM thread, we should introduce a new API, such as `handle_vm_collection_request`, and allows VMs to specify block vs non-block as an argument.
3. The hacks in the above code snippets should be removed.

PR #1586, opened 2026-09-14
Only block for gc if gc is triggered in harness_begin

qinsoon:
A workaround introduced with the LXR merge added `block_for_gc` in `harness_begin` to 'compensate' the skipped `block_for_gc` in `handle_user_collection_request`. However, for `NoGC`, when `harness_begin` is called, the `handle_user_collection_request` does not request GC, and the thread blocked in `block_for_gc` will never be waken up. This caused the hang when we run `NoGC` in our performance CI.

This PR adds a check `gc_triggered` to skip the block for `NoGC`.

Eventually we need to address the issue properly: https://github.com/mmtk/mmtk-core/issues/1585. Otherwise, we will trip on this again.


wks (approved):
LGTM

PR #1589, opened 2026-09-14
Add Wake N for the scheduler: wake up workers based on the number of new work packets.

qinsoon:
Currently we always wake up all the active workers if a new bucket is opened and is not empty. This is wasteful if the bucket only includes a small number of packets (e.g. one `Prepare`/`Release` bucket). Some workers may not get any work -- they are still waken up, and then parked. This PR only wakes up the same number of workers, as the number of work packets from new buckets.

I haven't measured the performance, and I am not sure if this actually improves performance. But the approach sounds reasonable to me.


wks (approved):
LGTM

PR #1591, opened 2026-09-14
Avoid acquiring lock and notify if no worker is parking

qinsoon:
This PR moves `WorkerParker` outside of `WorkerMonitorSync`. The main goal is to access `parked_workers` as an atomic without the lock. This allows us to avoid acquiring the lock and waking up all workers in `notify_work_available` -- which almost happens for every newly added work packet.


wks (approved):
LGTM

PR #1592, opened 2026-09-15
Inform binding and plan if a concurrent GC is done without a pause

qinsoon:
This PR adds hooks to inform the binding and the plan when a concurrent GC is likely done (without a pause). When a concurrent GC is in progress, `GcStatus` is `InConcurrentGC`. For concurrent marking (in concurrent Immix and LXR), we rely on a 'final pause' to set the status to `InPause`, and then `NotInGC`. LXR also has other concurrent work, such as lazy dec and lazy sweep. When those work packets are done, currently there is no way to end the `InConcurrentGC` status. This is wrong book keeping for MMTk.

For the binding side, if a mutator is for some reason waiting for a concurrent GC to finish, it will never get notified when the concurrent GC is done. E.g. a mutator calls `gc_disable` during concurrent work. `gc_disable` will fail, as the current status is `InConcurrentGC`, and the mutator will be blocked. If we don't notify the binding when concurrent work is done, the mutator will be blocked forever until a pause is triggered. This is a correctness issue.

The PR addresses both.


qinsoon:
The `msrv` check failed. From the logs, it looks like a dependency of `cargo-msrv` misbehaved and `cargo-msrv` (released 6 months ago) no longer built. It is not related with this PR, but it will block merging any PR (we require the check to pass). We could either temporarily add `msrv` to the ignored list of merge check, or temporarily allow bypass check to merge on Github.

wks (approved):
LGTM

PR #1594, opened 2026-09-15
Fix LOS mark bits

### LINKED ISSUE #1593: The LOS is not setting mark bit correctly
The LOS uses cyclic mark bits, but when setting the mark bit, it is always using the `old_bit | mark_state` to set the mark bit.  This works when mark_state is 1, but is wrong if mark_bit is 0.

```rust
fn test_and_mark(&self, object: ObjectReference, value: u8) -> bool {
        loop {
            // ...
            let old_value = VM::VMObjectModel::LOCAL_LOS_MARK_NURSERY_SPEC.load_atomic::<VM, u8>(object, ...);
            let mark_bit = old_value & mask;
            if mark_bit == value {
                return false;
            }
            // using LOS_BIT_MASK have side effects of clearing nursery bit
            if VM::VMObjectModel::LOCAL_LOS_MARK_NURSERY_SPEC
                .compare_exchange_metadata::<VM, u8>(
                    object,
                    old_value,
                    old_value & !LOS_BIT_MASK | value, // ERROR
                    None,
                    Ordering::SeqCst,
                    Ordering::SeqCst,
                )
                .is_ok()
            {
                break;
            }
        }
        true
    }
```

On the `// ERROR` line, if `value` is 0, it will be a no-op, and the mark bit will not be cleared (0 means marked).

I think the new value should simply be `value`.  The nursery bit should be cleared because tracing an object will promote the object to the mature space.  And the mark bit should be exactly the value of `value`.

```diff
-                    old_value & !LOS_BIT_MASK | value,
+                    value,
```

We noticed this problem when working with the CRuby binding.  We found that some nursery object have mark bits.  But if

1.  nursery objects must not be "marked", and
2.  all marked objects are kept alive,

then is it OK for the LOS to use sticky mark bit, like in StickyImmix?  This can eliminate the nursery bit, save metadata space, and potentially simplify the marking logic.  Can we use sticky mark bit while also using cyclic mark bit?  Not sure how much profit it will bring, but currently the granularity of `LOCAL_LOS_MARK_NURSERY_SPEC` is "2 bits per page", which doesn't seem to be a lot.

wks:
**DRAFT**:

-   [X] Testing with OpenJDK and CRuby binding.
-   [ ] Ensure LXR works.

Currently, the mark bit of the LOS is handled incorrectly in multiple places.

-   Objects are allocated with mark bit regardless whether
    `should_allocate_as_live()` is true or false.
-   `is_live` simply reads the mark bit.  The result will be wrong if
    the object is the current GC is a full-heap GC (the meaning of the
    mark bit has just been flipped), and the object is a nursery object
    but not marked.
-   `test_and_mark` fails to set the mark bit if the mark state is 0

In this PR, we make it clear that any object in the LOS can be in one of the three states at any time:

-   `Nursery`: It is a nursery object.
-   `MatureUnmarked`: It is a mature object, and has not been marked
    since the last major GC start.
-   `MatureMarked`: It is a mature object, and has been marked since the
    last major GC started.

(There is no "nursery marked" state.  A nursry object is promoted when it is marked.)

We design the bit pattern so that when flipping the meaning of the mark bit, `MatureUnmarked` and `MatureMarked` will swap, but `Nursery` remains `Nursery`.  Concretely,

| State          | nursery bit | mark bit      |
|----------------|-------------|---------------|
| Nursery        | 1           | ignored       |
| MatureUnmarked | 0           | != mark_state |
| MatureMarked   | 0           | == mark_state |

When allocating an object, if allocating as live, the object will be set to the `MatureMarked` state; otherwise set to the `Nursery` state.

`is_live` will simply check if the object is in the `MatureMarked` state.  All nursery objects are considered "not live".

In `test_andd_mark`, we simply atomically set the object state to `MatureMarked`.

Fixes: https://github.com/mmtk/mmtk-core/issues/1593


wks:
Rust 1.84 (MSRV) tests failed because one dependency `yoke-derive` no longer compiles.  We will address this issue by bumping our MSRV and pinned versions.  See https://github.com/mmtk/mmtk-core/pull/1596

src/policy/largeobjectspace.rs
```diff
@@ -525,46 +638,24 @@ impl<VM: VMBinding> LargeObjectSpace<VM> {
...
+    /// already in the marked state.  In either case, `old_state` is the old state before the atomic
+    /// operation.
+    fn test_and_mark(&self, object: ObjectReference) -> Result<u8, u8> {
+        let mark_state = self.mark_state;
+        VM::VMObjectModel::LOCAL_LOS_MARK_NURSERY_SPEC.fetch_update_metadata::<VM, u8, _>(
+            object,
+            Ordering::SeqCst,
+            Ordering::SeqCst,
+            |old_value| {
+                (!mark_nursery_bits_states::is_marked(old_value, mark_state)).then_some(mark_state)
```
qinsoon, line 653: I was unsure why we need `mod mark_nursery_bits_states`. I initially thought it tries to encapsulate the nursery bit (to make it clear that LXR does not use it). But this method is used by both LXR and non-rc plans, and it uses `mark_nursery_bits_states`. This makes me wonder why we need the mod `mark_nursery_bits_states`.

One alternative way is to introduce `LOSMarkState`, and provides two sets of method (normal methods and `rc_` prefixed methods, e.g. `test_and_mark` and `rc_test_and_mark`), for LXR and non-rc plans. Do you think that might be better?

qinsoon (commented):
Generally looks good to me.

src/policy/largeobjectspace.rs
wks: The primary intention of introducing a mod (or a type) is to encapsulate operations related to the two-bit state values.  An `enum MarkNurseryBitsState { Nursery, MatureMarked, MatureUnmarked }` type would be an overkill, and it couldn't encode the fact that `Nursery` has two representations: 0x10 and 0x11.  I could make a `fn decode(value: u8) -> MarkNurseryBitsState` to convert the two-bit value to the `enum`, but it looks unnecessary and wasteful.

I also considered adding `is_marked` and `is_nursery` methods to `LargeObjectSpace` itself, but I think the two-bit state is the subject (i.e. the state should be the `self`, not the `LargeObjectSpace`).  And the state they represent is only related to the two bits themselves and the `mark_state`, not other parts of `LargeObjectSpace`.

In the end, I chose something in the middle: a `mod mark_nursery_bits_states` that contains two methods that queries the meanings of the two-bit state.  I could make it `struct LOSMarkState;` and put the methods in `impl LOSMarkState`.  They are equivalent, but a mod (instead of a singleton type) should be enough.

> ...introduce LOSMarkState, and provides two sets of method...

That was my previous plan.  I had `mod tracing_states` and `mod rc_states`.  But in the end, I found that the set of states are the same (`Nursery`, `MatureMarked` and `MatureUnmarked`).  RC never uses the `Nursery` state, but the state transition is otherwise identical to tracing GC.  (Actually LXR uses the mark bit only during back-up tracing, so they are actually the same.)  The `test_and_mark` method is common.  Its semantics is just an atomic swap (an atomic store that returns the old value).  In both RC and tracing, `test_and_mark` should set the state to `MatureMarked`.  So I just let RC and tracing plans use the same set of states and methods.

qinsoon (approved):

